PR #34, opened 2024-08-26
Set/get chunk property test EAR-1189

dcherian:
One more commit from my testing spike last week. I didn't complete the stateful test bit but this could be useful with our new caching storage?


dcherian:
Seems not necessary any more.

PR #267, opened 2024-10-15
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

TODOs:
- [x] remove TEMP commit after verifying that we do or do not need provided custom action

### DIFF
--- a/icechunk/tests/test_s3_storage.rs
+++ b/icechunk/tests/test_s3_storage.rs
@@ -75,7 +75,7 @@ pub async fn test_chunk_write_read() -> Result<(), Box<dyn std::error::Error>> {
     assert_eq!(Bytes::from_static(b"ello"), back);
 
     let back = storage.fetch_chunk(&id, &ByteRange::to_offset(3)).await?;
-    assert_eq!(Bytes::from_static(b"hel"), back);
+    assert_eq!(Bytes::from_static(b"hel"), back);  // codespell:ignore
 
     let back = storage.fetch_chunk(&id, &ByteRange::bounded(1, 4)).await?;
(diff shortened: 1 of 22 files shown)


paraseba:
I like this! Let's make those checks pass and we can merge. Thank you @yarikoptic !

yarikoptic:
oops -- my absent knowledge of rust -- do not even know how to add a comment correctly, will fix shortly.

But as above single failure showed, you have no run of pre-commit via CI or pre-commit.ci service, so I will keep github action for now.

icechunk/tests/test_s3_storage.rs
```diff
@@ -75,7 +75,7 @@ pub async fn test_chunk_write_read() -> Result<(), Box<dyn std::error::Error>> {
     assert_eq!(Bytes::from_static(b"ello"), back);
 
     let back = storage.fetch_chunk(&id, &ByteRange::to_offset(3)).await?;
-    assert_eq!(Bytes::from_static(b"hel"), back);
+    assert_eq!(Bytes::from_static(b"hel"), back);  // codespell:ignore
```
paraseba, line 78:
```suggestion
    assert_eq!(Bytes::from_static(b"hel"), back); // codespell:ignore
```

yarikoptic: Thanks! I will take your help here ;)

paraseba:
Thank you @yarikoptic !

paraseba (approved):

PR #269, opened 2024-10-15
Fix stale link in readme to spec

asinghvi17:


paraseba (approved):
Thank you @asinghvi17

Issue #280, opened 2024-10-16
[Docs] errors in xarray page


callumrollo:
When working through the [xarray page of the docs](https://icechunk.io/icechunk-python/xarray), I encountered a few errors/omissions that added friction to my quick attempt at using icechunk + xarray:

- [x] import line imports modules `IcechunkStore`, `StorageConfig` from `icechunk`; later `icechunk.StorageConfig` is called, which fails as `icechunk` itself was not imported.
- [x] `ds = xr.tutorial.open_dataset('rasm')` requires the packages `pooch` and `netCDF4` to be installed, this should be included in the instructions
- [x] Final line states that `Notice that this second xarray.Dataset has a time dimension of length 18 whereas the first has a time dimension of length 36`, but the first has dimenstions `(y: 205, x: 275, time: 18)`. I think this is just a copy-paste error when making the docs

I will submit 3 PRs to remedy these. Looking forward to working with this package!

paraseba:
Closed by @callumrollo PRs. Thank you.

PR #281, opened 2024-10-16
[Docs] correct module calls to match imports

callumrollo:
Resolve part 1 of #280


paraseba (approved):

paraseba:
Thank you @callumrollo !

PR #287, opened 2024-10-16
Release 0.1.0-alpha.3 to the Rust world

paraseba:


icechunk-python/Cargo.toml
```diff
@@ -21,7 +21,7 @@ crate-type = ["cdylib"]
 bytes = "1.7.2"
 chrono = { version = "0.4.38" }
 futures = "0.3.30"
-icechunk = { path = "../icechunk", version = "0.1.0-alpha.1" }
+icechunk = { path = "../icechunk", version = "0.1.0-alpha.3" }
```
paraseba, line 24: this is weird right?

mpiannucci: These shoudl match, but with our ci setup they dont have to. 

This is just a weird requirement of cargo for local packages

mpiannucci (approved):

Issue #299, opened 2024-10-18
Separate out testing in pre-commit from actual linting


yarikoptic:
ATM pre-commit seems to run actual unit and integration testing as part of the 
```
  - repo: local
    hooks:
      - id: format-lint
        name: just
        description: Run the just pre-commit step
        entry: just pre-commit
        language: system
        pass_filenames: false
```
step.  That takes awhile, and also generally should not be needed if I e.g. modifying some basic documentation which does not effect functioning/tests, and thus should generally could be "excluded" explicitly in pre-commit step for testing.

Moreover, I cannot "lint" unless unittests pass, and they might fail seems due to lack of connectivity to some services, as for me they do ATM

```
running 1 test
test test_distributed_writes ... FAILED

failures:

---- test_distributed_writes stdout ----
Error: Ref(Storage(S3ListObjectError(DispatchFailure(DispatchFailure { source: ConnectorError { kind: Io, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })), connection: Unknown } }))))


failures:
    test_distributed_writes
```

PR #300, opened 2024-10-18
README minor tune up: syntax and other very minor typos etc

yarikoptic:
I had to commit with `--no-verify` since pre-commit was failing , filed a dedicated

- https://github.com/earth-mover/icechunk/issues/299

which I don't think relates to my changes, so hopefully they are lint-free ;)


paraseba (approved):
Thank you @yarikoptic !!

PR #303, opened 2024-10-19
Fix non-conformant "attributes" metadata

LDeakin:
I believe `"attributes": null` in array/group metadata is not conformant to the Zarr V3 specification. If empty, it must be omitted entirely or be `"attributes": {}`.

https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#attributes

> The following members are optional:
>
> `attributes`
> The value must be an object. The object may contain any key/value pairs, where the key must be a string and the value can be an arbitrary JSON literal. Intended to allow storage of arbitrary user metadata.


paraseba (approved):
Thank you @LDeakin !

PR #313, opened 2024-10-22
Read-Only mode when checking out snapshot or tag

### LINKED ISSUE #237: Checking out a tag or a snapshot_id should put the repo in RO mode.

dcherian:
Closes #237


icechunk-python/tests/test_timetravel.py
```diff
@@ -30,6 +30,7 @@ def test_timetravel():
     assert air_temp[200, 6] == 54
 
     store.checkout(branch="main")
+    store.set_mode("w-")
```
dcherian, line 33: This is important. When checking out a branch, we do not change the current mode.

So if you keep switching between a tag and a branch (say) , you'll need to explicitly switch to a writeable mode before making changes.
```
store = IcechunkStore.open_existing(mode="w");
store.checkout(tag=...) # sets mode="r"
store.checkout(branch="main") # preserves mode="r"
store.set_mode("w-") # needed
# make changes
```

icechunk/src/zarr.rs
```diff
@@ -396,8 +400,12 @@ impl Store {
 
     /// Checkout a specific version of the repository. This can be a snapshot id, a tag, or a branch tip.
     ///
-    /// If the version is a branch tip, the branch will be set as the current branch. If the version
-    /// is a tag or snapshot id, the current branch will be unset and the store will be in detached state.
+    /// If the version is a branch tip, the branch will be set as the current branch.
+    /// The current mode will be unchanged.
```
paraseba, line 404: I don't think people will understand the word "mode" here.

paraseba (approved):

mpiannucci:
So if you checkout a tag or branch you can't make local changes either?

dcherian:
> So if you checkout a tag or ~branch~ snapshot you can't make local changes either?

Yes. Not without explicitly switching the mode, and then any attempt to commit will still fail (no dangling commits are allowed). Some future `flush` API would need to be added to allow creating a new "dangling snapshot"

> How does this work if you were load the store into xarray?

Yeah if you keep switching between tags and branch tip you'll need to explicitly switch the mode. something like `ds.to_zarr(repo.store.with_mode("w-"))`

PR #317, opened 2024-10-23
Error when checking out invalid snapshot ID

### LINKED ISSUE #235: store.checkout should raise an error if snapshot_id does not exist
Icechunk let's you checkout a reference, even if it doesn't exist:

```python
store.checkout(snapshot_id="FOOOBARSPAM123456789")
```

If the snapshot does not exist, it would be much nicer to raise an error earlier.

dcherian:
Closes #235

```
ValueError: store error: unsuccessful repository operation: `snapshot not found: `F000BARSPAM123456780``
```

### DIFF
--- a/icechunk/src/repository.rs
+++ b/icechunk/src/repository.rs
@@ -1115,7 +1115,7 @@ async fn all_chunks<'a>(
 }
 
 pub async fn raise_if_invalid_snapshot_id(
-    storage: &(dyn Storage + Send + Sync),
+    storage: &(dyn Storage),
     snapshot_id: &SnapshotId,
 ) -> RepositoryResult<()> {
     storage
--- a/icechunk/src/repository.rs
+++ b/icechunk/src/repository.rs
@@ -143,6 +143,8 @@ impl RepositoryBuilder {
 pub enum RepositoryError {
     #[error("error contacting storage {0}")]
     StorageError(#[from] StorageError),
+    #[error("invalid snapshot ID: `{id}`")]
+    InvalidSnapshotId { id: SnapshotId },
     #[error("error in icechunk file")]
     FormatError(#[from] IcechunkFormatError),
     #[error("node not found at `{path}`: {message}")]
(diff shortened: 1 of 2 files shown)


icechunk/src/repository.rs
```diff
@@ -269,6 +271,16 @@ impl Repository {
         &self.config
     }
 
+    pub(crate) async fn raise_if_invalid_snapshot_id(
```
paraseba, line 274: I really like this ! So much that I would have it as a free function that takes a `&dyn Storage`

paraseba: I'd put the function in the same file.

paraseba (approved):

rabernat:
Could we get more colons into the error message please? 😆

icechunk/src/repository.rs
```diff
@@ -1112,6 +1114,17 @@ async fn all_chunks<'a>(
...
 }
 
+pub async fn raise_if_invalid_snapshot_id(
+    storage: &(dyn Storage + Send + Sync),
+    snapshot_id: &SnapshotId,
+) -> RepositoryResult<()> {
+    storage
+        .fetch_snapshot(snapshot_id)
+        .await
+        .map_err(|_| RepositoryError::SnapshotNotFound { id: snapshot_id.clone() })?;
```
dcherian, line 1124: This seemed the most sensible error to me.

icechunk/src/repository.rs
```diff
@@ -1112,6 +1114,17 @@ async fn all_chunks<'a>(
     Ok(existing_array_chunks.chain(new_array_chunks))
 }
 
+pub async fn raise_if_invalid_snapshot_id(
```
paraseba, line 1117: nitpicking: I think Storage is enough here? No need for Send + Sync?

dcherian: apparently required:
```
error[E0277]: `dyn icechunk::Storage` cannot be shared between threads safely
    --> icechunk-python/src/lib.rs:364:9
     |
364  | /         pyo3_asyncio_0_21::tokio::future_into_py(py, async move {
365  | |             do_checkout_snapshot(store, snapshot_id).await
366  | |         })
     | |__________^ `dyn icechunk::Storage` cannot be shared between threads safely
     |
     = help: the trait `std::marker::Sync` is not implemented for `dyn icechunk::Storage`, which is required by `{async block@icechunk-python/src/lib.rs:364:54: 364:64}: std::marker::Send`
     = note: required for `&dyn icechunk::Storage` to implement `std::marker::Send`
note: required because it's used within this `async` fn body
```

paraseba: But, I don't get it. It errors in an unrelated function we are not calling?

PR #318, opened 2024-10-23
list_refs

### LINKED ISSUE #136: Add python support for listing all refs
We have this on the Rust side via `list_refs`

dcherian:
Closes #136


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -3,6 +3,12 @@ import datetime
 from collections.abc import AsyncGenerator
 from typing import Any
 
+class Ref:
```
dcherian, line 6: Is this really the model we want to return to users?

Alternatives could be:
1. Arraylake style Pydantic models: `BranchName(str)` and `TagName(str)`
2. or, 
```
class RefType(Enum):
    BRANCH = auto()
    TAG = auto()

class Ref:
    name: str
    kind: RefType
```

icechunk-python/src/lib.rs
```diff
@@ -40,6 +40,38 @@ struct PyIcechunkStore {
...
+    fn from(reference: Ref) -> Self {
+        match reference {
+            Ref::Tag(name) => Self::Tag { name },
+            Ref::Branch(name) => Self::Branch { name },
+        }
+    }
+}
+
+#[pymethods]
+impl PyRef {
```
paraseba, line 60: beautiful

icechunk-python/src/lib.rs
```diff
@@ -432,6 +464,25 @@ impl PyIcechunkStore {
         Ok(snapshot_id.to_string())
     }
 
+    fn list_refs<'py>(
```
paraseba, line 467: can we also provide `async_list_refs`?

paraseba: can we add a python test that exercises this new function?

dcherian: Yes I can do both. I'd like some input on https://github.com/earth-mover/icechunk/pull/318#discussion_r1813703484 first before I tackle those.

icechunk-python/python/icechunk/_icechunk_python.pyi
paraseba: Oh sorry, forgot to comment. i like your approach best, but I'd like to see how easy/hard is to pattern match the result in python. Probably tests will show me that.

Other option would be to have separate functions, `list_tags` and `list_branches`, or maybe the three of them.

mpiannucci: I think that tags and branches can be in separate functions and that simplifies a lot from the Python user perspective

dcherian:
@mpiannucci has implemented this

PR #355, opened 2024-10-28
Fix python lint action

mpiannucci:
I am not sure how this slipped through..


mpiannucci:
The python rust formatting and linting used to happen in the rust ci which doesn't run if only files in the `icechunk-python` crate are changed. This runs rust checks on the python crate as well.

paraseba (approved):

PR #358, opened 2024-10-29
Fix example NOAA dataset

jbusecke:
Just noticed that the example of NOAA data did not work.


paraseba:
Thank you @jbusecke !

PR #360, opened 2024-10-29
Add era5 virtualizarr example

mpiannucci:
Just a single file for now, starting point for building large virtual version


TomNicholas:
In general I would encourage using `vds` vs `ds` to distinguish between datasets that contain `ManifestArrays` and those that don't. It's confusing otherwise because it's fundamentally ambiguous what the top-level type is (see https://github.com/zarr-developers/VirtualiZarr/issues/171).

li-em:
We can probably replace this one with https://app.earthmover.io/marketplace/695bff20622fd82a1ec88780 🙃

PR #154, opened 2024-10-07
Add `merge` method for combining changes from multiple stores

mpiannucci:
Still experimental, but putting in the open

PR #361, opened 2024-10-30
More control for merging stores together

mpiannucci:
in favor of #154 ..


icechunk-python/examples/dask_write.py
```diff
@@ -188,13 +188,15 @@ def update(args: argparse.Namespace) -> None:
...
-    worker_stores = client.gather(map_result)
+    worker_changes = client.gather(map_result)
 
     print("Starting distributed commit")
     # we can use the current store as the commit coordinator, because it doesn't have any pending changes,
     # all changes come from the tasks, Icechunk doesn't care about where the changes come from, the only
     # important thing is to not count changes twice
-    commit_res = store.distributed_commit("distributed commit", [ws.change_set_bytes() for ws in worker_stores])
+    for changes in worker_changes:
+        store.merge(changes)
```
dcherian, line 198: Thoughts on `store.merge(stores)` instead? Is there a benefit to merging the changeset bytes?

mpiannucci: I am ok with it, I was just thinking of use cases where the full stores aren't serialized but it's kind of a non issue

dcherian: I agree with it being a non-issue

icechunk/src/repository.rs
```diff
@@ -922,20 +907,18 @@ async fn get_existing_node<'a>(
     }
 }
 
-async fn distributed_flush<I: IntoIterator<Item = ChangeSet>>(
+async fn flush(
     storage: &(dyn Storage + Send + Sync),
-    change_sets: I,
+    change_set: &mut ChangeSet,
```
dcherian, line 912: I don't understand why this needs a mutable ref.

dcherian (approved):
Just that one comment about the mutable ref for my understanding.

Otherwise this seems OK to me!

icechunk/src/repository.rs
```diff
@@ -635,26 +635,15 @@ impl Repository {
...
+    /// Discard all uncommitted changes and return them as a `ChangeSet`
+    pub fn discard_changes(&mut self) -> ChangeSet {
+        std::mem::take(&mut self.change_set)
+    }
 
-        self.snapshot_id = new_snapshot_id.clone();
-        self.change_set = ChangeSet::default();
-        Ok(new_snapshot_id)
+    /// Merge a set of `ChangeSet`s into the repository without committing them
+    pub async fn merge(&mut self, changes: ChangeSet) -> RepositoryResult<()> {
```
paraseba, line 644: I don't think this can fail? let's turn it into `-> ()`

paraseba (approved):

icechunk/src/repository.rs
mpiannucci: It was a relic from the distributed flush! Good catch!

Issue #266, opened 2024-10-15
Support Microsoft Azure for the object store


paraseba:

kylebarron:
For both this and https://github.com/earth-mover/icechunk/issues/265, it seems like an `ObjectStore`-based API is ideal, correct?

In https://github.com/developmentseed/object-store-rs I've been developing pyo3 integration for the `object_store` crate, that we could reuse in the icechunk Python APIs. (That repo has both a user-facing Python API _and_ a [Rust-facing pyo3 integration](https://github.com/developmentseed/object-store-rs/tree/main/pyo3-object_store) (not yet well documented)).

I also heard vaguely from @jhamman that you originally used the `object_store` crate for all backends, but then hit some performance issues that made you switch to using `s3` directly. Can you elaborate on that? The `object_store` maintainers are very performance sensitive so I'd 1) be surprised that you could get better performance from using the s3 APIs natively and 2) if so the `object_store` maintainers would probably be interested in your use case. (If preferred, we can move this discussion to a separate thread).

geospatial-jeff:
> We had many random very weird errors with object_store when used with high concurrency. But this was the last week before launch, so we didn't have time to explore. The errors looked like malformed HTTP responses, very weird.

@paraseba I believe this could be due to a difference in how credentialization was implemented in the `object_store` version of the code compared to AWS SDK.  For deployments, object store will fall back to the node profile (credentials provider) when it finds no other form of authentication on the machine.  This is a remote network call (PUT request) to an AWS token endpoint that is sent before each HTTP request to the blob store.  At high concurrencies you may run into rate limits, account-level quotas, and other failures that cause malformed tokens.  I've seen this myself running both `boto3` and `gdal` at big scales.

The original `object_store` implementation would fall back to node profile if it hit this `else` block, and no credentials were configured in the environment - https://github.com/earth-mover/icechunk/blob/e3a119cddfcbe7b38173faeb0b4eb403bd69de7b/icechunk/src/storage/object_store.rs#L92-L94

The AWS SDK **does not** implement this fall back behavior unless you explicitly configure it in the code, and this was not configured when switching the code over to the AWS SDK.  Which would explain why this issue went away with the switch.

This is of course a lot of guesswork based on past experiences, it really depends how the code / dask cluster was deployed. 
 Local cluster on an EC2 instance?  Distributed cluster on k8s?  Running from your local machine?  All have different failure cases when it comes to credentialization.

kylebarron:
[...]

PR #362, opened 2024-10-30
Update to pyo3 0.22

kylebarron:
For https://github.com/earth-mover/icechunk/issues/266.

### Change list 

- Updates to `pyo3` 0.22 and `pyo3-async-runtimes` 0.22.
- pyo3 0.22 deprecated the default behavior of `Option<T>` defaulting to `None`, so now you have to explicitly specify this in `pyo3(signature)`
- Upgrading to `pyo3` 0.22 broke the `PyIcechunkStoreError` enum because the pyerr's don't implement `Debug`. An alternative way to fix this is to just remove the `derive(Debug)`, but I'd suggest storing _Rust error types_ in your Rust error enum, and then converting to `PyErr` as needed in the `From` impl.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1021,12 +1021,6 @@ dependencies = [
 
 [[package]]
 name = "heck"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
-
-[[package]]
-name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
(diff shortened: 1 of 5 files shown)


mpiannucci:
> but I'd suggest storing Rust error types in your Rust error enum, and then converting to PyErr as needed in the From impl.

Can you expand on how this is different from what we are doing now?

kylebarron:
> Can you expand on how this is different from what we are doing now?

In the case of `KeyNotFound` and `PyValueError`:
https://github.com/earth-mover/icechunk/blob/5aa9abeae3bb22deffa99ab30aa085233678bc0c/icechunk-python/src/errors.rs#L18-L19
https://github.com/earth-mover/icechunk/blob/5aa9abeae3bb22deffa99ab30aa085233678bc0c/icechunk-python/src/errors.rs#L26-L27
the [`#[from]` thiserror macro](https://docs.rs/thiserror/latest/thiserror/#details) I believes expands to hold those error types within the variant. 

This means that you're storing a _Python object_ within your Rust enum. (Each of those two rust structs wrap Python objects, a `PyAny`).

I think it can make sense to have one catch-all `PyError` variant in your enum, so that if you're in a place where you think "I really want a Python `IOError`", you can create a `PyIOError` and store that in the `PyIcechunkStoreError::PyError` variant. 

But in the case of `KeyNotFound` and `PyValueError` you're currently getting a sort of nested error description, where the error content is getting expanded into a string, but then _all_ of your errors are exposed to Python as a `ValueError`. That seems to not be what you want because you create your own custom exception type but it doesn't look like Python code will ever see it.

mpiannucci:
This makes sense. 

I am going to merge this in and do a follow up PR with the python error handling updates. 

Thank you for this!

mpiannucci (approved):

PR #368, opened 2024-11-03
Rust Icechunk learns how to collect garbage

paraseba:
We call garbage collection to the process where dangling objects are
deleted from object store.

We define a dangling object as a chunk, manifest, snapshot, attributes
or transaction log file that cannot be reached by navigating the
parent relationship starting from all possible refs.

There are currently two mechanisms that create dangling objects:

- Abandoning a session without committing it
- Resetting a branch leaving snapshots behind

In the future, we'll introduce more mechanisms that "generate" garbage,
with the objective of reducing storage costs. One example, would be
squashing commits when version resolution is not relevant.

Garbage collection is an inherently dangerous process. It's the only
time at which Icechunk actually deletes data from object store. As such,
it must be executed carefully.

There is an unavoidable race condition in garbage collection: Icechunk
has no way to distinguish a new object from a dangling one, if that
object was created after the garbage collection process has traced the refs.

To solve that issue, the garbage collection process only deletes objects
that have been created some time ago. Users can pass a timestamp as
configuration to the collection process. This timestamp must be  older
than the start time of the oldest possible writing session open. For
example, if the longest writing sessions last 48 hours, a safe timestamp
would be `now - 7 days`.

### DIFF
--- a/icechunk/src/ops/gc.rs
+++ b/icechunk/src/ops/gc.rs
@@ -1,0 +1,296 @@
+use std::{collections::HashSet, future::ready, iter};
+
+use chrono::{DateTime, Utc};
+use futures::{stream, Stream, StreamExt, TryStreamExt};
+use tokio::pin;
+
+use crate::{
+    format::{ChunkId, ManifestId, SnapshotId},
+    refs::{list_refs, RefError},
+    repository::ChunkPayload,
+    storage::ListInfo,
+    Storage, StorageError,
+};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Action {
+    Keep,
+    DeleteIfCreatedBefore(DateTime<Utc>),
+}
+
+#[derive(Debug)]
+pub struct GCConfig {
+    extra_roots: HashSet<SnapshotId>,
+    dangling_chunks: Action,
+    dangling_manifests: Action,
+    dangling_attributes: Action,
+    dangling_transaction_logs: Action,
+    dangling_snapshots: Action,
+}
+
+impl GCConfig {
+    pub fn new(
+        extra_roots: HashSet<SnapshotId>,
+        dangling_chunks: Action,
+        dangling_manifests: Action,
+        dangling_attributes: Action,
+        dangling_transaction_logs: Action,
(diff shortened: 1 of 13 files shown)


icechunk/src/ops/gc.rs
```diff
@@ -0,0 +1,296 @@
...
+            dangling_chunks,
+            dangling_manifests,
+            dangling_attributes,
+            dangling_transaction_logs,
+            dangling_snapshots,
+        }
+    }
+    pub fn clean_all(
+        chunks_age: DateTime<Utc>,
+        metadata_age: DateTime<Utc>,
```
rabernat, line 51: Curious about the rationale for having two different ages here.

paraseba: I thought, giving that metadata files are far less numerous than chunks, users would be willing to preserve them longer. If we someday support some form of reflog (or even without) it could be useful to have them around for forensics or understanding. It wouldn't allow to recover data, but you could at least recover "structure". For example, a reasonable thing to do could be to GC chunks older than 1 month and metadata older than 1 year.

icechunk/src/storage/mod.rs
```diff
@@ -106,4 +114,24 @@ pub trait Storage: fmt::Debug + private::Sealed {
         overwrite_refs: bool,
         bytes: Bytes,
     ) -> StorageResult<()>;
+
+    async fn list_chunks(
+        &self,
+    ) -> StorageResult<BoxStream<StorageResult<ListInfo<ChunkId>>>>;
```
mpiannucci, line 120: Can these impls be defaulted? The impls look the same mostly (except for the nested storage systems)

paraseba: great point, will try

paraseba: One tricky thing is that traits cannot be object safe if they have generic methods .... :facepalm:

mpiannucci: Ahhhhh right... not worth it then probably

mpiannucci (approved):

Issue #356, opened 2024-10-28
Add WASM bindings


mpiannucci:
Long term, there is a strong desire and need to run icechunk and access icechunk data fdirectly from nodejs, deno, or browser javascript runtimes. 

While today we can write bindings for node using [napi-rs](https://napi.rs/), but this wont scale to other javascript runtimes.

WASM is much harder because icechunk internally stands on the rust IO stack, so there is much work to be done feature splitting the core and making sure that internal IO pieces can be abstracted away

TomNicholas:
Is WASM the easiest way to get a javascript icechunk client? Could we also go via a [C interface](https://github.com/earth-mover/icechunk/issues/381)?

rabernat:
It appears that @jbms already has a native icechunk implementation in typescript. See https://github.com/google/neuroglancer/pull/718

mpiannucci:
The easiest way is to use napi-rs to create the bindings, it just can't run in the browser and only on node

mpiannucci:
So I looked through the crate and the current blockers for wasm support are:

* the Aws s3 client won't work (ok, we can use a js impl in wasm)
* we can't use any tokio runtime calls. Specifically `tokio::spawn` can't be used from wasm as it's always single threaded and it is used in icechunk a LOT. 

None of these issues exist for node support. But wasm allows for browser + deno + node support while napi only provides node support. Also of importance is that napi will be an order of magnitude or more faster than wasm. Any input on use cases would be helpful

Cc @kylebarron @sharkinsspatial

kylebarron:
In general, I don't think tokio can be used in wasm32-unknown-unknown. 

I don't know the target use cases you have, but I can't imagine that node support would be that useful. You don't have that many server-side use cases for array data in Node. Those users are all already using Python. 

> napi will be an order of magnitude or more faster than wasm

I thought the difference between wasm and native code was more like 2-3x in performance, not 10x.

mpiannucci:
> In general, I don't think tokio can be used in wasm32-unknown-unknown.

Some primitives can like `sync` but the runtime itself can't 

> I don't know the target use cases you have, but I can't imagine that node support would be that useful. You don't have that many server-side use cases for array data in Node.

I've had a few people reach out that are proxying their array data from a node server to their frontend with next or similar. I agree WASM should be the goal

mpiannucci:
Oh my god is this ever timely: 

https://napi.rs/blog/announce-v3

[...]

PR #371, opened 2024-11-04
Simplify Python binding of `StorageConfig`

kylebarron:
I'm starting to read through the storage binding for #266.

As it stands, you recreate the content of the struct in the Python binding as well, and then you convert the _content_ of the struct in the Python bindings to the _content_ of the struct in the Rust code.

In my opinion it's best to have as little logic as possible in the binding code and make it a slim wrapper around the underlying core Rust code. Minimizing the logic in the bindings also makes it simpler to make bindings to other languages in the future (https://github.com/earth-mover/icechunk/issues/356)

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -105,27 +105,7 @@ class StorageConfig:
     storage_config = StorageConfig.s3_from_config("bucket", "prefix", ...)
     ```
     """
-    class Memory:
-        """Config for an in-memory storage backend"""
 
-        prefix: str
-
-    class Filesystem:
-        """Config for a local filesystem storage backend"""
-
-        root: str
-
-    class S3:
-        """Config for an S3 Object Storage compatible storage backend"""
-
-        bucket: str
-        prefix: str
-        credentials: S3Credentials | None
-        endpoint_url: str | None
-        allow_http: bool | None
-        region: str | None
-
-    def __init__(self, storage: Memory | Filesystem | S3): ...
     @classmethod
     def memory(cls, prefix: str) -> StorageConfig:
         """Create a StorageConfig object for an in-memory storage backend with the given prefix"""
@@ -137,7 +117,14 @@ class StorageConfig:
         ...
 
     @classmethod
-    def s3_from_env(cls, bucket: str, prefix: str) -> StorageConfig:
+    def s3_from_env(
+        cls,
+        bucket: str,
+        prefix: str,
+        endpoint_url: str | None,
+        allow_http: bool = False,
+        region: str | None = None,
+    ) -> StorageConfig:
         """Create a StorageConfig object for an S3 Object Storage compatible storage backend
         with the given bucket and prefix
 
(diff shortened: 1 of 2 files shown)


icechunk-python/src/storage.rs
```diff
@@ -33,6 +30,22 @@ impl From<&PyS3Credentials> for StaticS3Credentials {
...
 
+impl From<PyS3Credentials> for StaticS3Credentials {
+    fn from(credentials: PyS3Credentials) -> Self {
+        StaticS3Credentials {
+            access_key_id: credentials.access_key_id,
+            secret_access_key: credentials.secret_access_key,
+            session_token: credentials.session_token,
+        }
+    }
+}
```
kylebarron, line 41: It's useful to have a value-to-value `From` impl instead of just an impl on `&PyS3Credentials` so that you can avoid the clones when not needed.

icechunk-python/src/storage.rs
```diff
@@ -51,62 +64,44 @@ impl PyS3Credentials {
...
+        Self(StorageConfig::LocalFileSystem { root })
     }
 
     #[classmethod]
     #[pyo3(signature = (
         bucket,
         prefix,
         endpoint_url = None,
-        allow_http = None,
+        allow_http = false,
```
kylebarron, line 87: `allow_http` can only be a bool, so why not default to `false` instead of defaulting to `None` and then unwrapping?

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -105,27 +105,7 @@ class StorageConfig:
...
-        """Config for an S3 Object Storage compatible storage backend"""
-
-        bucket: str
-        prefix: str
-        credentials: S3Credentials | None
-        endpoint_url: str | None
-        allow_http: bool | None
-        region: str | None
-
-    def __init__(self, storage: Memory | Filesystem | S3): ...
```
kylebarron, line 128: There's no `#[new]` method defined on `StorageConfig`, so it wasn't possible to directly construct this anyways.

icechunk-python/src/storage.rs
mpiannucci: no reason, just moving fast

icechunk-python/src/storage.rs
```diff
@@ -177,38 +168,21 @@ fn mk_credentials(config: Option<&PyS3Credentials>, anon: bool) -> S3Credentials
...
-                StorageConfig::S3ObjectStore {
-                    bucket: bucket.clone(),
-                    prefix: prefix.clone(),
-                    config: Some(s3_config),
-                }
-            }
-        }
+impl AsRef<StorageConfig> for PyStorageConfig {
+    fn as_ref(&self) -> &StorageConfig {
+        &self.0
```
kylebarron, line 185: This `AsRef` impl means that if you have a `PyStorageConfig` and you want to access an `&StorageConfig`, you can just call `.as_ref()`.

mpiannucci (approved):
Looks great! Thanks for cleaning this up

PR #388, opened 2024-11-14
Sync with zarr python 3 beta 2

### LINKED ISSUE #387: Sync with zarr python 3 beta 2

mpiannucci:
Closes #387 

This relies on a lot of the default functions from zarr's `Store` so it can most likely be further optimized in the future.


icechunk-python/python/icechunk/__init__.py
```diff
@@ -626,35 +603,40 @@ def supports_listing(self) -> bool:
     def supports_deletes(self) -> bool:
         return self._store.supports_deletes
 
-    def list(self) -> AsyncGenerator[str, None]:
```
mpiannucci, line 629: The change from async generator to async iterator doesnt change anything on the rust side, its the same prototype

icechunk-python/python/icechunk/__init__.py
```diff
@@ -444,10 +420,6 @@ def async_ancestry(self) -> AsyncGenerator[SnapshotMetadata, None]:
         """
         return self._store.async_ancestry()
 
-    async def empty(self) -> bool:
-        """Check if the store is empty."""
-        return await self._store.empty()
```
paraseba, line 449: what happened here? did this disappear from the abc?

icechunk-python/python/icechunk/__init__.py
```diff
@@ -594,7 +566,12 @@ async def delete(self, key: str) -> None:
         ----------
         key : strz
         """
-        return await self._store.delete(key)
```
paraseba, line 597: I think it would be better to implement this behavior in zarr.rs

icechunk/src/zarr.rs
```diff
@@ -1032,6 +1032,10 @@ async fn exists(key: &str, repo: &Repository) -> StoreResult<bool> {
     match get_key(key, &ByteRange::ALL, repo).await {
         Ok(_) => Ok(true),
         Err(StoreError::NotFound(_)) => Ok(false),
+        Err(StoreError::RepositoryError(RepositoryError::NodeNotFound {
+            path: _,
+            message: _,
+        })) => Ok(false),
```
paraseba, line 1038: great catch

icechunk-python/python/icechunk/__init__.py
mpiannucci: Do you mean the `delete_dir` behavior?

mpiannucci: #390

paraseba: Sorry, no, I mean the fact that delete doesn't fail if the key doesn't exist. I think it's a behavior the rust store should implement

icechunk-python/python/icechunk/__init__.py
mpiannucci: Its not the same, takes prefix now. Adding to backlog

mpiannucci: I added a simple implmenetation that just uses the rust `list_dir` function. I think that is right, but there is probably a more optimized way to do it?

paraseba: This is great, it will get fast once we make list_dir fast. Thank you

icechunk/src/zarr.rs
```diff
@@ -735,7 +740,14 @@ impl Store {
             Key::Chunk { node_path, coords } => {
                 let mut guard = self.repository.write().await;
                 let repository = guard.deref_mut();
-                Ok(repository.set_chunk_ref(node_path, coords, None).await?)
```
paraseba, line 738: beautiful thank you

paraseba (approved):

PR #395, opened 2024-11-15
Error when chunk coordinates are invalid

### LINKED ISSUE #175: Don't allow to write chunks to invalid chunk coordinates

emacollins:
Closes #175

### DIFF
--- a/icechunk/src/format/snapshot.rs
+++ b/icechunk/src/format/snapshot.rs
@@ -524,18 +524,8 @@ mod tests {
                     serde_json::Value::from(42),
                 )))),
             }],
-            storage_transformers: Some(vec![StorageTransformer {
-                name: "mytransformer".to_string(),
-                configuration: Some(HashMap::from_iter(iter::once((
-                    "foo".to_string(),
-                    serde_json::Value::from(42),
-                )))),
-            }]),
-            dimension_names: Some(vec![
-                Some("x".to_string()),
-                Some("y".to_string()),
-                Some("t".to_string()),
-            ]),
+            storage_transformers: None,
+            dimension_names: None,
         };
 
         
@@ -554,10 +544,10 @@ mod tests {
         let coord2 = ChunkIndices(vec![10, 11, 10]);
         let coord3 = ChunkIndices(vec![0, 0 ,0]);
 
-        assert_eq!(true, zarr_meta1.valid_chunk_coord(&coord1));
-        assert_eq!(false, zarr_meta1.valid_chunk_coord(&coord2));
+        assert!(zarr_meta1.valid_chunk_coord(&coord1));
+        assert!(!zarr_meta1.valid_chunk_coord(&coord2));
         
-        assert_eq!(true, zarr_meta2.valid_chunk_coord(&coord3));
+        assert!(zarr_meta2.valid_chunk_coord(&coord3));
 
     }
 }
--- a/icechunk/src/format/snapshot.rs
+++ b/icechunk/src/format/snapshot.rs
@@ -54,16 +54,23 @@ impl ZarrArrayMetadata {
     /// Returns an iterator over the maximum permitted chunk indices for the array.
     ///
     /// This function calculates the maximum chunk indices based on the shape of the array
-    /// and the chunk shape.
+    /// and the chunk shape, using (shape - 1) / chunk_shape. Given integer division is truncating,
+    /// this will always result in proper indices at the boundaries.
     ///
     /// # Returns
     ///
-    /// An iterator over the maximum permitted chunk indices.
-    fn max_chunk_indices_permitted(&self) -> impl Iterator<Item = u64> + '_ {
-        self.shape
+    /// A ChunkIndices type containing the max chunk index for each dimension.
+    fn max_chunk_indices_permitted(&self) -> ChunkIndices {
+
+        debug_assert_eq!(self.shape.len(), self.chunk_shape.0.len());
+
+        ChunkIndices(
+            self.shape
             .iter()
             .zip(self.chunk_shape.0.iter())
-            .map(|(s, cs)| ((s + cs.get() - 1)) / cs.get() - 1)
+            .map(|(s, cs)| if *s == 0 { 0 } else { ((s - 1) / cs.get()) as u32 })
+            .collect()
+        )
     }
 
     /// Validates the provided chunk coordinates for the array.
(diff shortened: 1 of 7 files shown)


icechunk/src/format/snapshot.rs
```diff
@@ -49,6 +49,55 @@ pub struct ZarrArrayMetadata {
...
+
+    /// Returns an iterator over the maximum permitted chunk indices for the array.
+    ///
+    /// This function calculates the maximum chunk indices based on the shape of the array
+    /// and the chunk shape.
+    ///
+    /// # Returns
+    ///
+    /// An iterator over the maximum permitted chunk indices.
+    fn max_chunk_indices_permitted(&self) -> impl Iterator<Item = u64> + '_ {
```
paraseba, line 62: I think chunk indices are u32:
```rust
pub struct ChunkIndices(pub Vec<u32>);
```

At the beginning of this method, it could be a good place to:
```rust
debug_assert_eq!(self.shape.len(), self.chunk_shape.0.len())
```
or similar.

icechunk/src/format/snapshot.rs
```diff
@@ -49,6 +49,55 @@ pub struct ZarrArrayMetadata {
...
+        self.shape
+            .iter()
+            .zip(self.chunk_shape.0.iter())
+            .map(|(s, cs)| ((s + cs.get() - 1)) / cs.get() - 1)
```
paraseba, line 66: The math looks a bit weird to me. Shouldn't it be just `s / cs.get() - 1`?

icechunk/src/format/snapshot.rs
```diff
@@ -49,6 +49,55 @@ pub struct ZarrArrayMetadata {
...
+    /// * `coord` - The chunk indices to validate.
+    ///
+    /// # Returns
+    ///
+    /// An `IcechunkResult` indicating whether the chunk coordinates are valid.
+    ///
+    /// # Errors
+    ///
+    /// Returns `IcechunkFormatError::ChunkCoordinatesNotFound` if the chunk coordinates are invalid.
+    pub fn valid_chunk_coord(&self, coord: &ChunkIndices) -> IcechunkResult<bool> {
```
paraseba, line 84: Actually, this function doesn't really fail. Things would be easier if you make it return a simple `bool`

icechunk/src/format/snapshot.rs
emacollins: Pushed a new commit to address this. I ended up returning a `ChunkIndices` type from this using `collect()` on the iterator and type casting the calculation result to u32. Not sure if there is a performance consideration with using collect prior to returning, but it seemed to make sense that the type returned be `ChunkIndices`. This way the actual validation function compares two ChunkIndices.

Shape and chunk shape were both u64, and I assume chunk indices are u32 since they will always be much smaller values. 

Also added the debug assert, which is causing tests in `repository.rs` to fail. I noticed this is because the chunk shape initialized is a single value, even when the shape is multiple. Do we assume that if a single chunk shape value is present, it applies to all dimensions? Or should this fail?

icechunk/src/format/snapshot.rs
emacollins: Yes I don't know why I left this expanded like this. Simplified it down to `(s - 1) / cs.get()`, which should be equivalent and also seems to handle all cases properly. 

I added tests to confirm this, but if there are additional edge cases I haven't thought of, let me know. Will continue to think on it.

icechunk/src/format/snapshot.rs
```diff
@@ -450,4 +502,62 @@ mod tests {
...
+            fill_value: FillValue::Float32(0f32),
+
+            codecs: vec![Codec {
+                name: "mycodec".to_string(),
+                configuration: Some(HashMap::from_iter(iter::once((
+                    "foo".to_string(),
+                    serde_json::Value::from(42),
+                )))),
+            }],
+            storage_transformers: Some(vec![StorageTransformer {
```
paraseba, line 527: I think you can pass `None` here to make it shorter

icechunk/src/format/snapshot.rs
```diff
@@ -450,4 +502,62 @@ mod tests {
...
+                name: "mytransformer".to_string(),
+                configuration: Some(HashMap::from_iter(iter::once((
+                    "foo".to_string(),
+                    serde_json::Value::from(42),
+                )))),
+            }]),
+            dimension_names: Some(vec![
```
paraseba, line 534: same here

icechunk/src/format/snapshot.rs
```diff
@@ -450,4 +502,62 @@ mod tests {
...
+            ]),
+            ..zarr_meta1.clone()
+
+        };
+
+        let coord1 = ChunkIndices(vec![9, 10, 9]);
+        let coord2 = ChunkIndices(vec![10, 11, 10]);
+        let coord3 = ChunkIndices(vec![0, 0 ,0]);
+
+        assert_eq!(true, zarr_meta1.valid_chunk_coord(&coord1));
```
paraseba, line 557: replace these by `assert!(...)` or `assert!(! ...)`

icechunk/src/repository.rs
```diff
@@ -424,9 +424,24 @@ impl Repository {
...
+                Err(RepositoryError::FormatError(IcechunkFormatError::ChunkCoordinatesNotFound { coords: coord }))
+            }
+                     
+        } else {
+            Err(RepositoryError::NotAnArray {
+                node: node_snapshot,
+                message: "getting an array".to_string(),
+            })
+        }
+
```
paraseba, line 444: maybe we can add a test for `Repository` too? that validates that the error is correctly bubbling up.

paraseba (commented):
This is looking really great @emacollins

icechunk/src/format/snapshot.rs
paraseba: > I ended up returning a ChunkIndices

perfect!

> I assume chunk indices are u32

yes, you cand valitade looking [here](https://github.com/earth-mover/icechunk/blob/ee982e714d25e1a5681f216141fe95a88f5b2f40/icechunk/src/format/mod.rs#L141)

> chunk shape initialized is a single value, even when the shape is multiple.

this is probably a bad test, we should fix it to match. We need to add more validations to the metadata to avoid this kind of thing, but it's not critical because zarr also validates.

> Do we assume that if a single chunk shape value is present, it applies to all dimensions

No, I don't think so, probably just a bad test, please change it to match.

emacollins: Thanks for answers! I will return to this PR next week after the holiday.

paraseba:
@emacollins I wonder if the CI failure is a real one ... maybe we are actually writing to an array out of bounds and your change is now flagging it? Do you think you can take a look at what that failing test is doing?

paraseba:
@emacollins I just looked at the code of that test that is failing. I think it's a bad test:
```rust
    let zarr_meta = ZarrArrayMetadata {
        shape: vec![SIZE as u64, SIZE as u64],
        data_type: DataType::Float64,
        chunk_shape: ChunkShape(vec![
            NonZeroU64::new(2).unwrap(),
            NonZeroU64::new(2).unwrap(),
        ]),
        chunk_key_encoding: ChunkKeyEncoding::Slash,
        fill_value: FillValue::Float64(f64::NAN),
        codecs: vec![],
        storage_transformers: None,
        dimension_names: None,
    };

```

Chunks are size 2, and we are writing to all coordinates. I think things will pass if you make the chunk size 1x1

paraseba:
@emacollins please let me know if you need help with that failing test

emacollins:
> @emacollins please let me know if you need help with that failing test

Just pushed the change to run the test again, but caught in some rebase conflicts headache...Will need to pick this up tomorrow. Apologies for the delay on getting this over the line.

paraseba:
No problem @emacollins ! Thank you!

paraseba (approved):
Fantastic work @emacollins

paraseba:
We just need a `cargo fmt --all` now and we can merge

emacollins:
@paraseba saw some additional tests failed on the last run, pushed a quick fix that should pass them

paraseba:
Awesome work @emacollins, thank you!

Feel free to join our [icechunk slack channel](https://join.slack.com/t/earthmover-community/shared_invite/zt-2cwje92ir-xU3CfdG8BI~4CJOJy~sceQ) if you are interested, and let me know if you want to work on some other issue.

Issue #381, opened 2024-11-07
Add C API


asinghvi17:
It would be great to have a C-accessible API for Icechunk, as a "lowest common denominator" for access from Julia, C, GDAL, etc. 

I am personally a contributor to Julia's Zarr implementation in https://github.com/JuliaIO/Zarr.jl, and would be happy to write a Julia integration for Icechunk given a C API from this end.  Interop from Julia to Rust directly is tricky (it's a similar problem as C++ has) so a C-API is the easiest way to go here.

I imagine it could be structured pretty simply, the store would likely have to be represented as a pointer.  C doesn't look to have many async capabilities but that can be handled on the calling language's end (Julia has async calls, for example, that live in a specific pool of threads).

Would be great to hear what might be needed to support this!

rabernat:
Yes, we think this would be a great addition and support it 💯 percent!

It would be great to find a contributor who has experience creating C bindings from Rust libraries to work on this. @asinghvi17 is that something you have any experience with?

meggart:
I have some limited experience in creating C bindings for a Rust library I have written 5 years ago for an internal project and for calling the lib from R, python, Julia and Matlab. However as this is quite some time ago I would need to check if what I have done back then is still idiomatic. 

I think a good step to get us started would be if you could share a small hello-world-icechunk example as a standalone Rust application, maybe one that just opens an existing public dataset and requests a chunk of data. I or someone else if there are volunteers could then try to create an MWE that starts wrapping the subset of functions called in this example, creates c headers automatically and calls the functions from Julia. Then we at least get a first idea of the problems we might face and the amount of work it would be to get to complete C bindings.

paraseba:
@meggart we have a couple of example applications in the `examples` [directory](https://github.com/earth-mover/icechunk/tree/main/icechunk-python/examples). [smoke_test.py](https://github.com/earth-mover/icechunk/blob/main/icechunk-python/examples/smoke-test.py) is a pretty trivial one in there. That would give you an understanding of the basic python bindings.

If you want something Rust only, we also have an `examples` dir. In there I'd recommend [this one](https://github.com/earth-mover/icechunk/blob/main/icechunk/examples/multithreaded_store.rs).

mpiannucci:
[...]

PR #399, opened 2024-11-19
Discussion: Creating a C Interface

meggart:
This is my first attempt of a proof-of-concept Rust -> C -> Julia roundtrip that was discussed in #381 where I just wrapped the first function calls of the basic example that create a repository and adds some groups. Whoever is interested, maybe you can have a quick look and let me know your opinions. A C header file is generated, but currently the c wrapper in icechunk-julia is manually generated. Of course, if one wants to roll out a full c lib for the storage interface there would be a few design decisions to make. The most burning questions I would have are the following:

1. To what extent would one expose rust structs through the C API as well. For example things like `RepositoryConfig` could be marked with `[repr(C)]` to get C memory layout which would make sharing them much more convenient. In the end only things wrapped in `Arc`s actually need to be hidden behind opaque pointers. 
2. Some system of automated error code translation would be nice, in the example here I only return a generic -1 in case of any error
3. Also important to note that any panic on the rust side leads to a segmentation fault on the caller side, to it is crucial to carefully unwrap every result
4. In this example I use calls to `block_on` to get out of the async machinery. However I have no idea how this would interact if a calling site actually tried to perform async calls into the library (probably it would just break?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,6 +52,55 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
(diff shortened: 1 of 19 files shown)


rabernat:
@meggart - just wanted to drop a note to say _THANK YOU_ for working on this. We are planning to give you some feedback very soon.

rabernat:
Hi @meggart! The Icechunk format and API have gone through a lot of churn in the past 2 months, but are now stabilizing. 

I appreciate the effort you have put in here, and I'm sorry that this has been superseded by these internal Icechunk changes. We would still love to support your work on this PR. Would you like to pick it back up again?

Issue #185, opened 2024-10-11
dask.array writing with distributed / multiprocessing schedulers


rabernat:
Icechunk does support distributed writing of Arrays. However,  currently Icechunk does not allow writing arrays via `dask.array.store` with distributed or multiprocessing schedulers.

This is because Icechunk must gather the results of each distributed write operation (metadata about the chunks that were written) back to the client in order to commit the transaction. Without this step, each worker's writes would be lost.

How it works today with regular Zarr

```mermaid
sequenceDiagram
    Client->>Worker0: Compute Chunk 0
    Client->>Worker1: Compute Chunk 1
    Client->>Worker2: Compute Chunk 2
    Worker0->>Object Store: Store Chunk 0
    Worker1->>Object Store: Store Chunk 1
    Worker2->>Object Store: Store Chunk 2
```

Once each worker has written their piece of data to storage, it's done.

Instead with Icechunk, it needs to look like this

```mermaid
sequenceDiagram
    Client->>Worker0: Compute Chunk 0
    Client->>Worker1: Compute Chunk 1
    Client->>Worker2: Compute Chunk 2
    Worker0->>Object Store: Store Chunk QYNXD
    Worker0->>Client:  Chunk 0 stored at QYNXD
    Worker1->>Object Store: Store Chunk PQM3C
    Worker1->>Client:  Chunk 1 stored at PQM3C
    Worker2->>Object Store: Store Chunk BUZXP
    Worker2->>Client:  Chunk 2 stored at BUZXP
    Note right of Client: Commit Snapshot
    Client->>Object Store: Write manifest
```

The dask.array.store code does not allow for returning metadata back to from each write process:

https://github.com/dask/dask/blob/20eeeda610260287a0dd50e4dd7b6a3cd8e007f3/dask/array/core.py#L1067

To overcome this, we need to either:
- Update dask.array to accommodate our scenario
- Write a custom dask function which generates the appropriate graph for an icechunk distributed write

This _does_ work with the dask threaded scheduler because the same icechunk store can be shared in memory between threads.

There are some parallels here to what is required to allow Dask to write iceberg tables (https://github.com/apache/iceberg/issues/5800).

dcherian:
Tracking list:

- [ ] https://github.com/pydata/xarray/pull/9695
- [ ] https://github.com/dask/dask/pull/11465
- [ ] #357

dcherian:
[...]

PR #405, opened 2024-11-21
Set store to read only after unpickling

### LINKED ISSUE #383: Zero values after saving to a local store with dask localcluster.
I am trying to combine multiple CMAQ datasets. I am following quickstart guide. When I save the dataset with dask local cluster and client the values are all zeros. Everything works as expected when not using dask.

```
cluster = LocalCluster(n_workers=32, threads_per_worker=2, memory_limit='16GB')
client = Client(cluster)

storage_config = StorageConfig.filesystem("./ice_cmaq")
store = IcechunkStore.create(storage_config)


def preprocess(ds):
    ds = ds.isel(TSTEP=slice(0, 24))  # Select the first 24 timesteps
    ds = ds.drop_vars('TFLAG')  # Drop the TFLAG data variable
    ds = ds.sel(LAY=0)  # Select LAY=0
    return ds

cmaq_ds = xr.open_mfdataset(cmaq_paths, preprocess=preprocess, concat_dim='TSTEP', combine='nested')
cmaq_ds[['O3', 'NO2']].to_zarr(store, zarr_format=3, consolidated=False)
store.commit("added cmaq")
```

dcherian:
Closes #383
xref #185

### DIFF
--- a/icechunk-python/python/icechunk/__init__.py
+++ b/icechunk-python/python/icechunk/__init__.py
@@ -1,4 +1,5 @@
 # module
+import contextlib
 from collections.abc import AsyncGenerator, AsyncIterator, Iterable
 from typing import Any, Self
 
@@ -33,6 +34,7 @@ __all__ = [
 
 class IcechunkStore(Store, SyncMixin):
     _store: PyIcechunkStore
+    _pickle_preserves_read_only: bool
 
     @classmethod
     async def open(cls, *args: Any, **kwargs: Any) -> Self:
@@ -89,6 +91,7 @@ class IcechunkStore(Store, SyncMixin):
                 "An IcechunkStore should not be created with the default constructor, instead use either the create or open_existing class methods."
             )
         self._store = store
+        self._pickle_preserves_read_only = False
 
     @classmethod
     def open_existing(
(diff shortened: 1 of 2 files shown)


icechunk-python/python/icechunk/__init__.py
```diff
@@ -33,6 +34,7 @@
 
 class IcechunkStore(Store, SyncMixin):
     _store: PyIcechunkStore
+    _pickle_preserves_read_only: bool
```
dcherian, line 37: I put this on this class, not PyIcechunkStore, since this is the one that handles pickling.

icechunk-python/python/icechunk/__init__.py
```diff
@@ -161,6 +166,18 @@ def __setstate__(self, state: Any) -> None:
...
+    def preserve_read_only(self) -> Generator[None, None, None]:
+        """
+        Context manager to allow unpickling this store preserving `read_only` status.
+        By default, stores are set to read-only after unpickling.
+        """
+        try:
+            self._pickle_preserves_read_only = True
+            yield
+        finally:
+            self._pickle_preserves_read_only = False
```
paraseba, line 179: shouldn't this be the previous value of the variable instead?

paraseba (approved):
It would be good to get a review from @mpiannucci too

icechunk-python/python/icechunk/__init__.py
dcherian: it's always only modified in this method, so the previous value is always False.

mpiannucci (approved):
Looks good to me

PR #407, opened 2024-11-23
codespell action upgrade

paraseba:


paraseba:
we were getting a warning about nodejs versions, this fixes it

PR #411, opened 2024-11-25
Azure blob suppport

qianqianwang-microsoft:
Adding support for azure blob storage using azure_storage and azure_storage_blobs crates. This PR is still a WIP.


kylebarron:
My own opinion is that the maintenance burden would be lower if icechunk used one API on top of the [`object_store` crate](https://docs.rs/object_store/latest/object_store/) rather than reimplementing object store operations individually for S3, GCS, and Azure.

rabernat:
@qianqianwang-microsoft - thanks so much for this PR! We are really excited to work with you to support Azure in Icechunk!

@kylebarron thanks for chiming in. I generally agree with you. We had some good discussion of this in https://github.com/earth-mover/icechunk/issues/266#issuecomment-2455881019. But that discussion stalled out (and @paraseba is OoO this week.)

We are very open to using an object_store-based approach in Icechunk, provided that:
- It can be as performant as our current solution (see [this comment](https://github.com/earth-mover/icechunk/issues/266#issuecomment-2455576007) for the specific benchmark)
- We can figure out how to type the configuration for each object store.

qianqianwang-microsoft:
@rabernat and @kylebarron, thank you for your feedback and for pointing me to the detailed discussion regarding Azure Blob support. I wasn’t aware that object_store already has mature support for Azure when I started this PR. It makes sense to use the object_store-based approach for adding support for Azure Blob. I will begin by working on the Rust part for supporting Azure Blob but keep the direct S3 support as is. Once I get the Rust compilation working, I’ll test it with the Python bindings together with @kylebarron. Please let me know if you have any concerns or feedback about the plan.

PR #415, opened 2024-11-26
Update CI linting checks

dcherian:


.github/workflows/python-check.yaml
```diff
@@ -71,15 +71,14 @@ jobs:
...
-      #     pip install icechunk['test'] --find-links dist --force-reinstall
-      #     ruff check
+      - name: pre-commit lint
+        shell: bash
+        working-directory: icechunk-python
+        run: |
+          set -e
+          source .venv/bin/activate
+          pip install pre-commit
+          SKIP=just-rust pre-commit run --all-files
```
dcherian, line 81: instead of running ruff explicitly (and leaving open the possibility of different versions between pre-commit and this file), we run pre-commit but skip the rust bits.

mpiannucci: Was the difference in versions the reason for sebas issues? I got it reproducing locally by passing preview to ruff but wasn't sure about that

mpiannucci (approved):

.github/workflows/python-check.yaml
dcherian: Yes I ran into it too but didn't understand why he was seeing it. Now I do

Issue #418, opened 2024-11-27
Comitting to store breaks virtual references after append


mpiannucci:
When multiple arrays exist in a root group, appending works correctly until a commit is made, then references are shuffled:

```python
root = zarr.Group.from_store(store=store, zarr_format=3)
time = root.require_array(
    name="time", shape=((2,)), chunk_shape=((1,)), dtype="i4"
)
lon = root.require_array(
    name="lon", shape=((1,)), chunk_shape=((1,)), dtype="i4"
)

# Set longitude
store.set_virtual_ref(
    "lon/c/0", "s3://testbucket/path/to/python/chunk-5", offset=0, length=4
)

store.set_virtual_ref(
    "time/c/0", "s3://testbucket/path/to/python/chunk-1", offset=0, length=4
)
store.set_virtual_ref(
    "time/c/1", "s3://testbucket/path/to/python/chunk-2", offset=1, length=4
)

assert (await store._store.get("lon/c/0")) == b"fift"
assert (await store._store.get("time/c/0")) == b"firs"
assert (await store._store.get("time/c/1")) == b"econ"

store.commit("Initial commit")

# resize the array and append a new chunk
time.resize((3,))

store.set_virtual_ref(
    "time/c/2", "s3://testbucket/path/to/python/chunk-3", offset=0, length=4
)

assert (await store._store.get("lon/c/0")) == b"fift"
assert (await store._store.get("time/c/0")) == b"firs"
assert (await store._store.get("time/c/1")) == b"econ"
assert (await store._store.get("time/c/2")) == b"thir"

# commit
store.commit("Append virtual ref")

assert (await store._store.get("lon/c/0")) == b"fift"
assert (await store._store.get("time/c/0")) == b"firs"
assert (await store._store.get("time/c/1")) == b"econ"
assert (await store._store.get("time/c/2")) == b"thir"

#        assert (await store._store.get("lon/c/0")) == b"fift"
#>       assert (await store._store.get("time/c/0")) == b"firs"
#E       AssertionError: assert b'fift' == b'firs'
#E         
#E         At index 2 diff: b'f' != b'r'
#E         Use -v to get more diff
#
# tests/test_virtual_ref.py:205: AssertionError
```

cc @abarciauskas-bgse

mpiannucci:
Also important to note that its not totally deterministic which refs are swapped, but one of `time/c/0` and `lon/c/0` are incorrect on every run

mpiannucci:
[...]

PR #419, opened 2024-11-27
Add test case for appending virtual refs

mpiannucci:
#418


mpiannucci:
Added to main in #421

Issue #422, opened 2024-11-30
New python API - design document


paraseba:
## New Icechunk Python API design

The following is pseudo-code of how a new Python API may look.

### Modules

We have different modules for sync and async interfaces. Most zarr users are not used to async
programming, so moving all async methods to a different module optimizes for clarity for the average user.

```python
import icechunk.sync
import icechunk.async
import zarr
```

We should evaluate making `icechunk.sync` just `icechunk`, or some type of alias.

### Free functions

```python
import icechunk.sync as icechunk

# TODO: define the types for arguments and return values for these functions
#       in a way that works for all configurations and object stores

storage = .....
icechunk.exists(storage)
repo: Repository = icechunk.create_repository(storage, ...)
repo: Repository = icechunk.open(storage, ...)
repo: Repository = icechunk.open_or_create(storage, ...)
```

Notice, there is no way to instantiate `Repository` in read only mode. This is by design, it would make the API more complex for very little value. As we'll see there is the concept of `Session` that can in fact be read-only.

### Ancestry data structures

```python
@dataclasss
class Commit
  msg: str
  created_at: datetime
  snapshot_id: str
```

### Repo management

Once we open a `Repository` by calling `open`,  `create_repository` or `open_or_create`, we can inspect it and manage it.

```python
import icechunk.sync as icechunk

repo = icechunk.open(...)

# exception if snapshot not found
repo.create_branch("dev", "some-snapshot-id")
# exception if snapshot not found
repo.reset_branch("dev", "some-snapshot-id")

# exception if snapshot not found
repo.create_tag("v1.0.5", "some-snapshot-id")

# returns the sequence of parent commits including some-snapshot-id
# exception if snapshot not found
repo.ancestry("some-snapshot-id") -> Iterator[Commit]

# returns branch names
repo.list_branches() -> Iterator[str]

# returns tag names
repo.list_tags() -> Iterator[str]

# returns the snapshot id pointed by a ref
# exception if ref not found
repo.get_ref(branch="main") -> str
repo.get_ref(tag="v1.0.5") -> str
```

### Read only sessions

Having a `Repository` instance allows the creation of read sessions. These are created using the `read_session` method, and they resolve to a snapshot id at the time of creation. If a branch is reset afterwards, or a tag deleted, the session doesn't change.

```python
import icechunk.sync as icechunk

repo = icechunk.open(...)
# exception if branch/tag/snapshot cannot be found
session = repo.read_session(branch="main")
[...]

Issue #221, opened 2024-10-14
Refactor Python API into separate Repo and Store objects 


rabernat:
The fact that we only have one python object--the `IcechunkStore` is causing some difficulty. This object needs to be a valid Zarr store but also support our own API methods.

I propose we evolve to a state more like Arraylake

- `icechunk.Repo` - with `create`, `open_existing`, `commit`, `reset`, etc.
- `icechunk.AsyncRepo` - with async versions of those
- `icechunk.ZarrStore` - the Zarr store interface. This should be accessible under the `.store` attribute for `Repo` and `AsyncRepo`. We disallow or discourage directly creating `ZarrStore`.

PR #432, opened 2024-12-03
[WIP] API Refactor

mpiannucci:
#422 #221


icechunk/src/session.rs
```diff
@@ -0,0 +1,1055 @@
...
+    /// Merge a set of `ChangeSet`s into the repository without committing them
+    pub async fn merge(&mut self, changes: ChangeSet) -> SessionResult<()> {
+        if self.read_only() {
+            return Err(SessionError::ReadOnlySession);
+        }
+        self.change_set.merge(changes);
+        Ok(())
+    }
+
+    pub async fn commit(
```
mpiannucci, line 418: `commit` is now consumed, so after committing you need to get a new session.

icechunk/src/store.rs
```diff
@@ -0,0 +1,978 @@
...
+    metadata::{
+        ArrayShape, ChunkKeyEncoding, ChunkShape, Codec, DataType, DimensionNames,
+        FillValue, StorageTransformer, UserAttributes,
+    },
+    repository::{get_chunk, ChunkPayload, RepositoryError, ZarrArrayMetadata},
+    session::{Session, SessionError},
+    zarr::{KeyNotFoundError, ListDirItem, StoreError, StoreOptions, StoreResult},
+};
+
+pub struct Store<'a> {
```
mpiannucci, line 36: This might cause major issues with python bindings, lifetimes arent allowed

icechunk/src/session.rs
```diff
@@ -0,0 +1,1055 @@
...
+    config: RepositoryConfig,
+    storage: Arc<dyn Storage + Send + Sync>,
+    virtual_resolver: Arc<dyn VirtualChunkResolver + Send + Sync>,
+    branch_name: Option<String>,
+    snapshot_id: SnapshotId,
+    change_set: ChangeSet,
+}
+
+impl Session {
+    pub fn create_readable_session(
```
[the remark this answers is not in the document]
mpiannucci: good catch, yeah thats a relic left over from when i tried to implement these as traits

icechunk/src/session.rs
```diff
@@ -0,0 +1,1055 @@
...
+        }
+    }
+
+    pub fn read_only(&self) -> bool {
+        self.branch_name.is_none()
+    }
+
+    pub fn store(&mut self) -> Store {
+        let read_only = self.read_only();
+        let mutexed = Arc::new(RwLock::new(self));
```
paraseba, line 101: this looks quite suspicious to me, but haven't thought about it enough. This when calling `store`, you give out the only possible mutable reference to the Session, so you no longer can make changes to the session until the `Store` is freed. Will have to wrap my head around it, not sure how commit would work for example.

mpiannucci: I really don't like this. I don't know how else to manage though. Need to think through the hierarchy

mpiannucci: OK the only mutable part is the `ChangeSet` I think, so we can in theory just `Arc<Mutex` the changeset and then the session is no longer locked. I think

icechunk/src/session.rs
```diff
@@ -0,0 +1,1065 @@
...
+        &self,
+        node: NodeId,
+        manifests: &[ManifestRef],
+        coords: &ChunkIndices,
+    ) -> RepositoryResult<Option<ChunkPayload>> {
+        // FIXME: use manifest extents
+        get_old_chunk(self.storage.as_ref(), node, manifests, coords).await
+    }
+
+    pub async fn list_nodes(
```
mpiannucci, line 185: @paraseba any ideas on how to fix this? This currently is not correct because the iterator grabs the lifetime of the RWLockGuard which is temporary, so returning an iterator does not compile

mpiannucci:
I like the approach here much better now, just have to figure out if there is a way to generator the listing generatively, because otherwise listing is (again) going to cause major issues

mpiannucci:
Rust crate is now compiling (except for tests)

icechunk/src/session.rs
```diff
@@ -0,0 +1,1975 @@
...
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_manifests_shrink() -> Result<(), Box<dyn Error>> {
+        let in_mem_storage =
+            Arc::new(ObjectStorage::new_in_memory_store(Some("prefix".into()))?);
+        let storage: Arc<dyn Storage + Send + Sync> = in_mem_storage.clone();
+        let repository =
+            Repository::create(RepositoryConfig::default(), storage, None).await?;
```
mpiannucci, line 1794: This shwos teh session usage. @paraseba is this the workflow you expected? with having to get a new session after each commit?

mpiannucci:
Closed in favor of #470

Issue #430, opened 2024-12-02
Virtual chunks improvements


paraseba:
# This doc has been moved to https://github.com/earth-mover/icechunk/pull/436

## What we need

1. Virtual chunks can be stored in all supported object stores
2. Multiple credentials sets for virtual chunk resolution
3. Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
4. Be able to know all the "places" pointed by a repo, without having to load every manifest in every version
5. Object level change detection (not serving a virtual chunk if the containing file changed)

### Chunks url styles

* Many chunks from a few objects, the prefix can be the file
  * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
  * Having between 10 and 10k chunks from the same file is not unusual
* Virtual zarr: each chunk comes from a different object, the prefix is not the object, we still need an extra string
  * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
  * There could be tens of millions of chunks (so different urls)

### Prefixes

Storing the full URL for each chunk duplicates a lot of data in the manifest, making it large. The idea is to try to extract common _prefixes_ and identify them with a short id. Then the manifest only needs to identify the prefix id + maybe an extra string + length + offset.

In the first url style above, the prefix could be the full file, `s3://some-bucket/some-prefix/some-file.nc`. In the second example, there are options for the prefix, it could be `s3://some-bucket/some-prefix/c/0/0/1`, or `s3://some-bucket/some-prefix/c/0/0` or `s3://some-bucket/some-prefix/c/0` or `s3://some-bucket/some-prefix/c`, with different numbers of chunks under it in each case.

Intelligence is needed to optimize what is consider a good prefix. If done badly, we could have:

* as many prefixes as chunks, for example by selecting `s3://some-bucket/some-prefix/c/0/0/1` in the second example above
* too large of an extra string in each chunk, for example, if we select `s3://some-bucket` as the prefix each chunk needs to add something like `some-prefix/c/0/0`.

### Should prefixes be per snapshot or global?

The advantage of per snapshot is that we don't need to load/parse all the prefixes if the current snapshot uses only a few. A case for this would be a virtual dataset that has been materialized in a newer version. If prefixes are per-snapshot we wouldn't need to load any prefixes.

[...]

PR #436, opened 2024-12-04
Virtual chunks design document

paraseba:
These are early thoughts, feel free to provide feedback and different ideas. Also, particularly important: am I missing requirements?

Closes: #430

### DIFF
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -332,7 +332,7 @@ The repo is stored in S3 us-east-1, all virtual chunks point to Google cloud sto
 
 #### Go to local MinIO for testing
 
-All virtual chunks point to S3 us-east-1, for testing purposes the chunks for an array in `s3://bucket/prefix/array` are also copied locally and served with MinIO on port 8080.
+All virtual chunks point to S3 us-east-1, for testing purposes the chunks for an array in `s3://bucket/prefix/array` are also copied locally and served with MinIO on port 9000.
 
 * When calling `set_virtual_ref` locations are passed with `s3://` protocol.
 * The icechunk repo instance is initialized with a configuration that creates a container with `name=minio-override`, `prefix=s3://bucket/prefix/array` and `endpoint_url=localhost:9000`
@@ -341,13 +341,14 @@ All virtual chunks point to S3 us-east-1, for testing purposes the chunks for an
 
 Ichecunk repo stored in S3 bucket `repo-bucket`. Virtual array `A` on Google cloud, bucket `virtual-a`. Virtual array `B` on Google cloud, bucket `virtual-b`. Virtual array `C` on s3 bucket `virtual-c`. All buckets have different credential sets.
 
-* When ingesting `virtual-a` VirtualiZarr is instructed to use protocol `modela`
-* When ingesting `virtual-b` VirtualiZarr is instructed to use protocol `modelb`
-* When ingesting `virtual-c` VirtualiZarr is instructed to use protocol `modelc`
+* When ingesting array `A` VirtualiZarr is instructed to use protocol `modela`. URLs will look something like: `modela://some-bucket/virtual-a`
+* When ingesting array `B` VirtualiZarr is instructed to use protocol `modelb`. URLs will look something like: `modelb://some-other-bucket/virtual-b`
+* When ingesting array `C` VirtualiZarr is instructed to use protocol `modelc`. URLs will look something like: `modelc://some-third-bucket-on-aws/virtual-c`
 * Persistent config is set up in the repo to create the following virtual chunk containers:
-  * `modela`: on Google storage platform with prefix `virtual-a`
-  * `modelb`: on Google storage platform with prefix `virtual-b`
-  * `modelc`: on S3 platform with prefix `virtual-c`
+  * `modela`: on Google storage platform with prefix `modela`
+  * `modelb`: on Google storage platform with prefix `modelb`
+  * `modelc`: on S3 platform with prefix `modelc`
+  * In this example we are using the same string for the container name and the prefix, but this doesn't need to be the case.
   * These containers can be set before or after the ingest (we recommend before and to pass `validate_containers=True`)
 * To open the repository users pass credentials associated with each virtual chunk container, in addition to the repository credentials.
 
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -13,7 +13,7 @@
 
 * Existing packages for reference generation (VirtualiZarr, Kerchunk, etc.) uses URLs to identify virtual chunk locations.
 * One common case is many virtual chunks from a few objects.
-  * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
+  * Example: `s3://some-bucket/some-prefix/some-file.nc` is referenced by 10k chunks.
   * Having between 10 and 10k chunks from the same object is not unusual
 * Other common use case for virtual zarr datasets: each chunk comes from a different object. This can happen, for example, with a virtual zarr dataset.
   * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -145,7 +145,7 @@ pub struct RepositoryConfig {
 }
 ```
 
-The `RepositoryConfig` object acts as a default, read during repository instantiation, and overrideable by users at runtime. Different fields of configuration may have different behaviors when overridden.
+The `RepositoryConfig` object acts as a default, read during repository instantiation, and overridable by users at runtime. Different fields of configuration may have different behaviors when overridden.
 
 A user can modify the `RepositoryConfig` for their current repo instance, but decide not to save the change to storage. This will be very common as users tweak the configuration during experimentation.
 
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -11,7 +11,7 @@
 
 ### Current virtual chunk URL styles
 
-* The open source stack seating on top of Icechunk (VirtualiZarr, etc.) uses URLs to identify virtual chunk locations.
+* Existing packages for reference generation (VirtualiZarr, Kerchunk, etc.) uses URLs to identify virtual chunk locations.
 * One common case is many virtual chunks from a few objects.
   * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
   * Having between 10 and 10k chunks from the same object is not unusual
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -15,7 +15,7 @@
 * One common case is many virtual chunks from a few objects.
   * Example: `s3://some-bucket/some-prefix/some-file.nc` is referenced by 10k chunks.
   * Having between 10 and 10k chunks from the same object is not unusual
-* Other common use case for virtual zarr datasets: each chunk comes from a different object. This can happen, for example, with a virtual zarr dataset.
+* The other common use case for virtual zarr datasets: each chunk comes from a different object. This can happen, for example, with a virtual zarr dataset.
   * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
   * There could be tens of millions of chunks (so tens of millions of different urls)
 
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -1,0 +1,163 @@
+# Improvements to virtual chunks handling
+
+## What we need
+
+1. Virtual chunks can be stored in all supported object stores
+2. Multiple credentials sets for virtual chunk resolution
+3. Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
+4. Be able to know all the "places" pointed by a repo, without having to load every manifest in every version
+5. Object level change detection (not serving a virtual chunk if the containing file changed)
+
+## Discussion
+
+### Chunks url styles
+
+* Many chunks from a few objects, the prefix can be the file
+  * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
+  * Having between 10 and 10k chunks from the same file is not unusual
+* Virtual zarr: each chunk comes from a different object, the prefix is not the object, we still need an extra string
+  * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
+  * There could be tens of millions of chunks (so different urls)
+
+### Prefixes
+
+Storing the full URL for each chunk duplicates a lot of data in the manifest, making it large. The idea is to try to extract common _prefixes_ and identify them with a short id. Then the manifest only needs to identify the prefix id + maybe an extra string + length + offset.
+
+In the first url style above, the prefix could be the full file, `s3://some-bucket/some-prefix/some-file.nc`. In the second example, there are options for the prefix, it could be `s3://some-bucket/some-prefix/c/0/0/1`, or `s3://some-bucket/some-prefix/c/0/0` or `s3://some-bucket/some-prefix/c/0` or `s3://some-bucket/some-prefix/c`, with different numbers of chunks under it in each case.
+
+Intelligence is needed to optimize what is consider a good prefix. If done badly, we could have:
+
+* as many prefixes as chunks, for example by selecting `s3://some-bucket/some-prefix/c/0/0/1` in the second example above
+* too large of an extra string in each chunk, for example, if we select `s3://some-bucket` as the prefix each chunk needs to add something like `some-prefix/c/0/0`.
+
+### Should prefixes be per snapshot or global?
+
+The advantage of per snapshot is that we don't need to load/parse all the prefixes if the current snapshot uses only a few. A case for this would be a virtual dataset that has been materialized in a newer version. If prefixes are per-snapshot we wouldn't need to load any prefixes.
+
+On the other hand, requirement 4 asks for the full list of "dependencies" for the dataset, and that means for all versions, which requires a global list of prefixes. Without a global list, we would need to fetch the list of prefixes of each snapshot.
+
+### Who creates/decides the prefixes
+
+Obviously, generating the list of prefixes given the list of virtual chunks is doable automatically. But it's a tricky optimization problem. One strategy could be to generate a trie with all chunk urls, and then use some heuristic to select prefixes using heavily populated nodes. It's tricky and potentially computationally expensive for large repos.
+
+An alternative is to let the user dictate the prefixes. This way, they can optimize them knowing beforehand what virtual chunks they will add. If they don't optimize, they get large manifests without prefix extraction. We still need to satisfy requirements 2 and 4, which would require automatic extraction of at least the bucket and platform.
+
+Additionally, the optimization problem can be treated as an extra process (similar to compaction or garbage collection), and ran offline on demand.
+
+### Templates instead of prefixes
+
+Using templates would be more efficient, example:
+
+* `s3://some-bucket/some-prefix/1-analysis-output-for-model-abc.nc`
+* `s3://some-bucket/some-prefix/2-analysis-output-for-model-abc.nc`
+
+If we use the template `s3://some-bucket/some-prefix/{}-analysis-output-for-model-abc.nc` each chunk could have a couple of extra characters only, filling the placeholders in the template.
+
+### Supporting multiple clouds
+
+When the user adds a virtual ref, they need to indicate the details of the platform to retrieve it. These include what cloud it is (so we know how to set up credentials), region, endpoint url, port and other settings. It looks a lot like our `S3Config` type (or an expanded version of that).
+
+### Supporting multiple credentials
+
+Credentials are a read time concern. When the user writes the virtual chunk, they don't necessarily know what credentials set will be used to read it.
+
+Credentials need to be specified later, when the chunks are going to be read. For example, by indicating in runtime configuration a set of credentials for each virtual chunk bucket/prefix/template.
+
+## Design
+
+### Explicit containers for virtual chunks
+
+* Each repo has a global list of `VirtualChunkContainers`
+
+```rust
+// this is pseudo-code
+struct VirtualChunkContainer {
+  name: String,  // something like model-output
+  platform: Platform,  // something like S3
+  url_template: String, // something like some-bucket/some-prefix/output/{}/{}.nc
+  default_template_arguments: Vec<String>,
+  cloud_client_config: Map<String, Any>,  // something like {endpoint_url: 8080}
+}
+```
+
+* Users can add to the list and edit members but not delete.
+
+### Global config
+
+* The list of chunk containers is stored in a new global `Config` object. There is a single configuration per repo, and all reads and writes are done reading the latest configuration. In the initial implementation, the "history" of configuration is present only for reference and user inspection by hand.
+
+TBD: how to store this information. One approach would be using the new PUT if-match S3 functionality, but this would be the first place where we overwrite objects.
+
+```rust
+// this is pseudo-code
+struct Config {
+  inline_chunk_threshold_bytes: u16,
+  virtual_chunk_containers: Vec<VirtualChunkContainer>,
+}
+```
+
+* The `Config` object can act as a configuration default when users don't override. But different fields of configuration may have different behaviors.
+
+### New virtual chunk ref
+
+We change from
+
+```rust
+pub struct VirtualChunkRef {
+    pub location: VirtualChunkLocation, // this is basically a String for the URL today
+    pub offset: ChunkOffset,
+    pub length: ChunkLength,
+}
+```
+
+To
+
+```rust
+pub struct VirtualChunkRef {
+    pub container_index: u16,
+    pub container_template_args: Vec<Option<String>>,
+    pub offset: ChunkOffset,
+    pub length: ChunkLength,
+    pub last_modified: Option<u32>,
+}
+```
+
+* The `last_modified` field contains an optional unix epoch in seconds. If set for a chunk, retrieval will fail if the file was modified after this time. We use this instead of an ETag or checksum to save space on the manifests. `u32` should be enough until year ~ 2100.
+
+### Reading virtual chunks
+
+* When a user create a repository instance, they can specify credentials for each `VirtualChunkContainer`, pointing to them by name.
+* They can also set default `VirtualChunkContainer` credentials, that will be used for any container not specifically set
+* If no credentials are specified the same as for the main `Storage` will be used (if platforms are the same)
+* On read of a virtual chunk:
+  * The `VirtualChunkContainer` is resolved from configuration.
+  * An object storage client is instantiated. We'll need a cache here, so we don't keep creating new instances for each chunk
+  * The client is initialized with the credentials passed (or defaulted).
+  * The container template is expanded with the `template_args` in the virtual chunks. If there are no enough template arguments or some are None, they are replaced by the default template arguments in the container. If there are too many, last ones are ignored.
+  * The chunk is fetched from the results URL.
+
+### Writing virtual chunks
+
+* Users must first update the repo configuration to include the needed `VirtualChunkContainers`.
+* When the user writes a virtual chunk ref, they pass in the `VirtualChunkRef` the index for the container and the template expansion arguments. An error is produced if the container is not found.
+* This is a bit inconvenient, having to declare the container before being able to write. An alternative would be to automatically create a container that includes only the bucket name as template. On writes we try to match all templates against the url, and use the first one that matches, if non does, we create a new container. We could add that as a new function in the future if it's not required on first version.
+
+## Check requirements are satisfied
+
+1. > Virtual chunks can be stored in all supported object stores
+
+   Now the virtual chunk containers inform what is the platform, and we can create the appropriate object store client at read time. These clients are cached for performance.
+2. > Multiple credentials sets for virtual chunk resolution
+
+    Now the user initializes a Repo instance by passing, if needed, a credential set for each virtual chunk container. If a chunk is read for which there are no credentials, we default to a common set of credentials for virtual chunks, and ultimately, to the Storage credentials if on the same platform.
+3. > Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
+  
+    For large prefixes, the template + arguments creates some savings. For example, a template for virtual zarr like `some-long-bucket-name/with-some-even-longer-prefix/and-more-nesting/some-long-array-name-for-the-dataset/c/{}/{}/{}`.
+   On the other hand, the template arguments vector has a 24 character overhead, on top of the values themselves. So it's not a lot of savings for smaller prefixes.
+5. > Be able to know all the "places" pointed by a repo, without having to load every manifest in every version
+
+    This can now be retrieved directly from the repository config. It's basically the set of virtual chunk containers (or a subset at least)
+
+6. > Object level change detection (not serving a virtual chunk if the containing file changed)
+
+    We optionally store the last-modified timestamp of the object containing each chunk. This is not optimal, a hash would be better, but it takes more space. Also not optimal, storing it per chunk, when many chunks come from the same object, but we don't have other easy place to store it if they are using containers that include more than one object. We could add more configuration to optimize this in the future.
--- a/design-docs/001-virtual-chunk-improvements.md
+++ b/design-docs/001-virtual-chunk-improvements.md
@@ -2,101 +2,205 @@
 
 ## What we need
 
-1. Virtual chunks can be stored in all supported object stores
-2. Multiple credentials sets for virtual chunk resolution
-3. Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
-4. Be able to know all the "places" pointed by a repo, without having to load every manifest in every version
-5. Object level change detection (not serving a virtual chunk if the containing file changed)
+1. Virtual chunks can be stored in all supported object stores. The same repo can have virtual chunks hosted in different object stores.
+2. Multiple credentials sets for virtual chunk resolution, accessing some chunks may require a different set than others.
+3. A mechanism to know all the "dependencies" of a repo. That is, all the "places" where its virtual chunks are hosted. This is important for users to know if they can delete old data, or if there are virtual datasets that depend on it.
+4. If the file hosting a virtual chunk changed after the chunk ref was written, we want to fail with a runtime error instead of serving corrupted data. This is because offset and length are usually not maintained during updates.
 
 ## Discussion
 
-### Chunks url styles
+### Current virtual chunk url styles
 
 * Many chunks from a few objects, the prefix can be the file
   * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
   * Having between 10 and 10k chunks from the same file is not unusual
-* Virtual zarr: each chunk comes from a different object, the prefix is not the object, we still need an extra string
+* Virtual zarr dataset: each chunk comes from a different object, the prefix is not the object, we still need an extra string
   * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
-  * There could be tens of millions of chunks (so different urls)
+  * There could be tens of millions of chunks (so tens of millions of different urls)
 
-### Prefixes
+### Supporting multiple virtual chunk locations
 
-Storing the full URL for each chunk duplicates a lot of data in the manifest, making it large. The idea is to try to extract common _prefixes_ and identify them with a short id. Then the manifest only needs to identify the prefix id + maybe an extra string + length + offset.
+Because of requirement 1, the same repo can contain virtual chunks in multiple clouds and buckets, potentially even bucket with the same name in different platforms.
 
-In the first url style above, the prefix could be the full file, `s3://some-bucket/some-prefix/some-file.nc`. In the second example, there are options for the prefix, it could be `s3://some-bucket/some-prefix/c/0/0/1`, or `s3://some-bucket/some-prefix/c/0/0` or `s3://some-bucket/some-prefix/c/0` or `s3://some-bucket/some-prefix/c`, with different numbers of chunks under it in each case.
+When the user adds a virtual ref, they will need to indicate the details of the platform to retrieve it. These include what cloud it is, region, endpoint url, etc. This means, the current approach of pointing to a chunk with a simple URL is not going to be enough. Not without encoding several new attributes in the URL, which seems hacky.
 
-Intelligence is needed to optimize what is consider a good prefix. If done badly, we could have:
+This all means, the following function may have to change:
 
-* as many prefixes as chunks, for example by selecting `s3://some-bucket/some-prefix/c/0/0/1` in the second example above
-* too large of an extra string in each chunk, for example, if we select `s3://some-bucket` as the prefix each chunk needs to add something like `some-prefix/c/0/0`.
+```python
+def set_virtual_ref(
+    self, key: str, location: str, offset: int, length: int
+) -> None: ...
+```
 
-### Should prefixes be per snapshot or global?
+`location` as a `str` is not enough.
 
-The advantage of per snapshot is that we don't need to load/parse all the prefixes if the current snapshot uses only a few. A case for this would be a virtual dataset that has been materialized in a newer version. If prefixes are per-snapshot we wouldn't need to load any prefixes.
-
-On the other hand, requirement 4 asks for the full list of "dependencies" for the dataset, and that means for all versions, which requires a global list of prefixes. Without a global list, we would need to fetch the list of prefixes of each snapshot.
-
-### Who creates/decides the prefixes
-
-Obviously, generating the list of prefixes given the list of virtual chunks is doable automatically. But it's a tricky optimization problem. One strategy could be to generate a trie with all chunk urls, and then use some heuristic to select prefixes using heavily populated nodes. It's tricky and potentially computationally expensive for large repos.
-
-An alternative is to let the user dictate the prefixes. This way, they can optimize them knowing beforehand what virtual chunks they will add. If they don't optimize, they get large manifests without prefix extraction. We still need to satisfy requirements 2 and 4, which would require automatic extraction of at least the bucket and platform.
-
-Additionally, the optimization problem can be treated as an extra process (similar to compaction or garbage collection), and ran offline on demand.
-
-### Templates instead of prefixes
-
-Using templates would be more efficient, example:
-
-* `s3://some-bucket/some-prefix/1-analysis-output-for-model-abc.nc`
-* `s3://some-bucket/some-prefix/2-analysis-output-for-model-abc.nc`
-
-If we use the template `s3://some-bucket/some-prefix/{}-analysis-output-for-model-abc.nc` each chunk could have a couple of extra characters only, filling the placeholders in the template.
-
-### Supporting multiple clouds
-
-When the user adds a virtual ref, they need to indicate the details of the platform to retrieve it. These include what cloud it is (so we know how to set up credentials), region, endpoint url, port and other settings. It looks a lot like our `S3Config` type (or an expanded version of that).
-
-### Supporting multiple credentials
+### Read time setup
 
 Credentials are a read time concern. When the user writes the virtual chunk, they don't necessarily know what credentials set will be used to read it.
 
-Credentials need to be specified later, when the chunks are going to be read. For example, by indicating in runtime configuration a set of credentials for each virtual chunk bucket/prefix/template.
+Because of requirements 1 and 2, when users open a repo they need to know what sets of credentials they need to provide to be able to access the chunks. This is not immediately obvious, Icechunk should provide a way to introspect a repo and inform the user what they need to provide. Then the user can setup the Repo instance appropriately with the credentials.
+
+This also applies to requirement 3, users need to know if a repo is going to be affected by, for example, the delete of an unrelated repo.
 
 ## Design
 
-### Explicit containers for virtual chunks
+### Virtual chunk containers
 
-* Each repo has a global list of `VirtualChunkContainers`
+We introduce a new datastructure, the `VirtualChunkContainer`:
 
 ```rust
 // this is pseudo-code
 struct VirtualChunkContainer {
-  name: String,  // something like model-output
-  platform: Platform,  // something like S3
-  url_template: String, // something like some-bucket/some-prefix/output/{}/{}.nc
-  default_template_arguments: Vec<String>,
-  cloud_client_config: Map<String, Any>,  // something like {endpoint_url: 8080}
+  name: String,                           // something like model
+  protocol: String,                       // something like s3
+  platform: Platform,                     // something like S3
+  region: Option<String>,                 // something like us-east-2
+  bucket_prefix: Option<String>,          // something like model-outputs/netcdf
+  cloud_client_config: Map<String, Any>,  // something like {endpoint_url: http://localhost:8080}
 }
 ```
 
-* Users can add to the list and edit members but not delete.
+* Each repo has a list of `VirtualChunkContainers` that is global (applies to all snapshots)
+* Users can add to the list and edit members but not delete. Names are unique within the repo.
+* Adding containers to a repo is optional but useful when there are virtual chunks.
+* `VirtualChunkContainers` serve several purposes:
+  * Using their `name` a user can setup credentials for a specific set of chunks, in a given object store and bucket prefix.
+  * Using their `protocol` and other properties, Icechunk knows how to build an object store client that can retrieve the virtual chunks.
+  * Users can review the list of `VirtualChunkContainers` for a repo to quickly understand what other buckets and prefixes contain data used by this repo.
+  * Users can review the list of `VirtualChunkContainers` for a repo to quickly understand what credentials they need to provide if they want to access all virtual chunks. They can still use the repo without providing those credentials, but they virtual chunk requests may fail.
 
 ### Global config
 
-* The list of chunk containers is stored in a new global `Config` object. There is a single configuration per repo, and all reads and writes are done reading the latest configuration. In the initial implementation, the "history" of configuration is present only for reference and user inspection by hand.
+The list of chunk containers is stored in a new global `Config` object. There is a single configuration per repo, and all operations use the last version of the configuration.
 
-TBD: how to store this information. One approach would be using the new PUT if-match S3 functionality, but this would be the first place where we overwrite objects.
+The configuration is stored as a JSON file in `repo-root/config.json`, and it's written using PUT if-match to ensure writers are aware of the latest updates.
+
+The global configuration solves also [other requirements](https://github.com/earth-mover/icechunk/issues/440) outside of this design doc. Icechunk repos need, also for other reasons, a way to store persistent default configuration.
 
 ```rust
 // this is pseudo-code
 struct Config {
   inline_chunk_threshold_bytes: u16,
+  ... more fields
+
   virtual_chunk_containers: Vec<VirtualChunkContainer>,
 }
 ```
 
-* The `Config` object can act as a configuration default when users don't override. But different fields of configuration may have different behaviors.
+The `Config` object acts as a default, overrideable by users at runtime. But different fields of configuration may have different behaviors on overrides.
+
+### New mechanism to set virtual references
+
+We add an optional `checksum` argument to `set_virtual_ref` (we need a better name for this argument).
+
+```python
+# this is pseudo-code
+
+class ETag:
+  value: str
+
+class LastModified:
+  value: datetime
+
+class Repository:  # the class with this method is going to change soon to be Session, for an unrelated reason
+  def set_virtual_ref(
+      self,
+      key: str,
+      location: str,
+      offset: int,
+      length: int,
+      checksum: ETag | LastModified | None = None,
+      validate_containers: bool = False
+  ) -> None: ...
+```
+
+`checksum` can be set to the in-object-store ETag of the file containing the virtual chunk, or to its last modified time. In the case of last modified time, the value
+doesn't need to be precise. The system will only fail to provide the chunk if the object's last-modified-at is larger than the passed value. So, passing `checksum = LastModified(datetime.now())` will usually be enough.
+
+This new argument will help us satisfy requirement 4.
+
+The interpretation of the `location` argument also changes. `location` must be an absolute url that includes a protocol. The protocol can be:
+
+* s3://
+* gcs://
+* tigris://  (is this the right protocol)
+* fs://  (is this right for local file system)
+* etc. for other supported object stores
+* arbitrary-virtual-chunk-container-name://  (example: models://)
+
+Depending on the protocol used, Icechunk will use a different mechanism to resolve the chunk, as explained below.
+
+If `validate_containers` is set to `True`, the function will fail if the protocol passed is not registered in a virtual chunk container. Using `validate_containers=False` is useful, if containers will be declared after the write.
+
+### Reading virtual chunks
+
+When a virtual chunk is requested, Icechunk will find in the `location` field of the stored reference, an absolute URL with a specific protocol. It will need to map this location to a given object store, bucket, path and set of credentials.
+
+The mapping is done by comparing the protocol and path to the set of configured virtual chunk containers, plus a set of default virtual chunk containers.
+
+This is how the mapping from location to cloud+bucket+prefix+credentials works:
+
+* Using the `location` protocol we find all containers that match the protocol
+* We sort the matching containers in order of descending prefix length
+* We find the first one that has a `bucket_prefix` that is a prefix of the `location` path (this will be the most specific container that matches)
+* If no containers match, the chunk request cannot be fulfilled and we fail with an error message indicating the missing config.
+* We use the rest of the data in the matched container to instantiate an object store client (or retrieve it from cache)
+* Credentials for the client must be passed by the user when the repository is instantiated, by indicating the container name to which they apply
+
+We have one default container for each supported object store, the container has protocol and name s3, gcs, tigris, etc. and has an empty `bucket_prefix` and no region configured. These default containers, just like any others set in the persistent repo configuration, can be overriden when the repository is instantiated.
+
+When a region is not indicated and the platform needs it, the logic is:
+
+* For the platform in which the Icechunk repo is stored, we use the same region as the repo
+* For all other platforms we use the default region for that platform
+
+#### Use case: basic usage
+
+All virtual chunks point to S3 us-east-1, single credential set.
+
+* When calling `set_virtual_ref` locations are passed with `s3://` protocol.
+* No persistent configuration needs to be set in the repo.
+* At runtime, the repo instance can be configured passing virtual chunk credentials for container "s3", the default container.
+* If Icechunk native data is also in s3 us-east-1, and the same credentials can be used, the previous point is not needed, the default will be using the same credentials as the ones for the repo.
+
+#### Use case: non default region
+
+All virtual chunks point to S3 sa-east-1, single credential set.
+
+* When calling `set_virtual_ref` locations are passed with `s3://` protocol.
+* A persistent configuration can be set creating a virtual chunk container with protocol s3 and region "sa-east-1", this overrides the default s3 container.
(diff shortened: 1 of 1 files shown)


paraseba:
@rabernat said:
> I'm very 👎 on creating a new mechanism for storing "config" outside of the existing snapshot system, particularly if it involves transactional updates. I would much rather create a new type of metadata file that is referenced from the snapshot. I'd like the entire state of the store to be self contained within a snapshot.

That's very similar to what I'm proposing here. I'm also proposing a new type of metadata file. But:

* The config needs to be global, not per snapshot.
* I don't want to have to serialize the request of the snapshot and the config, so the config_id cannot live in the snapshot.
* I could duplicate the file on every snapshot, renaming it to have the same id as the new snapshot, but, seems wasteful. Most of the time the config won't change.

So my current thinking is: single `configuration.msgpack` file at the root of the repo. It's updated using PUT if-matches. On reads, it gets loaded concurrently with the snapshot file.

paraseba:
@rabernat said:
> What if we created a manifest of every single virtual file that is referenced in the repo? How big could this be? it's hard to imagine more than, say 1_000_000 virtual files? 

My counterexample in the design document is a virtual zarr dataset. Every chunk is a different file. All those file are on the same platform, bucket, etc. But even without being as extreme, this still has the same problem no? All those prefixes in the files manifest are wasteful.

Also, we still need some kind of report on "dependencies", the concept of "containers". This last concept is user level in the sense that the user will want to provide different credential sets (potentially). Not on a per file basis, but on whatever they consider a container potentially. Users will want to know "what other datasets this dataset depends on?", containers is somewhat of a proxy to that.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
+# Improvements to virtual chunks handling
+
+## What we need
+
+1. Virtual chunks can be stored in all supported object stores
```
maxrjones, line 5: I know Rich Signell has used Kerchunk on local filesystems. Do you think virtual chunk storage needs to be limited to object stores?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+2. Multiple credentials sets for virtual chunk resolution
+3. Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
+4. Be able to know all the "places" pointed by a repo, without having to load every manifest in every version
+5. Object level change detection (not serving a virtual chunk if the containing file changed)
```
maxrjones, line 9: I wonder if a more specific requirement here would be helpful. What does "not serve" mean? Is it an error, warning, or configurable behavior?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+pub struct VirtualChunkRef {
+    pub container_index: u16,
+    pub container_template_args: Vec<Option<String>>,
+    pub offset: ChunkOffset,
+    pub length: ChunkLength,
+    pub last_modified: Option<u32>,
+}
+```
+
+* The `last_modified` field contains an optional unix epoch in seconds. If set for a chunk, retrieval will fail if the file was modified after this time. We use this instead of an ETag or checksum to save space on the manifests. `u32` should be enough until year ~ 2100.
```
maxrjones, line 125: Appreciating that using an optional unix epoch saves on space, what do you expect the impact of chosing unix epoch over ETag or checksum to be on read or write time?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
+# Improvements to virtual chunks handling
+
+## What we need
```
maxrjones, line 3: I understand that this design document pertains mostly to the `VirtualChunkContainer`, so I'm not sure if these are in scope but there are some interactions at the store level that may be worth mentioning. For example,

- Does not provide a barrier to distributed writes of virtual chunks (xref https://github.com/zarr-developers/VirtualiZarr/issues/308)
- Does not provide a barrier to combining virtual chunks of variable lengths in a single store (xref https://github.com/zarr-developers/VirtualiZarr/issues/12)
- Does not provide a barrier to combining virtual chunks with "inlined" data in a single store (xref https://github.com/zarr-developers/VirtualiZarr/issues/62)

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+  
+    For large prefixes, the template + arguments creates some savings. For example, a template for virtual zarr like `some-long-bucket-name/with-some-even-longer-prefix/and-more-nesting/some-long-array-name-for-the-dataset/c/{}/{}/{}`.
+   On the other hand, the template arguments vector has a 24 character overhead, on top of the values themselves. So it's not a lot of savings for smaller prefixes.
+5. > Be able to know all the "places" pointed by a repo, without having to load every manifest in every version
+
+    This can now be retrieved directly from the repository config. It's basically the set of virtual chunk containers (or a subset at least)
+
+6. > Object level change detection (not serving a virtual chunk if the containing file changed)
+
+    We optionally store the last-modified timestamp of the object containing each chunk. This is not optimal, a hash would be better, but it takes more space. Also not optimal, storing it per chunk, when many chunks come from the same object, but we don't have other easy place to store it if they are using containers that include more than one object. We could add more configuration to optimize this in the future.
```
maxrjones, line 163: > we don't have other easy place to store it if they are using containers that include more than one object

Both the redundant prefix and redundant timestamp issue seem similar in some ways to the redundant STAC information solved by [hydration/dehydration in pgstac](https://stac-utils.github.io/pgstac/contributing/#a-note-on-hydration-and-dehydration). I'm curious if there's anything to be learned from the benefits/downsides of that approach.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+* Many chunks from a few objects, the prefix can be the file
+  * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
+  * Having between 10 and 10k chunks from the same file is not unusual
+* Virtual zarr: each chunk comes from a different object, the prefix is not the object, we still need an extra string
+  * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
+  * There could be tens of millions of chunks (so different urls)
+
+### Prefixes
+
+Storing the full URL for each chunk duplicates a lot of data in the manifest, making it large. The idea is to try to extract common _prefixes_ and identify them with a short id. Then the manifest only needs to identify the prefix id + maybe an extra string + length + offset.
```
martindurant, line 24: Does this feel ever familiar to me

maxrjones (commented):
Super excited about this work, thanks for sharing the link in the icechunk slack!

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+Intelligence is needed to optimize what is consider a good prefix. If done badly, we could have:
+
+* as many prefixes as chunks, for example by selecting `s3://some-bucket/some-prefix/c/0/0/1` in the second example above
+* too large of an extra string in each chunk, for example, if we select `s3://some-bucket` as the prefix each chunk needs to add something like `some-prefix/c/0/0`.
+
+### Should prefixes be per snapshot or global?
+
+The advantage of per snapshot is that we don't need to load/parse all the prefixes if the current snapshot uses only a few. A case for this would be a virtual dataset that has been materialized in a newer version. If prefixes are per-snapshot we wouldn't need to load any prefixes.
+
+On the other hand, requirement 4 asks for the full list of "dependencies" for the dataset, and that means for all versions, which requires a global list of prefixes. Without a global list, we would need to fetch the list of prefixes of each snapshot.
```
martindurant, line 37: Parquet implements min/max on any column for each partition, so you know from the file metadatas alone what they contain, without loading all the data.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+Using templates would be more efficient, example:
+
+* `s3://some-bucket/some-prefix/1-analysis-output-for-model-abc.nc`
+* `s3://some-bucket/some-prefix/2-analysis-output-for-model-abc.nc`
+
+If we use the template `s3://some-bucket/some-prefix/{}-analysis-output-for-model-abc.nc` each chunk could have a couple of extra characters only, filling the placeholders in the template.
+
+### Supporting multiple clouds
+
+When the user adds a virtual ref, they need to indicate the details of the platform to retrieve it. These include what cloud it is (so we know how to set up credentials), region, endpoint url, port and other settings. It looks a lot like our `S3Config` type (or an expanded version of that).
```
martindurant, line 58: Credentials and config will not be the same for any given user. I don't think you can reasonably encode them for general use. As far as S3 goes, the only fields I would think would be up for consideration would be: anonymous, endpoint and region (the latter debatable). Of these, only anonymous is relevant for gcs/azure.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+  * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
+  * There could be tens of millions of chunks (so different urls)
+
+### Prefixes
+
+Storing the full URL for each chunk duplicates a lot of data in the manifest, making it large. The idea is to try to extract common _prefixes_ and identify them with a short id. Then the manifest only needs to identify the prefix id + maybe an extra string + length + offset.
+
+In the first url style above, the prefix could be the full file, `s3://some-bucket/some-prefix/some-file.nc`. In the second example, there are options for the prefix, it could be `s3://some-bucket/some-prefix/c/0/0/1`, or `s3://some-bucket/some-prefix/c/0/0` or `s3://some-bucket/some-prefix/c/0` or `s3://some-bucket/some-prefix/c`, with different numbers of chunks under it in each case.
+
+Intelligence is needed to optimize what is consider a good prefix. If done badly, we could have:
```
rabernat, line 28: Is intelligence really needed? Or would an off-the shelf compression algorithm handle it just fine?

I decided to test this with my favorite little storage library

```python
import zarr
import numpy as np

refix = 's3://some-bucket/some-prefix/c/0/'
paths = np.array([str(i) for i in range(10000)])
full_paths = np.array([prefix + p for p in paths])

store = zarr.MemoryStore()
group = zarr.group(store=store)

path_array = group.create("paths", shape=len(paths), chunks=len(paths), dtype=str, compressor=zarr.Zstd())
path_array[:] = paths

full_path_array = group.create("full_paths", shape=len(paths), chunks=len(paths), dtype=str, compressor=zarr.Zstd())
full_path_array[:] = full_paths

print(path_array.nbytes_stored)
print(full_path_array.nbytes_stored)
# 6579
# 7048
```

So storing the full paths and then using a zstd compressor on the raw bytes creates only 7% more data than just storing the last part of the path.

This is what compression algorithms do: find redundant patterns in data. Should we really spend time trying to implement our own compression scheme for prefixes?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+* Each repo has a global list of `VirtualChunkContainers`
+
+```rust
+// this is pseudo-code
+struct VirtualChunkContainer {
+  name: String,  // something like model-output
+  platform: Platform,  // something like S3
+  url_template: String, // something like some-bucket/some-prefix/output/{}/{}.nc
+  default_template_arguments: Vec<String>,
+  cloud_client_config: Map<String, Any>,  // something like {endpoint_url: 8080}
```
martindurant, line 79: The endpoint being different for some S3 than other S3 is totally a use case.

martindurant: Consider, a kerchunked original dataset on a public AWS S3, and additional updates on my own minio.

design-docs/001-virtual-chunk-improvements.md
martindurant, line 125: but actually you are after a (non-cryptographic) hash. Relying on times seems dangerous, even if we assume everything it UTC.

design-docs/001-virtual-chunk-improvements.md
rabernat, line 37: I think we should sharpen the requirement about what exactly these dependencies are. Does the prefix have to be part of it? That feels like a bit of a separate concern. The point of the dependencies is to know what credentials will be required and where on the internet the virtual chunks might live. For these purposes, I'd suggest it's enough to know

- The platform (e.g. S3)
- The bucket name
- [Optional] The region
- [Optional] The endpoint URL

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Should prefixes be per snapshot or global?
+
+The advantage of per snapshot is that we don't need to load/parse all the prefixes if the current snapshot uses only a few. A case for this would be a virtual dataset that has been materialized in a newer version. If prefixes are per-snapshot we wouldn't need to load any prefixes.
+
+On the other hand, requirement 4 asks for the full list of "dependencies" for the dataset, and that means for all versions, which requires a global list of prefixes. Without a global list, we would need to fetch the list of prefixes of each snapshot.
+
+### Who creates/decides the prefixes
+
+Obviously, generating the list of prefixes given the list of virtual chunks is doable automatically. But it's a tricky optimization problem. One strategy could be to generate a trie with all chunk urls, and then use some heuristic to select prefixes using heavily populated nodes. It's tricky and potentially computationally expensive for large repos.
```
rabernat, line 41: This is where simply defining prefix as `{protocol}://{bucket}` might save us a lot of trouble, and then we let a standard compression scheme handle the rest.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+
+An alternative is to let the user dictate the prefixes. This way, they can optimize them knowing beforehand what virtual chunks they will add. If they don't optimize, they get large manifests without prefix extraction. We still need to satisfy requirements 2 and 4, which would require automatic extraction of at least the bucket and platform.
```
rabernat, line 43: I think it needs to be automatic.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Supporting multiple clouds
+
+When the user adds a virtual ref, they need to indicate the details of the platform to retrieve it. These include what cloud it is (so we know how to set up credentials), region, endpoint url, port and other settings. It looks a lot like our `S3Config` type (or an expanded version of that).
+
+### Supporting multiple credentials
+
+Credentials are a read time concern. When the user writes the virtual chunk, they don't necessarily know what credentials set will be used to read it.
+
+Credentials need to be specified later, when the chunks are going to be read. For example, by indicating in runtime configuration a set of credentials for each virtual chunk bucket/prefix/template.
```
rabernat, line 64: I feel we need to do more work to figure out how we will be mapping Prefixes / Dependencies / Locations / VirtualChunkContainers (whatever we end up calling them) to credential sets.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+
+1. > Virtual chunks can be stored in all supported object stores
+
+   Now the virtual chunk containers inform what is the platform, and we can create the appropriate object store client at read time. These clients are cached for performance.
+2. > Multiple credentials sets for virtual chunk resolution
+
+    Now the user initializes a Repo instance by passing, if needed, a credential set for each virtual chunk container. If a chunk is read for which there are no credentials, we default to a common set of credentials for virtual chunks, and ultimately, to the Storage credentials if on the same platform.
+3. > Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
+  
+    For large prefixes, the template + arguments creates some savings. For example, a template for virtual zarr like `some-long-bucket-name/with-some-even-longer-prefix/and-more-nesting/some-long-array-name-for-the-dataset/c/{}/{}/{}`.
```
martindurant, line 155: Again, I need a lot of convincing on this in the presence of decent compression, say Zstd.

```python
strings = "".join([f"some-long-bucket-name/with-some-even-longer-prefix/and-more-nesting/some-long-array-name-for-the-dataset/c/{i}/{i}/{i}" for i in range(10000)])
out = cramjam.zstd.compress(strings.encode())
len(strings) # 1206670
len(out)   # 19052, 1.9 bytes per entry
```

design-docs/001-virtual-chunk-improvements.md
rabernat, line 79: Honestly I think the only two items we would ever want in `cloud_client_config` are:
- endpoint_url
- region

Everything else I can think of is associated with credentials.

Would it make sense to scope it down?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+## Design
+
+### Explicit containers for virtual chunks
+
+* Each repo has a global list of `VirtualChunkContainers`
+
+```rust
+// this is pseudo-code
+struct VirtualChunkContainer {
+  name: String,  // something like model-output
```
rabernat, line 75: What is `name` used for?

design-docs/001-virtual-chunk-improvements.md
martindurant, line 163: > but it takes more space

Take as many bytes as you feel necessary, It doesn't need to be a hash of the data, but a unique value from the store. 

and these are stored *per*file*, whereas there may be one or many chunks per file.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+  platform: Platform,  // something like S3
+  url_template: String, // something like some-bucket/some-prefix/output/{}/{}.nc
+  default_template_arguments: Vec<String>,
+  cloud_client_config: Map<String, Any>,  // something like {endpoint_url: 8080}
+}
+```
+
+* Users can add to the list and edit members but not delete.
```
rabernat, line 83: But does garbage collection potentially do this?

I don't think we can meaningfully distinguish between "users" and "admins" of a repo. Currently the only way to update a repo is to make a commit. If we start making out-of-band ways to change a repo, don't we open ourselves up to consistency problems?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Global config
+
+* The list of chunk containers is stored in a new global `Config` object. There is a single configuration per repo, and all reads and writes are done reading the latest configuration. In the initial implementation, the "history" of configuration is present only for reference and user inspection by hand.
+
+TBD: how to store this information. One approach would be using the new PUT if-match S3 functionality, but this would be the first place where we overwrite objects.
```
rabernat, line 89: Can you update Config and commit to a branch at the same time? What if a branch commit succeeds but Config update fails? Or vice versa? S3 makes no promises about transactions across multiple keys...that's what we have icechunk for.

This whole concept feels like a very slippery slope of nasty edge cases.

However, I see it may be necessary if we want to expose information that transcends an individual snapshot.

Here's an alternative approach: we could store the Config _directly inside the branch or tag file as JSON_.  Obviously this does not scale to large numbers of VirtualChunkContainers. But it would have these benefits:
- Does not increase the number of GET requests to open a store
- Guaranteed consistency between commit and Config

The downside is that, to gather all of the dependencies for a repo, we have to read all of the tag and branch files (note: we never have to read the branch _history_, only ever the _latest_ branch file). This number will probably be small for most repos.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+
+```rust
+// this is pseudo-code
+struct Config {
+  inline_chunk_threshold_bytes: u16,
+  virtual_chunk_containers: Vec<VirtualChunkContainer>,
+}
```
rabernat, line 96: In general, I really like the idea of repo-level config stored with the repo. Iceberg has this and it seems really useful. Could eventually put all kinds of stuff here.

design-docs/001-virtual-chunk-improvements.md
rabernat, line 125: This feels like an unnecessary space optimization. ETag is clearly meant for this purpose. ETag has several advantages over timestamp. In particular, if the file is overwritten with _the same data_, the ETag doesn't change. Also, no timezone headaches to worry about.

At least we could make this a configuration choice and leave it up to the user.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+    pub length: ChunkLength,
+    pub last_modified: Option<u32>,
+}
+```
+
+* The `last_modified` field contains an optional unix epoch in seconds. If set for a chunk, retrieval will fail if the file was modified after this time. We use this instead of an ETag or checksum to save space on the manifests. `u32` should be enough until year ~ 2100.
+
+### Reading virtual chunks
+
+* When a user create a repository instance, they can specify credentials for each `VirtualChunkContainer`, pointing to them by name.
```
rabernat, line 129: I see, `name` is used to associate credentials with containers.

Do we have a feeling for how many `VirtualChunkContainer`s we exect a typical repo to have?

Mixing the prefix into the `VirtualChunkContainer` will complicate this quite a bit. Credentials will generally be scoped to buckets, not prefixes. So if there are many `VirtualChunkContainer`s with the same bucket, they would have to redundantly specify credentials for each one.

This problem goes away if you basically make each `VirtualChunkContainer` 1:1 with a set of credentials.

design-docs/001-virtual-chunk-improvements.md
mpiannucci: No it doesn't, eventually we can support filesystems

design-docs/001-virtual-chunk-improvements.md
mpiannucci: Runtime error when the object has changed instead of getting invalid data

rabernat (commented):
Seba this is a really thoughtful and useful document for exploring this very tricky topic. Thanks so much for taking the time to write it out.

I think the proposal you have laid out will definitely work. I also think there is some room to clarify the requirements (what are we really optimizing for?) and possibly scope them down a bit.

The crux for me is defining the actual data model we will use, and your structs are a very useful way to do that. The key question appears to be, what exactly is "place" (c.f. req. 4)? Right now, I am concerned that VirtualChunkContainer couples together two separate requirements (4 and 3 -"optimized manifests") unnecessarily.

I personally like my idea of storing the config right in the branch / tag file. 😁 

I also still think it is worth considering storing each virtual file in a manifest of its own, and then referencing this from the chunk manifest.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
+# Improvements to virtual chunks handling
+
+## What we need
+
+1. Virtual chunks can be stored in all supported object stores
+2. Multiple credentials sets for virtual chunk resolution
```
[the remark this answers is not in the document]
maxrjones: I think it would be useful to create virtual stores that would allow people to open a data tree containing both raw data and overviews. The raw data would commonly be in a separate bucket from the overviews.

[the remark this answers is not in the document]
maxrjones: > Can you elaborate? Why? Could the overview be native writes to the repo?

To be fair, I'm not sure how common this is but the primary case in which the overviews cannot be native writes to the repo is when they already exist as Zarr V2 data.

A more common scenario would be continuously adding NetCDF references as virtual chunks along with continuously adding Zarr overviews. My understanding from previous conversations was that once native data is committed to a repository, all new commits will need to be native rather than virtual. Is that incorrect? Could one continuously add virtual chunks for the full resolution data and native chunks for the overviews?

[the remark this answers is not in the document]
mpiannucci: >  Could one continuously add virtual chunks for the full resolution data and native chunks for the overviews?

Icechunk does not place restrictions on which chunks are native and which are virtual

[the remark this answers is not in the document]
rabernat: > My understanding from previous conversations was that once native data is committed to a repository, all new commits will need to be native rather than virtual. Is that incorrect

That's incorrect. You can mix native and virtual chunks however you want.

In the scenario you describe, I'd want to make the overviews native writes and only use virtual chunks for the original full-resolution data.

I understand that my proposal (only one location for virtual chunks per repo) may be constraining, but that is deliberate.

design-docs/001-virtual-chunk-improvements.md
martindurant: > overwritten with the same data, the ETag doesn't change

This is only true (on S3) with a single PUT. For multi-part-uploads, the chunks sizes need to match too and maybe something else too.

design-docs/001-virtual-chunk-improvements.md
martindurant: > Credentials will generally be scoped to buckets, not prefixes.

All backends support intra-bucket permissioning. In addition, you might have the same bucket name at different endpoints.

design-docs/001-virtual-chunk-improvements.md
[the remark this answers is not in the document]
maxrjones: Thanks for the clarification about mixing native and virtual chunks.

> I understand that my proposal (only one location for virtual chunks per repo) may be constraining, but that is deliberate.

👍  Given the early design phase, it seemed worthwhile to detail what types of use-cases would be excluded in this design.

> 2. Virtual chunks within a repo can be stored in _one single object storage bucket in a specific cloud platform_. Mixing virtual references from multiple buckets in a single repo is not allowed.

For most of the data searchable via earthaccess, there are two storage buckets listed - a public and a private bucket (e.g., `s3://lp-prod-protected/MYD14.061`, `s3://lp-prod-public/MYD14.061`. Should this scenario require two separate virtual stores?

design-docs/001-virtual-chunk-improvements.md
TomNicholas: Isn't it already supported?

design-docs/001-virtual-chunk-improvements.md
TomNicholas: Should the Runtime Error be thrown at data-access time or store/session-opening time?

rabernat: It's not practical to do at store-opening time. That would mean checking the etag / modification time of every virtual file eagerly.

design-docs/001-virtual-chunk-improvements.md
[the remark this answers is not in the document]
TomNicholas: > Icechunk does not place restrictions on which chunks are native and which are virtual

Yes you should already be able to create "mixed" icechunk stores using xarray and virtualizarr. [Aimee's `append_dim` PR](https://github.com/zarr-developers/VirtualiZarr/pull/272) goes even further because it is one way to create a mixed _array_, where you could append native chunks to an array containing virtual chunks (see https://github.com/zarr-developers/VirtualiZarr/pull/272#issuecomment-2483603659).

---

> @TomNicholas - can you think of scenarios where you would want to mix multiple different buckets in a single repo?

If you wanted to use virtual references to create some kind of aggregated dataset, e.g. put a bunch of disparate model run datasets that together comprise CMIP6 into one virtual store.

Those model runs could be in different buckets (or even clouds...) from different data providers. This kind of scenario is a little more likely to occur across groups in a store rather than within a single group/array IMO - think a `DataTree` where each group came from a different source. I think its debatable whether that use case should be targeted by some kind of higher-level catalog instead though.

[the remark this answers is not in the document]
dcherian: yes there is an extremely thin line between "catalog" and "manifest" with arbitrary trees.

design-docs/001-virtual-chunk-improvements.md
paraseba: Yeah, we have support for filesystem meant mostly for tests. We won't drop it.

design-docs/001-virtual-chunk-improvements.md
paraseba: I changed the description to make it more clear.

design-docs/001-virtual-chunk-improvements.md
paraseba: I don't expect any performance impact on reads. The object store will retrieve the object only if the "didn't change" conditions are met without any observable latency difference. On writes, it's a different story, we need to discuss with @TomNicholas if the new proposal is something we can implement in VirtualiZarr and what the impact could be.

In any case, I'm making the checksum optional at the chunk level.

design-docs/001-virtual-chunk-improvements.md
paraseba: I dropped the requirement to do this optimization at this time. I think we can make it comfortable to 1.0 without it. In any case, this was an in-memory space optimization that we can add later.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
...
+### Chunks url styles
+
+* Many chunks from a few objects, the prefix can be the file
+  * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
+  * Having between 10 and 10k chunks from the same file is not unusual
+* Virtual zarr: each chunk comes from a different object, the prefix is not the object, we still need an extra string
+  * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
+  * There could be tens of millions of chunks (so different urls)
+
+### Prefixes
```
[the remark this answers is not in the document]
paraseba: We are currently not optimizing the on-disk format. This was meant as an in-memory optimization, Icechunk keeps full manifests in memory. But in my latest commit, I'm dropping this requirement for now, and we'll evaluate better ways to optimize the in memory datastructure independently of the on-disk format.

design-docs/001-virtual-chunk-improvements.md
paraseba: Absolutely, I think things are more clear in my latest commit. Credentials are an "Instance construction" concern, not a write time concern. Different users will instantiate the repository with different credentials sets. We just need a way to tell them what they need to provide, and a way for them to provide it.

design-docs/001-virtual-chunk-improvements.md
paraseba: Yes, you can have different containers, with different endpoints.

design-docs/001-virtual-chunk-improvements.md
paraseba: Its debatable I think. A few thoughts:
* We did hashes in Arraylake and they didn't get a lot of use, they are expensive to compute for example.
* Timezones are not a concern here, objects stores have a well defined and standardized format for their modified time headers.
* Doing last_modified is very cheap and easy at write time, for example, if you pass `now()` it will work most of the time. Etag you generally need to retrieve from the object store for each object. Imagine virtualizing a zarr store, you need to fetch the etag for every chunk.
* Last modified takes less space

In any case, this is very useful feedback. In my last commit I made it possible to provide timestamp, etag, or nothing.

design-docs/001-virtual-chunk-improvements.md
paraseba: Remember this was an optimization for the in-memory datastructure. Coupling it with the on-disk format was a mistake (as a consequence of Icechunk still not having a separation layer there). But even so, this is not the right way to optimize this. I dropped the requirement, we can absolutely do this optimization later. Thanks for the feedback on this, it changed my mind.

design-docs/001-virtual-chunk-improvements.md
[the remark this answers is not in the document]
paraseba: I think things look much better in the latest commit. Please take a look and tell me what you think. The solution seems much simpler, and still allows all kind of mixes of object stores, prefixes, credentials, etc.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,163 @@
+# Improvements to virtual chunks handling
+
+## What we need
+
+1. Virtual chunks can be stored in all supported object stores
+2. Multiple credentials sets for virtual chunk resolution
+3. Optimized manifests where the same prefix doesn't need to be repeated for every virtual chunk
```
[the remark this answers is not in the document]
paraseba: Good feedback, I was too close to the problem and failed to add this context. This requirement has been dropped for now, it's an in-mem optimization that we can do at a later time, if needed.

design-docs/001-virtual-chunk-improvements.md
paraseba: This is my failure to clarify that we were looking for an in-memory optimization. Currently Icechunk lacks a layer of separation between the in memory and on-disk datastructures (that will have to come soon) which made things even more confusing.

design-docs/001-virtual-chunk-improvements.md
paraseba: I think this is much more clear in my latest commit.

design-docs/001-virtual-chunk-improvements.md
paraseba: In principle it cannot be fully automated, it's not clear from a url in what platform, region, endpoint, etc a chunk lives. But I think the latest commit makes a much better job at making this easier. Configuration is optional and not needed in most normal cases.

design-docs/001-virtual-chunk-improvements.md
paraseba: I think `anonymous` will still be needed, but yes, take this as pseudocode, we'll figure out the exact details during coding.

design-docs/001-virtual-chunk-improvements.md
paraseba: I saw you figured it out. In my later commit this still survives, the name is a "handle" so users can link credentials sets to something. So they will say something like "for the `model` container use these credentials, for the `inputs` container these others"

design-docs/001-virtual-chunk-improvements.md
paraseba: In the latest commit, I'm allowing last_modified and etag (or nothing). What I *REALLY* like about `last_modified` is that you can "invent" it, you don't really need to fetch every etag for every object from the store. If you just pass `now` as last_modified, it will work in most cases.

design-docs/001-virtual-chunk-improvements.md
paraseba: Right, allowing more than one set of credentials for different paths of a bucket was part of the design (unfortunately as an unlisted requirement now that I think about it). It's also possible in my new commit.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,259 @@
...
+`checksum` can be set to the in-object-store ETag of the file containing the virtual chunk, or to its last modified time. In the case of last modified time, the value
+doesn't need to be precise. The system will only fail to provide the chunk if the object's last-modified-at is larger than the passed value. So, passing `checksum = LastModified(datetime.now())` will usually be enough.
+
+This new argument will help us satisfy requirement 4.
+
+The interpretation of the `location` argument also changes. `location` must be an absolute url that includes a protocol. The protocol can be:
+
+* s3://
+* gcs://
+* tigris://  (is this the right protocol)
```
rabernat, line 126: No. it's just S3 with a custom endpoint URL.

paraseba: There is also credentials... We may need to know it's tigris so we manage the generation of credentials differently from say localhost minio. Of course, we can always play games with the endpoint and figure out, but this seems more future-proof?

martindurant: > we manage the generation of credentials differently from say localhost minio

I recommend you keep to using ("s3") for everything that indeed matches that protocol, and the rest is in the configuration for the individual "containers" - isn't that the whole point? I don't think you want to maintain a potentially endless list of protocols for each and service that might have some icy data.

> generation of credentials

What does this mean?

By the way, how exactly are chunk URLs mapped to "containers", i.e., how do you know which container a particular URL belongs to?

paraseba: One downside of using `s3://` everywhere is that it forces users to actually create the VirtualChunkContainers, even for some very common object stores like Google's. Using the protocol would be a bit easier on them for the very common case of a simple repo on google or tigris. But I take your point, which ones are s3:// and which ones are a protocol is arbitrary. Will think more about it.

paraseba: >  how exactly are chunk URLs mapped to "containers", 

@martindurant a rough description of the algorithm is documented [here](https://github.com/earth-mover/icechunk/pull/436/files#diff-c698aed6bffbd27180af5e8495caefc5b69f58680485b27b1dbbbdbb0788bcf0R141)

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Reading virtual chunks
+
+When a virtual chunk is requested, Icechunk will find in the `location` field of the stored reference, an absolute URL with a specific protocol. It will need to map this location to a given object store, bucket, path and set of credentials.
+
+The mapping is done by comparing the protocol and path to the set of configured virtual chunk containers, plus a set of default virtual chunk containers.
+
+This is how the mapping from location to cloud+bucket+prefix+credentials works:
+
+* Using the `location` protocol we find all containers that match the protocol
```
martindurant, line 143: I don't know what this means

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,259 @@
...
+* We sort the matching containers in order of descending prefix length
+* We find the first one that has a `bucket_prefix` that is a prefix of the `location` path (this will be the most specific container that matches)
```
martindurant, line 145: What do buckets have anything to do with here?
Prefix is a string of any length, no?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,259 @@
...
+* If no containers match, the chunk request cannot be fulfilled and we fail with an error message indicating the missing config.
+* We use the rest of the data in the matched container to instantiate an object store client (or retrieve it from cache)
+* Credentials for the client must be passed by the user when the repository is instantiated, by indicating the container name to which they apply
+
+We have one default container for each supported object store, the container has protocol and name s3, gcs, tigris, etc. and has an empty `bucket_prefix` and no region configured. These default containers, just like any others set in the persistent repo configuration, can be overridden when the repository is instantiated.
```
martindurant, line 150: So, to be precise, you are envisioning a structure like:
```json
[{"name": "acontainer",
   "prefix": "gcs://bucket/path/prefix",
   "options": {"region": "US-EAST1"}},
 {...}
]
```
where the "name" handle is what the user passes to change the default?

and "prefix" is literal, it matches any paths that start with that string.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,259 @@
...
+
+When a region is not indicated and the platform needs it, the logic is:
+
+* For the platform in which the Icechunk repo is stored, we use the same region as the repo
+* For all other platforms we use the default region for that platform
```
martindurant, line 155: Oh, so you ALSO envision default options _per_protocol_ ? So you have some sort of options merging, from less specific to more specific? It will take care to allow the user to override at any level.

martindurant (commented):
Some comments on "container", which I think had better be renamed "prefixes", since hat;'s exactly what they are.

design-docs/001-virtual-chunk-improvements.md
martindurant: > One downside of using s3:// everywhere is that it forces users to actually create the VirtualChunkContainers, even for some very common object stores like Google's. 

?? google's storage does not use the s3 protocol. When I say "protocol", mean exactly that: a set of rules and conventions for how to communicate. S3's is [defined by AWS](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_Simple_Storage_Service.html) but implemented by many services. GCS is *only* used by google (as far as I know).

> a rough description of the algorithm

made some additional comments

design-docs/001-virtual-chunk-improvements.md
paraseba: I rewrote the line.

design-docs/001-virtual-chunk-improvements.md
paraseba: The chunk's URL is `s3://foo/bar/baz.nc`, we need to find a container that matches it to know what set of credentials and object store client configuration to use. If there are two containers that work for protocol s3, the one with `bucket_prefix = foo/bar` will match, instead of the one with `bucket_prefix = hello` or the one with `bucket_prefix = foo`.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -138,9 +138,9 @@ When a virtual chunk is requested, Icechunk will find in the `location` field of
 
 The mapping is done by comparing the protocol and path to the set of configured virtual chunk containers, plus a set of default virtual chunk containers.
 
-This is how the mapping from location to cloud+bucket+prefix+credentials works:
+This is how the mapping from a chunk location (url) to cloud+bucket+prefix+credentials works:
 
-* Using the `location` protocol we find all containers that match the protocol
+* Using the chunk location's protocol, we find all containers that have a matching `protocol` field
```
martindurant, line 143: It's just the first few characters of the prefix - why separate it out from the prefix in general? Sometimes the prefix might be "s3", sometimes "s3://bucket", sometimes "s3://bucket/path/to". Aside from knowing that these particular examples are all s3 protocol, I don't think that there's any difference in how you should treat them.

design-docs/001-virtual-chunk-improvements.md
martindurant: "We find the container with the longest prefix matching the given URL"

design-docs/001-virtual-chunk-improvements.md
paraseba: I like this. My remaining concern is, how to make it simple for the simple use case. Say, virtual chunks in google storage, repository in AWS. I don't want users to _have to_ create a container for this simple case. That's why I'm proposing to distinguish the platform by the url protocol (by default).

martindurant: Then you have two:
- "s3"
- "gs"
(and they happen to exactly look like what you would expect as protocols).

You could even have the default default be "", i.e., matches any prefix that's not matched by something more specific.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+1. Virtual chunks can be stored in all supported object stores. The same repo can have virtual chunks hosted in different object stores.
+2. Multiple credentials sets for virtual chunk resolution, accessing some chunks may require a different set than others.
+3. A mechanism to know all the "dependencies" of a repo. That is, all the "places" where its virtual chunks are hosted. This is important for users to know if they can delete old data, or if there are virtual datasets that depend on it.
+4. If the file hosting a virtual chunk changed after the chunk ref was written, we want to fail with a runtime error instead of serving corrupted data. This is because offset and length are usually not maintained during updates.
+
+## Discussion
+
+### Current virtual chunk URL styles
+
+* The open source stack seating on top of Icechunk (VirtualiZarr, etc.) uses URLs to identify virtual chunk locations.
```
rabernat, line 14:
```suggestion
* Existing packages for reference generation (VirtualiZarr, Kerchunk, etc.) uses URLs to identify virtual chunk locations.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+* One common case is many virtual chunks from a few objects.
+  * Example: `s3://some-bucket/some-prefix/some-file.nc` is pointed to by 10k chunks.
```
rabernat, line 16:
```suggestion
  * Example: `s3://some-bucket/some-prefix/some-file.nc` is referenced by 10k chunks.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+  * Having between 10 and 10k chunks from the same object is not unusual
+* Other common use case for virtual zarr datasets: each chunk comes from a different object. This can happen, for example, with a virtual zarr dataset.
```
rabernat, line 18:
```suggestion
* The other common use case for virtual zarr datasets: each chunk comes from a different object. This can happen, for example, with a virtual zarr dataset.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+  * Example: `s3://some-bucket/some-prefix/c/0/0/1` is pointed to by only 1 chunk. But there are many objects like it.
+  * There could be tens of millions of chunks (so tens of millions of different urls)
+
+### Supporting multiple virtual chunk locations
+
+Because of requirement 1, the same repo can contain virtual chunks in multiple buckets, object stores, and platforms, potentially even buckets with the same name in different platforms.
+
+When the user adds a virtual ref, they will need to indicate the details of the platform for Icechunk to be able to retrieve it. These includes what cloud it is, region, endpoint url, etc. This means, the current approach of pointing to a chunk with a simple URL is not going to be enough. Not without encoding several new attributes in the URL, which seems hacky. A URL such as `s3://foo/bar.nc`, in itself doesn't have enough details to reach the object. Even if the protocol part of that URL is identified with the wire protocol needed to talk to the object store, there are missing details such as the region, the endpoint, or the authentication mechanism.
```
rabernat, line 26:
```suggestion
When the user adds a virtual ref, they will need to specify all of these details for Icechunk to be able to retrieve it. These includes what cloud it is, region, endpoint url, etc. This means, the current approach of pointing to a chunk with a simple URL is not going to be enough. Not without encoding several new attributes in the URL, which seems hacky. A URL such as `s3://foo/bar.nc`, in itself doesn't have enough details to reach the object. Even if the protocol part of that URL is identified with the wire protocol needed to talk to the object store, there are missing details such as the region, the endpoint, or the authentication mechanism.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+```python
+def set_virtual_ref(
+    self, key: str, location: str, offset: int, length: int
+) -> None: ...
+```
+
+### Credentials
+
+Credentials are a read time concern. When the user writes the virtual chunk, they don't necessarily know what credentials set will be used to read it.
```
rabernat, line 38:
```suggestion
Credentials are a read-time concern. When the user writes the virtual chunk, they don't necessarily know what credentials will be used to read it.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+Because of requirements 1 and 2, when users open a repo they need to know what sets of credentials they have to provide to be able to access all the chunks. This is not immediately obvious, Icechunk should provide a way to introspect a repo and inform the user what credentials they need. Then the user can setup the Repo instance appropriately with the credentials.
```
rabernat, line 40:
```suggestion
Because of requirements 1 and 2, when users open a repo they need to know what sets of credentials they have to provide to be able to access all the chunks. Icechunk should provide a way to introspect a repo and inform the user what credentials they need. Then the user can set up the Repo instance appropriately with the credentials.

Icechunk itself should never store credentials in any of its metadata files.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+### Differences between object stores
+
+Many object stores support a subset of the S3 wire protocol to read and write objects. This seems to simplify the task, because Icechunk could talk to all those platforms with the same "client code". But there are many details that make this theoretical approach not work very well in practice.
+
+* The level of conformity with the protocol varies widely. Icechunk doesn't use very advanced features of the S3 protocol but this could change in the future. Even today, Icechunk already uses, for example, conditional requests (both writes and reads).
+* The ideal client settings for different platforms is different. Some object stores allow more aggressive retries than others, or more concurrency. In general, performance optimization needs to be targeted to a specific platform sometimes.
```
rabernat, line 47:
```suggestion
* The ideal client settings for different platforms may vary. Some object stores allow more aggressive retries than others, or more concurrency. In general, performance optimization needs to be targeted to a specific platform sometimes.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+* Encryption and checksumming capabilities changes widely too between platforms.
+* Error responses are not uniform, some tweaking is needed sometimes to extract the most relevant information.
+* Authentication protocols, and credential refresh varies widely.
+
+The more Icechunk knows about the specifics of the object store platform the better job it can do at fetching chunks. Since all the information for the virtual chunks is stored in the dataset at write time, it's important we store enough, to future proof for changes and improvements to the different platforms.
+
+All that said, good defaults and usability are very important. The ideal situation is: virtual chunks that work by default, and can be made as efficient as possible with a small amount of work.
+
```
rabernat, line 55: The section above is a fantastic motivation. Great work! 🙌

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+## Design
+
+### Checksum
+
+In order to support requirement 4, we add an optional `checksum` argument to `set_virtual_ref`.
```
rabernat, line 60:
```suggestion
In order to support requirement 4, we add an optional `verification` argument to `set_virtual_ref`.
```

I'd prefer to not call this `checksum` since in reality it is more general than that.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+If no `checksum` is provided, Icechunk will always attempt to retrieve the chunk, which is a potentially dangerous operation because there is no way to validate correctness.
+
+### Virtual chunk containers
+
+As discussed above, to correctly support requirements 1 and 2, we need to store information about the object store platform in which the chunk owner object is stored. Icechunk needs to link each virtual chunk ref to a platform, potentially a region, potentially an endpoint, etc. In contraposition, we want to maintain the rest of the open source stack as simple and unchanged as possible. The stack "speaks" URLs for virtual chunks, we maintain that, but we reinterpret what those URLs mean as explained in the following sections.
+
+We first introduce a new datastructure that maintain information about the different object stores used by the dataset, the `VirtualChunkContainer`:
```
rabernat, line 92:
```suggestion
We first introduce a new data structure that maintain information about the different object stores used by the dataset, the `VirtualChunkContainer`:
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+    pub anonymous: bool,
+    pub allow_http: bool,
+}
+```
+
+* `name`: is used exclusively for user interaction and it's not part of the chunk resolution algorithm, it gives a human readable way to refer to the container. Names must be unique per repo.
+* `object_store`: it gives Icechunk information on how to talk to the object store. Using this field Icechunk will be able to:
+  * Determine what protocol to use to fetch chunks from to the object store.
+  * Configure and optimize the client for the object store
+  * Provide defaults for other fields such as the `endpoin_url`.
```
rabernat, line 124:
```suggestion
  * Provide defaults for other fields such as the `endpoint_url`.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+* `prefix`: it will be used to find a matching container for a virtual chunk ref (see below). Must be unique per repo.
```
rabernat, line 125:
```suggestion
* `prefix`: used to find a matching container for a virtual chunk ref (see below). Must be unique per repo.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+The following examples will work for `set_virtual_ref` called with the given `location` and `validate_containers=True`:
+
+* `location=s3://foo/bar.nc` if there is a `VirtualChunkContainer` with `prefix=s3`.
+* `location=s3://foo/bar.nc` if there is a `VirtualChunkContainer` with `prefix=s3://`.
+* `location=s3://foo/bar.nc` if there is a `VirtualChunkContainer` with `prefix=s3://foo`.
+* `location=s3://foo/bar.nc` if there is a `VirtualChunkContainer` with `prefix=s3://foo/bar.nc`.
+* `location=tigris://foo/bar.nc` if there is a `VirtualChunkContainer` with `prefix=tigris`.
+* `location=models://foo/bar.nc` if there is a `VirtualChunkContainer` with `prefix=models`.
+
+The following examples will now work for `set_virtual_ref` called with the given `location` and `validate_containers=True`:
```
rabernat, line 167:
```suggestion
The following examples will not work for `set_virtual_ref` called with the given `location` and `validate_containers=True`:
```

Important change!

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+* `location=s3://foo/bar.nc` if there are no `VirtualChunkContainers`
+* `location=gcs://foo/bar.nc` if the only `VirtualChunkContainers` has `prefix=s3`
+
+But calling `set_virtual_ref` in the two last examples will still work with `validate_containers=False`.
```
rabernat, line 172: I'm not sure I understand the use case for `validate_containers=False`. When would this behavior be desirable?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+pub struct VirtualChunkRef {
+    pub location: VirtualChunkLocation,
+    pub offset: ChunkOffset,
+    pub length: ChunkLength,
+    pub checksum: Option<Checksum>,
+}
+```
+
+Using a `u32` for `SecondsSinceEpoch` gives us at least until year 2100.
```
rabernat, line 215: I'm imagining a future Y2.1K crisis where we regret this decision and Yottabytes of interplanetary data suddenly become invalid! 😆

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+* It's cheap compared to pulling a chunk
+* It gives flexibility on when the containers are created (for example after the writes by using `validate_containers=False`)
+* It allows runtime changes to the containers (for example pointing the virtual chunks to a different place during tests)
+
+`VirtualChunkContainer` is a `String` and not a `Url` to future proof the format. Some day we may want to do something different for other types of virtual refs, or to support new open source libraries.
+
+### Reading virtual chunks
+
+When a virtual chunk is requested we need to map the `VirtualChunkLocation` URL to one and only one `VirtualChunkContainer`. Then, from that container, we can instantiate an object store client, potentially with some user provided credentials.
```
rabernat, line 227:
```suggestion
When a virtual chunk is requested we need to map the `VirtualChunkLocation` URL to one and only one `VirtualChunkContainer`. Then, from that container, we can instantiate an object store client, potentially with some user-provided credentials.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+* Find the first container in the sorted list for which its `prefix` field is a prefix of the `VirtualChunkLocation` string.
+* If no containers satisfy the condition the fetch virtual chunk operation fails indicating the missing configuration.
+
+These algorithm ensures that we find the single most specific container that matches the chunk location. This allows for very flexible matching, for example, three containers with prefixes `models` and `models/dev` `models/prod` can coexist and be hosted in different object store platforms.
+
+Once the matching `VirtualChunkContainer` is found, its fields are used to build an object store client. Some smart defaults can be provided, for example:
+
+* Each `ObjectStorePlatform` has a default `endpoint_url`.
+* If the `ObjectStorePlatform` is the same as the repo is hosted on, we can use the same `region`.
+* `ObjectStorePlatform` have defaults regions, or may not need them.
```
rabernat, line 241:
```suggestion
* `ObjectStorePlatform` has defaults regions, or may not need them.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+### Default `VirtualChunkContainers`
+
+To optimize for usability for the most common use cases, a set of default `VirtualChunkContainers` is created in every repo. There will be default containers for the common supported platforms (s3, gcs, azure, tigris, etc.). The names and prefixes of these containers will be (`s3`, `gcs`, `azure`, `tigris`, etc). They will include no `region`, which in practice means they will use the same region as the repo or the platform default region.
```
rabernat, line 245: Are these defaults actually written to the config file? Or are they "built in"?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+
+```rust
+pub enum ObjectStoreCredentials {
+    FromEnv,
+    Anonymous,
+    Static {
+        access_key_id: String,
+        secret_access_key: String,
+        session_token: Option<String>,
+    },
```
rabernat, line 298: These are S3-style credentials. We'll need different structs for S3, GCS, and Azure protocols. This spec gives a good enumeration: https://github.com/unitycatalog/unitycatalog/blob/main/api/Models/TemporaryCredentials.md

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+An we will have to add at least one more variant once we implement credentials refresh.
+
+### Use case examples
+
+#### Basic usage
+
+All virtual chunks point to S3 us-east-1, single credential set.
+
+* When calling `set_virtual_ref` locations are passed with `s3://` protocol.
+* No persistent configuration needs to be set in the repo.
```
rabernat, line 311:
```suggestion
* No persistent configuration needs to be set in the repo. Default configuration is enough.
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+* At runtime, the repo instance can be configured passing credentials for container "s3", the default container.
+* If the repo data is also in s3 us-east-1, and the same credentials can be used, the previous point is not needed, the default will be using the same credentials as the ones for the repo.
```
rabernat, line 313: I don't understand why reusing the credentials requires it to be in the same region. This feels arbitrary.

Credentials are not scoped to regions or even buckets.

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+* When calling `set_virtual_ref` locations are passed with `gcs://` protocol.
+* A persistent configuration can be set creating a virtual chunk container with name gcs and region "africa-south1"
+* Credentials must be set at runtime for the custom protocol
+
+#### Go to local MinIO for testing
+
+All virtual chunks point to S3 us-east-1, for testing purposes the chunks for an array in `s3://bucket/prefix/array` are also copied locally and served with MinIO on port 8080.
+
+* When calling `set_virtual_ref` locations are passed with `s3://` protocol.
+* The icechunk repo instance is initialized with a configuration that creates a container with `name=minio-override`, `prefix=s3://bucket/prefix/array` and `endpoint_url=localhost:9000`
```
rabernat, line 338:
```suggestion
* The icechunk repo instance is initialized with a configuration that creates a container with `name=minio-override`, `prefix=s3://bucket/prefix/array` and `endpoint_url=localhost:8080`
```

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -0,0 +1,381 @@
...
+Ichecunk repo stored in S3 bucket `repo-bucket`. Virtual array `A` on Google cloud, bucket `virtual-a`. Virtual array `B` on Google cloud, bucket `virtual-b`. Virtual array `C` on s3 bucket `virtual-c`. All buckets have different credential sets.
+
+* When ingesting `virtual-a` VirtualiZarr is instructed to use protocol `modela`
+* When ingesting `virtual-b` VirtualiZarr is instructed to use protocol `modelb`
+* When ingesting `virtual-c` VirtualiZarr is instructed to use protocol `modelc`
+* Persistent config is set up in the repo to create the following virtual chunk containers:
+  * `modela`: on Google storage platform with prefix `virtual-a`
+  * `modelb`: on Google storage platform with prefix `virtual-b`
+  * `modelc`: on S3 platform with prefix `virtual-c`
+  * These containers can be set before or after the ingest (we recommend before and to pass `validate_containers=True`)
```
rabernat, line 351: Can you give an example of what the URLs look like in this scenario?

rabernat (approved):
Great work here!

Many minor questions / comments, and a few typos.

design-docs/001-virtual-chunk-improvements.md
paraseba: 9000 is MinIO default port, I'll fix the place where I used 8080 for the example.

design-docs/001-virtual-chunk-improvements.md
TomNicholas: I mean that's only 75 years, and GRIB is [already almost 40 years old](https://www.weather.gov/tg/histgate) and not going away anytime soon...

Issue #445, opened 2024-12-05
Add test suites from other projects


paraseba:
Arraylake + Zarr3 + Xarray test suites (probably create subtasks for these)

- [x] Zarr stateful tests (#437, #466)
- [x] Zarr property tests
- [x] Xarray backend tests (#423)
- [x] AL stateful tests (in progress, #453)
- [ ] AL `test_stateful_history_rewrite`

PR #453, opened 2024-12-06
Add stateful testing for version control ops

dcherian:
xref #445 

This ports `test_stateful_repo_ops.py` from Arraylake to Icechunk. It's really nice how many special cases I could delete!

This needs some upstream pushes to Zarr, so I'll leave it in draft for a while. It passes locally.

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -21,7 +21,6 @@ from hypothesis.stateful import (
     Bundle,
     RuleBasedStateMachine,
     Settings,
-    consumes,
     initialize,
     invariant,
     precondition,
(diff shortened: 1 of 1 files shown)


icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -0,0 +1,375 @@
...
+                self.sync_store.get(k, default_buffer_prototype()).to_bytes()
+            )
+            # FIXME: zarr omits this if None?
+            if "dimension_names" not in expected:
+                actual.pop("dimension_names")
+            actual_fv = actual.pop("fill_value")
+            expected_fv = expected.pop("fill_value")
+            if actual_fv != expected_fv:
+                # TODO: is this right? we are losing accuracy in serialization
+                np.testing.assert_allclose(actual_fv, expected_fv)
```
dcherian, line 348: This is the only issue I found,

```python
>           assert actual == expected
E           AssertionError: assert {'attributes'...bytes'}], ...} == {'attributes'...bytes'}], ...}
E
E             Omitting 9 identical items, use -vv to show
E             Differing items:
E             {'fill_value': 5.960464477539064e-08} != {'fill_value': 5.960464477539063e-08}
```

we are losing precision somewhere

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -0,0 +1,375 @@
...
+
+        for k in model_list:
+            # need to load to dict to compare since ordering of entries might differ
+            expected = json.loads(self.model[k].to_bytes())
+            actual = json.loads(
+                self.sync_store.get(k, default_buffer_prototype()).to_bytes()
+            )
+            # FIXME: zarr omits this if None?
+            if "dimension_names" not in expected:
+                actual.pop("dimension_names")
```
dcherian, line 343: wtf zarr

paraseba (approved):

PR #454, opened 2024-12-06
Bump to Zarr3b3

dcherian:


paraseba (approved):

icechunk-python/tests/test_zarr/test_array.py
```diff
@@ -122,8 +122,8 @@ def test_array_name_properties_no_group(
         store=store, shape=(100,), chunks=(10,), zarr_format=zarr_format, dtype="i4"
     )
     assert arr.path == ""
-    assert arr.name is None
-    assert arr.basename is None
```
dcherian, line 126: We should delete some of these tests, they don't really exercise icechunk and duplicate the Zarr property tests.

PR #457, opened 2024-12-06
Better deletion tracking in ChangeSet

dcherian:
Now track both Path and NodeId rather than just Path. This helps us handle tricky cases where a committed node is updated, deleted, and created again in the same session.

### DIFF
--- a/icechunk-python/tests/test_store.py
+++ b/icechunk-python/tests/test_store.py
@@ -1,0 +1,15 @@
+import icechunk
+import zarr
+
+
+async def test_store_clear() -> None:
+    store = icechunk.IcechunkStore.create(
+        storage=icechunk.StorageConfig.memory("test"),
+        config=icechunk.StoreConfig(inline_chunk_threshold_bytes=1),
+    )
+
+    zarr.group(store=store)
+    store.commit("created node /")
+    await store.clear()
+    zarr.group(store=store)
+    assert len([_ async for _ in store.list_prefix("/")]) == 1
--- a/icechunk/src/change_set.rs
+++ b/icechunk/src/change_set.rs
@@ -27,8 +27,8 @@ pub struct ChangeSet {
     updated_attributes: HashMap<NodeId, Option<UserAttributes>>,
     // FIXME: issue with too many inline chunks kept in mem
     set_chunks: HashMap<NodeId, HashMap<ChunkIndices, Option<ChunkPayload>>>,
-    deleted_groups: HashSet<Path>,
-    deleted_arrays: HashSet<Path>,
+    deleted_groups: HashSet<(Path, NodeId)>,
+    deleted_arrays: HashSet<(Path, NodeId)>,
 }
 
 impl ChangeSet {
@@ -36,11 +36,11 @@ impl ChangeSet {
         self.updated_arrays.keys()
     }
 
-    pub fn deleted_arrays(&self) -> impl Iterator<Item = &Path> {
+    pub fn deleted_arrays(&self) -> impl Iterator<Item = &(Path, NodeId)> {
         self.deleted_arrays.iter()
     }
 
-    pub fn deleted_groups(&self) -> impl Iterator<Item = &Path> {
+    pub fn deleted_groups(&self) -> impl Iterator<Item = &(Path, NodeId)> {
         self.deleted_groups.iter()
     }
 
@@ -64,11 +64,6 @@ impl ChangeSet {
     }
 
     pub fn add_group(&mut self, path: Path, node_id: NodeId) {
-        // when overwriting a group that exists in the base snapshot,
-        // zarr will delete the group first, and then add a new group
-        // get_group prioritizes `deleted_groups` over `new_groups`,
-        // so we must remove from `deleted_groups` here
-        self.deleted_groups.remove(&path);
         self.new_groups.insert(path, node_id);
     }
 
@@ -91,7 +86,7 @@ impl ChangeSet {
             }
             None => {
                 // it's an old group, we need to flag it as deleted
-                self.deleted_groups.insert(path);
+                self.deleted_groups.insert((path, node_id.clone()));
             }
         }
     }
@@ -126,11 +121,6 @@ impl ChangeSet {
         node_id: NodeId,
         metadata: ZarrArrayMetadata,
     ) {
-        // when overwriting a array that exists in the base snapshot,
-        // zarr will delete the array first, and then add a new group
-        // get_group prioritizes `deleted_arrays` over `new_arrays`,
-        // so we must remove from `deleted_arrays` here
-        self.deleted_arrays.remove(&path);
         self.new_arrays.insert(path, (node_id, metadata));
     }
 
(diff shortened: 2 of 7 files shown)


icechunk/src/change_set.rs
```diff
@@ -151,14 +141,24 @@ impl ChangeSet {
...
-        self.deleted_groups.contains(path)
-            || self.deleted_arrays.contains(path)
-            || path.ancestors().skip(1).any(|parent| self.is_deleted(&parent))
+    pub fn is_deleted(&self, path: &Path, node_id: &NodeId) -> bool {
+        dbg!(format!("checking is_deleted for {0} & {1}", path, node_id));
+        let key = (path.clone(), node_id.clone());
+        self.deleted_groups.contains(&key)
+            || self.deleted_arrays.contains(&key)
+            || path.ancestors().skip(1).any(|parent| {
+                self.get_group(&parent).is_some_and(|parent_node_id| {
```
dcherian, line 154: This is wrong, and an existing test fails. 

We are in `ChangeSet` and somehow need to know the `NodeId` for a node that is in the snapshot. How do I get that?

icechunk-python/tests/test_store.py
```diff
@@ -0,0 +1,15 @@
+import icechunk
+import zarr
+
+
+async def test_store_clear() -> None:
```
dcherian, line 5: hooray hypothesis!

icechunk/src/change_set.rs
paraseba: :thinking:  ... I think the problem is  with the `get_group` call right? That's what doesn't work... but it feels this is fixable? We need to know if a given path was deleted _in this changeset_, we shouldn't need the repo for that.

dcherian: We need to know if a *node* was deleted, not just the path (that's the issue :) )

icechunk/src/conflicts/basic_solver.rs
```diff
@@ -90,10 +90,10 @@ impl BasicConflictSolver {
                     ChunkDoubleUpdate{..} if self.on_chunk_conflict == VersionSelection::Fail
                 ) ||
                 matches!(conflict,
-                    DeleteOfUpdatedArray(_) if self.fail_on_delete_of_updated_array
+                         DeleteOfUpdatedArray{..} if self.fail_on_delete_of_updated_array
```
dcherian, line 93:
```suggestion
                     DeleteOfUpdatedArray{..} if self.fail_on_delete_of_updated_array
```

icechunk/src/conflicts/detector.rs
```diff
@@ -169,7 +169,8 @@ impl ConflictSolver for ConflictDetector {
         let deletes_of_updated_arrays = stream::iter(
             current_changes.deleted_arrays().map(Ok),
         )
-        .try_filter_map(|path| async {
+        .try_filter_map(|(path, _node_id)| async {
+            // FIXME: reuse node_id?
```
dcherian, line 173:
```suggestion
```

icechunk/src/conflicts/basic_solver.rs
```diff
@@ -90,10 +90,10 @@ impl BasicConflictSolver {
                     ChunkDoubleUpdate{..} if self.on_chunk_conflict == VersionSelection::Fail
                 ) ||
                 matches!(conflict,
-                    DeleteOfUpdatedArray(_) if self.fail_on_delete_of_updated_array
+                     DeleteOfUpdatedArray{..} if self.fail_on_delete_of_updated_array
```
dcherian, line 93:
```suggestion
                    DeleteOfUpdatedArray{..} if self.fail_on_delete_of_updated_array
```

icechunk-python/tests/test_store.py
paraseba: hooray!

icechunk/src/change_set.rs
```diff
@@ -151,14 +141,13 @@ impl ChangeSet {
...
+            self.deleted_arrays.insert((path, node_id.clone()));
         }
     }
 
-    pub fn is_deleted(&self, path: &Path) -> bool {
-        self.deleted_groups.contains(path)
-            || self.deleted_arrays.contains(path)
-            || path.ancestors().skip(1).any(|parent| self.is_deleted(&parent))
+    pub fn is_deleted(&self, path: &Path, node_id: &NodeId) -> bool {
+        let key = (path.clone(), node_id.clone());
```
paraseba, line 149: I hate that Rust forces you to clone here.

icechunk/src/repository.rs
```diff
@@ -355,8 +355,20 @@ impl Repository {
     /// Deletes of non existing groups will succeed.
     pub async fn delete_group(&mut self, path: Path) -> RepositoryResult<()> {
         match self.get_group(&path).await {
-            Ok(node) => {
-                self.change_set.delete_group(node.path, &node.id);
+            Ok(parent) => {
+                let nodes_iter: Vec<NodeSnapshot> = self.list_nodes().await?.collect();
```
paraseba, line 359: How about calling `filter` on the result of `list_nodes` passing `is_prefix_match`. That way, you can construct a much smaller `Vec`

icechunk/src/repository.rs
```diff
@@ -597,15 +609,8 @@ impl Repository {
...
-            self.list_nodes().await?.map(|node| (node.node_type(), node.path)).collect();
-
-        for (t, p) in to_delete {
-            match t {
-                NodeType::Group => self.delete_group(p).await?,
-                NodeType::Array => self.delete_array(p).await?,
-            }
-        }
+        #[allow(clippy::expect_used)]
+        self.delete_group(Path::try_from("/").expect("should not fail")).await?;
```
paraseba, line 613: Is this too clever? I think the root group doesn't need to exist, for example they can have a root array.

icechunk/src/repository.rs
```diff
@@ -931,6 +952,7 @@ pub async fn get_chunk(
     }
 }
 
+// Yields nodes in the base snapshot, applying any relevant updates in the changeset
```
paraseba, line 955: add another / so it shows up in rendered documentation.

icechunk/src/repository.rs
```diff
@@ -909,6 +914,22 @@ impl From<Repository> for ChangeSet {
     }
 }
 
+pub fn is_prefix_match(key: &str, prefix: &str) -> bool {
+    let tomatch =
+        if prefix != &String::from('/') { key.strip_prefix(prefix) } else { Some(key) };
```
paraseba, line 919: Can you avoid the allocation by saying `prefix != "/"` ?

paraseba: This code is tricky and somewhat hard to follow. I wonder, can we use `Path` methods for this? Path's should be able to tell you if they are parent/child right?

paraseba: That would also skip the somewhat expensive conversion from Path to String

icechunk/src/repository.rs
```diff
@@ -954,6 +976,8 @@ async fn updated_existing_nodes<'a>(
     Ok(updated_nodes)
 }
 
+// Yields nodes with the snapshot, applying any relevant updates in the changeset,
+// *and* new nodes in the changeset
```
paraseba, line 980: see above, add a "/"

paraseba (approved):
This is great Deepak

PR #466, opened 2024-12-10
Add committing to stateful test actions

dcherian:
xref #445 

Fixes another bug found by hypothesis, if the same path is in the snapshot, then deleted, then recreated we list both old and new chunk paths


icechunk/src/format/manifest.rs
```diff
@@ -175,21 +176,25 @@ impl FromIterator<ChunkInfo> for Manifest {
...
+            ChunkIndices(itertools::repeat_n(u32::MAX, self.ndim).collect());
         match &self.last_key {
             None => {
                 if let Some((k @ (_, coord), payload)) = self
                     .manifest
                     .chunks
                     .range((
                         Bound::Included((self.for_node.clone(), ChunkIndices(vec![]))),
-                        Bound::Unbounded,
+                        Bound::Included((self.for_node.clone(), upper_bound)),
```
dcherian, line 197: Why is `clone` not needed on `upper_bound`? Is it because both the `Bound` and `upper_bound` go out of scope at the same time?

icechunk-python/tests/test_can_read_old.py
```diff
@@ -193,7 +193,7 @@ async def test_icechunk_can_read_old_repo():
     ]
     assert sorted(
         [p async for p in store.list_dir("group2/group3/group4/group5/inner")]
-    ) == ["c", "zarr.json"]
+    ) == ["zarr.json"]
```
dcherian, line 196: this test never wrote a chunk, so `c` should not have been created.

icechunk/src/format/manifest.rs
```diff
@@ -215,15 +221,14 @@ impl Iterator for PayloadIterator {
...
                     .next()
                 {
-                    if k.0 == self.for_node {
-                        self.last_key = Some(k.clone());
-                        Some((coord.clone(), payload.clone()))
-                    } else {
-                        None
-                    }
+                    self.last_key = Some(k.clone());
+                    Some((coord.clone(), payload.clone()))
```
dcherian, line 231: With the `take_while` approach, we would return `ChunkInfo` here and unpack it up in `verified_node_chunk_iterator`. Worth it?

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -40,6 +94,17 @@ def add_array(
         assume(not np.iscomplexobj(array))
         super().add_array(data, name, array_and_chunks)
 
+    # TODO: port to zarr
+    @invariant()
+    def check_list_prefix_from_root(self) -> None:
```
dcherian, line 99: TODO: delete this

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -3,23 +3,77 @@
...
+
+# TODO: more before/after commit invariants?
+# TODO: add "/" to self.all_groups, deleting "/" seems to be problematic
 class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
+    def __init__(self, *args, **kwargs):
+        note("__init__")
+        super().__init__(*args, **kwargs)
+
+    @initialize()
+    def init_store(self) -> None:
```
dcherian, line 36: TODO: delete this too

icechunk/src/format/manifest.rs
paraseba: upper_bound can transfer ownership. Rust is smart enough to know that only one branch of the match  will execute, so you can move the variable. `self.for_node` on the other hand cannot, because it would partially destroy `self`.

paraseba (approved):

.github/workflows/python-check.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
           set -e
           python3 -m venv .venv
           source .venv/bin/activate
-          pytest -n 4
+          pytest -n 4 --hypothesis-verbosity=verbose -s
```
dcherian, line 105:
```suggestion
          pytest --hypothesis-verbosity=verbose -s
```

icechunk/src/change_set.rs
```diff
@@ -96,6 +96,10 @@ impl ChangeSet {
             .new_groups
             .iter()
             .filter(|(child_path, _)| child_path.starts_with(path))
+            // dcherian: sometimes we somehow end up deleting the same path twice
+            // I don't understand why, but hypothesis occasionally finds this behaviour
+            // It does not reproduce easily.
+            .filter(|(child_path, _)| child_path != &path)
```
dcherian, line 102: Here's a log that illustrates what can happen: (For some reason `"/5T/pE6Y.Q5O4dJ"` is deleted twice! the second time around that path is not in `new_groups` which means it gets added to `deleted_groups`)
```
Draw 6 (Group parent): '5T'
Adding group: path='5T/pE6Y.Q5O4dJ'
Checking 3 expected keys vs 3 actual keys
['zarr.json', '5T/zarr.json', '5T/pE6Y.Q5O4dJ/zarr.json']
['zarr.json', '5T/zarr.json', '5T/pE6Y.Q5O4dJ/zarr.json']
[icechunk/src/change_set.rs:63:9] &self = ChangeSet {
    new_groups: {
        Path(Utf8PathBuf {_encoding: "unix", inner: "/5T",},): f88bf7d7ac935e0d,
        Path(Utf8PathBuf {_encoding: "unix", inner: "/5T/pE6Y.Q5O4dJ",},): 66df7f9265bd04a8,
    },
    new_arrays: {},
    updated_arrays: {},
    updated_attributes: {
        f88bf7d7ac935e0d: Some(UserAttributes {parsed: Object {},},),
        66df7f9265bd04a8: Some(UserAttributes {parsed: Object {},},),
    },
    set_chunks: {},
    deleted_groups: {},
    deleted_arrays: {},
}
has uncommitted_changes: True

Draw 7 (Group deletion target): '5T'
Deleting group 'group_path='5T'', prefix='', group_name='5T' using delete
to delete: <Group <icechunk.store.IcechunkStore object at 0x106f4d9d0>>
[icechunk/src/change_set.rs:81:9] "deleting {} at {}" = "deleting {} at {}"
[icechunk/src/change_set.rs:81:9] &node_id = f88bf7d7ac935e0d
[icechunk/src/change_set.rs:81:9] &path = Path(
    Utf8PathBuf {_encoding: "unix", inner: "/5T",},
)
[icechunk/src/change_set.rs:81:9] "deleting {} at {}" = "deleting {} at {}"
[icechunk/src/change_set.rs:81:9] &node_id = 66df7f9265bd04a8
[icechunk/src/change_set.rs:81:9] &path = Path(Utf8PathBuf {_encoding: "unix", inner: "/5T/pE6Y.Q5O4dJ",},)
[icechunk/src/change_set.rs:81:9] "deleting {} at {}" = "deleting {} at {}"
[icechunk/src/change_set.rs:81:9] &node_id = 66df7f9265bd04a8
[icechunk/src/change_set.rs:81:9] &path = Path(Utf8PathBuf {_encoding: "unix", inner: "/5T/pE6Y.Q5O4dJ",},)
deleted   5T
to delete: <Group memory://4412194752>
deleted   5T
removed from self.all_groups which is now set()
Checking 1 expected keys vs 1 actual keys
['zarr.json']
['zarr.json']
[icechunk/src/change_set.rs:63:9] &self = ChangeSet {
    new_groups: {},
    new_arrays: {},
    updated_arrays: {},
    updated_attributes: {},
    set_chunks: {},
    deleted_groups: {
        (
            Path(Utf8PathBuf {_encoding: "unix", inner: "/5T/pE6Y.Q5O4dJ",},), 66df7f9265bd04a8,),
    },
    deleted_arrays: {},
}
has uncommitted_changes: True
[icechunk/src/change_set.rs:63:9] &self = ChangeSet {
    new_groups: {},
    new_arrays: {},
    updated_arrays: {},
    updated_attributes: {},
    set_chunks: {},
    deleted_groups: {
        (
            Path(Utf8PathBuf {_encoding: "unix", inner: "/5T/pE6Y.Q5O4dJ",},), 66df7f9265bd04a8,),
    },
    deleted_arrays: {},
}
```

icechunk/src/change_set.rs
```diff
@@ -75,14 +75,16 @@ impl ChangeSet {
         self.new_arrays.get(path)
     }
 
-    pub fn delete_group(&mut self, path: Path, node_id: &NodeId) {
+    pub fn delete_group(&mut self, path: Path, node_id: &NodeId, delete_children: bool) {
```
dcherian, line 78: Just pushing this change for discussion. I don't like it. It means an illegal state is representable (remember that we no longer check whether an ancestor of a requested node's path is in `deleted_groups`)

An alternative would be to require that deleting groups goes through `delete_dir` and not through `delete` for icechunk. Any `.delete("/path/to/group/zarr.json")` would fail asking the user to use `delete_dir("/path/to/group")`

icechunk/src/store.rs
```diff
@@ -361,7 +361,7 @@ impl Store {
                         Ok(guard.deref_mut().delete_array(node_path).await?)
                     }
                     NodeData::Group => {
-                        Ok(guard.deref_mut().delete_group(node_path).await?)
+                        Ok(guard.deref_mut().delete_group(node_path, false).await?)
```
dcherian, line 364: Zarr expects `delete` to only ever delete a single key, no children.

icechunk/src/session.rs
```diff
@@ -178,23 +178,7 @@ impl Session {
     /// Deletes of non existing groups will succeed.
     pub async fn delete_group(&mut self, path: Path) -> SessionResult<()> {
         match self.get_group(&path).await {
-            Ok(parent) => {
```
dcherian, line 181: @paraseba this is the error. Both this loop and `change_set.delete_group` are deleting children. 

So if there is `/group` and `/group/subgroup`. `nodes_iter` contains both those groups. IF `/group` appears first then `change_set.delete_group` will delete both groups. Then in the second iteration `change_set.delete_group("/group/subgroup")` will add an entry to `change_set.deleted_groups`.

dcherian: (this change is not right). I'll try filtering out any nodes added in the currnet session

icechunk/src/change_set.rs
```diff
@@ -75,45 +75,39 @@ impl ChangeSet {
         self.new_arrays.get(path)
     }
 
+    /// IMPORTANT: This method does not delete children. The caller
```
dcherian, line 78: I think this is the cleanest solution. `Session.delete_group` already builds a list of nodes to delete. We could filter those to not be children of any groups added in this session (and may some other similarly complex predicate), or we can just not delete children ...

icechunk-python/tests/test_can_read_old.py
paraseba: There is some weird behavior in here i think ... sometimes i need it sometimes i don't?

icechunk/src/change_set.rs
```diff
@@ -75,43 +75,13 @@ impl ChangeSet {
...
-        let arrays_to_delete: Vec<_> = self
-            .new_arrays
-            .iter()
-            .filter(|(child_path, _)| child_path.starts_with(path))
-            .map(|(k, (node, _))| (k.clone(), node.clone()))
-            .collect();
-
-        for (path, node) in arrays_to_delete {
-            self.delete_array(path, &node);
+        if self.new_groups.remove(&path).is_none() {
```
paraseba, line 82: but if it is in `new_groups` don't we need to take it out?

dcherian: Maybe this is too clever.... `.remove` will remove it returning `Some(foo)` . If it returns None, then we add it to `deleted_groups`

paraseba: afff sorry .. this is great

paraseba (approved):

Issue #467, opened 2024-12-11
Implement new Python API


paraseba:
As designed in https://github.com/earth-mover/icechunk/issues/422

PR #470, opened 2024-12-11
API Refactor

mpiannucci:
#467 

- [x] Rust code
- [x] Rust Tests
- [x] Rust Python Bindings
- [x] Python Code
- [x] Python Tests (only `test_distributed_writes.py` is failing (skipped for now)

Not included in this PR but still on list to do in follow ups:

* Documentation
* Virtual functionality (including new container and config) API in python 
* Repository Builder APIs in rust
* Rebase functionality in python
* Using sessions with context managers in python
* distributed write tests


mpiannucci:
The only current test failures are conflict resolution tests in `session`

mpiannucci:
First cut of the python API is in without tests

rabernat:
Lmk what sort of review would be helpful here.

mpiannucci:
I think the last peice is to figure out why the python `test_distributed_writes.py` deadlocks. All other tests seem good (except the xfailed virtual tests)

mpiannucci:
So we get a clippey warning now: 

```
https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_lock
```

Rock and a hard place.. We can either use a tokio RwLock to correctly hold the lock through await points, or we can hold over the await with the sync RwLock (https://docs.rs/tokio/1.24.2/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use). We are locking access to IO (storage layer) so i believe we should be using the tokio async aware mutex. The issue now is that is seemingly impossible to serialize and deserialize an async lock without running into issues where a tokio reactor is already in use. 

The implementation done on the `s3` storage implementation works well because the client is never `mut` after it is created, so i think the solution here means looking at improving the workflow of the `Session` layer so that we can be sure to only unlock for data access then release the lock for IO, since the IO calls themselves are never `mut`

Any input on this would be helpful

mpiannucci:
I think the way around this is to not make the store serde serializeable, but just add custom to_bytes and from_bytes methods that are async and call into serde json or rmp or whatever

icechunk/src/session.rs
```diff
@@ -0,0 +1,3288 @@
...
+pub struct Session {
+    config: RepositoryConfig,
+    storage: Arc<dyn Storage + Send + Sync>,
+    virtual_resolver: Arc<VirtualChunkResolver>,
+    branch_name: Option<String>,
+    snapshot_id: SnapshotId,
+    change_set: ChangeSet,
+}
+
+impl Hash for Session {
```
paraseba, line 97: we should make an effort to eliminate this and equality

icechunk/src/session.rs
```diff
@@ -0,0 +1,3288 @@
...
+
+    pub fn from_bytes(bytes: Vec<u8>) -> SessionResult<Self> {
+        rmp_serde::from_slice(&bytes).map_err(SessionError::DeserializationError)
+    }
+
+    pub fn as_bytes(&self) -> SessionResult<Vec<u8>> {
+        rmp_serde::to_vec(self).map_err(SessionError::SerializationError)
+    }
+
+    pub fn id(&self) -> String {
```
paraseba, line 153: hopefully we can eliminate this

icechunk/src/session.rs
```diff
@@ -0,0 +1,3288 @@
...
+        update_branch(
+            repo.storage().as_ref(),
+            "main",
+            conflicting_snap.clone(),
+            Some(&current_snap),
+            false,
+        )
+        .await?;
+
+        // TODO: We can't create writeable sessions from arbitrary snapshots anymore so not sure what to do about this?
```
paraseba, line 2681: it sucks a bit that we cannot, we should try to bring it back after merging

icechunk/src/storage/caching.rs
```diff
@@ -44,11 +51,27 @@ impl MemCachingStorage {
...
 impl private::Sealed for MemCachingStorage {}
 
+impl<'de> Deserialize<'de> for MemCachingStorage {
+    fn deserialize<D>(deserializer: D) -> Result<MemCachingStorage, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let backend: Arc<dyn Storage + Sync + Send> =
+            Deserialize::deserialize(deserializer)?;
+        Ok(Self::new_with_defaults(backend))
```
paraseba, line 69: This is quite dangerous ... we should preserve the cache sizes during serialization and recover them

icechunk/src/storage/s3.rs
```diff
@@ -126,17 +134,25 @@ pub async fn mk_client(config: Option<&S3Config>) -> Client {
...
+        Ok(S3Storage { client, config: config.clone() })
+    }
+
+    async fn get_client(&self) -> Arc<Client> {
+        let guard = self.client.read().await;
+        if let Some(client) = guard.as_ref() {
+            Arc::clone(client)
+        } else {
+            let client = Arc::new(mk_client(self.config.options.as_ref()).await);
+            *self.client.write().await = Some(Arc::clone(&client));
```
paraseba, line 149: why does this work? How can you lock for write if you are still holding to a read lock?

icechunk/src/store.rs
```diff
@@ -1443,12 +1077,31 @@ impl TryFrom<NameConfigSerializer> for ChunkKeyEncoding {
...
+
+    async fn create_memory_store_repository() -> Repository {
+        let storage = Arc::new(
+            ObjectStorage::new_in_memory_store(Some("prefix".into()))
+                .expect("failed to create in-memory store"),
+        );
+        Repository::create(None, None, storage, HashMap::new()).await.unwrap()
+    }
+
+    // async fn create_local_store_repository() -> Repository {
```
paraseba, line 1096: we are not using this?

icechunk/tests/test_s3_storage.rs
```diff
@@ -0,0 +1,190 @@
+use std::{collections::HashSet, sync::Arc};
+
```
paraseba, line 2: There is some rebase weirdness in here. I think I have renamed this file to test_storage, and added significantly to it

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1,39 +1,88 @@
...
+        branch: str | None = None,
+        tag: str | None = None,
+        snapshot_id: str | None = None,
+    ) -> PySession: ...
+    def writeable_session(self, branch: str) -> PySession: ...
+
+class PySession:
+    @classmethod
+    def from_bytes(cls, data: bytes) -> PySession: ...
+    def __eq__(self, value: object) -> bool: ...
```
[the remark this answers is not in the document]
mpiannucci: IDK if we need it @dcherian do we need eq for anything is python land? If not we can get rid of it (and id i think)

icechunk-python/src/errors.rs
```diff
@@ -18,10 +19,14 @@ use thiserror::Error;
...
 pub(crate) enum PyIcechunkStoreError {
+    #[error("storage error: {0}")]
+    StorageError(StorageError),
     #[error("store error: {0}")]
     StoreError(StoreError),
     #[error("repository error: {0}")]
-    RepositoryError(RepositoryError),
+    RepositoryError(#[from] RepositoryError),
+    #[error("session error: {0}")]
+    SessionError(SessionError),
```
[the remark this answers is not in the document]
mpiannucci: No because we manually implement it

icechunk/src/session.rs
mpiannucci: Its not a technical limitation, this was the design you and ryan proposed IIRC. Is there a pattern where we would create a writeable session detached from a branch?

icechunk/src/storage/s3.rs
mpiannucci: I have no idea, i will make it temporary though just in case

icechunk/src/storage/caching.rs
mpiannucci: I think this is what we are doing currently right? I dont think we serialize the Cacheing Storage at all?

paraseba: True, good point ... let me create a ticket

icechunk/src/session.rs
paraseba: Ryan doesn't like allowing "anonymous" writes. I do like them and think they are going to be useful. I feel Icechunk as a storage format should not dictate the user's workflow. But we can bring them back later.

icechunk-python/src/errors.rs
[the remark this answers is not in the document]
paraseba: ah got it

icechunk-python/src/store.rs
```diff
@@ -57,7 +57,7 @@ impl PyStoreConfig {
 
 #[pyclass(name = "PyStore")]
 #[derive(Clone)]
-pub struct PyStore(pub Arc<RwLock<Store>>);
```
mpiannucci: happy about this part

icechunk-python/src/store.rs
```diff
@@ -76,58 +76,48 @@ impl PyStore {
...
-    #[getter]
-    fn session_id(&self) -> PyIcechunkStoreResult<String> {
-        pyo3_async_runtimes::tokio::get_runtime().block_on(async move {
-            let session_id = self.0.read().await.session_id().await?;
-            Ok(session_id.to_string())
-        })
+        // If the stores were created from the same session they are equal
+        let session = self.0.session();
+        let other_session = other.0.session();
+        Arc::ptr_eq(&session, &other_session)
```
paraseba, line 90: I think you don't need the previous if? This second condition will also match in the first case

mpiannucci: Ah yeah that's a relic from before I got rid of the inner lock I didn't need

icechunk/src/storage/s3.rs
```diff
@@ -29,28 +29,28 @@ use futures::{
...
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(transparent)]
 pub struct S3Storage {
-    client: Arc<Client>,
-    prefix: String,
-    bucket: String,
+    config: S3Config,
+    #[serde(skip)]
+    client: RwLock<Option<Arc<Client>>>,
```
paraseba, line 53: Not a fan of the RwLock here. Why don't we use a OnceCell? That probably can replace both the RwLock and the Option.

Can you add a code comment explaining why we need the OnceCell to help readers? I understand it is because we don't serialize the client, so on deserialization it's initialized empty. But why can't we initialize it on deseralization and we skip the need for mutability completely?

design-docs/001-virtual-chunk-improvements.md
```diff
@@ -374,7 +374,7 @@ Ichecunk repo stored in S3 bucket `repo-bucket`. Virtual array `A` on Google clo
     Now the user initializes a Repo instance by passing, if needed, a credential set for each virtual chunk container. If a chunk is read for which there are no credentials, we default to the Storage credentials if on the same platform. Users can figure out the credentials they need to pass by checking the list of containers.
 
 3. > A mechanism to know all the "dependencies" of a repo
-  
+
```
paraseba, line 377: :pray:

icechunk/src/storage/s3.rs
mpiannucci: Done, this also lets me get rid of the `Arc`

paraseba (approved):

PR #472, opened 2024-12-11
Virtual chunk containers

paraseba:


icechunk/src/virtual_chunks.rs
```diff
@@ -279,14 +285,53 @@ impl ChunkFetcher for S3Fetcher {
...
+        match checksum {
+            Some(Checksum::LastModified(SecondsSinceEpoch(seconds))) => {
+                b = b.if_unmodified_since(aws_sdk_s3::primitives::DateTime::from_secs(
+                    *seconds as i64,
+                ))
+            }
+            Some(Checksum::ETag(etag)) => {
+                b = b.if_match(etag);
+            }
+            None => {}
```
mpiannucci, line 297: Glad the checksum is here and optional 👍

icechunk-python/src/lib.rs
```diff
@@ -122,6 +120,31 @@ impl From<SnapshotMetadata> for PySnapshotMetadata {
     }
 }
 
+#[derive(FromPyObject)]
+enum ChecksumArgument {
```
paraseba, line 124: @mpiannucci is this the right approach to take a python argument that can be any of these types?

mpiannucci (approved):
I really like this API. Its way better than what we have right now

icechunk-python/src/lib.rs
mpiannucci: Yes!

Issue #408, opened 2024-11-23
Design and implement python rebase API


paraseba:
We don't need to expose the full details, for now it's enough to have the `rebase` method, `ConflictDetector` and `BasicConflictSolver`

PR #483, opened 2024-12-18
Add design doc for rebase and commit python api improvements

mpiannucci:
This document is mostly straightforward, but the biggest question is what do we do when a rebase fails within a session context? 

[#408](https://github.com/earth-mover/icechunk/issues/408)

### DIFF
--- a/design-docs/002-rebase-commit-improvements.md
+++ b/design-docs/002-rebase-commit-improvements.md
@@ -18,7 +18,7 @@ However the `commit` call is not guaranteed to succeed, and an error is thrown i
 ```python
 try:
     commit_id = session.commit("yay")
-except Exception as e:
+except ConflictError as e:
     print(f"Commit failed: {e}")
 ```
 
@@ -27,13 +27,9 @@ To resolve this, the `rebase` method was added in the rust layer. This has not y
 ```python
 try:
     commit_id = session.commit("yay")
-except Exception as e:
-    if isinstance(e, ConflictError):
-        try:
-	          session.rebase(solver=BasicConflictSolver())
-	          commit_id = session.commit("merged")
-	      except Exception as e:
-			      print(f"Failed to rebase: {e}")
+except ConflictError as e:
+    session.rebase(solver=BasicConflictSolver())
+    commit_id = session.commit("merged")
 ```
 
 This is a common pattern that will need to be followed. Two notes on the errors returned for failures in the rust core:
@@ -59,7 +55,7 @@ try:
     commit_id = session.commit("merged")
 except RebaseFailedError as e:
     # just an example of what we could pull for data, the message would be more informative
-    print(f"Failed to rebase because of conflicts:")
+    print(f"Failed to rebase at snapshot {e.snapshot_id} because of conflicts:")
     for conflict in e.conflicts:
         print(f"{c.path}: {c.reason}")
 ```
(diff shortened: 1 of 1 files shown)


design-docs/002-rebase-commit-improvements.md
```diff
@@ -0,0 +1,79 @@
...
+except ConflictError as e:
+    # just an example of what we could pull for data, the message would be more informative
+    print(f"Commit failed: expected parent: {e.expected_parent}, actual parent: {e.actual_parent}")
+
+...
+
+try:
+    session.rebase(solver=BasicConflictSolver())
+    commit_id = session.commit("merged")
+except RebaseFailedError as e:
```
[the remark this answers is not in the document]
paraseba: To be clear, the above scenario will happen if you have a number of jobs stepping on each others feet. While one is rebasing (successfully) other is adding a new commit, etc.

design-docs/002-rebase-commit-improvements.md
```diff
@@ -0,0 +1,79 @@
...
+    print(f"Failed to rebase because of conflicts:")
+    for conflict in e.conflicts:
+        print(f"{c.path}: {c.reason}")
+```
+
+2. Create a context manager workflow that will create a scoped session and automatically commit and rebase (if needed) when context is finished (see below).
+
+```python
+with repo.open_writable_session("main", message="Update data", solver=BasicConflictSolver()) as session:
+	  store = session.store()
```
paraseba, line 71: We should still allow for something like `session.describe("commit msg")` because sometimes the commit message is constructed as the session moves forward.

maybe `solver -> conflict_solver`.

design-docs/002-rebase-commit-improvements.md
```diff
@@ -0,0 +1,79 @@
...
+	  # Do Stuff with the store
+
+# Session is finished, committed and rebased automatically
+```
+
+## Open Questions
```
paraseba, line 77: Should we have an async interface too? Are there any advantages?

design-docs/002-rebase-commit-improvements.md
paraseba: This context manager should probably implement the repeated rebase attempts I explained above. Probably there are extra arguments to tweak the behavior: number of retries, exponential backoff, etc.

design-docs/002-rebase-commit-improvements.md
```diff
@@ -0,0 +1,79 @@
...
+
+1. How do we handle the case where rebase fails within a session context? How do we avoid losing the user's work on failure in case they want to try a different strategy? Does this matter?
```
paraseba, line 79: The rust code is prepared to manage this case: it will rebase as far as possible, and when it can no longer make progress it will give you back a valid repo, with a valid base snapshot_id (the one of the latest rebased commit), and a valid ChangeSet.

So, we will be able to get a valid `session` after a failed rebase. The thing is how to surface it back to the user. Maybe in an exception that includes the session as an attribute?

```python
try:
    with repo.open_writable_session(.....) as session:
        ...
except RebaseFailed as e:
    do_something_with( e.session )

**Review by paraseba — APPROVED** [review]

File: design-docs/002-rebase-commit-improvements.md [hunk 5]
```diff
@@ -0,0 +1,79 @@
...
+    print(f"Commit failed: {e}")
+```
+
+To resolve this, the `rebase` method was added in the rust layer. This has not yet been added to the python API, but it could look like this:
+
+```python
+try:
+    commit_id = session.commit("yay")
+except Exception as e:
+    if isinstance(e, ConflictError):
```
rabernat, line 31:
```suggestion
except ConflictError as e:
```

design-docs/002-rebase-commit-improvements.md
```diff
@@ -0,0 +1,79 @@
...
+        try:
+	          session.rebase(solver=BasicConflictSolver())
+	          commit_id = session.commit("merged")
+	      except Exception as e:
+			      print(f"Failed to rebase: {e}")
```
rabernat, line 36: No need to try / except here, right? Just let it raise

design-docs/002-rebase-commit-improvements.md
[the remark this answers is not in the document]
rabernat: Arraylake currently does this retry loop automatically. Do we want to expose something similar here?

rabernat (approved):
I like this approach a lot!

I agree with Seba that a retry loop should probably be built into the high-level rebase logic, as it is with Arraylake.

PR #485, opened 2024-12-18
Python learns to rebase and report conflicts

### LINKED ISSUE #484: [Docs] API Reference is not up to date
For some reason, the API Reference section of the docs is missing a bunch of exported members like `Session`, `Repository` `RepositoryConfig`.

### LINKED ISSUE #487: When using the zarr store interface, only chunk conflicts can be resolved with the current `BasicConflictSolver`
The `BasicConflictSolver` has support for rebasing user attributes conflicts, however this currently doesn't work when updating the attributes through `Store` (and by extension, from python. This is because anytime the attributes are updated via the `Store` api, the entire node hierarchy is updated with `set`. 

To support updating and rebasing attributes via the `Store` api, we need to isolate the attributes update step [here](https://github.com/earth-mover/icechunk/blob/823ea545510e6417f399981625d7bc3cbb23f941/icechunk/src/store.rs#L565) to only update the attributes and not the full node metadata if only the attributes have been updated

mpiannucci:
#408 #483 

Closes #484 
Closes #487 

This does not implement the context manager api discussed in #483  , but it creates all the primitives to build it and allows users to use the rebase functionality


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -323,4 +326,50 @@ class StoreConfig:
...
+    ...
+
+class BasicConflictSolver(ConflictSolver):
+    """A basic conflict solver that allows for simple configuration of resolution behavior"""
+
+    def __init__(
+        self,
+        *,
+        on_user_attributes_conflict: VersionSelection = VersionSelection.UseOurs,
+        on_chunk_conflict: VersionSelection = VersionSelection.UseOurs,
```
rabernat, line 350: Would `Fail` potentially be a better / safer default, requiring an explicit overwrite for a more aggressive strategy?

icechunk-python/tests/test_timetravel.py
```diff
@@ -131,3 +133,42 @@ async def test_branch_reset():
...
+    root_a = zarr.group(store=store_a)
+    array_a = cast(zarr.Array, root_a["foo/bar/some-array"])
+    array_a.attrs["repo"] = 1
+    session_a.commit("update array")
+
+    root_b = zarr.group(store=store_b)
+    array_b = cast(zarr.Array, root_b["foo/bar/some-array"])
+    array_b.attrs["repo"] = 2
+
+    with pytest.raises(ValueError):
```
rabernat, line 164: Find myself wishing this was a more specific custom exception.

mpiannucci: Yeah definitely, that plan is included in the design doc

icechunk-python/python/icechunk/_icechunk_python.pyi
paraseba: The thing is, if you are going to say `Fail` everywhere, you shouldn't be calling `rebase` at all. In IC rebase is explicit, if you call it, you want to accept some kind of risk.

rabernat: Seba I'm not sure I follow. The current default Arraylake rebase behavior is to fail if there are any chunk conflicts or attribute conflicts. But you can still rebase, i.e. rewrite the snapshot lineage if there are no such conflicts.

This is desirable behavior in many scenarios and is a good default because it is the most "conservative" way to rebase.

rabernat: In fact we describe this exact scenario in the Arraylake docs: https://docs.earthmover.io/concepts/version-control-system#commit-process-and-conflict-resolution

icechunk-python/tests/test_timetravel.py
```diff
@@ -131,3 +132,42 @@ async def test_branch_reset():
...
+        session_b.commit("update array")
+
+    solver = icechunk.BasicConflictSolver(
+        on_user_attributes_conflict=icechunk.VersionSelection.UseOurs,
+    )
+
+    session_b.rebase(solver)
```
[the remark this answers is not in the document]
mpiannucci: The difference between this and the rust test is that when you modify attrs in python, it calls `set` on the entire array instead of just updaing the attrs, so the changeset has updated metadata instead of just updated attrs which seems to cause the failure

icechunk-python/python/icechunk/_icechunk_python.pyi
paraseba: I think the behavior you want is supported by the `ConflictDetector` in Rust. That's also a `ConflictSolver` but it doesn't attempt to "fix" things, it just fails on any conflicts. I think using that solver instead, you can recover the AL behavior. At least that was my intention when I built it. 

Then the `BasicConflictSolver` uses `ConflictDetector` output and tries to fix it according to user policies, in a pretty naif way, but if you only want "fast forward if absolutely no conflicts", the detector should be enough.

I'm OK with any behavior we want to implement as a default, and I would also favor the safest one. That's why I made `rebase` explicit and not a default, unlike in AL. Any form of rebase implies risk. For example, there may be no conflicts, but you still computed chunk A from an old version of chunk B that was since then updated. If you never wrote to B, there are no conflicts but you end up with bad data anyway.

In my view, we should give the full menu of options available to Python users: fail on conflicts detected, basic resolution strategies, and, in the future, more advanced and even custom strategies.

icechunk-python/src/errors.rs
```diff
@@ -71,12 +75,81 @@ impl From<SessionError> for PyIcechunkStoreError {
...
     fn from(error: PyIcechunkStoreError) -> Self {
         match error {
+            PyIcechunkStoreError::SessionError(SessionError::Conflict {
+                expected_parent,
+                actual_parent,
+            }) => ConflictError::new_err(PyConflictErrorData {
+                expected_parent: expected_parent.map(|s| s.to_string()),
+                actual_parent: actual_parent.map(|s| s.to_string()),
+            }),
+            PyIcechunkStoreError::SessionError(SessionError::RebaseFailed {
```
mpiannucci, line 85: We cant create exceptions with custom methods or properties from rust, so instead we pass a custom object to the args. Which prints a nice message, but unfortunately means the user needs to go to Error.args[0] to get the actual data properties. Not idea but it works fine. 

Should we repackage the errors with our own custom exceptions in the python world from here? We can just wrap it in `session.py`

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -323,4 +325,132 @@ class StoreConfig:
...
+        When the session is based on a branch, this is the snapshot ID of the branch tip. If this
+        error is raised, it means the branch was modified and committed by another session after
+        the session was created.
+        """
+        ...
+
+class PyConflictError(IcechunkError):
+    """An error that occurs when a conflict is detected"""
+
+    args: tuple[ConflictErrorData]
```
rabernat, line 406: Why is this just a one-item tuple? Wouldn't 
```suggestion
    conflicts: ConflictErrorData
```

be simpler?

mpiannucci: This wrapped in a python exception now, so this is hidden from UX

icechunk/src/store.rs
```diff
@@ -561,10 +561,29 @@ async fn set_array_meta(
...
-        // TODO: we don't necessarily need to update both
-        repo.set_user_attributes(path.clone(), array_meta.attributes).await?;
-        repo.update_array(path, array_meta.zarr_metadata).await?;
+    if let Ok(node) = repo.get_array(&path).await {
+        // Check if the user attributes are different, if they are update them
+        let existing_attrs = match node.user_attributes {
+            None => None,
+            Some(UserAttributesSnapshot::Inline(atts)) => Some(atts),
+            // FIXME: implement
+            Some(UserAttributesSnapshot::Ref(_)) => None,
```
mpiannucci, line 570: Does this ever happen? For now we just fallback to original impl, we update

paraseba: It's not implemented yet, so no, it won't happen

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1,14 +1,26 @@
...
 
 class RepositoryConfig:
+    """Configuration for an Icechunk repository"""
     def __init__(
         self,
         *,
         inline_chunk_threshold_bytes: int = 512,
         unsafe_overwrite_refs: bool = False,
-    ) -> None: ...
+    ) -> None:
```
paraseba, line 13: :heart:

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -323,4 +347,145 @@ class StoreConfig:
...
+
+    Where the `BasicConflictSolver` will attempt to resolve conflicts, the `ConflictDetector` will only detect them. This means
+    that during a rebase operation the `ConflictDetector` will raise a `RebaseFailed` error if any conflicts are detected, and
+    allow the rebase operation to be retried with a different conflict resolution strategy. Otherwise, if no conflicts are detected
+    the rebase operation will succeed.
+    """
+
+    def __init__(self) -> None: ...
+
+class IcechunkError(Exception):
```
paraseba, line 409: nice!

icechunk-python/python/icechunk/repository.py
```diff
@@ -19,12 +21,29 @@ def __init__(self, repository: PyRepository):
     def create(
         cls, storage: StorageConfig, *, config: RepositoryConfig | None = None
     ) -> Self:
+        """Create a new Icechunk repository.
+
+        If one already exists at the given storafe location, an error will be raised.
```
paraseba, line 26:
```suggestion
        If one already exists at the given store location, an error will be raised.
```

icechunk-python/src/conflicts.rs
```diff
@@ -0,0 +1,203 @@
...
+                path: path.to_string(),
+            },
+            Conflict::UserAttributesUpdateOfDeletedNode(path) => PyConflict {
+                conflict_type: PyConflictType::UserAttributesUpdateOfDeletedNode,
+                path: path.to_string(),
+            },
+            Conflict::ChunkDoubleUpdate { path, node_id: _, chunk_coordinates: _ } => {
+                PyConflict {
+                    conflict_type: PyConflictType::ChunkDoubleUpdate,
+                    path: path.to_string(),
```
paraseba, line 116: @mpiannucci Can we create a ticket to add the chunk_coordinates to the exception? Without the coordinates users will have a hard time understanding why they have a conflict in their algorithms. It doesn't need to happen in this PR.

paraseba (approved):
Amazing work @mpiannucci !!

icechunk-python/src/conflicts.rs
mpiannucci: Isnt the path the same as the coordinates? I will confirm..

mpiannucci: Ahhhh it is not. Sounds good that is important

PR #488, opened 2024-12-19
Python config design doc

paraseba:


design-docs/003-python-instantiation-api.md
```diff
@@ -0,0 +1,139 @@
...
+* We use the same types for Storage config and virtual chunk container store config
+* We use the same types for repo credentials and virtual chunk container credentials
+* We don't offer different config types for different object stores
+
+```python
+ObjectStorePlatform = Literal["S3", "GoogleCloudStorage", "Azure", "Tigris", "S3Compatible", "LocalFileSystem", "InMemory"]
+
+@dataclass
+class ObjectStoreConfig:
+    object_store: ObjectStorePlatform
```
mpiannucci, line 47: Is this missing prefix?

mpiannucci: I think the status quo is using urls to represent locations, and I think that is a better user experience personally

paraseba: ah, it is missing the prefix true. I moved it to VirtualChunkContainer, but then I lose it from the main object store config. Will rethink it, thank you.

paraseba: >  urls to represent locations

The problem is all the extra information: allow http? endpoint? does it support foo feature? etc

design-docs/003-python-instantiation-api.md
```diff
@@ -0,0 +1,139 @@
+# Python API for Repository instantiation and configuration
+
+## What we need
+
+1. A good interface to instantiate existing and new `Repositories`, including different types of `Storage` instances.
+    * For the case of existing repos, reading of configuration and snapshot should happen concurrently.
+1. A good interface to tune the configuration parameters of a `Repository`, both at runtime and in storage.
+1. A good interface to tune the configuration parameters of a `Store`.
```
dcherian, line 8:
```suggestion
1. A good interface to tune the configuration parameters of a `Repository`, both at runtime and in storage, such as `inline_chunk_threshold_bytes`, `unsafe_overwrite_refs`, and `virtual_chunk_containers`.
1. A good interface to tune the configuration parameters of a `Store`, such as `get_partial_values_concurrency`.
```

(had to look up which parameters were being referred to)

design-docs/003-python-instantiation-api.md
```diff
@@ -0,0 +1,139 @@
...
+* Should we move the `Store` config to the `Repository`?
+  * That may simplify things, and it would allow `store` to be a property instead of a function
+  * Of course, it would mean a `Repository` instance can only create `Stores` with the same config
+* Is a common `Credentials` type enough? Should every object store have its own `Credentials` type?
+  * For example, we could support refreshing for certain object stores and not others, or from environment.
+
+## Design
+
+* We unify repo and store config. The extra flexibility is not worth the complexity
+* We use the same types for Storage config and virtual chunk container store config
```
dcherian, line 38: hehe I just came to the same conclusion and then read this line. :+1:

design-docs/003-python-instantiation-api.md
paraseba: Maybe there is a usability feature to turn standard urls into `ObjectStoreConfig` instances.

dcherian: AFAICT there is no meaningful difference between ObjectStoreConfig and VirtualChunkContainer? The former just has no explicit name, and is used for all the "format" metadata files, and is the default location to store chunks. Should we merge the two?

paraseba: I'll think about it, and we can discuss. But part of your view is because I gave things a bad name. VirtualChunkContainer has a `prefix` field, a better name would be `url_prefix`. This is a string that will be matched against virtual chunks location url, it's not a bucket prefix.

The container name is also quite critical, you used it to provide credentials for that container.

mpiannucci: My thing is that the ObjectStoreConfig, is really actually S3 config. GCS and Azure have very different options available

design-docs/003-python-instantiation-api.md
```diff
@@ -0,0 +1,139 @@
...
+
+@dataclass
+class RepositoryConfig:
+    inline_chunk_threshold_bytes: int
+    unsafe_overwrite_refs: bool
+    virtual_chunk_containers: List[VirtualChunkContainer]
+
+    # gathering together the config for the repo and the store
+    get_partial_values_concurrency: int
+
```
dcherian, line 69: In our conversation, we discussed something like this (bringing it up again for discussion in the meeting)

```suggestion
    @classmethod
    def default() -> Self:
        # returns defaults from Rust.
        pass
```

design-docs/003-python-instantiation-api.md
```diff
@@ -0,0 +1,139 @@
...
+class VirtualChunkContainer:
+    name: str
+    prefix: str
+    store: ObjectStoreConfig
+
+@dataclass
+class RepositoryConfig:
+    inline_chunk_threshold_bytes: int
+    unsafe_overwrite_refs: bool
+    virtual_chunk_containers: List[VirtualChunkContainer]
```
dcherian, line 65: A dict would be nicer api for deletion and updates but then it wouldn't map on to the Rust Enum.

design-docs/003-python-instantiation-api.md
paraseba: I do like your `classmethod` better, changing it.

design-docs/003-python-instantiation-api.md
```diff
@@ -0,0 +1,167 @@
...
+```python
+ObjectStorePlatform = Literal["S3", "GoogleCloudStorage", "Azure", "Tigris", "S3Compatible", "LocalFileSystem", "InMemory"]
+
+class ObjectStoreConfigBase(ABC):
+  def object_store_name(self):
+     ...
+
+@dataclass
+class S3Config(ObjectStoreConfigBase):
+  object_store: ObjectStorePlatform
```
paraseba, line 51: ah this field must go

dcherian (approved):

PR #489, opened 2024-12-19
Store learns to `delete_dir`

### LINKED ISSUE #390: Add rust native `delete_dir` implementation
Currently we are falling back to using the zarr-python `delete_dir` implementation which is not efficient, and relies on recursively deleting keys.

dcherian:
Closes #390


icechunk/src/store.rs
```diff
@@ -337,6 +339,57 @@ impl Store {
...
+                // other cases are
+                // 1. delete("/path/to/array/c")
+                // 2. delete("/path/to/array/c/0/0")
+                let to_delete = self
+                    .list_prefix(prefix.trim_start_matches("/"))
+                    .await?
+                    .try_collect::<Vec<_>>()
+                    .await?;
+                for item in to_delete {
+                    self.delete(&item).await?;
```
mpiannucci, line 386: Are these guaranteed to exist when delete is called? Or should we just catch the error in case they are deleted out of order

mpiannucci (approved):

icechunk/src/store.rs
```diff
@@ -337,6 +339,57 @@ impl Store {
...
+                }) => Ok((NodeType::Group, node_path)),
+                Ok(NodeSnapshot {
+                    node_data: NodeData::Array(..),
+                    path: node_path,
+                    ..
+                }) => Ok((NodeType::Array, node_path)),
+                Err(err) => Err(err),
+            }
+        };
+
```
paraseba, line 366: I think it would be safer to lock the session for writes at the top. Otherwise you need to think about the edge case of somebody else deleting the node (or creating it) between the previous `get_node` and the following `delete_*`.

icechunk/src/store.rs
```diff
@@ -337,6 +339,57 @@ impl Store {
...
+        match res {
+            Ok((NodeType::Group, node_path)) => {
+                let mut guard = self.session.write().await;
+                Ok(guard.deref_mut().delete_group(node_path).await?)
+            }
+            Ok((NodeType::Array, node_path)) => {
+                let mut guard = self.session.write().await;
+                Ok(guard.deref_mut().delete_array(node_path).await?)
+            }
+            _ => {
```
paraseba, line 376: I think we need to match only for the `Err(NotFound)` (pseudocode)? Other types of errors (imagine missing credentials, etc) need to be bubbled up immediately.

paraseba (changes requested):
Great change!

Blocking merge only because I think we have a race condition, but happy to approve if you convince me otherwise.

icechunk/src/store.rs
dcherian: Yes that is the approach I took at first but `store.delete` will try to grab a write lock too. Now I realize I can solve that with a `self.delete_with_optional_locking`.

icechunk/src/store.rs
```diff
@@ -337,47 +337,60 @@ impl Store {
...
+                    node_data: NodeData::Array(..),
+                    path: node_path,
+                    ..
+                }) => Ok(guard.deref_mut().delete_array(node_path).await?),
+                Err(SessionError::NodeNotFound { .. }) => {
+                    // other cases are
+                    // 1. delete("/path/to/array/c")
+                    // 2. delete("/path/to/array/c/0/0")
+                    let prefix = prefix.trim_start_matches("/");
+                    let to_delete = guard
```
dcherian, line 367: why was `list_chunks_prefix` implemented with `try_stream!` and not something like this? 🤔

icechunk/src/store.rs
```diff
@@ -337,47 +337,60 @@ impl Store {
...
-                    Ok(_) => Ok(()),
-                    Err(SessionError::NodeNotFound { path: _, message: _ }) => {
-                        // When there is no chunk at the given key, we don't consider it an error, instead we just do nothing
-                        Ok(())
+        // TODO: Handling preceding "/" is ugly!
+        let path = format!("/{}", prefix.trim_start_matches("/"))
+            .try_into()
+            .map_err(|_| StoreError::BadKeyPrefix(prefix.to_owned()))?;
+
+        {
```
paraseba, line 350: do you need this extra scope?

icechunk/src/store.rs
```diff
@@ -337,47 +337,60 @@ impl Store {
...
+                Ok(NodeSnapshot {
+                    node_data: NodeData::Array(..),
+                    path: node_path,
+                    ..
+                }) => Ok(guard.deref_mut().delete_array(node_path).await?),
+                Err(SessionError::NodeNotFound { .. }) => {
+                    // other cases are
+                    // 1. delete("/path/to/array/c")
+                    // 2. delete("/path/to/array/c/0/0")
+                    let prefix = prefix.trim_start_matches("/");
```
paraseba, line 366: we already trimmed all "/" at the start

icechunk/src/store.rs
```diff
@@ -337,47 +337,60 @@ impl Store {
...
+                }) => Ok(guard.deref_mut().delete_group(node_path).await?),
+                Ok(NodeSnapshot {
+                    node_data: NodeData::Array(..),
+                    path: node_path,
+                    ..
+                }) => Ok(guard.deref_mut().delete_array(node_path).await?),
+                Err(SessionError::NodeNotFound { .. }) => {
+                    // other cases are
+                    // 1. delete("/path/to/array/c")
+                    // 2. delete("/path/to/array/c/0/0")
```
paraseba, line 365: 3. delete("/path/to/arr")
4. delete("non-existent-prefix")

icechunk/src/store.rs
```diff
@@ -337,47 +337,60 @@ impl Store {
...
+                    let prefix = prefix.trim_start_matches("/");
+                    let to_delete = guard
+                        .deref_mut()
+                        .all_chunks()
```
paraseba, line 369: Some thoughts:
1. This is only deleting chunks, isn't it? How about nodes?
2. If we are going to delete only chunks, we should be able to do it without all the extra `get_node` calls involved in `delete_key`
3. We should find a way to do a big delete without gathering every chunk into a vector in memory at once
4. It feels like we should have a way to get nodes based on a prefix, and then use the high level operations that delete nodes

icechunk/src/session.rs
```diff
@@ -256,6 +266,18 @@ impl Session {
...
     }
 
+    pub async fn delete_chunks(
+        &mut self,
+        node_path: &Path,
+        coords: Vec<ChunkIndices>,
+    ) -> SessionResult<()> {
+        let node = self.get_array(node_path).await?;
+        for coord in coords {
+            self.set_node_chunk_ref(node.clone(), coord, None).await?
```
dcherian, line 276: I don't see how to avoid this clone given that `set_node_chunk_ref` destructures `node`

icechunk/src/session.rs
```diff
@@ -179,6 +179,16 @@ impl Session {
         }
     }
 
+    pub async fn delete_node(&mut self, node: NodeSnapshot) -> SessionResult<()> {
```
dcherian, line 182: this is some minor DRY, because its used in `delete` and `delete_dir`

icechunk/src/session.rs
```diff
@@ -256,6 +266,18 @@ impl Session {
         Ok(())
     }
 
+    pub async fn delete_chunks(
+        &mut self,
+        node_path: &Path,
+        coords: Vec<ChunkIndices>,
```
dcherian, line 272: Should I make it take a stream instead?

icechunk-python/tests/test_zarr/test_store/test_icechunk_store.py
```diff
@@ -319,8 +319,6 @@ async def test_is_empty(self, store: IcechunkStore) -> None:
         assert await store.is_empty("spam/")
 
     async def test_delete_dir(self, store: IcechunkStore) -> None:
-        if not store.supports_deletes:
-            pytest.skip("store does not support deletes")
```
paraseba, line 323: :partying_face:

icechunk/src/session.rs
```diff
@@ -277,17 +299,27 @@ impl Session {
         data: Option<ChunkPayload>,
     ) -> SessionResult<()> {
         let node_snapshot = self.get_array(&path).await?;
+        self.set_node_chunk_ref(node_snapshot, coord, data).await
+    }
 
-        if let NodeData::Array(zarr_metadata, _) = node_snapshot.node_data {
+    // Helper function that accepts a NodeSnapshot instead of a path,
+    // this lets us do bulk sets (and deletes) without repeatedly grabbing the node.
+    async fn set_node_chunk_ref(
```
paraseba, line 307: beautiful

paraseba (approved):

PR #497, opened 2024-12-20
Refactor how we initialize and configure storage and repos

paraseba:


icechunk/src/config.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct S3CompatibleOptions {
+    pub region: Option<String>,
+    pub endpoint_url: Option<String>,
+    pub anonymous: bool,
+    pub allow_http: bool,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ObjectStoreConfig {
```
mpiannucci, line 16: The only thing I don't like is that this isn't extensible..for instance, for wasm we may need a different storage provider only for wasm, this means for every storage implementation we need to update this enum, there's no way for third parties to add their own storage implementations

paraseba: I agree. I feel this is high level API that should be overridable. We can add a more powerful mechanism later.

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -265,59 +194,53 @@ class S3Credentials:
...
-        """Config for an S3 Object Storage compatible storage backend"""
+class Credentials:
+    class FromEnv:
+        def __init__(self) -> None: ...
 
-        credentials: S3Credentials | None
-        endpoint_url: str | None
-        allow_http: bool | None
-        region: str | None
+    class Nonen:
```
mpiannucci, line 201: Possible typo?

icechunk-python/python/icechunk/repository.py
```diff
@@ -32,9 +30,7 @@ def create(
         return cls(PyRepository.create(storage, config=config))
 
     @classmethod
-    def open(
-        cls, storage: StorageConfig, *, config: RepositoryConfig | None = None
-    ) -> Self:
+    def open(cls, storage: Storage, config: RepositoryConfig | None = None) -> Self:
```
mpiannucci, line 33: Can we go back to having the config be a keyword only argument?

icechunk-python/python/icechunk/repository.py
```diff
@@ -18,9 +18,7 @@ def __init__(self, repository: PyRepository):
         self._repository = repository
 
     @classmethod
-    def create(
-        cls, storage: StorageConfig, *, config: RepositoryConfig | None = None
-    ) -> Self:
+    def create(cls, storage: Storage, config: RepositoryConfig | None = None) -> Self:
```
mpiannucci, line 21: same here wrt keyword args

icechunk-python/src/config.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+#[pyclass(name = "Storage")]
+#[derive(Clone, Debug)]
+pub struct PyStorage(pub Arc<dyn Storage + Send + Sync>);
+
+#[pyfunction]
+#[pyo3(signature = ( config, bucket=None, prefix=None, credentials=None))]
+pub(crate) fn make_storage(
```
mpiannucci, line 191: Can this be a classmethod on `Storage` instead?

mpiannucci (commented):
Some comments on making the python API more pythonic. I really like how `Storage` is its own thing users can create now

icechunk-python/python/icechunk/repository.py
paraseba: Yes, will change. Remind me why you prefer it, I'm indifferent I think

mpiannucci: It just makes it explicit since it's optional

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -183,77 +183,15 @@ class StorageConfig:
...
-        prefix: str,
-        endpoint_url: str | None,
-        allow_http: bool = False,
-        region: str | None = None,
-    ) -> StorageConfig:
-        """Create a StorageConfig object for an S3 Object Storage compatible storage
-        using anonymous access
-        """
-        ...
+    @staticmethod
```
mpiannucci, line 186: My very last nit is, can we make this a classmethod instead of staticmethod?

mpiannucci (approved):
Awesome stuff, can't wait to build the other storage layers for this, just one not important nit

Issue #492, opened 2024-12-19
`Repository.tag` is confusing


dcherian:
It returns the `SnapshotId` associated with  a tag name.

Perhaps we should call it `lookup_tag` and rename `.branch_tip` to `lookup_branch`?

mpiannucci:
I like that change. I chose the name tag and knew it was bad

jhamman:
The GitPython project has a pretty intuitive API. https://gitpython.readthedocs.io/en/stable/tutorial.html#examining-references

```python
repo.create_tag(ref: str)
repo.delete_tag(ref: str)
repo.tags -> Iterable[TagReference]
```

Issue #491, opened 2024-12-19
Add `Repository.delete_branch`


dcherian:
Noticed in https://github.com/earth-mover/icechunk/pull/453

paraseba:
Maybe `reset_branch` should take an `Option`?

PR #499, opened 2024-12-23
Refine Ref APIs on Repository

mpiannucci:
* Change `branch_tip` and `tag` functions to `lookup_branch` and `lookup_tag` #492 
* Add `delete_branch` methods #491


dcherian (approved):
Nice improvement.

I think we should consider whether we want to expose `Repo.branches -> tuple(Branch)` and `Repo.tags -> tuple(Tag)` in the python API similar to Arraylake.

PR #504, opened 2024-12-24
Virtual chunk location inspection

paraseba:


mpiannucci (commented):
Should we consider using a set for this? Since they won't be duplicated? Or is order guaranteed

paraseba:
> Should we consider using a set for this? Since they won't be duplicated? Or is order guaranteed

I'm only using a vector because I don't want to spend a lot of time for something maybe nobody uses soon. The right type would be an iterator, but that's tricky. There could be many items... we should switch to iterator at some point. I'll create a ticket

dcherian (approved):
Seems OK.

mpiannucci (approved):

PR #507, opened 2024-12-24
Join `RepositoryConfig` and `StoreConfig`

paraseba:


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -128,19 +128,17 @@ class PySession:
     @property
     def has_uncommitted_changes(self) -> bool: ...
     def discard_changes(self) -> None: ...
-    def store(self, config: StoreConfig | None = None) -> PyStore: ...
     def all_virtual_chunk_locations(self) -> list[str]: ...
+    def store(self) -> PyStore: ...
```
mpiannucci, line 132: Can we make this a getter?

paraseba: I was planning to do it in a separate PR, but I can do it in this one too.

mpiannucci (approved):

PR #512, opened 2024-12-27
Create specific credential types for each object store

paraseba:


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -253,15 +229,47 @@ class StaticCredentials:
...
+    storage_config = StorageConfig.memory("prefix")
+    storage_config = StorageConfig.filesystem("/path/to/root")
+    storage_config = StorageConfig.object_store("s3://bucket/prefix", vec!["my", "options"])
+    storage_config = StorageConfig.s3_from_env("bucket", "prefix")
+    storage_config = StorageConfig.s3_from_config("bucket", "prefix", ...)
+    ```
+    """
+
+    @staticmethod
+    def make_s3(
```
mpiannucci, line 263: I personally don't love the naming of `make_` and would prefer just simply `s3()`, `in_memory`, but I am open to alternatives if others like verbs

icechunk/src/storage/mod.rs
```diff
@@ -222,47 +222,30 @@ where
     s.try_filter_map(|info| async move { Ok(convert_list_item(info)) }).boxed()
 }
 
-pub async fn make_storage(
-    config: ObjectStoreConfig,
-    bucket: Option<String>,
+pub async fn make_s3_storage(
+    config: S3Options,
```
mpiannucci, line 226: Should this be in the s3 mod? And same for the object store impls?

paraseba: I'll move it. I thought it would  be easier to search in documentation if we put them here, but I think I can change that by `pub use`

icechunk-python/python/icechunk/_icechunk_python.pyi
paraseba: done

icechunk/src/storage/mod.rs
paraseba: I did this  but I didn't like it and revert it. Technicalities don't allow me to do something nice with `pub use`. These are such important methods, used as the first thing by any IC user, that I think they need to be very discoverable. I don't want to bury them inside many different classes the user would have to know about.

So I kept them here and made them public at the library level, so now you can say `icechunk::new_s3_storage`

mpiannucci (approved):

PR #521, opened 2024-12-30
Fix reset_branch op in stateful test

dcherian:


dcherian:
Actually let me think about this more.

mpiannucci (approved):

PR #528, opened 2024-12-30
Re-implement `list_dir`

### LINKED ISSUE #321: `list_dir` relies on `list_prefix` which traverses all chunks
This is a problem because it's much slower than it needs to be, and it's used by zarr to implement things like `tree`. We should make `list_dir` smarter.

dcherian:
Closes #321

### DIFF
--- a/icechunk-python/tests/test_zarr/test_stateful.py
+++ b/icechunk-python/tests/test_zarr/test_stateful.py
@@ -33,7 +33,7 @@ def chunk_paths(
         st.tuples(*tuple(st.integers(min_value=0, max_value=b - 1) for b in numblocks))
     )
     subset_slicer = (
-        slice(draw(st.integers(min_value=1, max_value=ndim))) if subset else slice(None)
+        slice(draw(st.integers(min_value=0, max_value=ndim))) if subset else slice(None)
     )
     return "/".join(map(str, blockidx[subset_slicer]))
 
@@ -153,6 +153,42 @@ class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
         note(f"deleting chunk {path=!r}")
         self._sync(self.model.delete(path))
         self._sync(self.store.delete(path))
+
+    @precondition(lambda self: bool(self.all_arrays) or bool(self.all_groups))
+    @rule(data=st.data())
+    def delete_dir(self, data) -> None:
+        array_or_group = data.draw(
+            st.one_of(
+                st.sampled_from(sorted(self.all_groups))
+                if self.all_groups
+                else st.nothing(),
+                st.sampled_from(sorted(self.all_arrays))
+                if self.all_arrays
+                else st.nothing(),
+            )
+        )
+        if data.draw(st.booleans()) and array_or_group in self.all_arrays:
+            arr = zarr.open_array(path=array_or_group, store=self.model)
+            path = data.draw(
+                st.one_of(
+                    st.sampled_from([array_or_group]),
+                    chunk_paths(ndim=arr.ndim, numblocks=arr.cdata_shape).map(
+                        lambda x: f"{array_or_group}/c/"
+                    ),
+                )
+            )
+        else:
+            path = array_or_group
+        note(f"delete_dir with {path=!r}")
+        self._sync(self.model.delete_dir(path))
+        self._sync(self.store.delete_dir(path))
+
+        matches = set()
+        for node in self.all_groups | self.all_arrays:
+            if node.startswith(path):
+                matches.add(node)
+        self.all_groups = self.all_groups - matches
+        self.all_arrays = self.all_arrays - matches
 
     @invariant()
     def check_list_prefix_from_root(self) -> None:
--- a/icechunk-python/tests/test_zarr/test_stateful.py
+++ b/icechunk-python/tests/test_zarr/test_stateful.py
@@ -12,6 +12,7 @@ from hypothesis.stateful import (
     run_state_machine_as_test,
 )
 
+import zarr
 from icechunk import Repository, in_memory_storage
 from zarr.core.buffer import default_buffer_prototype
 from zarr.testing.stateful import ZarrHierarchyStateMachine
@@ -22,6 +23,15 @@ from zarr.testing.strategies import (
 )
 
 PROTOTYPE = default_buffer_prototype()
+
+
+@st.composite
+def chunk_paths(draw: st.DrawFn, ndim: int, numblocks: tuple[int, ...]) -> str:
+    blockidx = draw(
+        st.tuples(*tuple(st.integers(min_value=0, max_value=b - 1) for b in numblocks))
+    )
+    subset = draw(st.integers(min_value=1, max_value=ndim))
+    return "/".join(map(str, blockidx[:subset]))
 
 
 # TODO: more before/after commit invariants?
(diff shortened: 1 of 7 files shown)


icechunk-python/tests/test_zarr/test_store/test_icechunk_store.py
```diff
@@ -301,6 +303,12 @@ async def test_set_if_not_exists(self, store: IcechunkStore) -> None:
 
     async def test_is_empty(self, store: IcechunkStore) -> None:
         assert await store.is_empty("")
+        await self.set(
```
dcherian, line 306: There are no implicit groups in Zarr V3, so I think this is fine. Anyone using the group API would have created these keys.

cc @jhamman

icechunk/src/store.rs
```diff
@@ -428,31 +423,90 @@ impl Store {
...
+                                })
+                            } else {
+                                None
+                            };
+                            Ok(res)
+                        })
+                        .try_collect::<HashSet<_>>()
+                        .await?
+                        .into_iter()
+                        .collect::<Vec<_>>()
```
dcherian, line 502: very yuck, I guess I could just create the stream in each arm of the match instead of returning a Vec

dcherian:
```
FAILED tests/test_can_read_old.py::test_icechunk_can_read_old_repo - AssertionError: assert ['c', 'zarr.json'] == ['zarr.json']
  
  At index 0 diff: 'c' != 'zarr.json'
  Left contains one more item: 'zarr.json'
  
  Full diff:
    [
  +     'c',
        'zarr.json',
    ]
```

I don't understand this, there is a manifest but no chunks were ever written?

icechunk/src/store.rs
```diff
@@ -428,31 +423,90 @@ impl Store {
...
-                // If the prefix is "/", get rid of it. This can happen when prefix is missing
-                // the trailing slash (as it does in zarr-python impl)
-                let rem = &s[idx..].trim_start_matches('/');
-                match rem.split_once('/') {
-                    Some((prefix, _)) => ListDirItem::Prefix(prefix.to_string()),
-                    None => ListDirItem::Key(rem.to_string()),
+        let session = Arc::clone(&self.session).read_owned().await;
+        let absolute_prefix =
+            if !prefix.starts_with("/") { &format!("/{}", prefix) } else { prefix };
+        let prefix = prefix.trim_end_matches("/");
```
paraseba, line 429: should you trim absolute_prefix instead?

icechunk/src/store.rs
```diff
@@ -428,31 +423,90 @@ impl Store {
...
+
+        #[allow(clippy::expect_used)]
+        let path =
+            Path::try_from(absolute_prefix).expect("could not create path from prefix");
```
paraseba, line 433: Let's replace by a not panic error

icechunk/src/store.rs
```diff
@@ -428,31 +423,90 @@ impl Store {
...
+        let results = match session.get_node(&path).await {
+            Ok(NodeSnapshot {
+                id: node_id,
+                node_data: NodeData::Array(.., manifests),
+                ..
+            }) => {
+                // if this is an array we know what to yield
+                let mut res = vec![ListDirItem::Key("zarr.json".to_string())];
+                if session.array_has_chunks_set(&node_id) || !manifests.is_empty() {
```
paraseba, line 442: hmm I don't think this condition is precise enough. It could have chunks in the manifests but all deleted by the session. I think we probably want to call some existing function that lists all chunks for the array and just verify if there are any.

icechunk/src/session.rs
```diff
@@ -317,6 +339,19 @@ impl Session {
         }
     }
 
+    pub async fn array_chunk_iterator<'a>(
+        &'a self,
+        path: &Path,
+    ) -> impl Stream<Item = SessionResult<ChunkInfo>> + 'a {
```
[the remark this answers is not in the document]
dcherian: apparently not

icechunk/src/store.rs
dcherian: yeah I was trying to be too clever about not downloading a manifest. That can be a future optimization.

icechunk/src/store.rs
dcherian: no that doesn't work becausse HashSet.into_iter() and Vec.into_iter() are different types :/

icechunk/src/store.rs
paraseba: Ahh I forgot about that! I think this is worth exploring because it can be a big optimization. Maybe if there are manifests and no chunk deletes....? We would have to think about edge cases ... but I like this idea.

icechunk-python/tests/test_zarr/test_store/test_icechunk_store.py
dcherian: `get_node("")` fails because there a root group was not explicitly created, and we go down the "array chunk list" path. I think this is another case where Icechunk is not a key-value store...

icechunk/src/store.rs
dcherian: OK I tried again.

dcherian: Now I've added a couple more deletion rules to the stateful tests in addition to a list_dir rule to gain more confidence.

dcherian: OK I think the one edge-case not addressed is when the user calls `delete_dir("path/to/array/c")`. This is not a deleted array, and we can't detect this without checking that every chunk of the array was deleted or never written in the first place :/

icechunk/src/session.rs
```diff
@@ -240,6 +242,26 @@ impl Session {
             .map(|node| self.change_set.update_array(node.id, metadata))
     }
 
+    pub async fn array_has_modified_chunks(
+        &self,
+        path: Path,
+        id: NodeId,
+    ) -> SessionResult<bool> {
+        let array = self.get_array(&path).await?;
```
paraseba, line 250: I think the caller always has executed get_array already? Maybe we should include in the arguments to this function the `NodeData`? I think with that you can also make this function return `bool`

icechunk/src/session.rs
```diff
@@ -240,6 +242,26 @@ impl Session {
             .map(|node| self.change_set.update_array(node.id, metadata))
     }
 
+    pub async fn array_has_modified_chunks(
```
paraseba, line 245: the name of this function confuses me, because it makes me think this returns true if the session (`self`) has modified any chunks in the array, but I  don't think that's what the function does.

icechunk/src/store.rs
```diff
@@ -428,31 +426,100 @@ impl Store {
...
+                path: node_path,
+                node_data: NodeData::Array(..),
+                ..
+            }) => {
+                // if this is an array we know what to yield
+                let mut res = vec![ListDirItem::Key("zarr.json".to_string())];
+                // The `c` prefix will exist when an array was created and a chunk was written
+                // either (1) in the changeset or (2) as part of a previous snapshot.
+                // It will not exist if (3) the array is recorded as deleted in the change_set.
+                if session.array_has_modified_chunks(node_path, node_id).await? {
```
paraseba, line 447: maybe we are working too hard for something that doesn't really matter? What happens if we always include the `/c`? I don't imagine anything will break.

icechunk/src/session.rs
```diff
@@ -240,6 +242,26 @@ impl Session {
...
+        path: Path,
+        id: NodeId,
+    ) -> SessionResult<bool> {
+        let array = self.get_array(&path).await?;
+        if let NodeSnapshot { node_data: NodeData::Array(_, manifests), .. } = array {
+            debug_assert_eq!(array.id, id);
+            let path_and_id = &(path, id);
+            Ok(!self.change_set.array_is_deleted(path_and_id)
+                && (self.change_set.array_has_modified_chunks(path_and_id)
+                    || !manifests.is_empty()))
```
paraseba, line 256: I think this misses the case of a session that deletes all chunks in an existing array. But see my comment below, maybe we are doing too much... maybe we can always return the `/c`

dcherian: yes... i had the same realization in https://github.com/earth-mover/icechunk/pull/528#discussion_r1900282037.

I've updated to always return `c`. We can get smarter if someone complains and really cares about it.

icechunk/src/session.rs
dcherian: function is now deleted.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -107,6 +122,73 @@ def clear(self) -> None:
...
+        note(f"delete_dir with {path=!r}")
+        self._sync(self.model.delete_dir(path))
+        self._sync(self.store.delete_dir(path))
+
+        matches = set()
+        for node in self.all_groups | self.all_arrays:
+            if node.startswith(path):
+                matches.add(node)
+        self.all_groups = self.all_groups - matches
+        self.all_arrays = self.all_arrays - matches
```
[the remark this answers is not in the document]
dcherian: I added an assert to the `list_prefix` invariant below to make sure the tracked state matches the `store` and `model`

icechunk/src/change_set.rs
```diff
@@ -48,6 +48,15 @@ impl ChangeSet {
         self.updated_attributes.keys()
     }
 
+    pub fn array_is_deleted(&self, path_and_id: &(Path, NodeId)) -> bool {
+        self.deleted_arrays.contains(path_and_id)
+    }
+
+    pub fn array_has_modified_chunks(&self, path_and_id: &(Path, NodeId)) -> bool {
```
paraseba, line 55: we can delete this function now right? Just because i don't like the name

icechunk/src/store.rs
```diff
@@ -136,8 +138,9 @@ impl Store {
     }
 
     pub async fn is_empty(&self, prefix: &str) -> StoreResult<bool> {
-        let res = self.list_dir(prefix).await?.next().await;
-        Ok(res.is_none())
+        let list = self.list_dir(prefix).await?;
+        pin_mut!(list);
+        Ok(list.next().await.is_none())
```
paraseba, line 143: is pin still needed now that we didn't change the signature?

paraseba (approved):

PR #535, opened 2025-01-03
Use multiple concurrent requests to download manifests

paraseba:
This requires:
* storing the manifests size in the snapshot
* configuring `Storage` instances with number and size of concurrent
  requests
* execute concurrent requests with the right ranges using tokio, then
  coalesce the pieces (which is currently not super optimal)

Currently only manifests are being pulled in this way, but the same
could be applied to chunks and snapshots


icechunk/src/storage/object_store.rs
```diff
@@ -217,6 +240,37 @@ impl ObjectStorage {
...
+            ));
+            let opts = GetOptions { range, ..Default::default() };
+            let path = path.clone();
+            let handle = tokio::spawn(async move { store.get_opts(&path, opts).await });
+            handles.push(handle)
+        }
+
+        let mut res = BytesMut::with_capacity(size as usize);
+
+        for h in handles {
```
mpiannucci, line 267: This isnt actually concurrent right? You need to join these futures, otherwise no work happens until `.await` is called

icechunk/src/storage/mod.rs
```diff
@@ -223,17 +238,44 @@ where
...
+    let num_parts = max(1, min(num_parts, max_parts as u64));
+    let equal_parts = num_parts - 1;
+    let equal_parts_size = size / num_parts;
+    let last_part_size = size - equal_parts * equal_parts_size;
+    let equal_requests =
+        iter::successors(Some((0, equal_parts_size)), move |(off, _)| {
+            Some((off + equal_parts_size, equal_parts_size))
+        });
+    let last_request = iter::once((equal_parts * equal_parts_size, last_part_size));
+    equal_requests.take(equal_parts as usize).chain(last_request)
```
dcherian, line 261: :clap:

icechunk/src/storage/mod.rs
```diff
@@ -300,5 +344,101 @@ pub fn new_gcs_storage(
...
+
+            // there are only two request sizes
+            let counts = res.iter().map(|(_, size)| size).counts();
+            prop_assert!(counts.len() <= 2); // only last element is smaller
+            if counts.len() > 1 {
+                // the smaller request size happens only once
+                prop_assert_eq!(counts.values().min(), Some(&1usize));
+            }
+
+            // there are no holes in the requests, nor bytes that are requested more than once
```
dcherian, line 409: very nice.

icechunk/src/storage/mod.rs
```diff
@@ -223,17 +238,44 @@ where
     s.try_filter_map(|info| async move { Ok(convert_list_item(info)) }).boxed()
 }
 
+/// Split an object request into multiple byte range requests
+///
+/// Returns tuples of (offset, size) for each request. It tries to generate the maximum number of
+/// requests possible, not generating more than `max_parts` requests, and each request not being
+/// smaller than `min_part_size`.
```
dcherian, line 245:
```suggestion
/// smaller than `min_part_size`. Note that the size of the last request is >= the preceding one.
```

(i assumed the other way, until I read the property test)

dcherian (commented):
split logic and tests LGTM:

icechunk/src/storage/mod.rs
```diff
@@ -223,17 +238,44 @@ where
...
+fn split_in_multiple_requests(
+    size: u64,
+    min_part_size: u64,
+    max_parts: u16,
```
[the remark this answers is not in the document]
paraseba: yes ... that's what i had, but it was so painful with all the wrapping and unwrapping and shit, just for an internal function. But you are right, I'll take a middle ground and change the types in the interface types to be nonzero.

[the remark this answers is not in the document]
dcherian: I thought as much

icechunk/src/storage/object_store.rs
paraseba: Switched to use `FuturesOrdered` which is what `join_all` uses but helps me avoid doubling memory. Good call!

I don't know if there will be any performance different compared to spawn. They said is good for "large" number of futures, but i don't know what that means ... we will have ~20 :shrug:

icechunk/src/storage/mod.rs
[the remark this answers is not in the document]
paraseba: done

icechunk/src/storage/mod.rs
paraseba: done

dcherian (approved):

mpiannucci (approved):

PR #540, opened 2025-01-04
Manifest compression

paraseba:
We use zstd for compressing all manifests. zstd showed significantly faster compression/decompression than other simple alternatives, with similar or better compression ratios.

Pending:
- [x] configurable compression level

For future:
- Configurable no compression
- Compress only larger files
- Metadata, content-type, etc


icechunk/src/repository.rs
```diff
@@ -73,6 +73,7 @@ pub type RepositoryResult<T> = Result<T, RepositoryError>;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Repository {
     config: RepositoryConfig,
+    storage_settings: storage::Settings,
```
mpiannucci, line 76: Why is this both in the config and here?

mpiannucci (approved):

icechunk/src/repository.rs
paraseba: Excellent question, i'll make a comment in the code. The settings in config are what the user passed, but most of the time they won't pass anything. `storage_settings` is the "merge" between what users passed and the defaults, so what the repo will use in future calls. We need to retain what the user passed (config) in case they call `save_config`, so I cannot override it with a default.

PR #553, opened 2025-01-09
Fetch chunks using concurrent requests

paraseba:
Only applies if the chunk is large enough. Same logic as with manifests.

This still doesn't apply to virtual chunks, only native. For no reason, we can follow up with virtual chunks in a different PR.

This commit also fixes a bug in chunks with offset != 0 (wich cannot be constructed in IC today)

### DIFF
--- a/icechunk/src/storage/mod.rs
+++ b/icechunk/src/storage/mod.rs
@@ -20,6 +20,7 @@ use std::{
     ffi::OsString,
     iter,
     num::{NonZeroU16, NonZeroU64},
+    ops::Range,
     path::Path,
     sync::Arc,
 };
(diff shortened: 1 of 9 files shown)


icechunk/src/storage/mod.rs
```diff
@@ -322,25 +323,27 @@ where
 
 /// Split an object request into multiple byte range requests
 ///
-/// Returns tuples of (offset, size) for each request. It tries to generate the maximum number of
+/// Returns tuples of Range for each request. It tries to generate the maximum number of
```
dcherian, line 326: ah i like this

dcherian (approved):

PR #555, opened 2025-01-09
Unpin zarr & xarray

dcherian:


mpiannucci:
We need to update the store to amtch the 3.0.0 api

icechunk-python/python/icechunk/store.py
```diff
@@ -134,7 +165,9 @@ async def get_partial_values(
         """
         # NOTE: pyo3 has not implicit conversion from an Iterable to a rust iterable. So we convert it
         # to a list here first. Possible opportunity for optimization.
-        result = await self._store.get_partial_values(list(key_ranges))
+        print(key_ranges)
```
dcherian, line 168:
```suggestion
```

mpiannucci:
The `ByteRequest` change doesnt line up with the rust version anymore

`Last` fetches up to the lenght of the chunk - `n` instad of fetching the last `n` chunks

icechunk-python/python/icechunk/store.py
```diff
@@ -12,6 +18,20 @@
     from icechunk import Session
 
 
+def _byte_request_to_tuple(
+    byte_request: ByteRequest | None,
+) -> tuple[int | None, int | None]:
+    if byte_request is None:
```
dcherian, line 24: Use the match, Luke.

mpiannucci: The problem is that these are dataclasses that i dont think i can send to rust? IDK

mpiannucci: oh you can match in pythonjnow lol

dcherian: python has match!

icechunk-python/python/icechunk/store.py
```diff
@@ -12,6 +18,20 @@
     from icechunk import Session
 
 
+def _byte_request_to_tuple(
```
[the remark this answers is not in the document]
dcherian: tuple is ok imo

icechunk-python/python/icechunk/store.py
```diff
@@ -93,22 +113,33 @@ async def get(
...
+            - SuffixByteRequest(int): Request the last int bytes. Note that here, int is the size of the request, not the byte offset. This is equivalent to bytes=-{int} as an HTTP header.
 
         Returns
         -------
         Buffer
         """
 
         try:
-            result = await self._store.get(key, byte_range)
+            if byte_range is None:
```
dcherian, line 137: isn't this handled in `_byte_range_to_tuple`?

icechunk/src/format/mod.rs
```diff
@@ -150,7 +150,7 @@ pub enum ByteRange {
     Bounded(Range<ChunkOffset>),
     /// All bytes from the given offset (included) to the end of the object
     From(ChunkOffset),
-    /// Last n bytes in the object
+    /// All bytes up to the last n bytes in the object
     Last(ChunkLength),
```
dcherian, line 154:
```suggestion
    Until(ChunkLength),
```

i think we should rename, last is just confusing now.

icechunk-python/pyproject.toml
```diff
@@ -16,7 +16,7 @@ classifiers = [
 license = { text = "Apache-2.0" }
 dynamic = ["version"]
 
-dependencies = ["zarr==3.0.0rc2"]
+dependencies = ["zarr>=3"]
```
paraseba, line 19: Is this the right way to specify it? should it be 3 but not for?

icechunk-python/pyproject.toml
```diff
@@ -39,7 +39,7 @@ test = [
   "ruff",
   "dask>=2024.11.0",
   "distributed>=2024.11.0",
-  "xarray@git+https://github.com/pydata/xarray.git@main",
+  "xarray>=2025.01.1",
```
paraseba, line 42: same question

icechunk/src/session.rs
```diff
@@ -1014,9 +1014,9 @@ async fn updated_nodes<'a>(
         .chain(change_set.new_nodes_iterator(manifest_id)))
 }
 
-async fn get_node<'a>(
+async fn get_node(
```
paraseba, line 1017: Why do you need all these new lifetimes @mpiannucci ? What breaks if you don't?

icechunk/src/format/mod.rs
```diff
@@ -150,8 +150,8 @@ pub enum ByteRange {
     Bounded(Range<ChunkOffset>),
     /// All bytes from the given offset (included) to the end of the object
     From(ChunkOffset),
-    /// Last n bytes in the object
-    Last(ChunkLength),
+    /// All bytes up to the last n bytes in the object
+    Until(ChunkLength),
```
paraseba, line 154: Let's keep Last too, so we don't break zarrs

icechunk/src/session.rs
mpiannucci: I got rid of them, rust 1.84 doesnt let clippy pass if you dont

icechunk-python/pyproject.toml
mpiannucci: presumably we want to support xarray forever

icechunk-python/pyproject.toml
dcherian: oh you want `"zarr>=3<4"`?

cc @jhamman for input

paraseba (approved):

PR #562, opened 2025-01-10
Metadata files download performance

paraseba:
- Most of the time seems consumed during parsing
- This change increases the number of concurrent requests for a 20M metadata file from 1 to 2.
- We improve the performance of parsing by using a buffer. Impact is small, estimated at ~40% of decompression time


icechunk/src/storage/mod.rs
```diff
@@ -323,28 +324,38 @@ where
...
-        move |range| Some(range.end..range.end + equal_parts_size),
-    );
-    let last_part_offset = range.start + equal_parts * equal_parts_size;
-    let last_request = iter::once(last_part_offset..last_part_offset + last_part_size);
-    equal_requests.take(equal_parts as usize).chain(last_request)
+    let size = max(0, range.end - range.start);
+    // we do a ceiling division, rounding always up
+    let num_parts = (size - 1) / ideal_req_size + 1;
+    // no more than max_parts, so we limit
+    let num_parts = max(1, min(num_parts, max_requests as u64));
```
dcherian, line 344: i don't see wht you need the `max(1, ...)` Isn't `let num_parts = min(num_parts, max_requests as u64);` enough?

icechunk/src/storage/mod.rs
```diff
@@ -323,28 +324,38 @@ where
...
-    let equal_requests = iter::successors(
-        Some(range.start..range.start + equal_parts_size),
-        move |range| Some(range.end..range.end + equal_parts_size),
-    );
-    let last_part_offset = range.start + equal_parts * equal_parts_size;
-    let last_request = iter::once(last_part_offset..last_part_offset + last_part_size);
-    equal_requests.take(equal_parts as usize).chain(last_request)
+    let size = max(0, range.end - range.start);
+    // we do a ceiling division, rounding always up
+    let num_parts = (size - 1) / ideal_req_size + 1;
```
dcherian, line 342: v.minor: could use the built-in method for clarity

```suggestion
    let num_parts = (size - 1).div_ceil(ideal_req_size);
```
https://doc.rust-lang.org/std/primitive.u64.html#method.div_ceil

icechunk/src/storage/mod.rs
```diff
@@ -323,28 +324,38 @@ where
...
+/// difference between the requests.
+///
+/// It tries to generate ceil(size/ideal_req_size) requests, but never exceeds max_requests.
+///
+/// ideal_req_size and max_requests must be > 0
 fn split_in_multiple_requests(
     range: &Range<u64>,
-    min_part_size: u64,
-    max_parts: u16,
+    ideal_req_size: u64,
```
dcherian, line 337: Don't we need to rename the config parameter too?

icechunk/src/storage/mod.rs
paraseba: What? is that a thing? i swear i searched for it ... with the wrong name i guess

icechunk/src/storage/mod.rs
paraseba: my soul rejects the notion of dividing by something i haven't checked is != 0

dcherian: hehehe i thought it was `NonZeroU64` at user-input time.

icechunk/src/storage/mod.rs
paraseba: done

icechunk/src/storage/mod.rs
paraseba: done

dcherian (approved):

PR #563, opened 2025-01-11
Solve deadlock exposed by python threads

paraseba:
Many of our rust functions called from python use some form of blocking. These functions need to release the Gil. Otherwise, if the lock that is to be released to make progress is modified by a different thread, we have a deadlock.

I don't know what the performance impact of this much calling to `allow_threads` is, but I think it's either that or disallowing python threads.


dcherian:
so one thread has the GIL but is blocked at the rust level by another thread that is waiting for the GIL? That's the issue?

dcherian (approved):
LGTM based on my reading of the docs. IMO it's better to get a review from @mpiannucci.

paraseba:
@dcherian  said:

> so one thread has the GIL but is blocked at the rust level by another thread that is waiting for the GIL? That's the issue?

I haven't worked on an explicit reproduction of a full flow, but imagine the following:

* Thread A takes a write lock on the store
* Thread A relases the Gil (for any reason, we released, or some low level operation does)
* Python runtime switches to thread B
* Thread B calls `block_on` waiting on a Read lock on the store. Notably it blocks without releasing the Gil
* Thread never has opportunity to run again and release the lock
* Deadlock

mpiannucci (approved):
I feel naive because it feels very heavy handed that the GIL is locked on every method call but that is indeed the case

PR #565, opened 2025-01-12
Better datastructure for manifest

paraseba:


icechunk/src/format/manifest.rs
```diff
@@ -177,64 +195,39 @@ impl Manifest {
...
-                    ))
-                    .next()
-                {
-                    self.last_key = Some(k.clone());
-                    Some((coord.clone(), payload.clone()))
-                } else {
-                    None
+                Some(last_key) => {
+                    if let Some((coord, payload)) =
+                        map.range((Bound::Excluded(last_key), Bound::Unbounded)).next()
```
dcherian, line 220: ah now that each hasmap is bounded to an array, we can get away with this :)

dcherian (approved):
I don't have any conceptual concerns per se, but what gives us the speedup? Avoiding all those clones?

PR #571, opened 2025-01-13
Disallow deleting main branch

paraseba:
Main branch is used to detect repository existence, so it shouldn't be deleted. It is also the default in several functions.

Closes: #515

### DIFF
--- a/icechunk/src/repository.rs
+++ b/icechunk/src/repository.rs
@@ -72,6 +72,8 @@ pub enum RepositoryError {
     IOError(#[from] std::io::Error),
     #[error("a concurrent task failed {0}")]
     ConcurrencyError(#[from] JoinError),
+    #[error("bain branch cannot be deleted")]
+    CannotDeleteMain,
 }
 
 pub type RepositoryResult<T> = Result<T, RepositoryError>;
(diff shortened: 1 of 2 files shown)


icechunk/src/repository.rs
```diff
@@ -378,8 +380,12 @@ impl Repository {
     /// This will remove the branch reference and the branch history. It will not remove the
     /// chunks or snapshots associated with the branch.
     pub async fn delete_branch(&self, branch: &str) -> RepositoryResult<()> {
-        delete_branch(self.storage.as_ref(), &self.storage_settings, branch).await?;
-        Ok(())
+        if branch != "main" {
```
mpiannucci, line 383: Should we use the default branch constant instead of hardcoding main?

mpiannucci (approved):
just one nit

icechunk/src/repository.rs
```diff
@@ -72,6 +72,8 @@ pub enum RepositoryError {
     IOError(#[from] std::io::Error),
     #[error("a concurrent task failed {0}")]
     ConcurrencyError(#[from] JoinError),
+    #[error("bain branch cannot be deleted")]
```
dcherian, line 75:
```suggestion
    #[error("main branch cannot be deleted")]
```

dcherian:
Could uncomment and add a `!= "main"` here, or I can do it later...

https://github.com/earth-mover/icechunk/blob/f3731cd554c159782561158cfa3f9c8fec6aeb7d/icechunk-python/tests/test_stateful_repo_ops.py#L348-L357

paraseba:
> Could uncomment and add a != "main" here, or I can do it later...

Great, I'll add it.

paraseba:
@dcherian I added the code in the stateful test, but it's failing for unrelated reasons. I'll leave it modified but commented out for you to take a look at some point.

icechunk/src/repository.rs
paraseba: done

icechunk/src/repository.rs
paraseba: done

Issue #510, opened 2024-12-26
add `validate_containers` option to `set_virtual_chunk_ref`


paraseba:
As documented in the design doc

PR #573, opened 2025-01-13
Add `validate_container` argument to set_virtual_ref

paraseba:
This argument, defaulted to `False` will make sure the location passed matches one of the existing virtual chunk containers.

Closes: #510

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -256,14 +256,7 @@ class PyStore:
         offset: int,
         length: int,
         checksum: str | datetime.datetime | None = None,
-    ) -> None: ...
-    async def async_set_virtual_ref(
-        self,
-        key: str,
-        location: str,
-        offset: int,
-        length: int,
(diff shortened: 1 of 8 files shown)


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -256,14 +256,7 @@ class PyStore:
         offset: int,
         length: int,
         checksum: str | datetime.datetime | None = None,
-    ) -> None: ...
-    async def async_set_virtual_ref(
```
paraseba, line 260: deleting because it no longer exists

icechunk-python/python/icechunk/store.py
```diff
@@ -203,34 +203,6 @@ async def set_if_not_exists(self, key: str, value: Buffer) -> None:
         """
         return await self._store.set_if_not_exists(key, value.to_bytes())
 
-    async def async_set_virtual_ref(
```
paraseba, line 206: this no longer exists, deleting.

mpiannucci (approved):

mpiannucci:
great stuff!

Issue #572, opened 2025-01-13
Can `ancestry` take a `branch` and/or `tag` kwarg in addition to `snapshot_id`?


dcherian:
Or should we have a different method to view the history for a branch/tag

PR #574, opened 2025-01-14
Better ancestry interface

paraseba:
Now users can use a snapshot, tag or branch to fetch history.

```python

class Repository:
  def ancestry(
      self,
      *,
      branch: str | None = None,
      tag: str | None = None,
      snapshot: str | None = None,
  ) -> list[SnapshotMetadata]: ...

```

Closes: #572


mpiannucci (approved):
One thought is that previous implementations could operate these types of queries as `ref=` instead of specifying branch, tag or snapshot. Do you see value in that?

paraseba:
> One thought is that previous implementations could operate these types of queries as `ref=` instead of specifying branch, tag or snapshot. Do you see value in that?

Yes. I went with this because it's the more "natural" operation in IC. To be able to distinguish between a tag and a a branch IC would have to try both and see what comes back. I still thing there is value in that, we should add that kind of API too.

PR #579, opened 2025-01-15
Split and download virtual chunks using concurrent requests

paraseba:
It applies only if the chunk is large, and respects the StorageSettings.


icechunk-python/tests/test_can_read_old.py
```diff
@@ -105,11 +105,17 @@ async def write_a_test_repo() -> None:
...
+    buffer_prototype = zarr.core.buffer.default_buffer_prototype()
+    chunk = await store.get("group1/big_chunks/c/0/1", prototype=buffer_prototype)
+    virtual_chunk_data_size = len(chunk.to_bytes())
+
     store.set_virtual_ref(
         "group1/big_chunks/c/0/0",
         "s3://testbucket/can_read_old/chunk-1",
         offset=0,
-        length=5 * 5 * 4,
+        length=virtual_chunk_data_size,
```
paraseba, line 118: This bug is the only reason why I need to regenerate the repository. It was a bug generating it, we add a virtual ref with the wrong size. The new code detects this.

mpiannucci (approved):

PR #584, opened 2025-01-15
Some initial work towards better python __repr__

paraseba:


mpiannucci:
deferring this review to @dcherian

paraseba:
@mpiannucci @dcherian This is far from done, but I think it's mergeable. I don't know what approach we will take for repr in enum classes for example. But this at least does part of the job.

mpiannucci (approved):

icechunk-python/src/config.rs
```diff
@@ -70,18 +71,62 @@ impl PyS3StaticCredentials {
...
         session_token: Option<String>,
         expires_after: Option<DateTime<Utc>>,
     ) -> Self {
         Self { access_key_id, secret_access_key, session_token, expires_after }
     }
+
+    pub fn __repr__(&self) -> String {
+        // TODO: escape
+        format!(
+            r#"S3StaticCredentials(access_key_id="{ak}", secret_access_key="{sk}", session_token={st}, expires_after={ea})"#,
```
dcherian, line 86: do we want this? A user might end up printing this in a notebook and committing it... Since it's static the expiration time might be quite long.

paraseba: maybe we replace with `*`?

PR #589, opened 2025-01-17
Mechanism to deserialize different spec versions

paraseba:
Also provides a mechanism do separate the serializable datastructure from the in memory one.

To understand the mechanism read the documentation at the top of `serializers/mod.rs`.

After this merges, I'll try to use the mechanism to implement the first backwards-compatible on-disk format change.

### DIFF
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -12,6 +12,10 @@ use crate::{
     format::{
         format_constants::{self, CompressionAlgorithmBin, FileTypeBin, SpecVersionBin},
         manifest::Manifest,
+        serializers::{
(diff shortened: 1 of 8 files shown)


mpiannucci (approved):
Looks amazing. One impl question: could say in icechunk alpha 13, the snapshot format changes but the manifest and others does not. Will all files require `current` to be the same or can some have a `current` of alpha 12 and others alpha 13?

paraseba:
> Looks amazing. One impl question: could say in icechunk alpha 13, the snapshot format changes but the manifest and others does not.

I will explore this type of question tomorrow. My first thought is: current re-exports what it wants. In your example, current for version 13 defines SnapshotSeralizer but reexports ManifestSerializer from version_12.

PR #593, opened 2025-01-19
Manifest split design document

paraseba:


design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+We define the closure of a set of arrays. In plain English, the closure of a
+set of arrays `A` is the set of all arrays that would need their manifest
+rewritten if we update all arrays in `A`. We are going to re-pack all arrays
+in the closure of the set of modified arrays.
+
+Example:
+
+Manifest 1 has arrays: a, b
+Manifest 2 has arrays: a, b, c
+Manifest 3 has arrays: c
```
dcherian, line 238: Feels like we'd gain a large amount of simplicity by disallowing that kind of packing.

For example, 
```
Manifest 1 has arrays: a, b
Manifest 2 has arrays: a, b
Manifest 3 has arrays: a  # b is over; c is not allowed
Manifest 4 has arrays: c
Manifest 5 has arrays: c
```

I'm not sure we gain a lot by allowing the  packing of `a,c` in Manifest 3 (assuming that matches the rules). Indeed we might be better off relaxing the threshold to fit `a` between 1,2 by being looser about enforcing `max-manifest-size`. Something like if the overflow of `a` in Manifest 3 is <20% of `max-manifest-size` then just redistribute over preceding manifests.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+
+    - coord2:
+        max-manifest-size: 10000
+                                  # overflow-to: default       # by default
+                                  # arrays-per-manifest: null  # by default
+                                  # cardinality: 1             # by default
+
+    - big-array:
+        max-manifest-size: null
+        arrays-per-manifest: 1
```
dcherian, line 132: I realized that `arrays-per-manifest`  is useful for building up the big ERA5 (or similar) dataset sequentially array-by-array.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+* When they need to, because their repo is very particular or very large, they can
+* It's not hard to implement (see algorithm below). At least not much harder
+than any other mechanism. The difficulties lie in generating the split
+manifests, and that is needed with any system. The extra work of classifying what arrays
+go where is simple.
+
+To understand how the feature works, we show an example configuration. Comments
+in the yaml file explain what the different settings do.
+
+```yaml
```
dcherian, line 104: I think we should write out some "optimal" configs for a few use-cases:
1. ERA5 ingest : 250 arrays with 10 million chunks each.
2. updating forecast dataset like HRRR.
3. A sparsely populated datacube where updates occur in specific geographic regions.

That might help constrain the design space.

dcherian (commented):
This looks pretty thorough to me but I am a little concerned about complexity.

I left a few more detailed comments near the end, will think more about this tomorrow

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+chunk-manifests:
+  sets:                           # the lists of manifest sets and their properties
+                                  # it's a list because order is important
+
+    - coord1:                     # a name for the set, used later with rules
+
+        max-manifest-size: 10000  # this set will contain manifests with <= 10k references
```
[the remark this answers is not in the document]
paraseba: I don't know, I think we don't need to? They can just setup a larger limit, we need to cut somewhere.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+
+        arrays-per-manifest: null # pack this many arrays in each manifest
```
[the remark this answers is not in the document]
paraseba: I'd set max-manifest-size to 1 if I knew I always read (and or write) that array independently of others, and I care about latency. A service that executes queries against the data could care for example. If you are servicing a couple of variables, you may want to say "this specific arrays, put them in their own manifest" to minimize latency.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+                                  # max-manifest-size and arrays-per-manifest are
+                                  # mutually exclusive
+
+        overflow-to: coord2       # if an array cannot fit in the set, send it to set coord2
```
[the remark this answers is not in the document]
paraseba: I'm not entirely sure we need it, but it's so easy to implement (si if/else in the algorithm pseudo-code). We can skip it, replace it, etc .... just a gathering of ideas for now, trying to go for maximum power before we cut

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+                                  # if no overflow-to is indicated it goes to
+                                  # the default set (see below)
+
+        cardinality: 2            # this set cannot have more than two manifests
```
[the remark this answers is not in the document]
paraseba: It's explained in the pseudo-code i think. If you have more arrays than can fit in the cardinality, you "overflow" them to `overflow-to`

design-docs/005-manifest-split.md
paraseba: Yes, that too, but also to minimize latency if you are only reading from a couple of variables, which I suspect is very common.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+                                # default always has cardinality = null
+
+  rules:                        # what arrays go to each manifest set, an ordered list
+
+    # each rule has a target (the manifest set) and a set of conditions that are and-ed
+    # rules are applied in the order declared, and they break on the first match
+
+    - path: .*/(latitude|longitude|time)  # an optional regex matching on path
+
+      metadata-chunks: [0, 500]           # arrays having number of chunks in this range
```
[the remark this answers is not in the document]
paraseba: (not committing this changes to apply the change everywhere)

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+    - metadata-chunks: [0, 5000]
+      path: .*
+      target: coordinates
+
+
+  # should we actually do this by default or is no-preload a better default?
+  preload:
+    max-manifest-size: 50000
+    max-manifests: 1
+    arrays:
```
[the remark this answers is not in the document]
paraseba: That is perfect.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+We define _manifest sets_. A manifest set is a group of individual array manifests that
+are stored together allowing a user to optimize for certain types of arrays. We let the user define
+the sets, but we offer good defaults that should work reasonably well in
+common cases.
+
+We let the user indicate _rules_ that assign arrays to
+a specific manifest set. These rules can be based on array paths or number
+of chunks. In the future we could add more power to the rules system.
+
+Other new feature we provide is the ability to preload certain manifests.
```
rabernat, line 82: Interesting.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+These manifests will be fetched in the background as soon as a session is
+created, speeding up interactive use. In the future, we may add the ability
+to prefetch manifests for certain branches/tags as soon as the Repository is
+open.
+
+Manifest sets, array rules, and manifest prefetch, are configured in the
+persistent configuration of the repository and can be overloaded on open as
+any other configuration value.
```
rabernat, line 90: How is the default configuration provided?

rabernat: And what happens if the _configuration is changed_? Will all of the manifests be rewritten?

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+arrays needs to be downloaded and parsed.
+* Each commit needs to rewrite the full manifest.
+* Space is wasted, both on disk and in memory, manifest become too large
+(~10MB per million virtual references, compressed on disk).
+
+## What we want
+
+* Icechunk to be fast for interactive usage, when a user is exploring a dataset
+* Small reads not to parse the whole manifest
+* Small writes not to rewrite the whole manifest
```
rabernat, line 32: Good set of requirements.

rabernat (commented):
This is an extremely thoughtful exploration of this complex issue.

My initial reaction was the same as Deepak's...this is creating a lot of complexity for us to implement and maintain. I'm trying to imagine the test suite for the features described here, and it's making my head hurt!

However, if you feel like you can see the path forward here, I'm happy to support it. I would encourage trying to whittle this down to the minimum possible set of capabilities needed to meet the requirements.

design-docs/005-manifest-split.md
paraseba: I think this will help significantly

design-docs/005-manifest-split.md
paraseba: I'll explain in the document, i missed this part. No, we won't rewrite until they are rewritten by a commit. In the future we could offer repacking the manifest as a explicit optimization operation.

design-docs/005-manifest-split.md
```diff
@@ -0,0 +1,406 @@
...
+        overflow-to: null
+
+  rules:
+    - metadata-chunks: [0, 5000]
+      path: .*
+      target: coordinates
+
+
+  # should we actually do this by default or is no-preload a better default?
+  preload:
```
dcherian, line 204: Actually, can we allow the user to preload the `coord1` set. Seems silly to duplicate info like the `path` regex in both `preload` and `rules`

paraseba: We could ... but currently the manifest sets are purely tacit. They don't persist in any way, it's purely a way to group them during flush. But of course, we could persist them. 

But, it feels less powerful? How about having a condition on manifest size too? so you could say "preload up to 5 small manifest" or "preload manifest for this array"

design-docs/005-manifest-split.md
dcherian: ### ERA5 bulk update
Goals:
1. Put coordinate arrays (time, latitude, longitude, level) in one manifest
2. One data var per manifest file, so that I can incrementally build up the archive (commit after 
   ingesting one var).
The `arrays-per-manifest: 1` config is great for this use case.
 ```yaml
 chunk-manifests:
  sets:
    - coordinates:                # a name for the set, used later with rules
        max-manifest-size: 10000  # this set will contain manifests with <= 10k references
                                  # large enough to hold all necessary time coordinate references
                                  # 60 years * (365*24) hours = 525,600 elements

        arrays-per-manifest: null # pack this many arrays in each manifest
                                  # max-manifest-size and arrays-per-manifest are
                                  # mutually exclusive
    - big-arrays:
        max-manifest-size: null
        arrays-per-manifest: 1
        cardinality: null

    - default:                  # default set must always exist, from defaults if needed

  rules:                        # what arrays go to each manifest set, an ordered list

    # each rule has a target (the manifest set) and a set of conditions that are and-ed
    # rules are applied in the order declared, and they break on the first match

    - path: .*/(latitude|longitude|time)  # an optional regex matching on path
      metadata-chunks: [0, 5000]          # arrays having number of chunks in this range
      target: coord1                      # arrays that match will go to this manifest set

    - metadata-chunks: [200000, null]    # huge arrays have their special manifest set
      target: big-array


  preload:                      # what manifests to asynchronously preload on session start
    max-manifest-size: 50000    # don't preload manifests larger than this
    max-manifests: 10           # don't preload more than this many manifests
    arrays:
      - path: .*/2m_temperature # will preload manifests for all arrays matching this regex
```

### GFS: Forecast-type appending dataset    
- start a new manifest every n time-steps?
- preload the most recent one? 
- definitely want multiple arrays in a single manifest file

 ```yaml
 chunk-manifests:
  sets:
    - coordinates:                # a name for the set, used later with rules
        max-manifest-size: 10000  # this set will contain manifests with <= 10k references
                                  # large enough to hold all necessary time coordinate references
                                  # 60 years * (365*24) hours = 525,600 elements

        arrays-per-manifest: null # pack this many arrays in each manifest
                                  # max-manifest-size and arrays-per-manifest are
                                  # mutually exclusive
    - appended-arrays:
        max-manifest-size: 4400   # 20 chunks per array for a single update, 4 updates per day
                                  # this gets us 5 days of updates in a single manifest file
        arrays-per-manifest: 10
        cardinality: null

    - default:                  # default set must always exist, from defaults if needed

  rules:                        # what arrays go to each manifest set, an ordered list

    # each rule has a target (the manifest set) and a set of conditions that are and-ed
    # rules are applied in the order declared, and they break on the first match

    - path: .*/(latitude|longitude|time|level)  # an optional regex matching on path
      metadata-chunks: [0, 5000]          # arrays having number of chunks in this range
      target: coord1                      # arrays that match will go to this manifest set

    - metadata-chunks: [200000, null]    # huge arrays have their special manifest set
      target: big-array


  preload:                      # what manifests to asynchronously preload on session start
    max-manifest-size: 50000    # don't preload manifests larger than this
    max-manifests: 10           # don't preload more than this many manifests
```

design-docs/005-manifest-split.md
dcherian: Another future use-case is to preload the most recent manifest for multiple arrays (in the case of an updating forecast dataset)

PR #595, opened 2025-01-19
Evict asset caches based on size

paraseba:
This breaks existing repos because we change the config names


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -77,32 +77,32 @@ class CachingConfig:
 
     def __init__(
         self,
-        snapshots_cache_size: int,
-        manifests_cache_size: int,
-        transactions_cache_size: int,
-        attributes_cache_size: int,
-        chunks_cache_size: int,
+        snapshots_nodes: int,
```
dcherian, line 80: I'd like the user-facing parameter names be easier to interpret (i didn't understand the units till I read the code). So how about
```
num_nodes
num_manifests
num_transaction_logs
nbytes_attributes
nbytes_chunks
```

icechunk/src/asset_manager.rs
```diff
@@ -31,19 +31,19 @@ use crate::{
 pub struct AssetManager {
     storage: Arc<dyn Storage + Send + Sync>,
     storage_settings: storage::Settings,
-    num_snapshots: u16,
-    num_manifests: u16,
-    num_transactions: u16,
-    num_attributes: u16,
-    num_chunks: u16,
+    num_snapshot_nodes: u64,
```
dcherian, line 34: these names are great!

icechunk-python/src/config.rs
```diff
@@ -492,15 +492,15 @@ impl From<&PyCompressionConfig> for CompressionConfig {
...
+    pub manifests_refs: u64,
     #[pyo3(get, set)]
-    pub transactions_cache_size: u16,
+    pub transactions_records: u64,
     #[pyo3(get, set)]
-    pub attributes_cache_size: u16,
+    pub attributes_bytes: u64,
     #[pyo3(get, set)]
-    pub chunks_cache_size: u16,
+    pub chunks_bytes: u64,
```
dcherian, line 503: Why switch from `u16` to `u64`? I guess 2**16 or 2**32 bytes is too small? A "docstring" comment saying we expect the user to set "reasonable values" would be useful

dcherian (approved):

PR #605, opened 2025-01-22
Efficient getsize in the Zarr store

paraseba:
This implements a natie version of:

```python
async def getsize(self, key: str) -> int: ...
```

Notably, it doesn't implement `getsize_prefix`. We think letting zarr-python do the loop that calls `getsize` can be enough performance.


paraseba:
Codespell will approve once we merge my manifest open PR that has all these fixes

icechunk-python/python/icechunk/store.py
```diff
@@ -332,3 +332,6 @@ def list_dir(self, prefix: str) -> AsyncIterator[str]:
         # listing methods should not be async, so we need to
         # wrap the async method in a sync method.
         return self._store.list_dir(prefix)
+
+    async def getsize(self, key: str) -> int:
```
dcherian, line 336: there is a `getsize_prefix` too: https://zarr.readthedocs.io/en/stable/api/zarr/abc/store/index.html#zarr.abc.store.Store.getsize_prefix

paraseba: yeah, see comment in the PR ... I want to try if we can skip implementing those. If performance is not there we can implement.

dcherian: AH ok. I'll push a benchmark, then approve.

dcherian:
Phew after some work:

```
-------- benchmark 'test_time_getsize_key era5-single': 2 tests --------
Name (time in us)                                       Median
------------------------------------------------------------------------
test_time_getsize_key[era5-single] (0034_main_3a)     110.2501 (1.05)
test_time_getsize_key[era5-single] (NOW)              104.7080 (1.0)
------------------------------------------------------------------------

-------- benchmark 'test_time_getsize_key gb-128mb': 2 tests --------
Name (time in us)                                    Median
---------------------------------------------------------------------
test_time_getsize_key[gb-128mb] (0034_main_3a)     106.5000 (1.05)
test_time_getsize_key[gb-128mb] (NOW)              101.2499 (1.0)
---------------------------------------------------------------------

-------- benchmark 'test_time_getsize_key gb-8mb': 2 tests --------
Name (time in us)                                  Median
-------------------------------------------------------------------
test_time_getsize_key[gb-8mb] (0034_main_3a)     105.6660 (1.05)
test_time_getsize_key[gb-8mb] (NOW)              100.8749 (1.0)
-------------------------------------------------------------------

------- benchmark 'test_time_getsize_prefix era5-single': 2 tests --------
Name (time in ms)                                         Median
--------------------------------------------------------------------------
test_time_getsize_prefix[era5-single] (0034_main_3a)     68.8355 (31.10)
test_time_getsize_prefix[era5-single] (NOW)               2.2133 (1.0)
--------------------------------------------------------------------------
```

So big difference in `getsize_prefix`. 

No difference in`getsize(key)` because I am only testing `getsize` on an `array.json` which is already in memory.

dcherian (approved):

icechunk-python/benchmarks/README.md
```diff
@@ -38,6 +38,56 @@ This takes about 5 minutes to run (depending on cloud scaling time).
...
+# Start on the PR branch
+git switch push-uknyqnpypzro
+# this will build the main branch, and re-create the datasets
+# And uses the pytest -k option to only run the new benchmarks
+python benchmarks/runner.py --pytest "-k getsize" main
+# It will print this to the screen
+# > pytest [...] --benchmark-save=main_3abfa48a --icechunk-prefix=benchmarks/main_3abfa48a/  benchmarks/
+# note the created prefix: main_(first-8-characters-of-commit), for convenienve export it
+export PREFIX=benchmarks/main_3abfa48a/
+pytest --benchmark-compare -k getsize --benchmark-group-by=group,func,param --benchmark-columns=median --benchmark-sort=name --icechunk-prefix=$PREFIX benchmarks/
```
dcherian, line 57: struggled with making a just command for this

PR #613, opened 2025-01-23
Better defaulting mechanism for configuration

paraseba:
Now configurations can be merged, this means users can pass only the configuration parameters they want to override, not the full object. With this, users don't have to fetch the configuration to be able to open a repository with some config changes.

Example:

```rust
Repository::open(
  ...,
  config=RepositoryConfig{inline_chunk_threshold_bytes=0, ..Default::default()})
```

Will override only `inline_chunk_threshold_bytes` and keep the rest of the configuration as stored in the repo (or the default config if no persistent configuration is found).

In addition, now we don't store the configuration for the repo on repo creation, unless it's different from the default. This allows us to know the user hasn't changed configuration, which will help with version evolution.

To achieve all this:

* Now all configuration parameters across all structs are optional
* We introduce a `merge` function for configurations


icechunk-python/tests/test_config.py
```diff
@@ -152,11 +156,18 @@ def test_can_change_deep_config_values() -> None:
...
+    config.compression = icechunk.CompressionConfig(level=8)
     config.compression.level = 2
-    config.caching.num_chunk_refs = 8
+    config.caching = icechunk.CachingConfig(num_chunk_refs=8)
     config.storage = storage.default_settings()
     config.storage.concurrency.ideal_concurrent_request_size = 1_000_000
 
+    assert (
+        repr(config)
+        == r"RepositoryConfig(inline_chunk_threshold_bytes=5, unsafe_overwrite_refs=True, get_partial_values_concurrency=42, compression=CompressionConfig(algorithm=None, level=2), caching=CachingConfig(num_snapshot_nodes=None, num_chunk_refs=8, num_transaction_changes=None, num_bytes_attributes=None, num_bytes_chunks=None), storage=StorageSettings(concurrency=StorageConcurrencySettings(max_concurrent_requests_for_object=5, ideal_concurrent_request_size=1000000)))"
```
mpiannucci, line 169: Nice!

mpiannucci (approved):

PR #621, opened 2025-01-24
Tag delete design document

paraseba:


design-docs/006-tag-delete.md
```diff
@@ -0,0 +1,91 @@
...
+tags to have the same name as a deleted tag.
+
+When it's time to resolve a tag `foo`, we fetch in parallel two keys:
+
+```
+refs/tag.foo/ref.json
+refs/tag.foo/ref.json.deleted
+```
+
+And wait for both to resolve. If `ref.json.deleted` is not found, we use
```
mpiannucci, line 82: 👍

design-docs/006-tag-delete.md
```diff
@@ -0,0 +1,91 @@
...
+somebody can edit `v42` and move it to a different snapshot,
+there are no longer any guarantees.
+
+## The problems
+
+* If a user makes a spelling mistake, or they assign a tag to the wrong
+snapshot, they cannot fix it. This is a minor problem, and we made
+an explicit trade-off here. The benefits justify this minor inconvenience.
+The will have to create a new, different tag.
+* In a world in which we want to expire on squash history, tags will have to
```
mpiannucci, line 23: Could all history except for tagged snapshots be expired?

paraseba: It could. One of the trickiest things in this topic, in my experience, is to decide how to define the multiple ways in which a user could want to expire. By commit age, by branches, by tags, and everything in between.

Initially I'm thinking something like "expire all commits older than X". If there are tags pointing to expirable snapshots, we would give them the option to fail or delete them.

design-docs/006-tag-delete.md
```diff
@@ -0,0 +1,91 @@
...
+from an Icechunk session initiated on a tag:
+```python
+repo = icechunk.Repository.open(...)
+session = repo.readonly_session(tag="v42")
+store = session.store
+```
+
+All data read from `store` will be immutable, it can be cached for as
+long as needed. This is only true if tags cannot be edited. If
+somebody can edit `v42` and move it to a different snapshot,
```
dcherian, line 14: if we allow delete, then we effectively allow this workflow. We should warn somewhere that any caching should rely on snapshot ID, not tag name.

design-docs/006-tag-delete.md
```diff
@@ -0,0 +1,91 @@
...
+  1. Create tag `foo` pointing to snapshot `ABC123`
+  1. Delete tag `foo`
+  1. Create tag `foo` pointing to snapshot `DEF456`
+
+  This is essentially tag edit, which we explicitly want to avoid, as
+  explained before.
+
+## Design
+
+We implement logic delete for tags. Tags will be flagged
```
dcherian, line 48:
```suggestion
We implement logical delete for tags. Tags will be flagged
```

design-docs/006-tag-delete.md
```diff
@@ -0,0 +1,91 @@
...
+refs/tag.foo/ref.json.deleted
+```
+
+And wait for both to resolve. If `ref.json.deleted` is not found, we use
+the contents of `ref.json` to find the snapshot. If `ref.json.deleted`
+is found we carry on as if `ref.json` was not found.
+
+### Optional improvement
+
+On tag creation, we may chose to offer different error messages
```
dcherian, line 88: yes sounds good.

dcherian (approved):

PR #625, opened 2025-01-24
Experimental support for Tigris

paraseba:
Tigris can now be used for Icechunk repositories and for virtual chunk containers.

Warning: Currently, if using Tigris for the repo storage, there are two behaviors that are broken:

* When creating a repository an assertions can sometimes trigger if the request is done from a region geographically far from where the data is. The repository is created successfully.
* There will be a delay between when commit is done and when it's visible in a branch. This breaks our consistency guarantees.

We are exploring ways to deal with the issues above.

### DIFF
--- a/icechunk-python/python/icechunk/__init__.py
+++ b/icechunk-python/python/icechunk/__init__.py
@@ -60,6 +60,7 @@ from icechunk.storage import (
     local_filesystem_storage,
     s3_storage,
     s3_store,
+    tigris_storage,
 )
 from icechunk.store import IcechunkStore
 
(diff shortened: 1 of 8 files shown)


mpiannucci (approved):
This loloks good, my one point is that this is confusing because tigris is really just s3 compatible like minio so it really isnt necessary to have this is it? Just convenient? Because on current main you can use the s3 client with tigris with the same level of support. 

I think we should probably document **why** you would want to use this over the s3 storage, since they are functionally the same

paraseba:
> Just convenient? 

In the storage layer yes, just for convenience, so people don't need to know Tigris endpoint url. On the virtual chunk container side, we have a default container that maps to the prefix `tigris://`, this change makes that work.

> I think we should probably document why you would want to use this over the s3 storage, since they are functionally the same

Let's wait for answer for Tigris and see. We may need to special case stuff based on incompatibilities.

PR #626, opened 2025-01-24
Icechunk learns how to delete tags.

paraseba:
See [design document](https://github.com/earth-mover/icechunk/blob/main/design-docs/006-tag-delete.md)


icechunk-python/tests/test_can_read_old.py
```diff
@@ -195,6 +197,9 @@ async def test_icechunk_can_read_old_repo() -> None:
         1:
     ]
 
+    with pytest.raises(ValueError, match="ref not found"):
```
mpiannucci, line 200: Sounds we make a new subclass of IcechunkError for this while working on this?

mpiannucci (approved):

icechunk-python/tests/test_can_read_old.py
paraseba: I think our whole error hierarchy needs a revamp. Let me delay this until we do that rfactoring.

PR #635, opened 2025-01-27
Uncache history

paraseba:
We had a cache in each snapshot with pointers to all the snapshot ancestry. This makes expiration harder, because we need to invalidate these caches.

We are replacing the cache by a single `parent_id` link. In the future, we may optimize this for performance.

Notice: python is still turning the stream of parents into a python list. This is now much worse because the new ancestry version is much slower. We'll need to change this to use an AsyncGenerator.


icechunk/src/format/snapshot.rs
```diff
@@ -157,95 +150,153 @@ pub struct AttributeFileInfo {
...
-    pub short_term_history: VecDeque<SnapshotMetadata>,
-
-    pub metadata: SnapshotMetadata,
-    pub started_at: DateTime<Utc>,
-    pub properties: SnapshotProperties,
-    pub(crate) nodes: BTreeMap<Path, NodeSnapshot>,
+    id: SnapshotId,
+    parent_id: Option<SnapshotId>,
+    flushed_at: DateTime<Utc>,
+    message: String,
```
dcherian, line 156: nice!

icechunk/src/format/snapshot.rs
```diff
@@ -157,95 +150,153 @@ pub struct AttributeFileInfo {
...
 
-    pub fn empty() -> Self {
-        let metadata =
-            SnapshotMetadata::with_message(Self::INITIAL_COMMIT_MESSAGE.to_string());
-        Self {
-            metadata,
-            ..Self::new(VecDeque::new(), 0, None, Default::default(), vec![], vec![])
-        }
+    pub fn initial() -> Self {
+        let properties = [("__root".to_string(), serde_json::Value::from(true))].into();
```
dcherian, line 256: Why not empty?

paraseba: I thought ... empty is not the main characteristic, the important part is that it is empty and has the commit message "repo initialized'

dcherian (approved):

icechunk/src/format/snapshot.rs
dcherian: Sorry I was talking about `properties` here, why not set it to `None`?

paraseba: I don't know... I thought we could start putting some stuff there, and reserving the `__` namespace.

PR #636, opened 2025-01-28
Implement expiration of snapshots

paraseba:
Also fix a bug where we were listing deleted tags.

Some minimal refactoring too.

### DIFF
--- a/icechunk/src/format/snapshot.rs
+++ b/icechunk/src/format/snapshot.rs
@@ -299,6 +299,20 @@ impl Snapshot {
         &self.attribute_files
     }
 
+    /// Cretase a new `Snapshot` with all the same data as `self` but a different parent
+    pub fn adopt(&self, parent: &Snapshot) -> Self {
+        Self {
+            id: self.id.clone(),
+            parent_id: Some(parent.id().clone()),
+            flushed_at: *self.flushed_at(),
+            message: self.message().clone(),
+            metadata: self.metadata().clone(),
+            manifest_files: self.manifest_files().clone(),
+            attribute_files: self.attribute_files().clone(),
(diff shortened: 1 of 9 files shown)


paraseba:
Some tasks that will come in future PRs:
* python bindings
* timestamp monotonicity validation
* snapshot timestamp check against object store clock

icechunk/src/format/snapshot.rs
```diff
@@ -299,6 +299,20 @@ impl Snapshot {
         &self.attribute_files
     }
 
+    /// Cretase a new `Snapshot` with all the same data as `self` but a different parent
+    pub fn adopt(&self, parent: &Snapshot) -> Self {
```
mpiannucci, line 303: I love the name

icechunk/src/ops/gc.rs
```diff
@@ -321,3 +321,124 @@ async fn gc_snapshots(
...
     Ok(storage.delete_snapshots(storage_settings, to_delete).await?)
 }
+
+/// Expire snapshots older than a threshold.
+///
+/// This only processes snapshots found by navigating `reference`
+/// ancestry. Any other snapshots are not touched.
+///
+/// The operation will edit in place the oldest non-expired snapshot,
+/// changing its parent to be the root of the repo.
```
mpiannucci, line 331: This means the initial commit snapshot is always the parent right?

mpiannucci (approved):

mpiannucci:
This is so cool and really easy to read

icechunk/src/ops/gc.rs
paraseba: Exactly, yes, the root of the repo will become the new parent of the first non-expired snapshot

PR #641, opened 2025-01-28
Python expiration and GC

paraseba:


mpiannucci (approved):
👏

rabernat:
In the interest of not growing our "documentation debt," would it be worthwhile to consider documenting these features via a narrative documentation page now, as part of the PR, while it's all fresh?

We could reuse the example from the design doc.

PR #645, opened 2025-01-29
Implement Python's Repository.async_ancestry

paraseba:
This allows to do more efficient ancenstry from python in an async context. Also pagination is enabled because we return an AsyncIterator and we take snapshot ids.

### DIFF
--- a/icechunk-python/tests/test_commit_properties.py
+++ b/icechunk-python/tests/test_commit_properties.py
@@ -1,5 +1,7 @@
 from datetime import UTC, datetime
 
+import pytest
+
 import icechunk as ic
 import zarr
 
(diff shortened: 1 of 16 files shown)


icechunk-python/tests/test_commit_properties.py
```diff
@@ -0,0 +1,33 @@
...
+    props = {
+        "string": "foo",
+        "true": True,
+        "none": None,
+        "int": 42,
+        "float": 42.0,
+        "list": ["hello", 42],
+        "object": {"foo": "bar", "baz": [1, 5], "inner": {"abc": None, "false": False}},
+    }
+    snapshot_id = session.commit("some commit", props)
```
paraseba, line 26: This is one of the points of the PR

icechunk-python/tests/test_timetravel.py
```diff
@@ -110,6 +115,17 @@ def test_timetravel() -> None:
     assert list(repo.ancestry(tag="v1.0")) == parents
     assert list(repo.ancestry(branch="feature-not-dead")) == parents
 
+    # check async ancestry works
+    assert list(repo.ancestry(snapshot=feature_snapshot_id)) == asyncio.run(
+        async_ancestry(repo, snapshot=feature_snapshot_id)
+    )
```
paraseba, line 121: This is the other point of the PR, we can retrieve ancestry using an async generator and starting on any arbitrary snapshot

mpiannucci (approved):

PR #647, opened 2025-01-29
Remove some cruft from the on disk datastuctures

paraseba:
This is a backwards incompatible change


icechunk/src/asset_manager.rs
```diff
@@ -559,7 +559,7 @@ async fn write_new_tx_log(
 
     let buffer = tokio::task::spawn_blocking(move || {
         let buffer = binary_file_header(
-            SpecVersionBin::V0_1_0Alpha12,
+            SpecVersionBin::V0dot1,
```
dcherian, line 562: why not current here?

paraseba: you are right, will change.

paraseba: done

dcherian (approved):

PR #649, opened 2025-01-30
Implement manifest preloading

paraseba:
This creates a pretty advanced mechanism to configure what manifest should be preloaded on session creation.

It can match arrays by name, path or size; and it can create complex conditions using `or` and `and` combinators.

By default, we preload manifests that have not more than 1k refs, up to a total of 10k refs, and corresponding to arrays with names matching any of the following regexes:

* `\bt\b|(time|min|hour|day|week|month|year)[0-9]*`
* `(z|nav_lev|gdep|lv_|[o]*lev|bottom_top|sigma|h(ei)?ght|altitude|depth|isobaric|pres|isotherm)[a-z_]*[0-9]*`
* `y|j|nlat|rlat|nj`
* `y?(nav_lat|lat|gphi)[a-z0-9]*`
* `x?(nav_lon|lon|glam)[a-z0-9]*`
* `x|i|nlon|rlon|ni`

The regex were taken from:

https://github.com/xarray-contrib/cf-xarray/blob/1591ff5ea7664a6bdef24055ef75e242cd5bfc8b/cf_xarray/criteria.py#L149-L160


paraseba:
Thank you @dcherian for showing me this list of regexes! Exactly what I needed

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -112,6 +112,59 @@ class CachingConfig:
     @staticmethod
     def default() -> CachingConfig: ...
 
+class ManifestPreloadCondition:
+    """Configuration for conditions under which manifests will preload on session creation"""
+
+    @staticmethod
+    def or_conditions(
```
mpiannucci, line 119: Can we add doc strings for all these methods?

paraseba: done

Issue #448, opened 2024-12-05
Chunk initialization information from IceChunk manifests


DahnJ:
# Context
In https://github.com/zarr-developers/zarr-specs/issues/300 I roughly outlined @SylveraIO's infrastructure to support incrementally populated zarr arrays and a potential road to aligning it with open-source initiatives like IceChunk and [VirtualiZarr](https://github.com/zarr-developers/VirtualiZarr).

The full details can be read in the issue. To sum up, there are two necessary pieces:
1. On-disk storage format for chunk initialization information
2. API to query this information

IceChunk manifests are a potential solution for 1. This issue aims to validate that and, if necessary, steer the manifests such that they can efficiently serve this purpose.

# Query pattern
The crucial query pattern here is
> "Which chunks have been initialized in this region of the array?"

This is equivalent to a `list_prefix` operation that lists only the keys in a specific range.

A possible first iteration of this idea is fetching the results of `list_prefix` and then filter.

However, that may be impractical. Our arrays have 100s of millions and up to billions of total chunks, with 10s of millions chunks initialized. 

More factors to keep in mind
- Manifests will be transferred over a network (in our case from S3)
- The chunks can be committed over a large number of individual commits (say 1e3 to 1e5 chunks per commit)

# Benchmark
This is an initial modestly-sized benchmark to have something to share, will follow up with more benchmarking.

```python
import icechunk
import zarr
from tqdm import tqdm

storage_config = icechunk.StorageConfig.filesystem("./icechunk-large")
store = icechunk.IcechunkStore.create(storage_config)

# 1e3*1e3=1e6 chunks in total
nchunks = int(1e3)
pixels_per_chunk = 10
shape = (nchunks * pixels_per_chunk, nchunks * pixels_per_chunk)
chunks = (pixels_per_chunk, pixels_per_chunk)

group = zarr.group(store)
array = group.create("my_array", shape=shape, chunks=chunks, dtype=int)
store.commit('initialize')

# initialize 1e3*1e2=1e5 chunks
for i in tqdm(range(1000)):
    for j in range(100):
        array[i*10, j*10] = 1

store.commit('write data')

# du -sh icechunk-large/manifests/*
# 3.6M	icechunk-large/manifests/B3DBQNQ0J08HN10JGW5G

# this takes about ~7s on my machine
len([a async for a in store.list_prefix('')])
```

paraseba:
@DahnJ a few questions:

[...]

PR #651, opened 2025-01-30
Fetch initialized chunks

paraseba:
A new `chunk_coordinates` function returns all initialized chunk coordinates in an array. Currently the returned type is an async iterator that yields batches of chunk coordinates. The batching  speeds up the python-rust context switch.

I have measured on my laptop that we can retrieve approximately 500k chunk coordinates per second with the default batch size.

I believe, but I haven't tested, that the performance is (amortized) independent of:

* number of commits
* number of non initialized chunks avaliable in the array
* number of arrays in the repo

Issue: #448


icechunk-python/tests/test_concurrency.py
```diff
@@ -70,6 +70,9 @@ async def test_concurrency() -> None:
                     write_to_store(array, x, y, barrier), name=f"write {x},{y}"
                 )
 
+    all_coords = {coords async for coords in session.chunk_coordinates("/array")}
+    assert all_coords == {(x, y) for x in range(N) for y in range(N)}
```
dcherian, line 74: could be nice to add a test where `y` goes to `N-1`

dcherian (approved):

icechunk-python/tests/test_concurrency.py
paraseba: done

PR #633, opened 2025-01-27
Upgrade `xarray.to_icechunk` and related docs

dcherian:
https://icechunk--633.org.readthedocs.build/en/633/icechunk-python/xarray/

1. Supports `region="auto"`
2. Support writing DataArrays
3. support `safe_chunks`
4. updated docs on xarray/dask/distributed write patterns

Closes #606
xref https://github.com/pydata/xarray/pull/9990

PR #652, opened 2025-01-31
Require opt-in to pickling writable Sessions and Stores.

### LINKED ISSUE #478: Re-enable Distributed Writes protections when pickling `IcechunkStore`s in python

dcherian:
This is uglier than it used to because we cannot deserialize to a read-only Session or Store by default anymore.
Well we could, but it would take some extra work.

I can add some docs once #633 goes in

Closes #478
xref https://github.com/earth-mover/icechunk/issues/185#issuecomment-2489713313


icechunk-python/tests/test_xarray.py
```diff
@@ -50,20 +50,23 @@ def create_test_data(
...
 
-        # if allow_distributed_write:
-        #     with session.allow_distributed_write():
-        #       to_icechunk(data, store=session.store, mode="w")
-        # else:
-        #     to_icechunk(data, store=session.store, mode="w")
+        if allow_pickling:
+            with session.allow_pickling():
+                to_icechunk(data, store=session.store, mode="w")
+                with xr.open_zarr(session.store, consolidated=False) as ds:
```
dcherian, line 63: this is not so nice now. 

The distributed-read piece needs to be within the context manager now, because we can't (easily) set read-only status when unpickling. We could do so I think, by creating a brand new Session with branch=None. Let's discuss tomorrow..

icechunk-python/python/icechunk/session.py
```diff
@@ -86,26 +88,49 @@ class Session:
...
+    @contextlib.contextmanager
+    def allow_pickling(self) -> Generator[None, None, None]:
+        """
+        Context manager to allow unpickling this store if writable.
+        """
+        try:
+            self._allow_pickling = True
+            yield
+        finally:
+            self._allow_pickling = False
```
paraseba, line 133: Should finally return to the previous value? or is previous value always False?

paraseba: lol I'm very consistent in the ways I bother people

paraseba (approved):

PR #660, opened 2025-01-31
Release version 0.1.0

paraseba:


icechunk-python/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "icechunk-python"
-version = "0.1.0-alpha.15"
+version = "0.1.0"
```
dcherian, line 3: we need to change pyproject.toml too. it is wrong currently.

Changelog.python.md
```diff
@@ -1,5 +1,35 @@
...
+  ```
+  This properties can be retrieved via `ancestry`.
+- New `chunk_coordinates` function to list all initialized chunks in an array.
+- It's now possible to delete tags. New tags with the same name won't be allowed to preserve the immutability of snapshots pointed by a tag.
+- Safety checks on distributed writes via opt-in pickling of the store.
+- More safety around snapshot timestamps, blocking commits if there is too much clock drift.
+- Don't allow creating repositories in dirty prefixes.
+- Experimental support for Tigris object store: it currently requires the bucket to be restricted to a single region to obtain the Icechunk consistency guarantees.
+- This version is the first candidate for a stable on-disk format. At the moment, we are not planning to change the on-disk format prior to releasing v1 but reserve the right to do so.
+
```
dcherian, line 20:
```suggestion
### Breaking Changes
- Users must now opt-in to pickling and unpickling of Session and IcechunkStore using the `Session.allow_pickling` context manager
- `to_icechunk` now accepts a Session, instead of an IcechunkStore
```

Changelog.python.md
```diff
@@ -1,5 +1,37 @@
...
+  ```
+  This properties can be retrieved via `ancestry`.
+- New `chunk_coordinates` function to list all initialized chunks in an array.
+- It's now possible to delete tags. New tags with the same name won't be allowed to preserve the immutability of snapshots pointed by a tag.
+- Safety checks on distributed writes via opt-in pickling of the store.
+- More safety around snapshot timestamps, blocking commits if there is too much clock drift.
+- Don't allow creating repositories in dirty prefixes.
+- Experimental support for Tigris object store: it currently requires the bucket to be restricted to a single region to obtain the Icechunk consistency guarantees.
+- This version is the first candidate for a stable on-disk format. At the moment, we are not planning to change the on-disk format prior to releasing v1 but reserve the right to do so.
+### Breaking Changes
```
dcherian, line 20:
```suggestion

### Breaking Changes
```

mpiannucci (approved):

Issue #664, opened 2025-02-01
Google links for docs are broken


DahnJ:
When I [google icechunk docs](https://www.google.com/search?q=icechunk+docs) (or just [icechunk](https://www.google.com/search?q=icechunk)) I get sent to old URL structure ([1](https://icechunk.io/overview/), [2](https://icechunk.io/icechunk-python/quickstart/)) that doesn’t contain anything useful. I suppose these could redirect to the `latest/` pages.

## Images

![Image](https://github.com/user-attachments/assets/bd7f856b-f613-4197-aadc-6cd33df21c46)

<img width="1305" alt="Image" src="https://github.com/user-attachments/assets/5d23f434-498d-4c54-b56b-880fc77623cd" />

jhamman:
Thanks for the report @DahnJ. We switched doc hosting services yesterday so I'm willing to bet Google will sort this out in the next 24 hours. If it's still an issue on Monday, we'll look into it and find a fix.

DahnJ:
I just noticed that all the [readme links](https://github.com/earth-mover/icechunk/tree/main?tab=readme-ov-file#documentation-and-resources) also lead to the old structure

jhamman:
I think this has resolved now. @DahnJ - let us know if you find any more outdated links on Google.

DahnJ:
All the docs links for me still point to the old URL. Maybe my google is just slower 🐌 

<img width="1195" alt="Image" src="https://github.com/user-attachments/assets/91d5e362-78d6-45c1-9e05-bb9dc9ddebee" />

PR #666, opened 2025-02-02
Add Storage specific docs

mpiannucci:
Also update links in readme #664

### DIFF
--- a/docs/docs/icechunk-python/storage.md
+++ b/docs/docs/icechunk-python/storage.md
@@ -1,0 +1,214 @@
+# Storage
+
+Icechunk can be configured to work with both object storage and filesystem backends. The storage configuration defines the location of an Icechunk store, along with any options or information needed to access data from a given storage type.
+
+### S3 Storage
+
+When using Icechunk with s3 compatible storage systems, credentials must be provided to allow access to the data on the given endpoint. Icechunk allows for creating the storage config for s3 in three ways:
+
+=== "From environment"
+
+    With this option, the credentials for connecting to S3 are detected automatically from your environment.
+    This is usually the best choice if you are connecting from within an AWS environment (e.g. from EC2). [See the API](./reference.md#icechunk.s3_storage)
+
+    ```python
+    icechunk.s3_storage(
+        bucket="icechunk-test",
+        prefix="quickstart-demo-1",
+        from_env=True
+    )
+    ```
+
+=== "Provide credentials"
+
+    With this option, you provide your credentials and other details explicitly. [See the API](./reference.md#icechunk.s3_storage)
+
+    ```python
+    icechunk.s3_storage(
+        bucket="icechunk-test",
+        prefix="quickstart-demo-1",
+        region='us-east-1',
+        access_key_id='my-access-key',
+        secret_access_key='my-secret-key',
+        # session token is optional
+        session_token='my-token',
+        endpoint_url=None,  # if using a custom endpoint
+        allow_http=False,  # allow http connections (default is False)
+    )
+    ```
+
+=== "Anonymous"
+
+    With this option, you connect to S3 anonymously (without credentials).
+    This is suitable for public data. [See the API](./reference.md#icechunk.s3_storage)
+
+    ```python
+    icechunk.s3_storage(
+        bucket="icechunk-test",
+        prefix="quickstart-demo-1",
+        region='us-east-1,
+        anonymous=True,
+    )
+    ```
+
+=== "Refreshable Credentials"
+
+    With this option, you provide a callback function that will be called to obtain S3 credentials when needed. [See the API](./reference.md#icechunk.s3_storage)
+
+    ```python
+    def get_credentials() -> S3StaticCredentials:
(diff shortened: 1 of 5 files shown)


mpiannucci:
Do we need a note about express zone? #667

docs/docs/icechunk-python/storage.md
```diff
@@ -0,0 +1,222 @@
...
+    )
+    ```
+
+=== "Refreshable Credentials"
+
+    With this option, you provide a callback function that will be called to obtain S3 credentials when needed. [See the API](./reference.md#icechunk.s3_storage)
+
+    ```python
+    def get_credentials() -> S3StaticCredentials:
+        return icechunk.S3StaticCredentials(access_key_id="xyz", secret_access_key="abc")
```
paraseba, line 60: In this example we should include the `expires_at` field (or whatever is called), so people understand the use of this mechanism. Maybe also add a paragraph explaining this is useful in contexts where you want to provide only temporary credentials.

docs/docs/icechunk-python/storage.md
```diff
@@ -0,0 +1,222 @@
...
+#### AWS Limitations
+
+- Icechunk is currently incompatible with [S3 Express One Zone](https://aws.amazon.com/s3/storage-classes/express-one-zone/). See [this issue](https://github.com/earth-mover/icechunk/issues/667) for more discussion.
+
+#### Tigris
+
+[Tigris](https://www.tigrisdata.com/) is available as a storage backend for Icechunk. Functionally this storage backend is the same as S3 storage, but with a different endpoint. Icechunk provides a helper function specifically for [creating Tigris storage configurations](./reference.md#icechunk.tigris_storage). There are a few things to be aware of when using Tigris:
+
+- The `endpoint_url` parameter must be set to the Tigris endpoint. This is defaulted to `https://fly.storage.tigris.dev` but may be different depending on your Tigris configuration.
+- Tigris is a globally distributed object store by default. There are currently some bugs with consistency when the store is distributed across multiple regions. For now, it is recommended to use a single region for your Tigris store.
```
paraseba, line 79: Let's soften the language here, maybe ~bugs~ caveats, or something.

For the last sentence I'd say something like:

For now, to get the all the consistency guarantees Icechunk offers, you will need to setup your Tigris bucket as restricted to a single region, and paste a screenshot of how to do it.

docs/docs/icechunk-python/storage.md
```diff
@@ -0,0 +1,222 @@
...
+    secret_access_key='minio123',
+    endpoint_url='http://localhost:9000',
+    allow_http=True,
+```
+
+A few things to note:
+
+1. The `endpoint_url` parameter is set to the URL of the Minio server.
+2. If the Minio server is running over HTTP and not HTTPS, the `allow_http` parameter must be set to `True`.
+3. Even though this is running on a local server, the `region` parameter must still be set to a valid region. By default Minio uses `us-east-1`.
```
paraseba, line 103: I don't know about the default, we should try, but probably you can pass anything for the single node MinIO.

mpiannucci: adding a link to this: https://github.com/minio/minio/discussions/15063

docs/docs/icechunk-python/storage.md
mpiannucci: done

docs/docs/icechunk-python/storage.md
mpiannucci: done, let me know what you think

docs/docs/icechunk-python/storage.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### Tigris
+
+[Tigris](https://www.tigrisdata.com/) is available as a storage backend for Icechunk. Functionally this storage backend is the same as S3 storage, but with a different endpoint. Icechunk provides a helper function specifically for [creating Tigris storage configurations](./reference.md#icechunk.tigris_storage). There are a few things to be aware of when using Tigris:
+
+- The `endpoint_url` parameter must be set to the Tigris endpoint. This is defaulted to `https://fly.storage.tigris.dev` but may be different depending on your Tigris configuration.
+- Tigris is a globally distributed object store by default. The caveat is that Tigris does not currently support the full consistency guarantees when the store is distributed across multiple regions. For now, to get the all the consistency guarantees Icechunk offers, you will need to setup your Tigris bucket as restricted to a single region. This can be done by setting the region in the Tigris bucket settings:
```
paraseba, line 85:
```suggestion
- Tigris is a globally distributed object store by default. The caveat is that Tigris does not currently support the full consistency guarantees when the store is distributed across multiple regions. For now, to get all the consistency guarantees Icechunk offers, you will need to setup your Tigris bucket as restricted to a single region. This can be done by setting the region in the Tigris bucket settings:
```

docs/docs/icechunk-python/storage.md
paraseba: great

paraseba (approved):

PR #680, opened 2025-02-05
Better `Storage` errors

paraseba:
Some conditions weren't really errors, but expected results. Changing those cases richer result types.

### DIFF
--- a/icechunk/src/storage/mod.rs
+++ b/icechunk/src/storage/mod.rs
@@ -75,12 +75,6 @@ pub enum StorageError {
     S3DeleteObjectError(#[from] SdkError<DeleteObjectsError, HttpResponse>),
     #[error("error streaming bytes from object store {0}")]
     S3StreamError(#[from] ByteStreamError),
-    #[error("cannot overwrite ref: {0}")]
-    RefAlreadyExists(String),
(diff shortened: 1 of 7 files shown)


icechunk/src/storage/mod.rs
```diff
@@ -192,6 +186,30 @@ impl Reader {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum UpdateConfigResult {
+    Updated { new_etag: ETag },
+    NotOnLatestVersion,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum GetRefResult {
```
dcherian, line 202: this one doesn't seem like a big improvement over a plain Result?

icechunk/src/storage/mod.rs
```diff
@@ -192,6 +186,30 @@ impl Reader {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FetchConfigResult {
```
dcherian, line 190: I'm also confused. Don't all of these have a generic "network failed" error possibility (for example)? How does that get recorded?

dcherian (approved):
I like it.

Issue #523, opened 2024-12-30
Add refreshable credentials for GCS


mpiannucci:
If required, add refreshable creds the same as for s3

PR #687, opened 2025-02-06
Add GCS Refreshable Credentials

mpiannucci:
#523 

cc @monodeldiablo @lindseynield 

- [x] Wire through the expiration to object_store's GCP crendentialing?
- [x] How do we store this config, so that this can be rebuilt reliably from serialized state
- [x] Docs


mpiannucci:
This refactor is pretty much done, but for some reason the storage tests are now failing so i messed something up!

docs/docs/icechunk-python/storage.md
```diff
@@ -175,6 +175,22 @@ Icechunk can be used with [Google Cloud Storage](https://cloud.google.com/storag
         )
         ```
 
+=== "Refreshable Credentials"
+
+    With this option, you provide a callback function that will be called to obtain Gcs credentials when needed. This is useful for workloads that depend on retrieving short-lived credentials from GCS or similar authority, allowing for credentials to be refreshed as needed without interrupting any workflows. This works at a lower level than the other methods, and accepts a bearer token and expiration time. These are the same credentials that are created for you when specifying the service account file, key, or ADC. [See the API](./reference.md#icechunk.gcs_storage)
```
monodeldiablo, line 180:
```suggestion
    With this option, you provide a callback function that will be called to obtain GCS credentials when needed. This is useful for workloads that depend on retrieving short-lived credentials from GCS or similar authority, allowing for credentials to be refreshed as needed without interrupting any workflows. This works at a lower level than the other methods, and accepts a bearer token and expiration time. These are the same credentials that are created for you when specifying the service account file, key, or ADC. [See the API](./reference.md#icechunk.gcs_storage)
```

icechunk-python/python/icechunk/storage.py
```diff
@@ -228,6 +229,8 @@ def azure_storage(
 
     Parameters
     ----------
+    account: str
+        The account where the repository will store its data
```
monodeldiablo, line 233:
```suggestion
        The account to which the caller must have access privileges
```

monodeldiablo (approved):

docs/docs/icechunk-python/storage.md
```diff
@@ -175,6 +175,22 @@ Icechunk can be used with [Google Cloud Storage](https://cloud.google.com/storag
         )
         ```
 
+=== "Refreshable Credentials"
```
paraseba, line 178: great docs here!

lindseynield (approved):

icechunk/src/asset_manager.rs
```diff
@@ -823,7 +823,7 @@ mod test {
     async fn test_dont_fetch_asset_twice() -> Result<(), Box<dyn std::error::Error>> {
         // Test that two concurrent requests for the same manifest doesn't generate two
         // object_store requests, one of them must wait
-        let storage: Arc<dyn Storage + Send + Sync> = new_in_memory_storage()?;
+        let storage: Arc<dyn Storage + Send + Sync> = new_in_memory_storage().await?;
```
paraseba, line 826: there is probably some documentation we need to change like this too?

paraseba (approved):

icechunk/src/asset_manager.rs
mpiannucci: Only the rust side is async for this, i wrapped it on the python side for now

kylebarron:
In case you're curious, I'm taking a similar approach in obstore: https://github.com/developmentseed/obstore/pull/234 (the PR works; I just haven't implemented a "token cache" yet, so that the tokens only get fetched as they're close to expiration)

PR #689, opened 2025-02-07
Better error reporting and logs

paraseba:
We use the tracing and tracing-subscriber crates for this.

Error types now include the tracing span where they were captured. Converting from an error type to other must be done using `into` to preserve the span.

Errors are much richer now, including the span where the error was generated. To transfer this information to Python we use the exception message.

If for some reason python users don't want to connect the tracing mechanism, they can set the environment variable `ICECHUNK_NO_LOGS`. But this means exceptions won't include the span information in their messages.

There is a new optional `feature` in the Rust library that adds log enabling. `icechunk-python` uses the feature and enables logs in `__init__.py`.

Code using Icechunk as a Rust library is at liberty of calling the new Rust function `initialize_tracing` or not.

Logs are controlled by the `ICECHUNK_LOG` environment variable as described [here](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives).

Logs are sent to stdout and are not integrated with python logging.


icechunk-python/Cargo.toml
```diff
@@ -37,6 +37,7 @@ serde_json = "1.0.137"
 async-trait = "0.1.85"
 typetag = "0.2.19"
 serde = { version = "1.0.217", features = ["derive", "rc"] }
+miette = { version = "7.5.0", features = ["fancy"] }
```
paraseba, line 40: this is used purely to generate the exception message string, with a nice report of all the chain of exception causes, some colors, etc

icechunk-python/python/icechunk/__init__.py
```diff
@@ -134,3 +136,5 @@
     "s3_store",
     "tigris_storage",
 ]
+
+initialize_logs()
```
paraseba, line 140: Is this ok? Is this too much? I need to initialize the logging system, but I only want to do that in the "application", not in the Rust library, because users of the library may want to do something different. Since we don't have the entrypoint to the python process, I'm doing this on import of icechunk.

Thoughts?

icechunk-python/src/errors.rs
```diff
@@ -77,24 +82,32 @@ impl From<SessionError> for PyIcechunkStoreError {
...
+                ..
             }) => PyRebaseFailedError::new_err(PyRebaseFailedData {
                 snapshot: snapshot.to_string(),
                 conflicts: conflicts.iter().map(PyConflict::from).collect(),
             }),
             PyIcechunkStoreError::PyKeyError(e) => PyKeyError::new_err(e),
             PyIcechunkStoreError::PyValueError(e) => PyValueError::new_err(e),
             PyIcechunkStoreError::PyError(err) => err,
-            _ => IcechunkError::new_err(error.to_string()),
+            error => {
```
paraseba, line 102: this generates a message for the python exception that captures the full list of "causes" and the full span captured by the error

icechunk-python/src/lib.rs
```diff
@@ -22,11 +24,20 @@ use errors::{
     IcechunkError, PyConflictError, PyConflictErrorData, PyRebaseFailedData,
     PyRebaseFailedError,
 };
+use icechunk::initialize_tracing;
 use pyo3::prelude::*;
 use repository::{PyGCSummary, PyRepository, PySnapshotInfo};
 use session::PySession;
 use store::PyStore;
 
+#[pyfunction]
```
paraseba, line 33: Should we have no logs by default instead?

icechunk/Cargo.toml
```diff
@@ -57,3 +63,6 @@ tempfile = "3.15.0"
 
 [lints]
 workspace = true
+
+[features]
```
paraseba, line 67: we have an optional feature now, that creates a function to enable tracing. Apps using Icechunk as a rust library don't want this, they want to do their own tracing initialization.

icechunk/src/error.rs
```diff
@@ -0,0 +1,45 @@
+use std::fmt::Display;
+
+use tracing_error::SpanTrace;
+
+#[derive(Debug)]
```
paraseba, line 5: New "base" for all rich Icechunk errors. We store a cause error and the span where it happened. The spans is similar to a stack trace, but more semantic, not all functions are in the list, and they can maintain more information, such as their arguments and arbitrary extra data.

icechunk/src/format/manifest.rs
```diff
@@ -41,16 +41,29 @@ pub enum VirtualReferenceError {
...
     #[error("the checksum of the object owning the virtual chunk has changed ({0})")]
     ObjectModified(String),
     #[error("error retrieving virtual chunk, not enough data. Expected: ({expected}), available ({available})")]
     InvalidObjectSize { expected: u64, available: u64 },
-    #[error("error parsing virtual reference {0}")]
+    #[error("error parsing virtual reference")]
     OtherError(#[from] Box<dyn std::error::Error + Send + Sync>),
 }
 
+pub type VirtualReferenceError = ICError<VirtualReferenceErrorKind>;
```
paraseba, line 54: we use the new machinery in error.rs to create actual errors and propagate the spans

icechunk/src/lib.rs
```diff
@@ -46,3 +47,23 @@ mod private {
...
     pub trait Sealed {}
 }
+
+#[cfg(feature = "logs")]
+pub fn initialize_tracing() {
+    use tracing_error::ErrorLayer;
+    use tracing_subscriber::{
+        layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Layer, Registry,
+    };
+
```
paraseba, line 57: We have two `Layers`. One keeps track of the spans to feed the `ICError` instances. The other is the one spitting logs to stdout. Filtering only applies to the second one.

paraseba:
@dcherian we'll need to check how this impacts our benchmarks.

icechunk-python/src/lib.rs
dcherian: Yes I think that's "standard" behaviour? Particularly since we aren't plugged in to the python logging system.

dcherian: Though what are the available log levels? What is the default?

icechunk/src/lib.rs
dcherian: Can you convert that to a comment in the code please?

icechunk-python/python/icechunk/__init__.py
dcherian: If the default is to print nothing to stdout, it seems like this could be gated on the presence of the `ICECHUNK_LOG` env var.

icechunk/src/storage/object_store.rs
```diff
@@ -301,6 +307,7 @@ impl Storage for ObjectStorage {
             Err(err) => Err(err.into()),
         }
     }
+    #[instrument(skip(self, _settings, config, etag))]
```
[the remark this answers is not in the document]
paraseba: Added, it was already present in S3

icechunk-python/src/lib.rs
dcherian: foudn the answer here: https://docs.rs/tracing-core/0.1.33/tracing_core/metadata/struct.Level.html

icechunk-python/python/icechunk/__init__.py
paraseba: Currently, the default is to ERROR level logging. That's what you get if you don't define the ICECHUNK_LOG variable. The issue with making it easy no to to call this is that hen, error messages don't get the call stack. The mechanism for rich error messages depends on logs being enabled. The can be enabled with no logging, but we have to call the initialize function

icechunk-python/src/lib.rs
paraseba: default is ERROR, available goes from trace to fatal iirc

icechunk-python/python/icechunk/__init__.py
dcherian: OK in that case I support enabling logs always. The errors are great now.

icechunk/src/lib.rs
paraseba: done

dcherian (approved):

paraseba:
Example log:
image

PR #713, opened 2025-02-10
Design doc: multiple manifests per array

dcherian:
Ready for 30% feedback. I haven't thought about implementation yet, will look at that by tomorrow lunch time

### DIFF
--- a/design-docs/008-array-multiple-manifests.md
+++ b/design-docs/008-array-multiple-manifests.md
@@ -1,0 +1,46 @@
+# Multiple manifests per array
+
+## Current Status (Icechunk v0.1.2)
+
+1. A single array has a single manifest, regardless of size. [PR 598](https://github.com/earth-mover/icechunk/pull/598)
+2. Manifests are preloaded using regexes: [PR 649](https://github.com/earth-mover/icechunk/pull/649)
+3. The yaml config described below has not been implemented yet, but the current status is equivalent to:
+```yaml
+chunk-manifests:
+  - default:
+    arrays-per-manifest: 1
+    max-manifest-size: null
+    cardinality: null
+```
+
+## Next
+Our next job is to split the manifest files for large arrays, while providing the power-user with a lot of configurability.
+How might we optimize the splitting of a manifest file?
+1. User: "I expect to only read the most recent data most of the time." Split along "time" so that N days of data are in a single manifest file. For ERA5 for example, we might split by a time frequency: a year's worth of references, or a month's worth of references, in a single manifest.
+2. User: "I expect to generally read a small number of vertical levels (<5) for a dataset with O(50) levels.". Split so that a single level's references are in a single file.
+
+As has been pointed out, this is similar to sharding, and we general want to align the manifest shards with expected read patterns.
+Some considerations
+1. Zarr does not enforce dimension names, so we must allow splitting by axes too. `dimension_names=null` and dimension_names=`['time', null]` are allowed.
+2. Manifest extents must form non-overlapping bounding boxes, so using a `dict` to specify shard sizes seems like the right API.
+3. There needs to be a priority order for dimensions. So for example, I may always want different vertical levels in different manifest files, and only split those files once there are many timesteps in that. Thus we use a sequence of key-value pairs in the YAML instead of a map (which is [not guaranteed to be ordered](https://yaml.org/spec/1.2.2/#3221-mapping-key-order)).
+
+```yaml
+rules:
+  - path: ./2m_temperature  # 3D variable: (null, latitude, longitude)
+    manifest-split-sizes:
+      - 0: 120
+    max-num-refs: null  # (default) must be null if manifest-split-dims is provided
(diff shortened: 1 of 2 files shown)


design-docs/008-array-multiple-manifests.md
```diff
@@ -0,0 +1,46 @@
...
+
+As has been pointed out, this is similar to sharding, and we general want to align the manifest shards with expected read patterns.
+Some considerations
+1. Zarr does not enforce dimension names, so we must allow splitting by axes too. `dimension_names=null` and dimension_names=`['time', null]` are allowed.
+2. Manifest extents must form non-overlapping bounding boxes, so using a `dict` to specify shard sizes seems like the right API.
+3. There needs to be a priority order for dimensions. So for example, I may always want different vertical levels in different manifest files, and only split those files once there are many timesteps in that. Thus we use a sequence of key-value pairs in the YAML instead of a map (which is [not guaranteed to be ordered](https://yaml.org/spec/1.2.2/#3221-mapping-key-order)).
+
+```yaml
+rules:
+  - path: ./2m_temperature  # 3D variable: (null, latitude, longitude)
```
paraseba, line 30: is this a regex? should we allow them to use the same settings for multiple arrays? is that a common use case?

design-docs/008-array-multiple-manifests.md
```diff
@@ -0,0 +1,46 @@
...
+      - "level": 1  # alternatively 0: 1
+      - "time": 12  #           and 1: 12
+  - path: ./temperature
+    manifest-split-sizes:
+      - "level": 1
+      - "time": 8760  # ~1 year
+      - "latitude": null  # for unspecified, default is null, which means never split.
+  - path: ./*   # the default rules
+    manifest-split-sizes: null  # defer to max-num-refs
+    max-num-refs: 100_000  # split after at least this many references
```
paraseba, line 45: do we still gather the chunks in some order?

design-docs/008-array-multiple-manifests.md
dcherian: Yes, it follows from the previous proposal. I can add a comment to clarify, though I guess it's in the "default" setting below.

If anything I'd bet it's common to use the same settings for arrays of the same dimensionality (e.g. `[time, lat, lon]`, or `[time, level, lat, lon]`

design-docs/008-array-multiple-manifests.md
dcherian: yeah C-order, so in `[time, vertical, lat, lon]` we are splitting along the outer (or earlier) dimensions depending on the value.

kthyng:
@dcherian Is this change related to subchunking within a reference file? I've been trying this out in kerchunk and it has been useful. Also is there a possibility of having multiple representations of the same model output within the same reference file? That is, be able to access the same reference file but use one part of it if accessing a time series and another if accessing a single time step. Thanks for any info!

design-docs/008-array-multiple-manifests.md
paraseba: I think maybe we shouldn't do any splitting by default, and we shouldn't support `max-num-refs`. If you want splitting, tell us how.

dcherian: Fine by me.

dcherian:
> Is this change related to subchunking within a reference file?

No, it's about splitting up very large manifests

> That is, be able to access the same reference file but use one part of it if accessing a time series and another if accessing a single time step. 

Presumably these are separated as different arrays or arrays with the same name but in different groups. So that's already supported.

kthyng:
> it's about splitting up very large manifests

I guess I mean being both User 1 and User 2 from your text, at different times. So if I want to read in a time series at one depth I would access one reference file (User 1), and another if I want to read all of a few time steps (User 2). Is that the idea?

design-docs/008-array-multiple-manifests.md
```diff
@@ -0,0 +1,46 @@
...
+    cardinality: null
+```
+
+## Next
+Our next job is to split the manifest files for large arrays, while providing the power-user with a lot of configurability.
+How might we optimize the splitting of a manifest file?
+1. User: "I expect to only read the most recent data most of the time." Split along "time" so that N days of data are in a single manifest file. For ERA5 for example, we might split by a time frequency: a year's worth of references, or a month's worth of references, in a single manifest.
+2. User: "I expect to generally read a small number of vertical levels (<5) for a dataset with O(50) levels.". Split so that a single level's references are in a single file.
+
+As has been pointed out, this is similar to sharding, and we general want to align the manifest shards with expected read patterns.
```
dcherian, line 22: address repeated dimension names.

dcherian:
> I guess I mean being both User 1 and User 2 from your text, at different times. So if I want to read in a time series at one depth I would access one reference file (User 1), and another if I want to read all of a few time steps (User 2). Is that the idea?

Ah, this is the "multiply chunked arrays" idea, no that is not the focus here. It is an idea I like a lot though

PR #714, opened 2025-02-10
[Draft] Icechunk CLI Design Document

DahnJ:
## Icechunk CLI Design Document
Design document for https://github.com/earth-mover/icechunk/issues/461

Written as a design document, but this is just my suggestion – feel free to disagree and offer alternatives.

## TODOs
The configuration section is underdeveloped and the next part I will work on.

Issue #461, opened 2024-12-09
Icechunk could use a command line interface


paraseba:
Examples:
* create a new repo
* list history of a repo
* list refs
* check configuraiton
* invoke administrative tasks (garbage collection, compaction, etc)
* get statistics
* potentially do simple reads and writes

hendrikmakait:
This sounds like a fun project to work on while getting acquainted with `icechunk`, I'd love to work on this. Do you have any preferences on how this should be implemented (e.g., Rust vs. Python, choice of CLI toolkit)?

paraseba:
Very happy if you want to take this @hendrikmakait !

We want it in Rust. No preference of CLI toolkit. I'd say start small, with one operation in a PR that we can look at, and then we keep adding.

I think a challenge is going to be how to pass all the arguments needed to define a repo. What object store, location, credentials, etc.

Should be fun!

mpiannucci:
[clap](https://docs.rs/clap/latest/clap/) is the best cli toolkit for rust IMO, so that may be a good place to start! 

It can also start out in the `bin` directory of the main `icechunk` create as a bundled binary.

paraseba:
> It can also start out in the bin directory of the main icechunk create as a bundled binary.

+1

We'll need to see how to distribute it easily for people who install the python library, but that seems like an easy problem.

paraseba:
Amazing news @DahnJ ! Feel free to ping me in slack if you want pointers. I guess I would start with a short [design document](https://github.com/earth-mover/icechunk/tree/main/design-docs) explaining what functionality we want to include, how users are going to pass arguments via command line, how it's going to be used from python, etc.

One tricky thing, as I mention above, is that you need a bunch of information to "point" to an icechunk repo. You need the object store details (potentially including things like region, endpoint_url, etc), you need credentials, you need a prefix. It sounds like a lot to pass in the command line. Maybe we should thing if it's worth having a config file for the CLI, where you can alias all this set of parameters with a single name. So I can say `--repository foo` and `ic-cli` pulls all the details  from config.

For reference, this is the list of arguments needed to create an instance of S3 Storage:
```python
def s3_storage(
    *,
    bucket: str,
    prefix: str | None,
    region: str | None = None,
    endpoint_url: str | None = None,
    allow_http: bool = False,
    access_key_id: str | None = None,
    secret_access_key: str | None = None,
    session_token: str | None = None,
    expires_after: datetime | None = None,
    anonymous: bool | None = None,
[...]

PR #716, opened 2025-02-11
[Ready for review] Icechunk CLI Initial Implementation

DahnJ:
## Icechunk CLI Implementation
Initial implementation for https://github.com/earth-mover/icechunk/issues/461

Design doc: #714 

Implements
- Python integration
- Serializable configuration
- Works against local filesystem and S3
- Operations
   - `icechunk repo create`
   - `icechunk snapshot list`
   - `icechunk config init`
   - `icechunk config add`
   - `icechunk config list`


Cargo.lock
```diff
@@ -1,21 +1,21 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 4
+version = 3
```
paraseba, line 3: hmm I think you may want to update your rust installation (personally I'm on rustc and cargo 1.83)

Cargo.lock
```diff
@@ -1,21 +1,21 @@
...
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.0"
```
paraseba, line 16: I wonder if your system ran something that updates all dependencies?

icechunk-python/src/lib.rs
```diff
@@ -84,6 +103,7 @@ fn _icechunk_python(py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<PyDiff>()?;
     m.add_function(wrap_pyfunction!(initialize_logs, m)?)?;
     m.add_function(wrap_pyfunction!(spec_version, m)?)?;
+    m.add_function(wrap_pyfunction!(cli_entrypoint, m)?)?;
```
paraseba, line 106: I think you will want to also add the new function in the __init__ file and in the .pyi file. Youll see all other function in there too.

icechunk/src/cli/config.rs
```diff
@@ -0,0 +1,21 @@
...
+use crate::config::{RepositoryConfig, ObjectStoreConfig, Credentials}
+
+pub struct RepoLocation {
+    bucket: String,
+    prefix: String,
+}
+
+pub struct RepositoryDefinition {
+    location: RepoLocation,
+    object_store_config: ObjectStoreConfig,
```
paraseba, line 12: What to put in here is going to be one of the tricky bits.

paraseba (commented):
This is really great @DahnJ !

We don't need to wait until it's totally done to merge. lmk when you think you have reached a good milestone and we can merge in that state.

Cargo.lock
DahnJ: synced back with `main` in https://github.com/earth-mover/icechunk/pull/716/commits/1dbd8b5f3117496a8790fca2fd31c1f7a8383fa8

icechunk-python/src/lib.rs
DahnJ: So it seems necessary to add the function here to install it through `[project.scripts]`, but I'm not sure init/pyi is needed – this should never be called from Python. What do you think?

DahnJ:
I've added a new version. This works well with me both from the Python binary and Rust one.

Make sure to edit the bucket in `default.yaml` directly or through editing `default.rs` which generates it.

```bash
cd icechunk
cargo install --path .

# Local repo
❯ icechunk repo create localrepo
✅ Created repository RepositoryAlias("localrepo")
❯ icechunk repo list localrepo
SnapshotInfo { id: 14883a60e82269cadf100ec8, parent_id: None, flushed_at: 2025-02-15T10:10:15.952248Z, message: "Repository initialized", metadata: {"__root": Bool(true)} }


# S3 repo
❯ icechunk repo create s3repo
✅ Created repository RepositoryAlias("s3repo")
❯ icechunk repo list s3repo
SnapshotInfo { id: 1e075c18c0c4e81ff2fb48e1, parent_id: None, flushed_at: 2025-02-15T10:09:43.722002Z, message: "Repository initialized", metadata: {"__root": Bool(true)} }
```

It's still WIP, but wanted to check that I'm going in the right direction with the config structure here.

icechunk/default.yaml
```diff
@@ -0,0 +1,31 @@
+repos:
```
paraseba, line 1: can we move this file src/bin/icechunk?

icechunk/src/bin/icechunk/main.rs
```diff
@@ -0,0 +1,8 @@
+use clap::Parser;
+use icechunk::cli::interface::{run_cli, IcechunkCLI};
+
+#[tokio::main]
+async fn main() {
+    let cli = IcechunkCLI::parse();
+    let _ = run_cli(cli).await;
```
paraseba, line 7: This should probably return an error, something like:
```rust
fn main() -> Result<(), Box<dyn Error>>
```
we will want the status code to be != 0 if some error happened

icechunk/src/cli/config.rs
```diff
@@ -0,0 +1,125 @@
+use std::{collections::HashMap, path::PathBuf, str::FromStr};
+
+use serde::{Deserialize, Serialize};
+
+use crate::config::{RepositoryConfig, S3Options, S3StaticCredentials};
+
+// Redefine to remove the Refreshable field
```
paraseba, line 7: It seems like quite a bit of work just to remove that field. Remember that you'll need also to do the same for GCP and other credential types. Can we live with `Refreshable` even if we fail on it?

icechunk/src/cli/default.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn test_generate_repos() {
+        let s3_repo = create_s3_repo();
+        let local_repo = create_local_repo();
+
+        let mut repos = Repositories { repos: HashMap::new() };
+
+        repos.repos.insert(RepositoryAlias("s3repo".to_string()), s3_repo);
+        repos.repos.insert(RepositoryAlias("localrepo".to_string()), local_repo);
+
+        let path = "default.yaml";
```
paraseba, line 51: I don't know ... this test is writing code to the repos itself. Maybe move it to a temp file?

icechunk/src/cli/interface.rs
```diff
@@ -0,0 +1,140 @@
...
+    let storage = get_storage(repo).await?;
+    let config = Some(repo.get_config().clone());
+
+    let repository = Repository::open(config, Arc::clone(&storage), HashMap::new())
+        .await
+        .context(format!("❌ Failed to open repository {:?}", list_cmd.repo))?;
+
+    let snapshot_id =
+        repository.lookup_branch("main").await.context("❌ Failed to lookup branch")?;
+    let ancestry = repository.ancestry(&VersionInfo::SnapshotId(snapshot_id)).await?;
```
paraseba, line 113: you can pas the branch directly, without having to do `lookup_branch` first: you can create a `VersionInfo` from a branch name.

icechunk/src/cli/interface.rs
```diff
@@ -0,0 +1,140 @@
...
+    pin!(ancestry);
+    while let Some(snapshot_info) = ancestry.next().await {
```
paraseba, line 115: navigating ancestry in Icechunk is (currently) very slow. I'd recommend putting a limit of say 10 parents, and then let users override that default.

You may want to look at the `TryStreamExt` trait for useful methods in `Stream`, so you don't have to code the explicit loop. There is `try_map`, `take` and all what you would expect from an iterator type.

icechunk/src/cli/interface.rs
```diff
@@ -0,0 +1,140 @@
...
+        println!("{:?}", snapshot_info?);
```
paraseba, line 116: I recommend printing object ids (like `SnapshotId`) using its `to_string` method, instead of the `Debug` instance. It has a nicer, shorter presentation, and it's the characters that go into the object store path.

paraseba (commented):
Looking great @DahnJ !!!!!!

paraseba:
@DahnJ one idea to think about: it could be good to write all the logic in the CLI as a library crate in the project, and then have the binary only dispatching to the library using clap. That way, we can test the CLi very easily, and other libraries can reuse it in the future.

icechunk/src/cli/default.rs
DahnJ: Absolutely, was just a quick way of generating something.

I have added a rudimentary `config init` in https://github.com/earth-mover/icechunk/pull/716/commits/4d7afdaa2115772accc78ba12ad9f8cef65c4ad3

icechunk/src/cli/interface.rs
DahnJ: https://github.com/earth-mover/icechunk/pull/716/commits/18d4d087105a1a40775da01fa76b598d26127265

icechunk/src/cli/interface.rs
DahnJ: Is this along the lines of what you were suggesting? https://github.com/earth-mover/icechunk/pull/716/commits/18d4d087105a1a40775da01fa76b598d26127265

icechunk/default.yaml
DahnJ: I have now removed the file altogether (https://github.com/earth-mover/icechunk/pull/716/commits/5b03fbd9d6c39eedbd11ffaad4f3db4c0f305a76).

Should we instead decide on a standard location for the config, not relative to the repo? How about `$HOME/.config/icechunk/cli_config.yaml`, obtained through [dirs::config_dir()](https://crates.io/crates/dirs)?

icechunk/src/cli/config.rs
DahnJ: Initially I wanted to but the `Refreshable` field contains a trait object for which I couldn't simply derive `PartialEq`, preventing me from testing these configs. Having this subset-enum seemed like the simplest solution. Do you think there's a better one?

DahnJ:
Implemented a rudimentary interactive `config init` and `config list`


```bash
❯ icechunk config init
Enter alias: s3repo
Select repository type: S3
Enter bucket: mybucket
Enter prefix: myprefix
Enter region: eu-west-1
❯ icechunk config list
repos:
  s3repo: !S3
    location:
      bucket: mybucket
      prefix: myprefix
    object_store_config:
      region: eu-west-1
      endpoint_url: null
      anonymous: false
      allow_http: false
    credentials: FromEnv
    config:
      inline_chunk_threshold_bytes: null
      unsafe_overwrite_refs: null
      get_partial_values_concurrency: null
      compression: null
      caching: null
      storage: null
      virtual_chunk_containers: null
      manifest: null
```

icechunk/default.yaml
paraseba: love it

icechunk/src/cli/config.rs
paraseba: Yeah, you lose `PartialEq` but it may be the better than maintaining another parallel hierarchy.

 We do a lot of `assert!(matches!(...))` in tests when we cannot use `PartialEq`

icechunk/src/cli/interface.rs
paraseba: Great, and probably you no longer need to pin it

icechunk/src/cli/interface.rs
paraseba: We probably should implement `__repr__` for `PySnapshotInfo`

icechunk/Cargo.toml
```diff
@@ -55,6 +55,9 @@ tracing-subscriber = { version = "0.3.19", features = [
 tracing = "0.1.41"
 err-into = "1.0.1"
 serde_yaml_ng = "0.10.0"
+clap = { version = "4.5", features = ["derive"] }
+anyhow = "1.0.95"
+dialoguer = "0.11.0"
```
paraseba, line 60: What would be the impact of creating another library crate in the workspace and putting everything cli related there? Projects that want to depend on Icechunk as a Rust library, wouldn't need to bring in all these dependencies.

paraseba (commented):
this is great!

icechunk/src/cli/config.rs
DahnJ: Ah, nice, I didn't know about `assert!(matches!(...))`. Changed in https://github.com/earth-mover/icechunk/pull/716/commits/7cdd08768fab1796069d0ca144a4320b1dff3043

icechunk/src/cli/interface.rs
DahnJ: https://github.com/earth-mover/icechunk/pull/716/commits/1c857e645bd4694a10a9024466e8716b48b5c844

icechunk/default.yaml
DahnJ: Did that here https://github.com/earth-mover/icechunk/pull/716/commits/8f8d89fd1e97f79aa226a938cc7348495bec0bdb

On Mac, this actually [defaults to](https://docs.rs/dirs/latest/dirs/fn.config_dir.html) `$HOME/Library/Application Support`.

icechunk/src/bin/icechunk/main.rs
DahnJ: Improved error handling in https://github.com/earth-mover/icechunk/pull/716/commits/1724fcd2fabb4b53d17382693e9ba30bc623d9ab

DahnJ:
> @DahnJ one idea to think about: it could be good to write all the logic in the CLI as a library crate in the project, and then have the binary only dispatching to the library using clap. That way, we can test the CLi very easily, and other libraries can reuse it in the future.

Regarding library crate, I offered an alternative in https://github.com/earth-mover/icechunk/pull/716#discussion_r1966644283, let me please know what you think.

Regarding testing, made the interface more testable + added tests in https://github.com/earth-mover/icechunk/pull/716/commits/3137aceeb06d0e246e95623d8228d7d55e20316d.
- Sadly the dialoguer functions might not be [easily testable](https://github.com/console-rs/dialoguer/issues/95).

DahnJ:
@paraseba as far as I am concerned, this is near a mergeable state. My goal with this PR is to provide something well-designed enough for others to be able to start contributing. Happy to make it good enough for Icechunk's standards though, let me know what you think.

dcherian:
Thanks @DahnJ . Seba is on vacation this week, so feedback here will be a bit delayed.

Thanks for your patience.

icechunk-python/src/lib.rs
```diff
@@ -27,10 +27,43 @@ use errors::{
...
+            std::process::exit(1);
+        }
+    }
+}
+
+#[cfg(not(feature = "cli"))]
+#[pyfunction]
+fn cli_entrypoint(_py: Python) -> PyResult<()> {
+    println!("Must install the optional `cli` feature to use the Icechunk CLI.");
+    Ok(())
```
paraseba, line 64: Oh such a nice detail!

icechunk-python/Cargo.toml
```diff
@@ -38,6 +38,13 @@ async-trait = "0.1.85"
 typetag = "0.2.19"
 serde = { version = "1.0.217", features = ["derive", "rc"] }
 miette = { version = "7.5.0", features = ["fancy"] }
+clap = { version = "4.5", features = ["derive"], optional = true }
+anyhow = { version = "1.0.95", optional = true }
+dialoguer = { version = "0.11.0", optional = true }
+dirs = { version = "6.0.0", optional = true }
```
paraseba, line 44: @DahnJ why do we need all of these in `icechunk-python` ? Isn't it enough to have them in the `icechunk` crate?

DahnJ: You're right, we don't need them. Only clap is needed since it's used in `icechunk-python`, the rest can be installed through the feature https://github.com/earth-mover/icechunk/pull/716/commits/446622b32a36381a35843d6ce9fa52f9e70b0d7f

DahnJ: Should be all good now ✅

paraseba: can we drop the extra deps?

paraseba:
I merged all of this in a [separate PR](https://github.com/earth-mover/icechunk/commit/70a7d4827be650c860d6fb7d7fa2b700c64fc477) to easily apply a couple of fixes. Thank you @DahnJ !

PR #728, opened 2025-02-11
Add optional object store backend for s3

mpiannucci:
Useful for testing for now, not sure if it belongs in main but it will make integration testing easier...


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -623,6 +623,8 @@ class Storage:
         bucket: str,
         prefix: str | None,
         credentials: AnyS3Credential | None = None,
+        *,
+        use_object_store: bool = False,
```
paraseba, line 627: why don't we create a new function that we don't document to create this?

mpiannucci: Cool I'll do that instead

icechunk-python/python/icechunk/__init__.py
```diff
@@ -65,6 +65,7 @@
     gcs_storage,
     in_memory_storage,
     local_filesystem_storage,
+    s3_object_store_storage,
```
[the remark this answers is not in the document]
mpiannucci: On second thought it can just be imported directly from `storage` and i think that is hidden enough

dcherian (approved):

PR #735, opened 2025-02-17
Skip sharding codec in test

paraseba:
Zarr 3.0.3 included the ShardingCodec in hypothesis strategies

We don't seem to work with it.


dcherian (approved):
Since we only ever receive encoded bytes, I don't see why this would fail. but the sharding codec does seem a bit buggy.

paraseba:
> Since we only ever receive encoded bytes, I don't see why this would fail. but the sharding codec does seem a bit buggy.

yeah ... can this be a bug in our offset/length handling? I don't think so?

Issue #730, opened 2025-02-14
Improving session API


rabernat:
The following works

```python
repo.writable_session("main")
```

but this does not

```python
repo.readonly_session("main")
```

you have to do

```python
repo.readonly_session(branch="main")
```

I understand why (you can only have a writable session on a branch), but the API feels weirdly asymmetrical.

---

Additionally, I think both these functions should have a default `branch='main'` so that you can use them without any arguments. I imagine 95% of users will never create other branches, so this would be a reasonable default.

paraseba:
What would be a good order for finding the ref if you say `readonly_session("main")` ? I guess we can go tag -> branch -> error?

I'm not a fan of `repo.readonly_session()`. I this situation I think being explicit is worth the extra 6 characters. Here is a scenario where it would be very easy to make a mistake:

```python
repo = ....
session = repo.writable_session("develop")
...  
...
session.commit(...)

# now I want to verify my work

repo.readonly_session()
```

User forgot to pass the "develop" branch in the last call.

paraseba:
> I think we could make it `readonly_session(branch, *, tag=None, snapshot=None)`. So that we don't have to guess.

but that would require `readonly_session(None, tag="foo")` right? How about 

```python
reandonly_session(branch=None, *, tag=None, snapshot=None)
```

or we can do `"main"` as the default, but I feel that's a bit much as I explained above. If the default is `None` and you call without arguments. we can fail explaining the error

PR #737, opened 2025-02-17
Repository.readonly_session default

paraseba:
Now the following works:

```python
repo = ...
session = repo.readonly_session("main")
```

which is more symmetric with `writable_session`

Closes: #730


dcherian (approved):
mind doing a find-replace in the docs please?

PR #741, opened 2025-02-17
Remote benchmark tweaks

dcherian:
@paraseba 

```
python benchmarks/runner.py --pytest="-k open" --where s3 main
```

works for me. It will *print* results to screen for now. The pytest kwarg simply runs a set of short test. Delete to run the whole thing

### DIFF
--- a/icechunk-python/benchmarks/most_recent.sh
+++ b/icechunk-python/benchmarks/most_recent.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+
+echo $(ls -t ./.benchmarks/**/* | head -n 1)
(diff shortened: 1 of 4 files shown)


icechunk-python/benchmarks/most_recent.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+
+echo $(ls -t ./.benchmarks/**/* | head -n 1)
+pytest-benchmark compare --group=group,func,param --sort=fullname --columns=median --name=normal `ls -t ./.benchmarks/**/* | head -n 1`
```
dcherian, line 4: this is so dumb. but this is where we could upload to S3 for example. 

I confirmed with Coiled that their `--sync` option will not work for us.

paraseba (approved):
thank you Deepak

Issue #690, opened 2025-02-07
Parsing of zarr array metadata limits icechunk usability


jbms:
Icechunk thoroughly parses the zarr array metadata, even though as far as I understand it needs very little, if any, information from the array metadata.

In particular, because it parses data_type and fill_value, it won't work with data type extensions for which support has not been specifically added to icechunk, even though as far as I can see icechunk could just leave `data_type` and `fill_value` as `serde::Json` and avoid this needless limitation.

PR #749, opened 2025-02-18
Treat zarr metadata as a blob (mostly)

### LINKED ISSUE #391: Support complex `fill_value` with NaNs and Infs
there are xfailed tests

paraseba:
We were parsing too much of Zarr metadata. Icechunk currently is only interested in the array size and chunk sizes. It may become interested in the dimension names at some point. But still, we were parsing the whole metadata, storing internally as parsed object and then formatting it back to json.

We did this when the project started, imagining we may need more from the metadata. For example, we thought we could need to incorporate the codec pipeline in Icechunk.

With this patch, we now only extract the parts of the zarr metadata we care about. And we preserve the original blob of metadata as is, in a new user_data byte array. We return this blob in metadata requests.

If, in the future, we need more from the metadata, we can parse it and add it to the storage.

Simpler and less code. It works with zarr extensions, it's more resilient to zarr spec changes.

There is a price to this: we are no longer separating the user attributes from the rest of the metadata. The only impact of this, is we no longe can treat conflicts in user attributes separate from the rest of the zarr metadata.

If we consider this important in the short term, we can add it back by parsing more of the metadata blobs.

Also in this change:

- No more AttributeFile. We'll implement it when we need it
- Better snapshot serialization

Closes #391 
Closes: #690 

[on-disk breaking change]


icechunk/src/format/snapshot.rs
```diff
@@ -1,74 +1,51 @@
...
-    Inline(UserAttributes),
-    Ref(UserAttributesRef),
-}
-
-#[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub enum NodeType {
-    Group,
-    Array,
+pub struct DimensionShape {
+    array_length: u64,
```
dcherian, line 20:
```suggestion
    dim_length: u64,
```

I think this would be less confusing in the long run. `array.size` means "number of elements in the array" in numpy, so I was confused at first.

icechunk/src/format/snapshot.rs
```diff
@@ -1,74 +1,51 @@
...
-    pub chunk_key_encoding: ChunkKeyEncoding,
-    pub fill_value: FillValue,
-    pub codecs: Vec<Codec>,
-    pub storage_transformers: Option<Vec<StorageTransformer>>,
-    pub dimension_names: Option<DimensionNames>,
+impl DimensionShape {
+    pub fn new(array_length: u64, chunk_length: NonZeroU64) -> Self {
+        Self { array_length, chunk_length: chunk_length.get() }
+    }
+    pub fn array_length(&self) -> u64 {
```
dcherian, line 28: what is the advantage of doing this instead of setting the fields to `pub`? Is it that we only allow gets and disallow sets this way?

icechunk/src/format/snapshot.rs
```diff
@@ -87,35 +64,83 @@ impl ZarrArrayMetadata {
...
+            } else {
+                ((dim_shape.array_length - 1) / dim_shape.chunk_length) as u32
+            }
+        })
+    }
 }
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum DimensionName {
```
dcherian, line 91: Just for understanding: why pick this over `Option`?

icechunk/src/format/snapshot.rs
```diff
@@ -147,14 +172,23 @@ impl<'a> From<gen::ManifestRef<'a>> for ManifestRef {
...
+}
 
-    fn try_from(value: gen::ArrayNodeData<'a>) -> Result<Self, Self::Error> {
-        // TODO: is it ok to call `bytes` here? Or do we need to collect an iterator
-        let meta = rmp_serde::from_slice(value.zarr_metadata().bytes())?;
-        let manifest_refs = value.manifests().iter().map(|m| m.into()).collect();
-        Ok(Self::Array(meta, manifest_refs))
+impl<'a> From<gen::ArrayNodeData<'a>> for NodeData {
+    fn from(value: gen::ArrayNodeData<'a>) -> Self {
+        let dimension_names = value
```
dcherian, line 186: wow is there no other way to deep-copy this?

icechunk/src/change_set.rs
```diff
@@ -4,38 +4,39 @@ use std::{
...
+        snapshot::{ArrayShape, DimensionName, NodeData, NodeSnapshot},
         ChunkIndices, NodeId, Path,
     },
-    metadata::UserAttributes,
     session::SessionResult,
 };
 
-#[derive(Clone, Debug, PartialEq, Default, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ArrayData {
```
dcherian, line 21: Wondering why not share this with `NodeData::Array` like before. But I guess this struct is so small, that flatter is nicer.

icechunk/src/change_set.rs
```diff
@@ -44,8 +45,12 @@ impl ChangeSet {
         self.deleted_groups.iter()
     }
 
-    pub fn user_attributes_updated_nodes(&self) -> impl Iterator<Item = &NodeId> {
-        self.updated_attributes.keys()
+    pub fn updated_arrays(&self) -> impl Iterator<Item = &NodeId> {
```
dcherian, line 48: Is this really "nodes_with_updated_attributes"?

icechunk/src/change_set.rs
```diff
@@ -44,8 +45,12 @@ impl ChangeSet {
...
+        self.updated_arrays.keys().chain(self.updated_groups.keys())
+    }
+
+    pub fn updated_groups(&self) -> impl Iterator<Item = &NodeId> {
+        self.updated_groups.keys().chain(self.updated_groups.keys())
```
dcherian, line 53:
```suggestion
        self.updated_groups.keys()
```

icechunk/src/session.rs
```diff
@@ -3840,6 +3665,7 @@ mod tests {
 
         let err = ds2.rebase(&solver).await.unwrap_err();
 
+        dbg!(&err);
```
dcherian, line 3668:
```suggestion
```

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1482,10 +1478,8 @@ class ConflictType(Enum):
             A zarr metadata update conflicts with an existing zarr metadata update
         ZarrMetadataUpdateOfDeletedArray: int
             A zarr metadata update is attempted on a deleted array
-        UserAttributesDoubleUpdate: int
-            A user attribute update conflicts with an existing user attribute update
-        UserAttributesUpdateOfDeletedNode: int
-            A user attribute update is attempted on a deleted node
+        ZarrMetadataUpdateOfDeletedGroup: int
```
dcherian, line 1481: these should all. be `tuple[int]` but why that choice?

icechunk/src/format/snapshot.rs
paraseba: Yeah, i want to protect from having 0 chunk length, and yet, i don't want to handle the NonZero everywhere

dcherian (approved):
I looked at change_set, snapshot most closely. I really like this simplification. Just some minor comments. A couple of helper methods in ChangeSet look a bit wonky (I left a comment)

icechunk/src/format/snapshot.rs
paraseba: I just thought it was more clear ... DimensionName not set has a significant meaning, it's not just that they didn't pass it, they intentionally skipped it.

icechunk/src/change_set.rs
paraseba: I thought it wasn't 100% the same? I think user_data is in different places?

icechunk/src/change_set.rs
dcherian:
```
self.updated_arrays.keys().chain(self.updated_groups.keys())
``` 

Why are we chaining `updated_groups` here?

paraseba: :facepalm: great catch ...

It shows the quality of our conflict testing

icechunk-python/python/icechunk/_icechunk_python.pyi
paraseba: I have no idea actually ... @mpiannucci maybe?

PR #755, opened 2025-02-19
Switch to conditional PUT for branch updates

paraseba:
We no longer do the sequence of write-only branch versions.


mpiannucci (approved):
Very cool. We need gcs test but I can do that after this is in

icechunk/src/refs.rs
```diff
@@ -560,13 +435,15 @@ mod tests {
         mut f: F,
     ) -> ((Arc<dyn Storage>, R), (Arc<dyn Storage>, R, TempDir)) {
         let mem_storage = new_in_memory_storage().await.unwrap();
+        println!("Using mem storage");
```
dcherian, line 438: oops these snuck in

PR #767, opened 2025-02-21
Manifest Splitting

dcherian:
- [x] does the config get serialized properly?
- [x] Add ManifestSplitCondition.AnyArray
- [x] python tests for Or, And
- [x] Add docs
- [x] real-world benchmark; test with ERA5
- [ ] ~add ndim based condition (3D vs 4D)~ (if someone asks for it)

Minimal docs here: https://icechunk--767.org.readthedocs.build/en/767/icechunk-python/performance/

I rewrote the ERA5 manifests to put 1 year per manifest (~9000 chunks); This gets us 3X speedup.

image


icechunk/src/session.rs
```diff
@@ -1376,14 +1376,40 @@ impl<'a> FlushProcess<'a> {
...
+        let shards = ManifestShards::from_edges(vec![vec![0, 10, 20], vec![0, 10, 20]]);
+
+        // for it in chunks {}
+
+        // if manifest-split-size is set then we know extents, this is then duplicated between snapshot and config
+        //  1. Should we record the "real" extent or the theoretical-max extent (as set in the config).
+        //  2.
+        // if not, we aggregate_extents
+        // this is a choice, we could just always aggregate extents
         let chunks = aggregate_extents(&mut from, &mut to, chunks, |ci| &ci.coord);
```
dcherian, line 1396: one q here: Now that the extents are defined by the shard sizes, then this function is not needed? Unless we want to record "true" extents for some reason.

icechunk/src/format/manifest.rs
```diff
@@ -37,9 +33,77 @@ impl ManifestExtents {
...
+pub struct ManifestRef {
+    pub object_id: ManifestId,
+    pub extents: ManifestExtents,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ManifestShards(Vec<ManifestExtents>);
+
+impl ManifestShards {
+    pub fn default(ndim: usize) -> Self {
```
dcherian, line 59: I don't like this, but it is certainly tied to `ndim`.

icechunk/src/format/manifest.rs
```diff
@@ -37,9 +33,77 @@ impl ManifestExtents {
...
+        Self(vec![ManifestExtents(repeat_n(0..u32::MAX, ndim).collect())])
```
dcherian, line 60: I guess I can do 
```suggestion
        Self(vec![ManifestExtents(repeat_n(0.., ndim).collect())])
```

icechunk/src/format/manifest.rs
```diff
@@ -37,9 +33,77 @@ impl ManifestExtents {
...
+    // This can be used at write time to split manifests based on the config
+    // and at read time to choose which manifest to query for chunk payload
+    pub fn which(&self, coord: &ChunkIndices) -> Result<usize, IcechunkFormatError> {
+        // shard_range[i] must bound ChunkIndices
+        // 0 <= return value <= shard_range.len()
+        // it is possible that shard_range does not include a coord. say we have 2x2 shard grid
+        // but only shard (0,0) and shard (1,1) are populated with data.
+        // A coord located in (1, 0) should return Err
+        // Since shard_range need not form a regular grid, we must iterate through and find the first result.
+        // ManifestExtents in shard_range MUST NOT overlap with each other. How do we ensure this?
```
paraseba, line 85: We probably should have a better datastructure to maintain this invariant.

icechunk/src/format/manifest.rs
```diff
@@ -186,28 +251,42 @@ impl Manifest {
 
         let mut all = all.iter().peekable();
 
-        let mut array_manifests = Vec::with_capacity(1);
+        let mut array_manifests = Vec::with_capacity(shards.len());
+        // why peek?
+        // this seems to handle multiple nodes, but we only send one in?
```
paraseba, line 256: because we need to gather chunks for each node (this handles the case of chunks coming from multiple arrays to the same manifest). So peek allows me to know if I should continue or if I can finalize the current node.

icechunk/src/session.rs
paraseba: I think actual extents is useful...

icechunk/src/session.rs
```diff
@@ -1516,7 +1598,17 @@ async fn flush(
 
     for (node_path, node_id) in flush_data.change_set.new_arrays() {
         trace!(path=%node_path, "New node, writing a manifest");
-        flush_data.write_manifest_for_new_node(node_id, node_path).await?;
+        // FIXME: grab the config
+        let config = ManifestShardingConfig::default();
```
dcherian, line 1602: Need advice on how to thread the repo config down here

paraseba: The Session has the `RepositoryConfig`, I think it's OK to thread the sharding config through the few functions: do_commit, flush, etc. It's not great, those functions already have many args, but I think we can still stick a couple more before we refactor.

icechunk/src/config.rs
```diff
@@ -128,6 +130,82 @@ impl CachingConfig {
...
+                        .unwrap_or(false)
+                })
+                .unwrap_or(false),
+        }
+    }
+}
+
+// FIXME: isn't this really another condition?
+#[derive(Debug, Hash, PartialEq, Eq, Serialize, Deserialize, Clone)]
+pub enum ShardDimCondition {
```
dcherian, line 184: I didn't fold this into `ManifestShardCondition` because that matches on `path` but these conditions match on axis and dim-name.

icechunk/src/config.rs
```diff
@@ -128,6 +130,82 @@ impl CachingConfig {
...
+    DimensionName(String),
+    // TODO: Since dimension name can be null,
+    // i don't think we can have DimensionName(r"*") catch the "Any" case
+    Any,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
+pub struct ManifestShardingConfig {
+    // TODO: need to preserve insertion order of conditions, so hashmap doesn't work
+    pub shard_sizes: Vec<(ManifestShardCondition, Vec<(ShardDimCondition, u32)>)>,
```
dcherian, line 195: We'd need to add a new crate to have a hashmap that preserves insertion order. So i just went with tuples.

icechunk/src/format/manifest.rs
dcherian: :+1: Let's chat about how to go about designing this datastructure.

icechunk/src/session.rs
```diff
@@ -1361,60 +1365,96 @@ impl<'a> FlushProcess<'a> {
...
+        // TODO: what is a good capacity
+        let ref_capacity = 8_192;
+        sharded_refs.insert(0, Vec::with_capacity(ref_capacity));
+        for chunk in chunks {
+            let shard_index = shards.which(&chunk.coord)?;
+            sharded_refs
+                .entry(shard_index)
+                .or_insert_with(|| Vec::with_capacity(ref_capacity))
+                .push(chunk);
+        }
```
dcherian, line 1400: I am attempting to convert this method to accept a `impl Stream<Item=SessionResult<ChunkInfo>>` but I don't see how to convert this groupby logic.

icechunk/src/config.rs
```diff
@@ -128,6 +130,82 @@ impl CachingConfig {
...
+}
+
+//```yaml
+//rules:
+//  - path: ./2m_temperature  # regex, 3D variable: (null, latitude, longitude)
+//    manifest-split-sizes:
+//      - 0: 120
+//  - path: ./temperature  # 4D variable: (time, level, latitude, longitude)
+//    manifest-split-sizes:
+//      - "level": 1  # alternatively 0: 1
```
dcherian, line 149: needs validation. E.g. do these dimensions exist? does that axis number make sense?

icechunk/src/config.rs
```diff
@@ -206,20 +284,33 @@ static DEFAULT_MANIFEST_PRELOAD_CONDITION: OnceLock<ManifestPreloadCondition> =
...
     pub preload: Option<ManifestPreloadConfig>,
+    pub sharding: Option<ManifestShardingConfig>,
 }
 
 static DEFAULT_MANIFEST_PRELOAD_CONFIG: OnceLock<ManifestPreloadConfig> = OnceLock::new();
+static DEFAULT_MANIFEST_SHARDING_CONFIG: OnceLock<ManifestShardingConfig> =
+    OnceLock::new();
 
 impl ManifestConfig {
     pub fn merge(&self, other: Self) -> Self {
```
dcherian, line 295: merging is for merging a user-defined value with the library's default value.

icechunk/src/config.rs
```diff
@@ -206,20 +284,33 @@ static DEFAULT_MANIFEST_PRELOAD_CONDITION: OnceLock<ManifestPreloadCondition> =
...
-        Self { preload: other.preload.or(self.preload.clone()) }
+        Self {
+            preload: other.preload.or(self.preload.clone()),
+            // FIXME: why prioritize one over the other?
+            sharding: other.sharding.or(self.sharding.clone()),
```
dcherian, line 299: TODO: this could be overwrite instead. We need to careful about ordering after merge.

icechunk/src/format/manifest.rs
```diff
@@ -37,9 +37,79 @@ impl ManifestExtents {
...
+impl ManifestShards {
+    pub fn default(ndim: usize) -> Self {
+        Self(vec![ManifestExtents(repeat_n(0..u32::MAX, ndim).collect())])
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn from_edges(iter: impl IntoIterator<Item = Vec<u32>>) -> Self {
```
dcherian, line 69: TODO: add a doctest, and docstring

```
from_edges(Iterator<[0, 10, 20], [0, 10, 20]>) -> [[0..10, 0..10], [0..10, 10..20], [10..20, 0..10], [10..20, 10..20]]
```

Decision: validate on writes. don't validate on read.
Undefined behaviour if manifestextents is overlapping.

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -482,20 +482,109 @@ class ManifestPreloadConfig:
         """
         ...
 
+class ManifestShardCondition:
+    """Configuration for conditions under which manifests will preload on session creation"""
+
+    @staticmethod
```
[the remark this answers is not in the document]
dcherian: The issue is that not all Preload conditions make sense here : e.g. `True` and `NumRefs`. And I'm not sure how to model that in Rust.

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -482,20 +482,109 @@ class ManifestPreloadConfig:
...
+
+    class DimensionName:
+        """Shard along specified named dimension."""
+        def __init__(self, regex: str) -> Self: ...
+
+    class Any:
+        """Shard along any other unspecified dimension."""
+        def __init__(self) -> Self: ...
+
+ShardSizes: TypeAlias = list[ManifestShardCondition, [ShardDimCondition, int]]
```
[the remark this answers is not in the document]
dcherian: I don't understand what you mean.

icechunk/src/session.rs
```diff
@@ -1492,6 +1565,88 @@ impl<'a> FlushProcess<'a> {
     }
 }
 
+impl ShardDimCondition {
+    fn matches(&self, axis: usize, dimname: Option<String>) -> bool {
+        match self {
+            ShardDimCondition::Axis(ax) => ax == &axis,
+            ShardDimCondition::DimensionName(regex) => dimname
+                .map(|name| {
+                    Regex::new(regex)
```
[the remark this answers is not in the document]
dcherian: Once per array: https://github.com/earth-mover/icechunk/blob/56158c5d79b9e31ddfe1eeb0e4b0a881e8e63387/icechunk/src/session.rs#L1687

This call translates these conditions to ManifestShards, which is then used in the hot loop

icechunk/src/session.rs
```diff
@@ -705,19 +739,28 @@ impl Session {
...
-                    return Ok(Some(payload.clone()));
-                }
-                Err(IcechunkFormatError {
-                    kind: IcechunkFormatErrorKind::ChunkCoordinatesNotFound { .. },
-                    ..
-                }) => {}
-                Err(err) => return Err(err.into()),
+        let shards = ManifestShards::from_extents(
+            manifests.iter().map(|m| m.extents.clone()).collect(),
+        );
```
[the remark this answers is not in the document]
dcherian: We could cache it on `NodeData::Array` but I'm definitely assuming that there's not much impact for 10-100 shards.

icechunk-python/python/icechunk/_icechunk_python.pyi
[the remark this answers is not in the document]
dcherian: Ah ok, no. It is the "shard size" or "manifest chunk size". Precedence is given by order of appearance in the list.

icechunk/src/session.rs
```diff
@@ -1375,52 +1421,99 @@ impl<'a> FlushProcess<'a> {
...
+                            .or_insert_with(|| Vec::with_capacity(ref_capacity))
+                            .push(chunk);
+                        sharded_refs
+                    })
+                },
+            )
+            .await?;
+
+        for i in 0..shards.len() {
+            if let Some(shard_chunks) = sharded_refs.remove(&i) {
```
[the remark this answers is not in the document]
dcherian: Switched to a direct iteration.

icechunk-python/src/config.rs
```diff
@@ -937,27 +940,225 @@ impl From<ManifestPreloadConfig> for PyManifestPreloadConfig {
...
+        Self::PathMatches { regex }
+    }
+    #[staticmethod]
+    pub fn name_matches(regex: String) -> Self {
+        Self::NameMatches { regex }
+    }
+
+    pub fn __repr__(&self) -> String {
+        use PyManifestShardCondition::*;
+        match self {
```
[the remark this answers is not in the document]
dcherian: good catch, fixed.

icechunk-python/src/config.rs
```diff
@@ -937,27 +940,225 @@ impl From<ManifestPreloadConfig> for PyManifestPreloadConfig {
         })
     }
 }
+#[pyclass(name = "ManifestShardCondition", eq)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
```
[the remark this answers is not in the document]
dcherian: See https://github.com/earth-mover/icechunk/pull/767#discussion_r2021810894

dcherian:
**Local** benchmarks as of late last night:

### Read: 4x speedup

```
python benchmarks/runner.py --pytest '-m read_benchmark' main 137f2834
```

benchmark code: one array with shape 500_000_000, chunks=1000, shard_size=100_000 so 5 shards with 100_000 chunk refs each. Reading one element (so really one chunk request).
```python
    def fn():
        repo = ic.Repository.open(
            storage=synth_dataset.storage,
            config=ic.RepositoryConfig(manifest=ic.ManifestConfig(preload=preload)),
        )
        ds = xr.open_zarr(
            repo.readonly_session("main").store,
            group=synth_dataset.group,
            chunks=None,
            consolidated=False,
        )
        subset = ds.isel(synth_dataset.chunk_selector)
        subset[synth_dataset.load_variables].compute()
```
Parameterized over:
- sharding : either `unsharded` or `5 shards`
- preload: `off` or `default` (makes no difference)

```
------------------ benchmark 'xarray-read test_time_xarray_read_chunks_cold_cache': 6 tests ------------------
Name (time in ms)                                                                             Median
--------------------------------------------------------------------------------------------------------------
test_time_xarray_read_chunks_cold_cache[large-manifest-sharded-default] (this PR)       10.1482 (1.0)
test_time_xarray_read_chunks_cold_cache[large-manifest-sharded-off] (this PR)           10.2042 (1.01)
test_time_xarray_read_chunks_cold_cache[large-manifest-unsharded-default] (main)     43.5519 (4.29)
test_time_xarray_read_chunks_cold_cache[large-manifest-unsharded-default] (this PR)     45.1441 (4.45)
test_time_xarray_read_chunks_cold_cache[large-manifest-unsharded-off] (main)         43.3678 (4.27)
test_time_xarray_read_chunks_cold_cache[large-manifest-unsharded-off] (this PR)         45.2304 (4.46)
--------------------------------------------------------------------------------------------------------------
```

### Write: 10% slowdown on commit.

benchmark: This benchmarks _only_ `session.commit` after setting 500_000 virtual chunk refs

sharding: None (default) or shard size = 10_000 refs, so 50 shards in total.


even for the default case of writing a single shard. But this is only 0.5s so I don't think it matters much.
```
---------------- benchmark 'refs-write test_write_sharded_refs': 3 tests ----------------
Name (time in ms)                                                        Median
-----------------------------------------------------------------------------------------
test_write_sharded_refs[no-sharding-large-1d] (main)           491.2220 (1.0)
test_write_sharded_refs[no-sharding-large-1d] (this PR)           545.9192 (1.11)
test_write_sharded_refs[shard-size-10_000-large-1d] (this PR)     555.0974 (1.13)
-----------------------------------------------------------------------------------------
```

icechunk/src/session.rs
```diff
@@ -1375,52 +1456,81 @@ impl<'a> FlushProcess<'a> {
...
 
+    async fn write_manifest_from_iterator(
+        &mut self,
+        node_id: &NodeId,
+        chunks: impl Stream<Item = SessionResult<ChunkInfo>>,
+        splits: ManifestSplits,
+    ) -> SessionResult<()> {
+        // TODO: think about optimizing writes to manifests
+        // TODO: add benchmarks
+        let split_refs = chunks
```
paraseba, line 1479: I wonder is there is a way to do this using less memory. For example, should we "classify" only the ChunkIndices instead of the whole ChunkInfo which is much larger? Then we use the ChangeSet or the previous manifest to get the contents of each index. Not sure if it works, but something to think about.

But bottom line, maybe keeping track of the shard each chunk belongs to in the manifest is not hard and that would help a lot with transient memory.

icechunk/src/session.rs
```diff
@@ -1375,52 +1456,81 @@ impl<'a> FlushProcess<'a> {
...
             asset_manager,
             change_set,
             parent_id,
+            config,
             manifest_refs: Default::default(),
             manifest_files: Default::default(),
         }
     }
 
+    async fn write_manifest_from_iterator(
```
paraseba, line 1471: _manifests_  (pural) would be better

icechunk/src/session.rs
```diff
@@ -1708,6 +1899,7 @@ async fn fetch_manifest(
 /// available in `from` and `to` arguments.
 ///
 /// Yes, this is horrible.
+#[allow(dead_code)]
```
paraseba, line 1902: we no longer use this shit?

icechunk/src/store.rs
```diff
@@ -935,6 +935,12 @@ async fn exists(key: &str, session: &Session) -> StoreResult<bool> {
         Key::Chunk { node_path, coords } => {
             match session.get_chunk_ref(&node_path, &coords).await {
                 Ok(r) => Ok(r.is_some()),
+                Err(SessionError {
+                    // this chunk belongs to a shard that has not been written
+                    // or is invalid
+                    kind: SessionErrorKind::InvalidIndexForSplitManifests { .. },
+                    ..
+                }) => Ok(false),
```
paraseba, line 943: should this really return false? I feel this is an error not a 'this key doesn't exist' response?

icechunk/tests/test_gc.rs
```diff
@@ -145,13 +160,13 @@ pub async fn do_test_gc(
     )
     .await?;
     assert_eq!(summary.chunks_deleted, 10);
-    assert_eq!(summary.manifests_deleted, 1);
+    assert_eq!(summary.manifests_deleted, 110);
```
paraseba, line 163: awesome

icechunk-python/tests/test_can_read_old.py
```diff
@@ -282,4 +393,5 @@ async def test_icechunk_can_read_old_repo() -> None:
 
     # we import here so it works when the script is ran by pytest
 
-    asyncio.run(write_a_test_repo())
+    # asyncio.run(write_a_test_repo())
```
paraseba, line 396: we want both right?

icechunk/src/session.rs
dcherian: Yeah, i do a pass to group the references in to a manifest shard, so I just accumulate in that pass. I can delete.

icechunk/src/session.rs
```diff
@@ -1362,10 +1407,46 @@ pub fn construct_valid_byte_range(
     }
 }
 
+#[derive(Default)]
```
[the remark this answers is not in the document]
dcherian: I use it to initialize the HashMap with a default: (line 1485) `split_refs.entry(index).or_default().update(chunk)` . IIRC clippy suggested it haha.

icechunk/src/session.rs
```diff
@@ -1362,10 +1407,46 @@ pub fn construct_valid_byte_range(
...
+
+impl SplitManifest {
+    fn update(&mut self, chunk: ChunkInfo) {
+        if self.from.is_empty() {
+            debug_assert!(self.to.is_empty());
+            debug_assert!(self.chunks.is_empty());
+            // important to remember that `to` is not inclusive, so we need +1
+            let mut coord = chunk.coord.0.clone();
+            self.to.extend(coord.iter().cloned().map(|n| n + 1));
+            self.from.append(&mut coord);
```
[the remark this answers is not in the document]
dcherian: [Append "Moves all the elements of `other` into `self`, leaving `other` empty."](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) I was surprised by it too...

icechunk/src/format/snapshot.rs
```diff
@@ -87,6 +100,15 @@ impl ArrayShape {
     }
 }
 
+// Implement indexing for immutable access
+impl Index<usize> for ArrayShape {
```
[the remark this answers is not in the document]
dcherian: I think that's what the trait [requires](https://doc.rust-lang.org/std/ops/trait.Index.html), so our hands are tied, no?

I could add a `get` method instead that returns Option

icechunk/src/format/manifest.rs
```diff
@@ -37,9 +37,80 @@ impl ManifestExtents {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct ManifestSplits(Vec<ManifestExtents>);
+
+impl ManifestSplits {
+    /// Used at read-time
+    pub fn from_extents(extents: Vec<ManifestExtents>) -> Self {
+        Self(extents)
+    }
```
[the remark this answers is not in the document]
dcherian: we chatted about this a while ago, and decided that undefined behaviour for a badly written dataset is fine.  For example, we are expecting these extents to not overlap with each other.

i can assert that it is not-empty though

icechunk/src/repository.rs
```diff
@@ -1127,6 +1150,357 @@ mod tests {
...
+        dimension_names: &Option<Vec<DimensionName>>,
+        split_config: &ManifestSplittingConfig,
+        storage: Option<Arc<dyn Storage + Send + Sync>>,
+    ) -> Result<Repository, Box<dyn Error>> {
+        let backend: Arc<dyn Storage + Send + Sync> =
+            storage.or(Some(new_in_memory_storage().await?)).expect("foo");
+        let storage = Arc::clone(&backend);
+
+        let man_config = ManifestConfig {
+            preload: Some(ManifestPreloadConfig {
```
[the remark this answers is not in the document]
dcherian: `None` means default, and the current code means turn preloading off, correct?

icechunk/src/store.rs
dcherian: This is some weird Zarr thing but I can't reproduce. commenting it out for now.

dcherian: Ah  this case (invalid coordinate) is now handled inside `get_old_chunk` . I refactored the code earlier, and this particular error handling isn't needed anymore. I will delete.

icechunk/src/repository.rs
```diff
@@ -1127,6 +1150,357 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_manifest_splitting_complex_config() -> Result<(), Box<dyn Error>> {
+        let shape = ArrayShape::new(vec![(25, 1), (10, 1), (3, 1), (4, 1)]).unwrap();
+        let dimension_names = Some(vec!["t".into(), "z".into(), "y".into(), "x".into()]);
+        let temp_path: Path = "/temperature".try_into().unwrap();
+
+        let split_sizes = vec![
+            (
+                ManifestSplitCondition::PathMatches { regex: r".*".to_string() },
+                vec![(ManifestSplitDimCondition::DimensionName("t".to_string()), 12)],
```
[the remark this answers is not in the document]
dcherian: I added `ManifestSplitDim { condition, num_chunks }`. I didn't fully wire it up to python, so the user types:
```python
    conditions = {
                ManifestSplitDimCondition.DimensionName("longitude"): 3,
                ManifestSplitDimCondition.DimensionName("latitude"): 3,
                # this next one gets overwritten by DimensionName above.
                ManifestSplitDimCondition.Axis(1): 13,
                ManifestSplitDimCondition.Any(): 1,
            }
```
, it gets converted to lists with (basically) `map(list, conditions.items())`, and then converted to the new ManifestSplitDim for use in the Rust code.

icechunk/src/format/manifest.rs
dcherian: Ok I don't need the default any more. It was an artifact that appeared because I implemented the core logic before wiring up the config. Now the default gets set when parsing the config using the Array Metadata
image

icechunk/src/format/snapshot.rs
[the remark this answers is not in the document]
paraseba: yes, that is what i was trying to say, a get to get an option would be safer.

icechunk/src/repository.rs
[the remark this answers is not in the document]
paraseba: No, we have a complex default condition for preloading. Remember you gave me the regexes to match usual coordinate arrays?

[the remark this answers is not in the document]
dcherian: Does this code mean that `preload: None` then we get the complex default:
https://github.com/earth-mover/icechunk/blob/2023752b148bad67e24a3ec70254393e16e6c3f9/icechunk/src/config.rs#L218-L223

icechunk/src/format/snapshot.rs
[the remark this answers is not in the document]
dcherian: turns out, I might not be using this any more.

docs/.readthedocs.yaml
```diff
@@ -8,7 +8,8 @@ build:
   jobs:
      install:
       - which mamba
-      - cd icechunk-python && maturin build && pip install ../target/wheels/*.whl && cd ../docs
+
+      - cd icechunk-python && maturin build && pip install "$(ls ../target/wheels/*.whl | head -n 1)[docs]" && cd ../docs
```
ianhi, line 12: I wonder why moving th einstall to pip is so much faster. I would have suspected that it would be slower?

dcherian: pip is quite good, just can be bad at compiled dependencies.

icechunk/src/repository.rs
```diff
@@ -1127,6 +1150,357 @@ mod tests {
...
+
+        let session = repo.writable_session("main").await?;
+        let actual =
+            split_config.get_split_sizes(&session.get_node(&temp_path).await?)?;
+        assert_eq!(actual, expected);
+
+        Ok(())
+    }
+
+    // #[tokio::test]
```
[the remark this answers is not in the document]
paraseba: ping

paraseba (approved):

docs/docs/icechunk-python/performance.md
```diff
@@ -0,0 +1,83 @@
+# Performance
+
+!!! info
+
+    This is advanced material, and you are unlikely to need it. Icechunk aims to provide an excellent experience out of the box.
```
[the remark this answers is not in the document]
dcherian: Not sure, what number to put down. More than a million?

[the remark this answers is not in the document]
paraseba: more than a million sounds good to me

Issue #676, opened 2025-02-04
`repo.readonly_session` has parameter `snapshot`, not `snapshot_id`


DahnJ:
The docs imply that `repo.readonly_session` has a parameter `snapshot_id`, when in fact the parameter is [named `snapshot`](https://github.com/earth-mover/icechunk/blob/e1afdae7570ad28e0461a8cb03a50f7a1e830e72/icechunk-python/python/icechunk/_icechunk_python.pyi#L340).

In [Version Control](https://icechunk.io/en/latest/icechunk-python/version-control/#time-travel):

```python
session = repo.readonly_session(snapshot_id="BSHY7B1AGAPWQC14Q18G")
```

In [Quickstart](https://icechunk.io/en/latest/icechunk-python/quickstart/#explore-version-history), there's a related syntax error:

```python
earlier_session = repo.readonly_session(snapshot=snapshot_id=hist[1].id)
```

While this is a docs issue, I would also consider whether the parameter name should be unified in the API:
- `snapshot` is used for `readonly_session`, `ancestry` and `async_ancestry` methods
- `snapshot_id` is used elsewhere

If it's helpful, I am happy to make a PR with whichever change you think is best.

paraseba:
@DahnJ it would be great to unify. I think `snapshot_id: str` is what makes most sense, so people don't try to pass the items that `ancestry` returns.

DahnJ:
Sounds good, made the change in https://github.com/earth-mover/icechunk/pull/770

PR #770, opened 2025-02-22
feat: unify API on snapshot_id

DahnJ:
Aligning the parameter names: always using `snapshot_id` for the string id

Closes https://github.com/earth-mover/icechunk/issues/676

Made the changes in
- `ancestry`
- `async_ancestry`
- `diff`
- `readonly_session`

There are a few more errors that could be changed ([1](https://github.com/DahnJ/icechunk/blob/feat/dj-unify-session-id/icechunk/src/refs.rs#L120), [2](https://github.com/DahnJ/icechunk/blob/feat/dj-unify-session-id/icechunk/src/session.rs#L96)) and some internal uses ([example](https://github.com/DahnJ/icechunk/blob/feat/dj-unify-session-id/icechunk/src/storage/logging.rs#L207)) that I didn't change. Let me know if they should also be aligned.

### DIFF
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -46,7 +46,7 @@ repo.readonly_session("dev")
 # still possible:
 repo.readonly_session(tag="v0.1")
 repo.readonly_session(branch="foo")
-repo.readonly_session(snapshot="NXH3M0HJ7EEJ0699DPP0")
+repo.readonly_session(snapshot_id="NXH3M0HJ7EEJ0699DPP0")
 ```
 
 - Icechunk is now more resilient to changes in Zarr metadata spec, and can handle Zarr extensions.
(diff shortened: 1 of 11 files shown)


mpiannucci:
Changing the RefData would have broken format stability, so good call reverting that.

I think you got all the important points, the user facing API def the most important here

mpiannucci:
Changing the RefData would have broken format stability, so good call reverting that.

I think you got all the important points, the user facing API def the most important here

DahnJ:
CI failing due to [newly unmaintained ring dependency](https://github.com/rustsec/advisory-db/issues/2227).

mpiannucci:
Ya saw that yesterday... once I fix that later, I'll get this merged in. Thanks a lot!

PR #777, opened 2025-02-24
Start of executable docs

ianhi:
By making docs executable we will be able to automatically pick up errors. Currently I'm doing this using [`markdown-exec`](https://pawamoy.github.io/markdown-exec/) because that was the only way to integrate into the existing markdown docs while also preserving the tabbing that exists in several pages. This has the downside of currently being a bit verbose in adding to the start of each executable codeblock. However I think this is fixable with some issues I've opened upstream (https://github.com/pawamoy/markdown-exec/issues/77,  https://github.com/pawamoy/markdown-exec/issues/76#issuecomment-2679233674). There are two alternatives:

1.  [`mkdocs-jupyter`](https://mkdocs-jupyter.danielfrg.com/)

Woudl allow for writing documentation in jupyter notebooks. however, we would lose some of the pymdownx extensions (e.g. tabbing) 

2. Switching the docs build to sphinx and using [myst](https://mystmd.org/) would allow for writing everthign in jupyter as well as preserving tab behavior. However, this would require swapping the docs framework.

### Have doc build fail on unexpected error:

markdown-exec currently warns on an unexpected error. We can make the docs fail by passing `--strict` to `mkdocs build` (https://github.com/pawamoy/markdown-exec/issues/75)

Alternatively `mkdocs-jupyter` provides this already, or we can achieve the same with myst/jupyterbook


ianhi:
Ahh annoyingly this requires readthedocs to know how to install the dev version of icechunk, otherwise it won't be able to properly execute the docs. may take some figuring out to make that happen in readthedocs with poetry etc

docs/docs/icechunk-python/quickstart.md
```diff
@@ -57,7 +67,7 @@ However, you can also create a repo on your local filesystem.
 
 === "Local Storage"
 
-    ```python
+    ```python exec="on" session="quickstart" source="above"
     import icechunk
     storage = icechunk.local_filesystem_storage("./icechunk-local")
```
dcherian, line 72: How about 
```python
storage = icechunk.local_filesystem_storage(tempfile.mkdtemp()))
```

so we don't have to clear the local folder above?

docs/.readthedocs.yaml
```diff
@@ -14,7 +15,11 @@ build:
       - poetry config virtualenvs.create false
     post_install:
       # Install deps and build using poetry
-      - . "$READTHEDOCS_VIRTUALENV_PATH/bin/activate" && cd docs && poetry install
+      - . "$READTHEDOCS_VIRTUALENV_PATH/bin/activate" && cd docs && poetry install && cd ../icechunk-python && maturin develop && cd ../docs
```
dcherian, line 18: we can directly install from github with pip too if we know the commit ID. example:
```
git+https://github.com/earth-mover/icechunk.git@{self.full_commit}#subdirectory=icechunk-python
```

This will require maturin in the env, which we seem to have.

ianhi:
With the current strategy we need an upstream change before this works. I put in a PR: https://github.com/pawamoy/markdown-exec/pull/80

otherwise can't finish the version control page which has codeblocks that intentional raise execeptions.

dcherian:
> can't finish the version control page which has codeblocks that intentional raise execeptions.

Can we skip this one for now and continue with the rest?

ianhi:
> Can we skip this one for now and continue with the rest?

sure

docs/docs/icechunk-python/dask.md
```diff
@@ -113,20 +126,33 @@ Notably the ``compute`` kwarg is not supported.
...
+<!-- Waiting for fix of https://github.com/earth-mover/icechunk/issues/789  to execute-->
+
 ```python
 import icechunk.xarray
 import xarray as xr
 
+icechunk_session = icechunk_repo.writable_session("main")
 # Assuming you have a valid writable Session named icechunk_session
-dataset = xr.tutorial.open_dataset("rasm", chunks={"time": 1}).isel(time=slice(24))
+with icechunk_session.allow_pickling():
```
dcherian, line 137: this should only be needed for the read down below.

docs/docs/icechunk-python/quickstart2.ipynb
dcherian, line 1: delete this?

docs/docs/icechunk-python/version-control.md
```diff
@@ -46,55 +45,51 @@ Now that we have a `Repository` with a `main` branch, we can modify the data in
...
+print(session.commit(message="Update foo attribute on root group"))
 ```
 
 With a few snapshots committed, we can take a look at the ancestry of the `main` branch:
 
-```python
+
+```python exec="on" session="version" source="material-block" result="code"
 for snapshot in repo.ancestry(branch="main"):
     print(snapshot)
```
dcherian, line 78:
```suggestion
print(list(repo.ancestry(branch="main"))
```
perhaps?

docs/docs/icechunk-python/version-control.md
```diff
@@ -103,141 +98,138 @@ Now that we've created a new snapshot, we can time-travel back to the previous s
...
-# 'S3QY2RDQQTRYFGJDTB6G'
+print(session.commit(message="Update foo attribute on root group"))
 ```
 
-With these branches created, the hierarchy of the repository now looks like below.
+With these branches created, the hierarchy of the repository now looks like below. 
 
-```mermaid
+```python exec="on" result="mermaid" session="version"
+main_commits = [s.id[:6] for s in list(repo.ancestry(branch='main'))]
```
dcherian, line 139: We've wanted a method to do this for quite long!

docs/docs/icechunk-python/dask.md
```diff
@@ -5,23 +5,16 @@ However, because of how Icechunk works, it's not possible to use the existing [`
...
+Start with an icechunk store and dask arrays.
 
-First let's start a distributed Client and create an IcechunkStore.
-
-```python
-# initialize a distributed Client
-from distributed import Client
-
-client = Client()
-
```
ianhi, line 20: noteable change here is that the first example doesn't use a client, because it was not working without `allow_pickling` so i only used the client after that was described

docs/docs/icechunk-python/dask.md
```diff
@@ -113,20 +126,31 @@ Notably the ``compute`` kwarg is not supported.
...
-dataset = xr.tutorial.open_dataset("rasm", chunks={"time": 1}).isel(time=slice(24))
+icechunk_session = icechunk_repo.writable_session("main")
+dataset = xr.tutorial.open_dataset(
+    "rasm",
+    chunks={"time": 1}).isel(time=slice(24)
+    )
 
-icechunk.xarray.to_icechunk(dataset, session)
+# `to_icechunk` takes care of "allow_pickling" for you
+icechunk.xarray.to_icechunk(dataset, icechunk_session, mode="w")
```
ianhi, line 141: had to add the mode to avoid an error.

docs/docs/icechunk-python/quickstart.md
```diff
@@ -57,9 +57,10 @@ However, you can also create a repo on your local filesystem.
 
 === "Local Storage"
 
-    ```python
+    ```python exec="on" session="quickstart" source="above"
```
ianhi, line 60:
```suggestion
    ```python exec="on" session="quickstart" source="material-block"
```

docs/docs/icechunk-python/version-control.md
```diff
@@ -103,141 +98,138 @@ Now that we've created a new snapshot, we can time-travel back to the previous s
 
     It's important to note that because the `zarr Store` is read-only, we need to pass `mode="r"` to the `zarr.open_group` function.
 
-```python
-session = repo.readonly_session(snapshot_id="BSHY7B1AGAPWQC14Q18G")
+```python exec="on" session="version" source="material-block" result="code"
+session = repo.readonly_session(snapshot_id=list(repo.ancestry(branch="main"))[1].id)
```
ianhi, line 102: Is there a nicer way to get the second to last commit? equivalent to `HEAD~1`?

dcherian: `next(next(repo.ancestry(branch="main")))`?

docs/docs/icechunk-python/version-control.md
ianhi: You mean creating the mermaid diagram from python? it would be amazing if it could just auto generate the tree for you in a notebook. Would just have to bundle `mermaid.js`

docs/mkdocs.yml
```diff
@@ -165,6 +168,8 @@ markdown_extensions:
   - pymdownx.emoji:
       emoji_index: !!python/name:material.extensions.emoji.twemoji
       emoji_generator: !!python/name:material.extensions.emoji.to_svg
+  - toc:
+      permalink: "#"
```
ianhi, line 172: Also added permalinks that show up when you hover. unrelated to executable changes

docs/docs/icechunk-python/version-control.md
ianhi: renders poorly unfortunately. As one really long line:

image

docs/docs/icechunk-python/parallel.md
```diff
@@ -59,9 +59,9 @@ def write_timestamp(*, itime: int, session: Session) -> None:
 
 Now execute the writes.
 
-```python exec="on" session="parallel" source="material-block" result="code"
+<!-- ```python exec="on" session="parallel" source="material-block" result="code" -->
+```python
```
ianhi, line 63: this example runs fine for me locally, but ends up not writing anything to store when running on readthedocs.

docs/docs/icechunk-python/version-control.md
dcherian: Ah too bad. we should work on a better repr for that ancestry object.

docs/docs/icechunk-python/version-control.md
dcherian, line 102:
```suggestion
prev_snapshot = next(next(repo.ancestry(branch="main")))
session = repo.readonly_session(snapshot_id=prev_snapshot)
```

docs/docs/icechunk-python/parallel.md
```diff
@@ -16,38 +16,38 @@ including those executed remotely in a multi-processing or any other remote exec
...
+repo = Repository.create(local_filesystem_storage(tempfile.TemporaryDirectory().name))
 session = repo.writable_session("main")
 ```
 
 We will orchestrate so that each task writes one timestep.
 This is an arbitrary choice but determines what we set for the Zarr chunk size.
 
-```python
+```python exec="on" session="parallel" source="material-block" result="code"
 chunks = {1 if dim == "time" else ds.sizes[dim] for dim in ds.Tair.dims}
```
dcherian, line 33: can we turn off output for this one?
image

there are a couple more places we'll need to do that on this page.

dcherian:
Would be nice to add https://docs.astral.sh/ruff/integrations/#mdformat to format these codeblocks.

docs/docs/icechunk-python/parallel.md
dcherian: OK that is concerning. Does it work when you build the docs locally too?

ianhi: yes, `mkdocs build` runs fine for me. I can push a version with this reactivated so we can lookat the error.

ianhi:
@dcherian the failure on readthe docs:

https://readthedocs.org/projects/icechunk/builds/27378323/

image


maybe somethign weird is going if RTD limits the number of threads available?

> Would be nice to add https://docs.astral.sh/ruff/integrations/#mdformat to format these codeblocks.

Sadly it seems mdformat is incompatible with the mkdocs admonitions. Making changes like this:

Screenshot 2025-03-03 at 2 28 14 PM

dcherian:
That error is concerning. Presumably the `wait` isn't working, I am not sure why.

---

Could try https://github.com/KyleKing/mdformat-mkdocs but not worth spending too much time on.

dcherian:
The multiprocessing piece fails too:
```
concurrent.futures.process._RemoteTraceback: 
  """
    File "/home/docs/.asdf/installs/python/3.13.0/lib/python3.13/multiprocessing/queues.py", line 262, in _feed
      obj = _ForkingPickler.dumps(obj)
    File "/home/docs/.asdf/installs/python/3.13.0/lib/python3.13/multiprocessing/reduction.py", line 51, in dumps
      cls(buf, protocol).dump(obj)
      ~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  _pickle.PicklingError: Can't pickle <function write_timestamp at 0x7f36e02d3420>: attribute lookup write_timestamp on _code_block_session_parallel_n6_ failed
  """
```

Can you open an issue about these failures? I think we can move on here.

dcherian (approved):

ianhi:
I think I can get mdformat working. but it makes lots of small changes that will make the diff a bit confusing. So lets make that a folllowup PR

PR #778, opened 2025-02-25
Make repositories pickelable

mpiannucci:


mpiannucci:
Yes, that is already done for `Session`

PR #779, opened 2025-02-25
Add multi-processing write example

dcherian:
cc @ianhi


ianhi:
We need to explain to the reader why the code on the page is not standalone executable (I believe because of the need to wrap in `__main__` checks?). After that then we can eliminate some of the setup code in the examples and just show the core loop.

Also if we are going to keep the script separate from the main text we should execute it as part of the tests.

dcherian:
> We need to explain to the reader why the code on the page is not standalone executable 

I added a line.

> Also if we are going to keep the script separate from the main text we should execute it as part of the tests.

tracking here: https://github.com/earth-mover/icechunk/issues/780

ianhi (approved):

PR #798, opened 2025-03-04
Support for read-only Storage

paraseba:
Why:

* We may implement `Storage` for certain read-only s3-like, such as HTTP
* We'll need this for [Tigris support](https://github.com/earth-mover/icechunk/blob/main/design-docs/009-tigris-support.md)


mpiannucci (approved):

mpiannucci:
Nice!

Issue #786, opened 2025-02-27
Include bytes collected in garbage collection summary


rabernat:
Right now a GC summary looks something like this

```
{'chunks_deleted': 6000, 'manifests_deleted': 301, 'snapshots_deleted': 301}
```

It would be great if we could also surface the number of bytes cleaned up.

paraseba:
@rabernat now that this is done, it would be really cool to add it to your demos that run GC

rabernat:
Will do! (But I need a release.)

PR #799, opened 2025-03-04
Garbage collection results expose number of bytes deleted

paraseba:
Closes: #786

### DIFF
--- a/icechunk-python/src/repository.rs
+++ b/icechunk-python/src/repository.rs
@@ -317,26 +317,44 @@ impl PyDiff {
 #[derive(Debug, PartialEq, Eq, Default)]
 pub struct PyGCSummary {
     #[pyo3(get)]
-    pub chunks_deleted: usize,
+    pub bytes_deleted: u64,
     #[pyo3(get)]
-    pub manifests_deleted: usize,
+    pub chunks_deleted: u64,
     #[pyo3(get)]
-    pub snapshots_deleted: usize,
+    pub manifests_deleted: u64,
     #[pyo3(get)]
-    pub attributes_deleted: usize,
+    pub snapshots_deleted: u64,
     #[pyo3(get)]
-    pub transaction_logs_deleted: usize,
+    pub attributes_deleted: u64,
+    #[pyo3(get)]
+    pub transaction_logs_deleted: u64,
 }
 
 impl From<GCSummary> for PyGCSummary {
     fn from(value: GCSummary) -> Self {
         Self {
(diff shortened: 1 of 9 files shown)


rabernat (commented):
As part of this PR, it would be great if `GCSummary` could become a bit more ergonomic for the user. As a new user, it was very hard for me to figure out how to work with this object.

For example, a `to_dict()` method would be very convenient. So would a `__repr__` which explicitly shows all of the fields.

[This](https://github.com/earth-mover/icechunk-nasa/blob/2d85a05ffed5c5dcfc90279cfbbee9616867ffed/continuous-update-demo/writer.py#L142) is what I had to do to work around these issues.

paraseba:
Not having a `__repr__` sucks, but, why did you have to go to such extremes? Is it a documentation problem? The class has accessors for all fields.

rabernat:
Yes but it was hard to me to understand _what the fields actually were_.

This is a minor nit I suppose.

rabernat:
My point of reference is a Dataclass or a Pydantic model, which is easier to introspect, has a built in repr, and can easily convert to a dict. This object feels like a Dataclass, but without any of the syntactic sugar I'm used to.

paraseba:
@rabernat as a step forward in this PR, I added the `repr` implementation to `GCSummary`

icechunk/tests/test_gc.rs
```diff
@@ -142,6 +142,7 @@ pub async fn test_gc() -> Result<(), Box<dyn std::error::Error>> {
     assert_eq!(summary.chunks_deleted, 10);
     assert_eq!(summary.manifests_deleted, 1);
     assert_eq!(summary.snapshots_deleted, 1);
+    assert!(summary.bytes_deleted > summary.chunks_deleted);
```
mpiannucci, line 145: clever!

icechunk-python/src/repository.rs
```diff
@@ -340,6 +343,21 @@ impl From<GCSummary> for PyGCSummary {
     }
 }
 
+#[pymethods]
+impl PyGCSummary {
+    pub fn __repr__(&self) -> String {
+        format!(
+            r#"GCSummary(bytes_deleted={bytes}, chunks_deleted={chunks}, manifests_deleted={manifests}, snapshots_deleted={snapshots}, attributes_deleted={atts}, transaction_logs_deleted={txs})"#,
```
mpiannucci, line 350: ❤️

mpiannucci (approved):
Thansk for adding the repr!

PR #805, opened 2025-03-04
Add `force_path_style` S3 config to force path-style addressing

### LINKED ISSUE #793: Default `region` for S3-compatible Storage
### What happened?

We now have multiple reports of failing to create an icechunk repo when `region` is not set. 

This seems to be required to use the `aws_sdk_s3` crate:
https://github.com/earth-mover/icechunk/blob/26fbceb3a5a6abf01af03bf5740a1036dd9c6de1/icechunk/src/storage/s3.rs#L74-L86

But `region` may not be needed for object stores that provide an S3-compatible interface.

We may not want to always set a default `region` since this might cause confusing errors (for.e.g with real S3). So should Icechunk set a default fallback `region` when `endpoint_url` is set? Is the `endpoint_url` condition good enough to catch S3-compatible storage systems?

dcherian:
Closes #793

I chose `force_path_style` to match the AWS SDK. 

The other option is to use `addressing_style: Literal["path", "bucket", "auto"]` following [boto](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#addressing-style)


icechunk/src/config.rs
```diff
@@ -22,17 +22,19 @@ pub struct S3Options {
     pub endpoint_url: Option<String>,
     pub anonymous: bool,
     pub allow_http: bool,
+    pub force_path_style: Option<bool>,
```
paraseba, line 25: Can we keep this a `bool` but instruct serde to use "false" if not found? I think we can do that. we need to be very careful with backwards compatibility

icechunk/src/storage/s3.rs
```diff
@@ -119,7 +119,9 @@ pub async fn mk_client(config: &S3Options, credentials: S3Credentials) -> Client
 
     let mut s3_builder = Builder::from(&aws_config.load().await);
 
-    if config.allow_http {
+    if config.force_path_style.is_some_and(|x| x)
```
paraseba, line 122: I usually do `unwrap_or(false)` but see my previous comment on trying to make this an actual bool to match the rest

icechunk-python/python/icechunk/storage.py
```diff
@@ -110,7 +113,12 @@ def s3_storage(
         from_env=from_env,
         get_credentials=get_credentials,
     )
-    options = S3Options(region=region, endpoint_url=endpoint_url, allow_http=allow_http)
+    options = S3Options(
```
paraseba, line 116: I think you'll need to modify the .pyi file too

icechunk/src/config.rs
dcherian: The current behaviour is to force path style when `allow_http` is True, so i think we nee d an Option?

paraseba: I think we can break that behavior in favor of the simplicity of a bool? What do you think?

In any case, we'll need a test to check that we can still read a persisted old configuration, because I think this class goes into the virtual chunks configuration. I think if you do nothing it won't be backwards compatible, you probably need to flag it as "use default if not present".

I'd also add a comment explaining on what version the field was added

dcherian: I used `test_can_read_old`. Is that enough?

paraseba: i think test can read is enough, because it has a config file with S3Options.

paraseba (approved):

PR #806, opened 2025-03-05
Initial support for tigris consistency

paraseba:
In the default mode, we add the necessary headers to every request to get Tigris to behave with high consistency.

We offer initial support for eventually consistent behavior, for readonly datasets. This will require a bit more of work to improve safety, because Tigris seems to very aggressively cache objects.


icechunk-python/python/icechunk/storage.py
```diff
@@ -198,6 +199,10 @@ def tigris_storage(
         The region to use in the object store, if `None` a default region will be used
     endpoint_url: str | None
         Optional endpoint where the object store serves data, example: http://localhost:9000
+    eventually_consistent: bool
+        If set to True it will return a Storage instance that is read only but can read from the
```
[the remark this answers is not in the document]
dcherian: I'm not sure `eventually_consistent` is the best name. Perhaps `read_only_nearest_region`?

icechunk/src/storage/s3.rs
```diff
@@ -117,8 +159,20 @@ pub async fn mk_client(config: &S3Options, credentials: S3Credentials) -> Client
...
+    if !extra_read_headers.is_empty() || !extra_write_headers.is_empty() {
+        s3_builder = s3_builder.interceptor(ExtraHeadersInterceptor {
+            extra_read_headers,
+            extra_write_headers,
+        })
+    }
+
+    if config.allow_http {
+        s3_builder = s3_builder.force_path_style(true);
+    }
```
dcherian, line 174: need to merge `main`

dcherian (approved):
Makes sense to me. 

This interceptor pattern is something...
image

icechunk/src/storage/s3.rs
paraseba: oh I thought i did ... did I fuck up the merge? will check

paraseba: fixed

Issue #836, opened 2025-03-17
Calculate repo size


jhamman:
### Problem

It would be useful to have a simple method for calculating the total size of an Icechunk repository (across all snapshots). Today, this can (mostly) be accomplished by retrieving the s3 metadata for all objects in a repository but for large repositories, this will be much slower than using the manifests.

### Proposed Solution

Something like

```
def repo.getsize(self, snapshot: str | None = None, include_virtual_chunks: bool = True) -> int:
```

The two flags above are:
- `snapshot`: whether to return the size for a single snapshot or all snapshots.
- `include_virtual_chunks`: whether to include virtual chunks in the size calculation. When set to False, the size should be the same as the total size of all objects in the `chunks` directory.

paraseba:
Inline and virtual chunks are a bit harder to deal with. In particular, it's hard to count them uniquely. How about we start by counting only native chunks @jhamman ?

jhamman:
Thanks for knocking this out quickly @paraseba !

PR #841, opened 2025-03-17
New `total_chunks_storage` function

paraseba:
This new method in the `Repository` class computes the total space used by chunks in all snapshots. Only native, pointed chunks are included.

Closes: #836

### DIFF
--- a/icechunk/src/ops/stats.rs
+++ b/icechunk/src/ops/stats.rs
@@ -1,0 +1,75 @@
+use futures::TryStreamExt as _;
+use std::{collections::HashSet, sync::Arc};
+use tokio::pin;
+
+use crate::{
+    Storage,
+    asset_manager::AssetManager,
+    format::{IcechunkFormatError, IcechunkFormatErrorKind, manifest::ChunkPayload},
+    repository::RepositoryResult,
+    storage,
+};
+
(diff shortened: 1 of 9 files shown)


icechunk/src/ops/stats.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    let mut seen_manifests = HashSet::new();
+    let mut seen_chunks = HashSet::new();
+    let mut size = 0;
+
+    pin!(all_snaps);
+    while let Some(snap_id) = all_snaps.try_next().await? {
+        let snap = asset_manager.fetch_snapshot(&snap_id).await?;
+
+        for manifest_id in snap.manifest_files().map(|mf| mf.id) {
```
mpiannucci, line 39: Is it worth extracting this to be able to get the size of a specific snapshot? Or is that not quite relevant. Wondering if it could be helpful for deciding when to gc

mpiannucci (approved):
Approved with one question

icechunk/src/ops/stats.rs
paraseba: Yeah, that would be useful, but I don't have an immediate use case. I can add it a bit down the road if that's OK

Issue #723, opened 2025-02-11
slow perf with object_store (GCS)


dcherian:
Here are two dask performance reports for writing the "public" ERA5 dataset using Coiled:
1. [S3](https://icechunk-public-data.s3.us-east-1.amazonaws.com/reports/era5-ingest-s3-v01-2025-02-05+14%3A33%3A19.826694.html)
2. [GCS](https://icechunk-public-data.s3.us-east-1.amazonaws.com/reports/ERA5-ingest-gcs-v01-2025-02-06+18%3A24%3A51.824219.html)

For identical amount of data, with S3 it took 15 minutes, with Google Cloud it took 2.5 hours.

On S3, individual write tasks took 11-12sec. While it takes 55s on Google Cloud

mpiannucci:
> On S3, individual write tasks took 11-12sec. While it takes 55s on Google Cloud

Is this on a single worker? Or is this across many processes that icechunk has been pickled across?

dcherian:
Multiple workers, but in theory they should be getting unpickled once per worker, so later tasks should not see this overhead?

kylebarron:
Also fwiw every time you construct a new `ObjectStore`, you instantiate a _new_ connection pool. Two different `ObjectStore` instances won't share the same connection pool, and that would really slow requests down.

In case you ever come back to implementing `object_store` support, I've published [`pyo3-object_store`](https://docs.rs/pyo3-object_store/latest/pyo3_object_store/), so it's easy to expose Python classes that give you rust-native `Arc<dyn ObjectStore>` instances.

mpiannucci:
We are using object store for gcs and azure now. I think the slowdowns are from pickling which creates new connection pools, because there is no other way? Does that make sense?

kylebarron:
Are you saying that your non-`object_store` solution is able to pickle while reusing existing old connection pools? How is that possible?

mpiannucci:
I'm not sure, we haven't tracked it down, but so far we have a big speed difference when writing between s3 using aws sdk and gcs using object store. 

Our next step is to compare s3 with aws and object store to try and compare more directly to figure out what is going on

dcherian:
> On S3, individual write tasks took 11-12sec. While it takes 55s on Google Cloud

I now see 30-40s for individual write tasks on Google Cloud. The chunk sizes are small: 1MB-ish.

mpiannucci:
[...]

PR #845, opened 2025-03-18
Don't do multipart uploads for chunks

paraseba:
Some leftover code was doing unconditional multipart uploads of chunks, in the object store that use `object_store` (gcs, local, etc).

This was producing for usual chunk sizes ~ 3x performance penalty.

We should add multipart back, but only for large objects.

Closes: #723


mpiannucci (approved):

dcherian (approved):

kylebarron:
Interesting. What was the average size of uploads that went through this code path?

paraseba:
Size was small compared to what is recommended for multipart (~ 100 MB)

PR #849, opened 2025-03-19
Minor object-store backend cleanup

kylebarron:
I was curious about your general implementation from #723, and two things I thought could be better (at least provide simpler maintenance, although probably not better performance)

- `mk_object_store` is never async, and so the method doesn't have to be async and the trait doesn't have to be async-capable

That said, no strong feelings if you don't want to merge this.

Nothing immediately came to mind about why the object_store backend could be slower, other than if you're unintentionally re-creating the store multiple times. (I will say that storing a `OnceCell` reference in the store config "looks" weird to me, but maybe just because I haven't needed to use that pattern before)

### DIFF
--- a/icechunk/src/storage/object_store.rs
+++ b/icechunk/src/storage/object_store.rs
@@ -35,10 +35,7 @@ use std::{
     path::{Path as StdPath, PathBuf},
     sync::Arc,
 };
-use tokio::{
-    io::AsyncRead,
-    sync::{Mutex, OnceCell},
-};
+use tokio::{io::AsyncRead, sync::OnceCell};
 use tokio_util::compat::FuturesAsyncReadCompatExt;
 use tracing::instrument;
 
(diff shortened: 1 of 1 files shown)


mpiannucci:
In the case of our custom GCP token provider, if I understand the links correctly it would be cached in the second link, and the first link would never be hit right?

kylebarron:
Actually thanks for making me double check this; you were right.

- If you pass in a credential [via `with_credentials`](https://github.com/apache/arrow-rs/blob/660a3ac22a8ef8601acf4548d65146bc623f653a/object_store/src/gcp/builder.rs#L392-L396)
- Then that [overrides any other credential provider](https://github.com/apache/arrow-rs/blob/660a3ac22a8ef8601acf4548d65146bc623f653a/object_store/src/gcp/builder.rs#L478-L479)
- Which [gets set on the underlying client](https://github.com/apache/arrow-rs/blob/660a3ac22a8ef8601acf4548d65146bc623f653a/object_store/src/gcp/builder.rs#L551)
- And then the [raw `get_credential` method gets called directly](https://github.com/apache/arrow-rs/blob/660a3ac22a8ef8601acf4548d65146bc623f653a/object_store/src/gcp/client.rs#L240).

mpiannucci:
Thanks for confirming! It took a bit to figure that out originally so nice to know we were on the right path. 

Thank you for this!

mpiannucci (approved):

PR #860, opened 2025-03-24
Add optional default commit metadata to `Repository`

mpiannucci:
This allows users/systems to define static persistent metadata to following sessions


lindseynield (approved):
Python pieces look good!

icechunk/src/repository.rs
```diff
@@ -136,6 +139,8 @@ pub struct Repository {
     asset_manager: Arc<AssetManager>,
     virtual_resolver: Arc<VirtualChunkResolver>,
     virtual_chunk_credentials: HashMap<ContainerName, Credentials>,
+    #[serde(skip)]
```
dcherian, line 142: why skip this field? Will it survive pickling then?

PR #863, opened 2025-03-25
Fix repo serialization with default commit metadata

mpiannucci:


dcherian (approved):
Seems OK, but not confident around locks.

icechunk/src/repository.rs
```diff
@@ -139,8 +139,7 @@ pub struct Repository {
     asset_manager: Arc<AssetManager>,
     virtual_resolver: Arc<VirtualChunkResolver>,
     virtual_chunk_credentials: HashMap<ContainerName, Credentials>,
-    #[serde(skip)]
-    default_commit_metadata: Arc<RwLock<Option<SnapshotProperties>>>,
+    default_commit_metadata: Arc<Mutex<Option<SnapshotProperties>>>,
```
dcherian, line 142: Why Mutex over RwLock here?

mpiannucci: Yeah exactly. And here we can't use the async versions because of serialization. I can use rwlock if @paraseba would prefer tho

paraseba, line 142: Why do we need locking? Can't the `set_default_commit_metadata` just take a `&mut self`? I think that would be the cleanest option, then `PyRepository` is the one that needs locking to call this method on the repo, which seems also right.

mpiannucci: Because we dont currently lock `PyRepository` at all and its a big change to introduce it. We use internal mutability everywhere

mpiannucci: Sorry thats what i meant. I was trying to avoid having to lock the `Repository` but i can do if we want it

paraseba: We can merge and I can give it a try tomorrow if you are busy. But, i'd prefer if we don't introduce internal mutability into the Repo.

mpiannucci: i can do it in this pr then! Thanks for the feedback

mpiannucci:
Need to do python next but have to switch off

mpiannucci:
Ok @paraseba let me know what you think

icechunk-python/src/repository.rs
```diff
@@ -578,12 +594,20 @@ impl PyRepository {
...
         py.allow_threads(move || {
             let version = args_to_version_info(branch, tag, snapshot_id, None)?;
             let ancestry = pyo3_async_runtimes::tokio::get_runtime()
-                .block_on(async move { repo.ancestry_arc(&version).await })
+                .block_on(async move {
+                    let (snapshot_id, asset_manager) = {
+                        let lock = self.0.read().await;
+                        (
+                            lock.resolve_version(&version).await?,
+                            Arc::clone(lock.asset_manager()),
```
paraseba, line 606: nit: you could do the `Arc::clone` once the lock is released.

icechunk-python/src/repository.rs
```diff
@@ -832,10 +880,11 @@ impl PyRepository {
         py.allow_threads(move || {
             let result =
                 pyo3_async_runtimes::tokio::get_runtime().block_on(async move {
+                    let lock = self.0.read().await;
```
paraseba, line 883: Can we call `expire` without holding the lock?

icechunk-python/src/repository.rs
```diff
@@ -877,10 +926,11 @@ impl PyRepository {
                         delete_object_older_than,
                         Default::default(),
                     );
+                    let lock = self.0.read().await;
```
paraseba, line 929: same here

icechunk-python/src/repository.rs
```diff
@@ -897,10 +947,11 @@ impl PyRepository {
         py.allow_threads(move || {
             let result =
                 pyo3_async_runtimes::tokio::get_runtime().block_on(async move {
+                    let lock = self.0.read().await;
```
paraseba, line 950: same here, can we release the lock before calling the expensive function?

icechunk/src/repository.rs
```diff
@@ -60,7 +59,8 @@ pub enum RepositoryErrorKind {
     FormatError(IcechunkFormatErrorKind),
     #[error(transparent)]
     Ref(RefErrorKind),
-
+    #[error("failed to obtain mutex lock")]
+    SyncError,
```
paraseba, line 63: I think this is no longer needed?

icechunk/src/repository.rs
```diff
@@ -382,17 +381,17 @@ impl Repository {
...
-    pub async fn set_default_commit_metadata(
-        &self,
-        metadata: Option<SnapshotProperties>,
-    ) {
-        let mut guard = self.default_commit_metadata.write().await;
-        *guard = metadata;
+    pub fn set_default_commit_metadata(
+        &mut self,
+        metadata: SnapshotProperties,
+    ) -> RepositoryResult<()> {
```
paraseba, line 387: let's make this one return `()`, it can't fail.

icechunk/src/repository.rs
```diff
@@ -382,17 +381,17 @@ impl Repository {
...
+        self.default_commit_metadata = metadata;
+        Ok(())
     }
 
     #[instrument(skip_all)]
-    pub async fn default_commit_metadata(&self) -> Option<SnapshotProperties> {
-        self.default_commit_metadata.read().await.clone()
+    pub fn default_commit_metadata(&self) -> RepositoryResult<SnapshotProperties> {
+        Ok(self.default_commit_metadata.clone())
```
paraseba, line 394: Same here, I don't think we need to  return a `RepositoryResult`

paraseba (commented):
Looks great @mpiannucci !! Only minor comments.

icechunk-python/src/repository.rs
mpiannucci: How? We need the lock to get the reference to the asset manager right?

paraseba: lock -> get the `&Arc<AssetManager>` -> unlock -> Clone

mpiannucci: Aha, I didn't know you could keep a reference after locking

paraseba: Ah, I didn't really think about that... I'm not sure now

mpiannucci: Yeah i have to clone from a scope where the lock is held. Which i handled by making an inner scope so it is released immediately after cloning the asset manager

mpiannucci:
Scoped down the locks where it seemed possible/appropriate. Hope it makes sense!

icechunk/src/repository.rs
```diff
@@ -382,17 +379,13 @@ impl Repository {
...
-        let mut guard = self.default_commit_metadata.write().await;
-        *guard = metadata;
+    pub fn set_default_commit_metadata(&mut self, metadata: SnapshotProperties) {
+        self.default_commit_metadata = metadata;
     }
 
     #[instrument(skip_all)]
-    pub async fn default_commit_metadata(&self) -> Option<SnapshotProperties> {
-        self.default_commit_metadata.read().await.clone()
+    pub fn default_commit_metadata(&self) -> SnapshotProperties {
```
paraseba, line 387: I think in here I'd prefer to return `&SnapshotProperties`. That way the caller can clone it only if they need to.

paraseba (approved):

PR #868, opened 2025-03-26
Add expiration to stateful test

### LINKED ISSUE #867: Add `repo.lookup_snapshot(snapshot_id: str) -> SnapshotInfo`
I'm trying to add `expire_snapshots` to `test_stateful_repo_ops`, and find myself wanting to know the `written_at` time for the commit I just made.  

We could return `SnapshotInfo` from `commit` but that seems like a biggish break.

We could also return `SnapshotInfo` from `lookup_tag` and `lookup_branch`.

dcherian:
Closes #867 

Adds both `expire_snapshots` and `garbage_collect`.

`expire_snapshots` is basically an independent reimplementation of the rust logic.

`garbage_collect` is very simple, and only deletes expires snapshots.

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -397,6 +397,14 @@ class VersionControlStateMachine(RuleBasedStateMachine):
     @rule(name=simple_text, commit=commits, target=branches)
     def create_branch(self, name: str, commit: str) -> str:
         note(f"Creating branch {name!r}")
+
+        #### FIXME: get rid of this once the bug is fixed.
+        try:
+            self.repo.lookup_snapshot(commit)
+        except IcechunkError:
+            assume(False)
+        #####
+
         # we can create a tag and branch with the same name
         if name not in self.model.branches and commit in self.model.commits:
             self.repo.create_branch(name, commit)
@@ -413,6 +421,14 @@ class VersionControlStateMachine(RuleBasedStateMachine):
     @rule(name=simple_text, commit_id=commits, target=tags)
     def create_tag(self, name: str, commit_id: str) -> str:
         note(f"Creating tag {name!r} for commit {commit_id!r}")
+
+        #### FIXME: get rid of this once the bug is fixed.
+        try:
+            self.repo.lookup_snapshot(commit_id)
+        except IcechunkError:
+            assume(False)
+        #####
+
         if (
             name in self.model.created_tags
             or name in self.model.tags
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -84,13 +84,13 @@ class ExpireInfo:
 class CommitModel:
     id: str
     written_at: datetime.datetime
-    state: dict[str, Any]
+    store: dict[str, Any]
     parent_id: str
 
     @classmethod
     def from_snapshot_and_store(cls, snap: SnapshotInfo, store: dict[str, Any]) -> Self:
         return cls(
-            id=snap.id, written_at=snap.written_at, parent_id=snap.parent_id, state=store
+            id=snap.id, written_at=snap.written_at, parent_id=snap.parent_id, store=store
         )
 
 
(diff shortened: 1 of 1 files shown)


icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -186,22 +206,84 @@ def checkout_branch(self, ref: str) -> None:
...
+
+        # model this exactly like icechunk does.
+        # Start with named refs, and walk the ancestry back to the root.
+        snaps_to_delete = set()
+        tag_iter = map(operator.attrgetter("commit_id"), self.tags.values())
+        for commit_id in itertools.chain(self.branches.values(), tag_iter):
+            snap = self.commits[commit_id]
+            if snap.written_at < older_than:
+                # FIXME: This mimics the Rust code, but is not intuitive behaviour
+                continue
```
dcherian, line 252: IMO this is very confusing.

icechunk-python/tests/test_timetravel.py
```diff
@@ -307,3 +311,66 @@ async def test_default_commit_metadata() -> None:
...
+    root = zarr.group(store=session.store, overwrite=True)
+    root.create_group("child2")
+    c = session.commit("c")
+
+    expired = repo.expire_snapshots(
+        repo.lookup_snapshot(b).written_at + timedelta(microseconds=1),
+        delete_expired_branches=True,
+    )
+    assert expired == {a, b}
+    assert "branch" not in repo.list_branches()
```
dcherian, line 365: this line fails; presumably because of caches.

icechunk-python/tests/test_timetravel.py
```diff
@@ -307,3 +311,66 @@ async def test_default_commit_metadata() -> None:
...
+
+    for snap in (a, b):
+        repo.lookup_snapshot(snap)
+        repo.garbage_collect(
+            repo.lookup_snapshot(snap).written_at + timedelta(microseconds=1)
+        )
+        with pytest.raises(ValueError):
+            repo.lookup_snapshot(snap)
```
dcherian, line 373: this fails too.

icechunk-python/tests/test_timetravel.py
```diff
@@ -307,3 +311,67 @@ async def test_default_commit_metadata() -> None:
...
+    repo.create_tag("0", a)
+    repo.create_tag("1", b)
+    assert repo.list_tags() == {"0", "1"}
+
+    expired = repo.expire_snapshots(
+        repo.lookup_snapshot(b).written_at + timedelta(microseconds=1),
+        delete_expired_tags=True,
+    )
+    assert expired == {a, b}
+    assert repo.list_tags() == {}
```
dcherian, line 341: This also fails.

dcherian:
OK I think this is finally done 🥳 .

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -106,35 +64,61 @@ def list_prefix(self, prefix: str) -> None:
...
+    expired_snapshots: set[str]
+    deleted_branches: set[str]
+    deleted_tags: set[str]
+
+
+@dataclass
+class CommitModel:
+    id: str
+    written_at: datetime.datetime
+    state: dict[str, Any]
```
paraseba, line 87: I was curious about what the state was gonna be. Should we call this store?

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -186,22 +173,108 @@ def checkout_branch(self, ref: str) -> None:
...
-    def list_prefix(self, prefix: str) -> None:
+    def list_prefix(self, prefix: str) -> tuple[str, ...]:
         assert prefix == ""
         return tuple(self.store)
 
+    def refs_iter(self):
+        tag_iter = map(operator.attrgetter("commit_id"), self.tags.values())
+        return itertools.chain(self.branches.values(), tag_iter)
+
+    def expire_snapshots(
```
paraseba, line 202: I think all of this is great, but, we are re-implementing the algorithms in the tests, potentially with the same bugs we are blind to. I wonder if in the future we can define properties of the GCed or Expired repo, and try to track less in the model.

paraseba (approved):

icechunk-python/tests/test_stateful_repo_ops.py
dcherian: I landed up here because I'm doing an exact equality check for snapshot ids that were expired because only expires snaphsots that are reachable. 

I guess I can search the "forest" and assert that `set(IC expired IDs) < set(all possible expirable IDs)`

PR #870, opened 2025-03-27
`expire_ref` can now edit snapshot pointed by refs

paraseba:
This is a small deviation from the design document. But there is no objective reason why not to do this. By looking at the tests cases you'll see this expiration algorithm is more aggressive, it frees more snapshots, but we believe it's sound.

This is


icechunk/src/format/snapshot.rs
```diff
@@ -349,7 +350,7 @@ impl Snapshot {
 
         let message = builder.create_string(&message);
         let parent_id = parent_id.map(|oid| generated::ObjectId12::new(&oid.0));
-        let flushed_at = Utc::now().timestamp_micros() as u64;
+        let flushed_at = flushed_at.unwrap_or_else(Utc::now).timestamp_micros() as u64;
```
dcherian, line 353: so we preserve `written_at` now, even when editing? Am I understanding this correctly?

paraseba: yes, this is the original time at which the commit was done

paraseba: we were violating one of IC invariants: snapshots flushed_at are monotonically descending in `ancestry`

dcherian (approved):

PR #875, opened 2025-03-28
Remove cache entries during GC

dcherian:


icechunk/src/ops/gc.rs
```diff
@@ -350,6 +374,7 @@ async fn gc_transaction_logs(
         .filter_map(move |tx| {
             ready(tx.ok().and_then(|tx| {
                 if config.must_delete_transaction_log(&tx) && !keep_ids.contains(&tx.id) {
+                    asset_manager.remove_cached_snapshot(&tx.id);
```
paraseba, line 377: transaction log?

icechunk/src/ops/gc.rs
```diff
@@ -286,6 +305,7 @@ async fn gc_chunks(
 
 #[instrument(skip(storage, storage_settings, config, keep_ids), fields(keep_ids.len = keep_ids.len()))]
 async fn gc_manifests(
+    asset_manager: &Arc<AssetManager>,
```
paraseba, line 308: do you need these to be Arcs? Or can they be &?

paraseba (changes requested):

icechunk/src/ops/gc.rs
dcherian: Looks like i can do `&`.

paraseba (approved):

Issue #844, opened 2025-03-18
Fail if a Tigris object store path is provided to S3 compatible


lindseynield:
Tigris object stores should use `tigris_storage()` for Storage instance creation: https://icechunk.io/en/latest/icechunk-python/reference/#icechunk.tigris_storage. 

If a user tries to create a Storage instance for a Tigris bucket with S3 compatible/`s3_storage` (which is the legacy workflow), Icechunk should fail and tell the user to use `tigris_storage()` instead.

paraseba:
We should also thoroughly document why this function needs to be used

PR #889, opened 2025-04-02
Fail when creating Storage for Tigris using s3_compatible

paraseba:
Closes: #844


dcherian (approved):
Nice!

dcherian:
Actually we should mention the constructor name specifically to help the user

paraseba:
> Actually we should mention the constructor name specifically to help the user

I never know how to do this kind of thing properly. The constructor name is in python, subject to change, etc. :shrug:

PR #891, opened 2025-04-03
Better `Debug` instances and __repr__ methods.

paraseba:
Python's `SnapshotInfo` now also includes the Manifests info.


icechunk-python/src/repository.rs
```diff
@@ -161,28 +171,52 @@ impl From<PySnapshotProperties> for SnapshotProperties {
...
 impl From<SnapshotInfo> for PySnapshotInfo {
     fn from(val: SnapshotInfo) -> Self {
-        PySnapshotInfo {
+        Self {
             id: val.id.to_string(),
             parent_id: val.parent_id.map(|id| id.to_string()),
             written_at: val.flushed_at,
             message: val.message,
             metadata: val.metadata.into(),
+            manifests: val.manifests.into_iter().map(|v| v.into()).collect(),
```
dcherian, line 192: Isn't this potentially a really big list? that will now be listed our with `ancestry`?

dcherian: Maybe we should have `SnapshotInfo.list_manifests()` instead?

paraseba: yes, but, ancestry is very slow anyway, since it fetches each snapshot from object store. Gathering the few manifests cannot add much

paraseba: scratch, my latest about "print". What I should have said is: this objects get populated only by ancestry, which is very slow, and they are mostly used and discarded, I don't imagine gathering a few hundred extra objects will become a problem

dcherian: ancestry prints snapshotinfo though. My concern is `ancestry`'s repr being hard to read. Though perhaps we should just build a better repr

dcherian: An updating weather forecast repo might have O(100) manifests, one per array, for example.

paraseba: Ah I understand now. My goal with this was trying to give us a bit more insight to debug, using the `lookup_snapshot` method. I like the idea of having a dedicated ancestry representation, that could be text in the console, html in a notebook, etc

paraseba: > An updating weather forecast repo might have O(100) manifests in each snapshot, one per array, for example.

But we are not fetching those manifests, only surfacing the data that is already in the snapshot, maybe 30 bytes per snapshot or so.

paraseba: But this definitely has an impact on the __repr__ for snapshots. Do you want me to skip manifests there? we can do that

dcherian: Yes, I'm only concerned about the repo, because we use it for ancestry right now.

Yes, let's just comment it out from the repr for now.

paraseba: Done in https://github.com/earth-mover/icechunk/pull/891/commits/f9baf050bac147aaf30672a15053ff4564e05fca

dcherian (approved):

PR #894, opened 2025-04-03
Release version v0.2.12

paraseba:


icechunk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "icechunk"
-version = "0.2.10"
+version = "0.2.11"
```
dcherian, line 3: shall we get this one back in sync?

paraseba: I don't think so, that would force us to either skip releases on rust or release versions without changes. These things are going to get out of sync eventually anyway. As the Python library grows, there will be python only releases.

dcherian (approved):

PR #903, opened 2025-04-10
S3 Storage supports setting storage class

paraseba:
Object store based Storage instances are not supported yet, object_store doesn't have a way to set the Storage Class.

Closes: #871


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -617,6 +620,19 @@ class StorageSettings:
             Don't write metadata fields in Icechunk files.
             This is only useful in object stores that don't support the feature.
             Use at your own risk.
+
+        storage_class: str | None
```
mpiannucci, line 624: Can we specify this is only supported with s3 for now?

mpiannucci (approved):

Issue #620, opened 2025-01-24
Bring back async interface for `Repository` and `Session`


mpiannucci:
In the large refactor we focused on simplicity and removed the async duplications of methods. Now that we are mostly happy with the API we should bring back the async interface for repository operations by creating an `async` module that exposes `AsyncRepository` and `AsyncSession` that mirrors the existing functionality but offers async for relevant methods.

PR #914, opened 2025-04-15
First pass at async api options

mpiannucci:
First pass at design for bringing async API back to python #620 

This currently makes no decisions but starts to lay out options to iterate on.

### DIFF
--- a/design-docs/010-async-python-api.md
+++ b/design-docs/010-async-python-api.md
@@ -1,0 +1,112 @@
+# Async Python API
+
+The Icechunk rust API for `Repository` and `Session` are both async using `tokio`. Originally, the python API was also async before the transition to separate `Repository`,` Session`, and `Store` classes. 
+
+These changes were originally made to ease the typical python developer experience which may not be running from within an async context. However, Icechunk has many applications that may require an async runtime such as use within web servers. In these cases, blocking the main thread for 200 ms to perform IO is not acceptabl. 
+
+This design document seeks to plan out the ability to perform async lifecycle functions from python, specifically in the `Repository` and `Session` interfaces. 
+
+## API Options
+
+There are a few different ways this interface can be achieved, we will iterate them here and add links to external references where appropriate. 
+
+### Separate Classes
+
+We will have a separate `asyn` module within the `icechunk` python module, leaving an API listing like this: 
+
+```
+icechunk
+├── asyn
+│   ├── async_repository.py
+│   ├── async_session.py
+│   └── __init__.py
+├── repository.py
+├── session.py
+└── __init__.py
+```
+
+Looking at the `async_repository.py` file, it would have the following structure: 
+
+```python
+class AsyncRepository:
+    async def fetch_config(self) -> RepositoryConfig | None:
+        ...
+
+...
+class Repository:
+    def fetch_config(self) -> RepositoryConfig | None:
+        ...
+```
+
+#### Alternative Implementation
+
+This approach leaves a few alternatives to maaximize reuse of code between the `Repository` and `AsyncRepository` classes. 
+
+```python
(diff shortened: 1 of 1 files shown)


mpiannucci:
Tagging @TomNicholas @kylebarron in case they have some insight or opinion!

design-docs/010-async-python-api.md
```diff
@@ -0,0 +1,112 @@
+# Async Python API
+
+The Icechunk rust API for `Repository` and `Session` are both async using `tokio`. Originally, the python API was also async before the transition to separate `Repository`,` Session`, and `Store` classes.
+
+These changes were originally made to ease the typical python developer experience which may not be running from within an async context. However, Icechunk has many applications that may require an async runtime such as use within web servers. In these cases, blocking the main thread for 200 ms to perform IO is not acceptabl.
```
paraseba, line 5:
```suggestion
These changes were originally made to ease the typical python developer experience which may not be running from within an async context. However, Icechunk has many applications that may require an async runtime such as use within web servers. In these cases, blocking the main thread for 200 ms to perform IO is not acceptable.
```

paraseba (commented):
This is great  @mpiannucci . It very clearly explains the options.

I don't think anybody will like "Classes with Async methods", or I don't at least (but it's great to keep it in the doc as an option).

I feel the option with `_async_repo` and python doing the blocking is very compelling because of its simplicity and small code footprint. But, I'm concerned about having more async stuff in python, I have fought those async loops way more than I want to remember. For example, we would need to explore how well this interacts with multithreaded python code.

The everything-in-rust option is by far what takes more code, but also what feels safer. I'm afraid to ask, but ... how about writing a macro that can generate both versions of every method?

mpiannucci:
> But, I'm concerned about having more async stuff in python, I have fought those async loops way more than I want to remember. For example, we would need to explore how well this interacts with multithreaded python code.

This is my exact worry as well.

A macro actual my might be the best way to do this, because the inner async closure will be the same for both sync and async and the only thing that changes are the call signatures and the runtime wrapper. I'll look into that

paraseba:
>  the inner async closure will be the same for both sync and async and the only thing that changes are the call signatures and the runtime wrapper.

Exactly.

Maybe we can do a small proof of concept and see how it goes. The other concern with this is that we still need to create both python classes calling to the proper Rust object. I wonder if there are any tricks we play there without killing performance.

kylebarron:
> I don't think anybody will like "Classes with Async methods", or I don't at least (but it's great to keep it in the doc as an option).

FWIW this is my preferred option 😄. E.g. in obstore the same operations are synchronous with `def get_range` or async with `async def get_range_async`.

I don't think there should be two separate classes that represent the same underlying concept and data on the rust side. E.g. for obstore the act of creating an S3Store is totally indifferent to whether you plan to use it synchronously or asynchronously. And the underlying Rust data (`Arc<dyn ObjectStore>`) is the same whether or not you use async methods on it. So I think it would be more confusing to have two separate classes that do the same thing. 

Even when you might want an async constructor, I prefer having a single class. E.g. in https://github.com/kylebarron/arro3/pull/313 I'm updating my `ParquetFile` API. But since the underlying data is the same in a sync or async case, I prefer having alternate constructors: [`open` and `open_async`](https://github.com/kylebarron/arro3/pull/313/files#diff-2bb3d3e59d23dcb919ce704de77aa44d6a7e639432fffd9a827d3f99f7b958a8R220-R238).

I also prefer `fetch_config_async` instead of `async_fetch_config` in your example so you get quicker narrowing of autocompletion in IDE type hinting.

paraseba:
Thanks @kylebarron . This gives me a new appreciation for that option.

PR #915, opened 2025-04-16
Multipart upload for S3 native Storage instances

paraseba:
Implements multipart upload for objects (by default) larger than 100MB. All object types are supported.

object_store instances not implemented yet.

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -597,6 +597,7 @@ class StorageSettings:
         storage_class: str | None = None,
         metadata_storage_class: str | None = None,
         chunks_storage_class: str | None = None,
+        minimum_size_for_multipart_upload: int | None = None,
     ) -> None:
         """
         Create a new `StorageSettings` object
@@ -636,6 +637,10 @@ class StorageSettings:
             Store chunk objects using this object store storage class.
             Currently not supported in GCS.
             Defaults to storage_class.
+
+        minimum_size_for_multipart_upload: int | None
+            Use object store's multipart upload for objects larger than this size in bytes.
+            Default: 100 MB if None is passed.
         """
         ...
     @property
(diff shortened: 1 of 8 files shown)


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -636,6 +637,9 @@ class StorageSettings:
             Store chunk objects using this object store storage class.
             Currently not supported in GCS.
             Defaults to storage_class.
+
+        minimum_size_for_multipart_upload: int | None
+            Use object store's multipart upload for objects larger than this size in bytes.
```
mpiannucci, line 642: Can we add the default to the docstring?

icechunk/src/storage/s3.rs
```diff
@@ -308,6 +308,135 @@ impl S3Storage {
...
+                    .upload_id(upload_id)
+                    .bucket(self.bucket.clone())
+                    .key(key)
+                    .part_number(idx)
+                    .body(body)
+                    .send()
+                    .await
+                    .map(|res| (idx, res))
+            })
+            .collect::<FuturesOrdered<_>>();
```
mpiannucci, line 379: 👍

mpiannucci (approved):

mpiannucci:
Just one tiny nit. Very clean

icechunk-python/python/icechunk/_icechunk_python.pyi
paraseba: good call! done

icechunk/src/storage/s3.rs
```diff
@@ -308,6 +308,135 @@ impl S3Storage {
...
+        content_type: Option<impl Into<String>>,
+        metadata: I,
+        storage_class: Option<&String>,
+        bytes: &Bytes,
+    ) -> StorageResult<()> {
+        let mut multi = self
+            .get_client()
+            .await
+            .create_multipart_upload()
+            // We would like this, but it fails in MinIO
```
dcherian, line 326: could gate these conditions on `endpoint_url.is_some()`?

paraseba: Not really, different object stores can run in different hostnames. But I think, for now at least, the defaults seem to be OK

paraseba:
This PR is waiting on Tigris to solve a checksum issue when a partial object is requested using the latest S3 SDK.

icechunk/src/storage/s3.rs
dcherian: Right but for S3, endpoint_url will be None, so we can opt-in to the extra checks then?

PR #918, opened 2025-04-16
Allow  to run Rust tests with logs

paraseba:
Remember to pass `--nocapture` when you run the tests.

By default levels >= Warn are shown.

Example run with info level:

```
RUST_LOG=icechunk=info cargo test --all --all-features -- --nocapture
```

To make this work we need to replace all `test` and `tokio::test` attribute macros by our own.


test_macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "test_macros"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
```
dcherian, line 6: is this a separate workspace to keep the dependencies separate?

dcherian (approved):
ah this is a great unlock. Can we add the magic command to `just`?

test_macros/Cargo.toml
paraseba: no, it's a separate crate. One limitation of Rust macros, this type of macro at least, is that you need to declare them in a separate crate to be able to use them.

icechunk/Cargo.toml
```diff
@@ -63,9 +63,15 @@ assert_fs = { version = "1.1.2", optional = true }
 flatbuffers = "25.2.10"
 
 [dev-dependencies]
+icechunk-macros = { path = "../icechunk-macros" }
```
mpiannucci, line 66:
```suggestion
icechunk-macros = { path = "../icechunk-macros" , version = "0.1.0" }
```

mpiannucci (approved):
One change

paraseba:
Good catch @mpiannucci , thank you

Issue #919, opened 2025-04-18
API for automatic rebasing on commit


rabernat:
# Context

Currently the commit process is extremely careful in order to guarantee serializability of transactions against a store. Here are some of the properties of how it works today

- When a session starts, it records the parent commit ID for its branch
- When that session commits, it checks that the tip that branch still matches the expected parent ID (in a race-condition-safe way)
  - If so, the commit succeeds
  - If not, the commit fails with a `ConflictError` (perhaps the wrong name for this error), even if there are no actual conflicts between chunks, metadata etc
- The user has the option to detect the `ConflictError` and attempt a rebase, using various rebase strategies
- If the rebase fails, the user can retry the entire transaction from the start of the session. 

As @paraseba argued on Slack, this conservative approach is necessary to ensure full consistency between what is read and what is written, e.g. in an application like a bank-account ledger.

> ...there are not only write conflicts, there are also read conflicts we cannot detect today. Even more, there are side-channels to read that we will never be able to detect (example, a second repo instance). There may be no chunk conflicts in the writes, but the user computed their chunk value based on the values of an old neighboring chunk

However, many users don't care about that level of consistency, and just want to be able to do concurrent, uncoordinated writes to different parts of their repos (e.g. writing to different groups within a hierarchy, updating different parts of an array in parallel). The current API is cumbersome for them.

# Proposal

We should add a way to "automatically rebase on commit". There are two parameters that the user needs to specify:

- `rebase_strategy: ConflictSolver` - how to perform the rebase. I think `ConflictDetector` is a good default (if we have a default).
- `rebase_retries: int` - How many times to try rebasing before giving up.

There are two places we could add this

1. Add to existing `session.commit` method. We could add `rebase_strategy=None` as the default (to preserve the current behavior), but allow specifying a rebase strategy, which would trigger a retry loop.
2.  A a new `session.rebase_and_commit` method.

I have a weak preference for 1.

Either way, excellent documentation is needed to allow users to use this feature correctly.

paraseba:
I like adding `rebase_strategy=None`.

mpiannucci:
+1 to option 1

rabernat:
Now if only we could convince the maintainers to implement it... 😆

paraseba:
> Now if only we could convince the maintainers to implement it... 😆

I have a contact

PR #920, opened 2025-04-18
Commit learns how to rebase in case of conflict

paraseba:
`Session.commit` accepts new arguments:

```
rebase_with: ConflictSolver | None = None,
rebase_tries: int = 5
```

It uses the `ConflictSolver` to call `Session.rebase` in a loop, up to `rebase_tries` times, and try to commit after each rebase.

Closes: #919


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1066,7 +1066,13 @@ class PySession:
     @property
     def store(self) -> PyStore: ...
     def merge(self, other: PySession) -> None: ...
-    def commit(self, message: str, metadata: dict[str, Any] | None = None) -> str: ...
+    def commit(
+        self,
+        message: str,
+        metadata: dict[str, Any] | None = None,
+        rebase_with: ConflictSolver | None = None,
+        rebase_tries: int = 5,
```
rabernat, line 1074:
```suggestion
        rebase_tries: int = 1000,
```

In my experience, 5 is almost never going to be enough. If you invoke this option, I think you want to to basically keep trying until it truly hits a conflict.

icechunk-python/python/icechunk/session.py
```diff
@@ -250,7 +250,13 @@ def merge(self, other: Self) -> None:
         """
         self._session.merge(other._session)
 
-    def commit(self, message: str, metadata: dict[str, Any] | None = None) -> str:
+    def commit(
+        self,
+        message: str,
+        metadata: dict[str, Any] | None = None,
+        rebase_with: ConflictSolver | None = None,
+        rebase_tries: int = 5,
```
rabernat, line 258:
```suggestion
        rebase_tries: int = 1000,
```

icechunk-python/src/session.rs
```diff
@@ -180,23 +180,35 @@ impl PySession {
...
-                    .await
-                    .commit(message, metadata.map(|m| m.into()))
-                    .await
-                    .map_err(PyIcechunkStoreError::SessionError)?;
+                let mut session = self.0.write().await;
+                let snapshot_id = if let Some(solver) = rebase_with {
+                    session
+                        .commit_rebasing(
+                            solver.as_ref(),
+                            rebase_tries.unwrap_or(5),
```
rabernat, line 201:
```suggestion
                            rebase_tries.unwrap_or(1000),
```

rabernat (approved):
API looks good! But I would massively increase the number of retries for the default.

I recommend updating docs as part of this PR if possible.

icechunk-python/src/session.rs
```diff
@@ -180,23 +180,35 @@ impl PySession {
...
-                    .map_err(PyIcechunkStoreError::SessionError)?;
+                let mut session = self.0.write().await;
+                let snapshot_id = if let Some(solver) = rebase_with {
+                    session
+                        .commit_rebasing(
+                            solver.as_ref(),
+                            rebase_tries.unwrap_or(1_000),
+                            message,
+                            metadata,
+                            |_| async {},
```
mpiannucci, line 204: Im guessing these are left for future work in python?

mpiannucci (approved):

icechunk-python/src/session.rs
paraseba: I'm currently using them for Rust tests, if they become useful we can expose them to python.

paraseba:
Default retries and documentation updated.

PR #921, opened 2025-04-18
Credential refresh can be initialized with the first set of credentials

paraseba:
`s3_storage` style functions got a new argument:

```
resolve_get_credentials: bool = False
```

If True is passed, pickled repos and sessions will use the result returned the first time `get_credentials` was called, instead of having to run it again.

This speeds up short distributed tasks, since they no longer need to spend time fetching credentials. Tasks will reuse the credentials gathered by the "coordinator"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1701,6 +1701,7 @@ dependencies = [
  "miette",
  "pyo3",
  "pyo3-async-runtimes",
+ "rand 0.9.1",
  "serde",
  "serde_json",
  "thiserror 2.0.12",
--- a/icechunk-python/Cargo.toml
+++ b/icechunk-python/Cargo.toml
@@ -39,6 +39,7 @@ typetag = "0.2.20"
 serde = { version = "1.0.219", features = ["derive", "rc"] }
 miette = { version = "7.5.0", features = ["fancy"] }
 clap = { version = "4.5", features = ["derive"], optional = true }
+rand = "0.9.0"
 
 [features]
 cli = ["clap", "icechunk/cli"]
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -1260,8 +1260,13 @@ class S3Credentials:
         ----------
         pickled_function: bytes
             The pickled function to use to provide credentials.
+        current: S3StaticCredentials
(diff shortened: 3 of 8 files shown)


rabernat:
I don't understand the name `resolve_get_credentials`. `cache_credentials` maybe?

Also, this seems like a good default. Why make the default False?

mpiannucci:
> Also, this seems like a good default. Why make the default False?

Security I would guess, in case you don't creds pickled over the wire

mpiannucci:
I think we will need the same option for GCS as well

paraseba:
> I don't understand the name `resolve_get_credentials`. `cache_credentials` maybe?

yeah, god, that name sucks. I was trying to imply that the first set of creds is cached for its expiration time, but not beyond that.                                                                             `cache_credentials` works better, I change to that if there are no other ideas.

> Also, this seems like a good default. Why make the default False?

I'm open to make it the default. Just for security. I think it depends on how frequent it is for coordinator/worker communication to be unencrypted. I bet that's not uncommon, because of the pain involved in setting up ssl.

paraseba:
@mpiannucci 

> I think we will need the same option for GCS as well

I think I did that? Did I miss it somewhere?

mpiannucci:
> @mpiannucci 
> 
> 
> 
> > I think we will need the same option for GCS as well
> 
> 
> 
> I think I did that? Did I miss it somewhere?

You did! I commented to early :( sorry!!!!!

mpiannucci (approved):
Looks good, took me a minute to remember. this isnt directly pickled, its pickled by rust when you bundle a session i beklieve right?

paraseba:
> Looks good, took me a minute to remember. this isnt directly pickled, its pickled by rust when you bundle a session i believe right?

yes exactly, the description in terms of pickling is purely a user level thing, they don't need to know the details. In reality, it is only the Storage instance that contains this info when it gets pickled

rabernat:
Just wanted to note that my experiments last night suggested this did not work as expected, and that pickled / scattered Repos were still calling `credential_refresh_func` rather than using cached credentials.

paraseba:
@rabernat is this even after you remove the access_key and secret? Passing only `get_credentials` and `scatter* = True` arguments?

rabernat:
Correct

paraseba:
@rabernat I'll investigate

PR #925, opened 2025-04-21
Add raster datacube ingest

dcherian:
This is really rough, but I would also like to be done with it :)

🤞🏾 the notebook renders in the website.


lindseynield (approved):
I didn't know `merge_sessions` existed Very cool :) 

I did not run through this whole notebook myself, but the logic seems right. I think the amount of description you have is enough; we will run through this very quickly during the workshop and folks can reach out if they have more specific questions

PR #934, opened 2025-04-28
Raster notebook: Use only text repr

dcherian:
This should render much better


dcherian:
This is still weird, but better

PR #935, opened 2025-04-29
Update local_filesystem_storage parameter in docs

maxrjones:
The parameter name changed at some point from `prefix` to `path`.


paraseba:
Thank you @maxrjones !

PR #938, opened 2025-05-07
Initial HTTP support for Virtual Refs

mpiannucci:


paraseba (commented):
Can we document or link to documentation on what the options are?

mpiannucci:
Yeah I can actually make them more strongly typed

mpiannucci:
I think this is good to merge

paraseba:
> I think this is good to merge

I'll take a look today @mpiannucci

betolink:
This feature would be amazing to have! we are testing VirtualiZarr integrations and one of the most recurrent questions  has been "can we test it with Icechunk?" and our data doesn't allow egress with S3(NASA Earth). thanks for all the work @paraseba @mpiannucci !

icechunk/src/storage/object_store.rs
```diff
@@ -756,6 +757,61 @@ impl ObjectStoreBackend for LocalFileSystemObjectStoreBackend {
...
+    ) -> Result<Arc<dyn ObjectStore>, StorageError> {
+        let builder = HttpBuilder::new().with_url(&self.url);
+
+        // Add options
+        let builder = self
+            .config
+            .as_ref()
+            .unwrap_or(&HashMap::new())
+            .iter()
+            .fold(builder, |builder, (key, value)| builder.with_config(*key, value));
```
paraseba, line 802: @mpiannucci can you use the `Settings` you get as argument to configure the retry attempts this store will do? You can base it on the S3 or GCS examples in this same file

icechunk/src/storage/object_store.rs
```diff
@@ -756,6 +757,61 @@ impl ObjectStoreBackend for LocalFileSystemObjectStoreBackend {
...
+                    .collect::<Vec<_>>()
+                    .join(", "))
+                .unwrap_or("None".to_string())
+        )
+    }
+}
+
+#[typetag::serde(name = "http_object_store_provider")]
+impl ObjectStoreBackend for HttpObjectStoreBackend {
+    fn mk_object_store(
```
paraseba, line 790: We would need to override the `can_write` method to return false.

icechunk/src/virtual_chunks.rs
```diff
@@ -58,6 +62,9 @@ impl VirtualChunkContainer {
             (ObjectStoreConfig::LocalFileSystem(..), _) => {
                 Err("in memory storage does not accept credentials".to_string())
             }
+            (ObjectStoreConfig::Http(_), _) => {
+                Err("http storage does not support credentials".to_string())
```
paraseba, line 67: Should we support basic auth at least? Is there a reason we cannot or simply not implementing in this PR? If so, please add a TODO in the code and a ticket

icechunk/src/virtual_chunks.rs
```diff
@@ -122,8 +129,25 @@ pub fn mk_default_containers() -> HashMap<ContainerName, VirtualChunkContainer>
...
+        (
+            "http".to_string(),
+            VirtualChunkContainer {
+                name: "http".to_string(),
+                url_prefix: "http".to_string(),
+                store: ObjectStoreConfig::Http(HashMap::new()),
+            },
+        ),
+        (
+            "https".to_string(),
```
paraseba: I guess we don't need to add `https` because `http` will match..

icechunk/src/virtual_chunks.rs
mpiannucci: I dont want to do it in this PR, Ill add a ticket

mpiannucci: done

icechunk/src/storage/object_store.rs
mpiannucci: done

icechunk/src/storage/object_store.rs
mpiannucci: done

icechunk/src/virtual_chunks.rs
mpiannucci: done

PR #948, opened 2025-05-13
[WIP] Optimize split manifest writes.

dcherian:


icechunk/src/change_set.rs
```diff
@@ -24,20 +25,81 @@ pub struct ArrayData {
...
+            } else {
+                false
+            }
+        })
+    }
+}
+
 #[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub struct ChangeSet {
+    splitting: ManifestSplittingConfig,
```
dcherian, line 82: record the splitting config when writable session is created

icechunk/src/session.rs
```diff
@@ -1156,6 +1157,52 @@ async fn node_chunk_iterator<'a>(
...
+            let old_splits = ManifestSplits::from_extents(
+                manifests.iter().map(|m| m.extents.clone()).collect(),
+            );
+            // splits created when change_set was modified
+            let new_splits = change_set.splits(&node.id);
+
+            // FIXME: determine if splits are compatible. This is true iff every split
+            // in old_splits is fully contained within exactly one split in new_splits
+            // DISCUSS
+            let compatible = true;
```
dcherian, line 1184: This bit is messy, because we record the "actual" extents in the snapshot file, not the "theoretical extents". I am pretty sure we can dependably assert that the two extents are compatible.

icechunk/src/session.rs
```diff
@@ -1636,69 +1663,65 @@ impl ManifestSplitDimCondition {
 }
 
 impl ManifestSplittingConfig {
-    pub fn get_split_sizes(&self, node: &NodeSnapshot) -> SessionResult<ManifestSplits> {
```
dcherian, line 1639: Only the indent has changed

icechunk/src/change_set.rs
```diff
@@ -24,20 +25,81 @@ pub struct ArrayData {
...
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct SplitManifest {
+    from: Vec<u32>,
+    to: Vec<u32>,
+    // It's important we keep these sorted, we use this fact in TransactionLog creation
+    chunks: BTreeMap<ChunkIndices, Option<ChunkPayload>>,
+}
+
+impl SplitManifest {
+    pub fn update(&mut self, coord: ChunkIndices, data: Option<ChunkPayload>) {
```
paraseba, line 43: I wonder if  this will  be fast enough not to slow down massive virtual dataset ingestion. Maybe keeping a

`range: Vec<(u32,u32)>`

 would be slightly faster?

icechunk/src/change_set.rs
```diff
@@ -166,34 +276,60 @@ impl ChangeSet {
         node_id: NodeId,
         coord: ChunkIndices,
         data: Option<ChunkPayload>,
-    ) {
+    ) -> SessionResult<()> {
+        let cached_splits = self.splits.get(&node_id).expect("logic bug");
```
paraseba, line 280: let's put more info on the expect call, so we know where the error was triggered if somebody reports this error

icechunk/src/change_set.rs
```diff
@@ -166,34 +276,60 @@ impl ChangeSet {
...
+            .flatten()
     }
 
     /// Drop the updated chunk references for the node.
     /// This will only drop the references for which `predicate` returns true
     pub fn drop_chunk_changes(
         &mut self,
         node_id: &NodeId,
-        predicate: impl Fn(&ChunkIndices) -> bool,
+        predicate: impl Fn(&ChunkIndices) -> bool + Clone,
```
paraseba, line 327: This is so weird, never had to do it .... but I can't think of a better alternative. Maybe `+ Copy` works and is more efficient?

icechunk/src/change_set.rs
```diff
@@ -24,20 +25,81 @@ pub struct ArrayData {
...
+    // It's important we keep these sorted, we use this fact in TransactionLog creation
+    chunks: BTreeMap<ChunkIndices, Option<ChunkPayload>>,
+}
+
+impl SplitManifest {
+    pub fn update(&mut self, coord: ChunkIndices, data: Option<ChunkPayload>) {
+        if self.from.is_empty() {
+            debug_assert!(self.to.is_empty());
+            debug_assert!(self.chunks.is_empty());
+            // important to remember that `to` is not inclusive, so we need +1
```
paraseba, line 47: Hm ... I think we need to distinguish between setting and deleting chunks?

icechunk/src/change_set.rs
dcherian: yeah Copy works too, I wasn't sure which to pick and why.

icechunk/src/change_set.rs
dcherian: TODO: look for an existing data structure here?

icechunk/src/change_set.rs
```diff
@@ -100,11 +165,54 @@ impl ChangeSet {
...
+    pub fn update_array(
+        &mut self,
+        node_id: &NodeId,
+        path: &Path,
+        array_data: ArrayData,
+    ) -> SessionResult<()> {
+        // FIXME: This is an issue, if a dimension name is changed so that the `splits` are changed
+        // for this array. we are in a bad state.
+        // DISCUSS
+        self.update_cached_splits(
```
dcherian, line 210: TODO: could re-process if the splits end up changing

icechunk/src/change_set.rs
```diff
@@ -193,38 +212,50 @@ impl ChangeSet {
...
+                split.retain(|coord, _| !predicate(coord));
+            }
         }
     }
 
     pub fn array_chunks_iterator(
         &self,
         node_id: &NodeId,
         node_path: &Path,
+        extent: Option<ManifestExtents>,
```
dcherian: I will have to learn some lifetime things to make this `Option<&ManifestExtents>`.

EDIT: haven't tackled this yet

icechunk/src/change_set.rs
```diff
@@ -247,26 +300,24 @@ impl ChangeSet {
...
+    // ) {
+    //     self.set_chunks = chunks
+    // }
 
     /// Merge this ChangeSet with `other`.
     ///
     /// Results of the merge are applied to `self`. Changes present in `other` take precedence over
     /// `self` changes.
     pub fn merge(&mut self, other: ChangeSet) {
-        // FIXME: this should detect conflict, for example, if different writers added on the same
```
dcherian: TODO: revert this

icechunk/src/conflicts/detector.rs
```diff
@@ -145,7 +145,6 @@ impl ConflictSolver for ConflictDetector {
                     None
                 } else {
                     let conflicting: HashSet<_> = changes
-                        .keys()
```
dcherian, line 148: changed the iterator

icechunk/src/session.rs
```diff
@@ -199,6 +248,8 @@ pub struct Session {
     snapshot_id: SnapshotId,
     change_set: ChangeSet,
     default_commit_metadata: SnapshotProperties,
+    // This is an optimization so that we needn't figure out the split sizes on every set.
+    splits: HashMap<NodeId, ManifestSplits>,
```
dcherian, line 220: Stored on Session since we'll need to grab splits for arrays created in a previous session, but being modified in this one & the change_set doesn't have enough info (it receives NodeId, not NodeSnapshot IIRC)

icechunk/src/session.rs
```diff
@@ -411,6 +465,10 @@ impl Session {
         user_data: Bytes,
     ) -> SessionResult<()> {
         self.get_array(path).await.map(|node| {
+            // Q: What happens if we set a chunk, then change a dimension name, so
+            //   that the split changes.
+            // A: We ignore it. splits are set once for a node in a session, and are never changed.
+            // self.cache_splits(&node.id, path, &shape, &dimension_names);
```
dcherian: We could be stricter and error here if the splits change.

icechunk/src/session.rs
```diff
@@ -484,9 +567,14 @@ impl Session {
...
     ) -> SessionResult<()> {
-        if let NodeData::Array { shape, .. } = node.node_data {
+        if let NodeData::Array { shape, dimension_names, .. } = node.node_data {
             if shape.valid_chunk_coord(&coord) {
-                self.change_set.set_chunk_ref(node.id, coord, data);
+                let splits = self
+                    .get_splits(&node.id, &node.path, &shape, &dimension_names)
+                    // FIXME: this clone is a workaround for two mutable borrows
+                    // on self.change_set
+                    .clone();
```
dcherian, line 556: ☝🏾 not sure how to avoid this clone. there's a mutable borrow on self (and by extension self.change_set) and then another mutable borrow on `self.change_set` in the next line

icechunk/src/session.rs
```diff
@@ -1162,20 +1272,22 @@ async fn verified_node_chunk_iterator<'a>(
     snapshot_id: &'a SnapshotId,
     change_set: &'a ChangeSet,
     node: NodeSnapshot,
+    extent: Option<ManifestExtents>,
```
dcherian: Important: All iterators have gain `Option<ManifestExtents>` to do some filtering to extents=. This is where I'll need to add lifetimes to switch to `Option<&ManifestExtents>`

icechunk/src/session.rs
```diff
@@ -1493,81 +1584,109 @@ impl<'a> FlushProcess<'a> {
...
-                    .and_modify(|v| v.push(new_ref.clone()))
-                    .or_insert_with(|| vec![new_ref]);
-            }
+    async fn write_manifest_from_iterator(
+        &mut self,
+        chunks: impl Stream<Item = SessionResult<ChunkInfo>>,
+    ) -> SessionResult<Option<ManifestRef>> {
+        let mut from = vec![];
+        let mut to = vec![];
+        let chunks = aggregate_extents(&mut from, &mut to, chunks, |ci| &ci.coord);
```
dcherian, line 1625: I had to bring this back.

AFAICT there's no way to update the extents when chunks are deleted without scanning the whole thing. To do so, we'll need to keep track of "how many chunks have chunk index 'i'", decrement that counter when deleting chunks, and then construct a bbox from that histogram when needed. This would be a format change.

icechunk/src/session.rs
```diff
@@ -1636,69 +1807,65 @@ impl ManifestSplitDimCondition {
...
-                //   - path: *
-                //     manifest-split-size:
-                //     - t : 10
-                //     - y : 2
-                let mut already_matched: HashSet<usize> = HashSet::new();
-
-                #[allow(clippy::expect_used)]
-                let split_sizes = self
-                    .split_sizes
+    pub fn get_split_sizes(
```
dcherian, line 1809: no logic changes, just the signature. Arguably, it should've been this way earlier 🤦🏾‍♂️

icechunk/src/session.rs
```diff
@@ -1923,6 +2082,63 @@ async fn fetch_manifest(
     Ok(asset_manager.fetch_manifest(manifest_id, manifest_info.size_bytes).await?)
 }
 
+/// Map the iterator to accumulate the extents of the chunks traversed
```
dcherian, line 2103: no logic changes

icechunk/src/format/manifest.rs
```diff
@@ -52,10 +56,32 @@ impl ManifestExtents {
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    pub fn intersection(&self, other: &Self) -> Option<Self> {
+        debug_assert_eq!(self.len(), other.len());
+        let ranges = std::iter::zip(self.iter(), other.iter())
+            .map(|(a, b)| max(a.start, b.start)..min(a.end, b.end))
+            .collect::<Vec<_>>();
+        if any(ranges.iter(), |r| r.end < r.start) { None } else { Some(Self(ranges)) }
```
paraseba: should the condition be `r.end <= r.start`? Aren't these `[)` intervals? One empty set gives an empty intersection

icechunk/src/change_set.rs
```diff
@@ -24,15 +23,14 @@ pub struct ArrayData {
...
 #[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub struct ChangeSet {
     new_groups: HashMap<Path, (NodeId, Bytes)>,
     new_arrays: HashMap<Path, (NodeId, ArrayData)>,
     updated_arrays: HashMap<NodeId, ArrayData>,
     updated_groups: HashMap<NodeId, Bytes>,
-    // It's important we keep these sorted, we use this fact in TransactionLog creation
-    // TODO: could track ManifestExtents
-    set_chunks: BTreeMap<NodeId, BTreeMap<ChunkIndices, Option<ChunkPayload>>>,
+    set_chunks: BTreeMap<NodeId, HashMap<ManifestExtents, SplitManifest>>,
```
paraseba, line 33: I don't imagine how using the ManifestExtents as key helps. I'll continue to read, but it may be good to have a comment here explaining how the datastructure works. The way I imagine it, given a ChunkIndices you need to decide which Manifest it belongs to, but that sounds like a search anyway, how does the HashMap help?

icechunk/src/change_set.rs
```diff
@@ -60,17 +58,18 @@ impl ChangeSet {
 
     pub fn chunk_changes(
```
paraseba: with this new signature, a better name could be `changed_chunks`

icechunk/src/change_set.rs
```diff
@@ -166,23 +165,43 @@ impl ChangeSet {
         node_id: NodeId,
         coord: ChunkIndices,
         data: Option<ChunkPayload>,
+        splits: &ManifestSplits,
     ) {
+        #[allow(clippy::expect_used)]
+        let (_, extent) = splits.which_extent_and_index(&coord).expect("logic bug. Trying to set chunk ref but can't find the appropriate split manifest.");
```
paraseba: I think this is what i meant in my previous comment, isn't the index enough if you saved the SplitManifests in a vector instead of a HashMap?

icechunk/src/session.rs
```diff
@@ -160,31 +161,79 @@ impl From<VirtualReferenceError> for SessionError {
 
 pub type SessionResult<T> = Result<T, SessionError>;
 
+// Returns the index of split_range that includes ChunkIndices
+// This can be used at write time to split manifests based on the config
+// and at read time to choose which manifest to query for chunk payload
+pub fn which_extent_and_index<'a>(
+    iter: impl Iterator<Item = &'a ManifestExtents>,
+    coord: &ChunkIndices,
+) -> SessionResult<(usize, ManifestExtents)> {
```
paraseba: I feel this should return an `Option` instead of a `SessionResult`

icechunk/src/session.rs
```diff
@@ -160,31 +161,79 @@ impl From<VirtualReferenceError> for SessionError {
...
+    // ManifestExtents in split_range MUST NOT overlap with each other. How do we ensure this?
+    // ndim must be the same
+    // debug_assert_eq!(coord.0.len(), split_range[0].len());
+    // FIXME: could optimize for unbounded single manifest
+    // Note: I don't think we can distinguish between out of bounds index for the array
+    //       and an index that is part of a split that hasn't been written yet.
+    iter.enumerate()
+        .find(|(_, e)| e.contains(coord.0.as_slice()))
+        .map(|(i, e)| (i, e.clone()))
+        .ok_or(
```
paraseba: e can probably save cycles using `ok_or_else`, to avoid the allocation and the `clone`. But see my comments about:
* returning an Option
* returning purely an index

icechunk/src/change_set.rs
```diff
@@ -166,23 +165,43 @@ impl ChangeSet {
...
         node_id: &NodeId,
         coords: &ChunkIndices,
     ) -> Option<&Option<ChunkPayload>> {
-        self.set_chunks.get(node_id).and_then(|h| h.get(coords))
+        if let Some(node_chunks) = self.set_chunks.get(node_id) {
+            which_extent_and_index(node_chunks.keys(), coords).ok().and_then(
+                |(_, extent)| node_chunks.get(&extent).and_then(|s| s.get(coords)),
+            )
+        } else {
+            None
```
paraseba: nitpicking: `else {None}` feels weird when i see it, something like `if condition then True else False`. But maybe you just didn't want to pile so many nested `and_then`

icechunk/src/change_set.rs
```diff
@@ -166,23 +165,43 @@ impl ChangeSet {
         node_id: NodeId,
         coord: ChunkIndices,
         data: Option<ChunkPayload>,
+        splits: &ManifestSplits,
     ) {
+        #[allow(clippy::expect_used)]
```
paraseba, line 243: Help me understand this, when would `which_extent_and_index` return an error? Are we expecting users to always have a config that allows infinite extents? If so, this shouldn't be an expect but an error, the program shouldn't die.

icechunk/src/change_set.rs
```diff
@@ -166,23 +165,43 @@ impl ChangeSet {
...
+            .or_insert_with(|| {
+                let mut h = HashMap::<
+                    ManifestExtents,
+                    BTreeMap<ChunkIndices, Option<ChunkPayload>>,
+                >::with_capacity(splits.len());
+                h.entry(extent.clone())
+                    // TODO: this is duplicative. I can't use `or_default` because it's
+                    // nice to create the HashMap using `with_capacity`
+                    .or_default()
+                    .insert(coord, data);
```
paraseba: Note to self: we are inserting `Nones` in the ManifestSplits when chunks are deleted. Then at write time we'll need to check if the Manifest is empty before writing it.

icechunk/src/change_set.rs
```diff
@@ -193,38 +212,50 @@ impl ChangeSet {
...
             return Either::Left(iter::empty());
         }
         match self.set_chunks.get(node_id) {
             None => Either::Left(iter::empty()),
-            Some(h) => Either::Right(h.iter()),
+            Some(h) => Either::Right(
+                h.iter()
+                    // FIXME: review this
+                    .filter(move |(manifest_extent, _)| {
+                        extent.is_none() || Some(*manifest_extent) == extent.as_ref()
```
paraseba: weird it looks like if extent was an option to a reference this would be even easier. But I can help figure it out. That said, I still want to understand why you are using the ManifestExtent as a "key" in all these places. Will continue reading.

icechunk/src/change_set.rs
```diff
@@ -235,6 +266,28 @@ impl ChangeSet {
...
+        &self,
+        node_id: &NodeId,
+        node_path: &Path,
+    ) -> impl Iterator<Item = (&ManifestExtents, &SplitManifest)> + use<'_> {
+        if self.is_deleted(node_path, node_id) {
+            return Either::Left(iter::empty());
+        }
+        match self.set_chunks.get(node_id) {
+            None => Either::Left(iter::empty()),
+            Some(h) => Either::Right(h.iter()),
```
paraseba: Note to self: What happens if the SplitManifest is full of Nones (deleted chunks)? I'll find out when I read the session code

icechunk/src/session.rs
```diff
@@ -160,31 +161,79 @@ impl From<VirtualReferenceError> for SessionError {
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Overlap {
+    Complete,
+    Partial,
+    None,
+}
+
+/// Important: this is not symmetric.
```
paraseba: maybe adding it with a different name to the `impl` would help make the asymmetry more clear?

icechunk/src/session.rs
```diff
@@ -160,31 +161,79 @@ impl From<VirtualReferenceError> for SessionError {
...
+pub fn overlaps(us: &ManifestExtents, them: &ManifestExtents) -> Overlap {
+    debug_assert!(us.len() == them.len());
+
+    let mut overlaps = vec![];
```
paraseba: I think this algorithm could be simpler/faster:
* iterate over coords
* initialize result = Complete
* if no overlap in the coord return None
* if partial overlap in the coord assign result = Partial
* When iteration ends return result

icechunk/src/session.rs
```diff
@@ -220,6 +271,7 @@ impl Session {
             snapshot_id,
             change_set: ChangeSet::default(),
             default_commit_metadata: SnapshotProperties::default(),
+            splits: Default::default(),
```
paraseba, line 244: Note to self: Interesting, I wonder how this thing will get initialized with the info from previous sessions

icechunk/src/session.rs
paraseba: I'm not sure that is the problem. I think this has to do with the lifetime of the `splits` argument to `set_chunk_ref`

icechunk/src/change_set.rs
dcherian: Yes I agree, an alternate one would be:
```
set_chunks: BTreeMap<NodeId, Vec<SplitManifest>
```
where
```
pub struct SplitManifest {
    extents: ManifestExtents,
    chunks: BTreeMap<ChunkIndices, Option<ChunkPayload>>,
}
```

icechunk/src/change_set.rs
dcherian: Yes, we rely on `verified_node_chunk_iterator` to do the "join" between the changeset and the manifest on disk for a specified split, if it returns nothing then no manifest is written for that split.

icechunk/src/format/manifest.rs
dcherian: Should we disallow empty ManifestExtents? I thought we were doing that, but apparently not.

icechunk/src/change_set.rs
dcherian: In general, I'm treating the `Vec<ManifestExtents>` constructed from the splitting config as a reference frame.

 For example, in `flush` I loop over `Vec<ManifestExtents>` and then ask for an iterator of chunks for each `ManifestExtents` and then I do the lookup on the Changeset. Since I'm doing a lookup by `ManifestExtents` I went with a map...

Since we store the "actual extents" on disk, we need to use a `Vec<ManifestExtents>` constructed from the splitting config as a reference frame to keep everything straight.

Does this help?

icechunk/src/format/manifest.rs
```diff
@@ -52,10 +64,49 @@ impl ManifestExtents {
...
+
+    pub fn union(&self, other: &Self) -> Self {
+        debug_assert_eq!(self.len(), other.len());
+        Self::from_ranges_iter(
+            zip(self.iter(), other.iter())
+                .map(|(a, b)| min(a.start, b.start)..max(a.end, b.end)),
+        )
+    }
+
+    pub fn overlap_with(&self, other: &Self) -> Overlap {
```
dcherian, line 94: Moved here to 'suggest' the asymmetry.

icechunk/src/change_set.rs
```diff
@@ -114,6 +119,30 @@ impl ChangeSet {
...
+            let mut new_manifests =
+                HashMap::<ManifestExtents, SplitManifest>::with_capacity(
+                    new_splits.len(),
+                );
+            for (old_extents, chunks) in manifests.into_iter() {
+                // FIXME: these extents had better be compatible
+                // test_cases: increasing size of (multiple) dimensions
+                //             decreasing size of (multiple) dimensions
+                //
+                new_splits
```
dcherian: yeah this is ugly. Since we track extents on the change set, we need to update them for any `resize` 🤦🏾‍♂️

icechunk/src/change_set.rs
dcherian: It shoudl only ever be a logic bug

icechunk/src/change_set.rs
dcherian: Yes we rely on `verified_node_chunks_iterator` to filter those out.

icechunk/src/session.rs
dcherian: Now moved to `ManifestExtents.overlap_with(other: &ManifestExtents)`

icechunk/src/change_set.rs
dcherian: Actually, I have renamed it now. We only need the keys.

icechunk/src/change_set.rs
```diff
@@ -275,15 +344,14 @@ impl ChangeSet {
...
-                    self.set_chunks.insert(node, other_chunks);
-                }
+        for (node, other_splits) in other.set_chunks {
+            let manifests = self.set_chunks.entry(node).or_insert_with(|| {
+                HashMap::<ManifestExtents, SplitManifest>::with_capacity(
+                    other_splits.len(),
+                )
+            });
+            for (extent, their_manifest) in other_splits {
+                manifests.entry(extent).or_default().extend(their_manifest)
```
dcherian, line 399: this is readable but seems silly for the default case where we can just insert `their_manifest`.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -44,10 +47,10 @@ def chunk_paths(
 class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
     def __init__(self, repo: Repository) -> None:
         self.repo = repo
-        store = repo.writable_session("main").store
+        store = LoggingStore(repo.writable_session("main").store, log_level=logging.INFO + 1)
```
dcherian: I'll revert this, was useful for debugging

icechunk/src/change_set.rs
```diff
@@ -24,15 +23,18 @@ pub struct ArrayData {
...
     updated_arrays: HashMap<NodeId, ArrayData>,
     updated_groups: HashMap<NodeId, Bytes>,
-    // It's important we keep these sorted, we use this fact in TransactionLog creation
-    // TODO: could track ManifestExtents
-    set_chunks: BTreeMap<NodeId, BTreeMap<ChunkIndices, Option<ChunkPayload>>>,
+    set_chunks: BTreeMap<NodeId, HashMap<ManifestExtents, SplitManifest>>,
+    // This map keeps track of any chunk deletes that are
+    // outside the domain of the current array shape. This is needed to handle
+    // the very unlikely case of multiple resizes in the same session.
+    deleted_chunks_outside_bounds: BTreeMap<NodeId, HashSet<ChunkIndices>>,
```
dcherian, line 37: This seems to be the cleanest solution to me.

This commit tried tracking all deleted chunks separately (https://github.com/earth-mover/icechunk/pull/948/commits/5af247e6979bea6abc92805cf02da6d049410d03) but it's very messy.

icechunk/src/repository.rs
```diff
@@ -1506,10 +1560,8 @@ mod tests {
                     .await?
             }
 
-            add += (axis_size as u32).div_ceil(expected_split_sizes[ax]) as usize
-                - 1 * ((ax > 0) as usize);
-            dbg!(&ax, &add);
-            total_manifests += add;
+            total_manifests +=
```
dcherian, line 1807: number went down!

icechunk/src/repository.rs
```diff
@@ -1245,7 +1320,7 @@ mod tests {
             )
             .await?;
         session.commit("last split", None).await?;
-        total_manifests += 2; // FIXME: this should be +1 once writes are optimized
+        total_manifests += 1;
```
dcherian, line 1479: number went down!

icechunk/src/repository.rs
```diff
@@ -1187,6 +1188,80 @@ mod tests {
...
+        session
+            .update_array(
+                &array_path,
+                shape3.clone(),
+                dimension_names.clone(),
+                array_def.clone(),
+            )
+            .await?;
+        session.commit("second commit", None).await?;
+        assert_manifest_count(&storage, 2).await;
```
dcherian, line 1293: test for #604

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -47,7 +49,7 @@ def __init__(self, repo: Repository) -> None:
         store = repo.writable_session("main").store
         super().__init__(store)
 
-    @precondition(lambda self: self.store.session.has_uncommitted_changes)
+    @precondition(lambda self: self.store._store.session.has_uncommitted_changes)
```
dcherian, line 52:
```suggestion
    @precondition(lambda self: self.store.session.has_uncommitted_changes)
```

icechunk/tests/test_gc.rs
```diff
@@ -164,7 +167,7 @@ pub async fn do_test_gc(
     )
     .await?;
     assert_eq!(summary.chunks_deleted, 10);
-    assert_eq!(summary.manifests_deleted, 110);
+    assert_eq!(summary.manifests_deleted, 1);
```
paraseba, line 173: remind me why this changed, maybe even add a comment in the code

icechunk/src/session.rs
```diff
@@ -160,31 +163,44 @@ impl From<VirtualReferenceError> for SessionError {
 
 pub type SessionResult<T> = Result<T, SessionError>;
 
+// Returns the index of split_range that includes ChunkIndices
+// This can be used at write time to split manifests based on the config
+// and at read time to choose which manifest to query for chunk payload
+/// It is useful to have this act on an iterator (e.g. get_chunk_ref)
+/// The find method on ManifestSplits is simply a helper.
+pub fn find_coord<'a, I>(mut iter: I, coord: &'a ChunkIndices) -> Option<ManifestExtents>
```
paraseba: I don't understand why this doesn't return a `Option<&'a ManifestExtents>`

icechunk/src/session.rs
```diff
@@ -160,31 +163,44 @@ impl From<VirtualReferenceError> for SessionError {
...
+    // Note: I don't think we can distinguish between out of bounds index for the array
+    //       and an index that is part of a split that hasn't been written yet.
+    iter.find(|e| e.contains(coord.0.as_slice())).cloned()
+}
+
+pub fn position_coord<'a, I>(iter: I, coord: &'a ChunkIndices) -> Option<usize>
+where
+    I: Iterator<Item = &'a ManifestExtents>,
+{
+    enumerate(iter).find(|(_, e)| e.contains(coord.0.as_slice())).map(|x| x.0)
```
paraseba: why not combining this function an the previous one into a single function that returns both things? As long as you manage to return a & in the previous function it should be OK from performance perspective

icechunk/src/session.rs
```diff
@@ -475,6 +500,39 @@ impl Session {
...
+
+    fn get_splits(
+        &mut self,
+        node_id: &NodeId,
+        path: &Path,
+        shape: &ArrayShape,
+        dimension_names: &Option<Vec<DimensionName>>,
+    ) -> &ManifestSplits {
+        if !self.splits.contains_key(node_id) {
+            self.cache_splits(node_id, path, shape, dimension_names);
```
paraseba: Performance of this method is critical, it gets called at least on `set_chunk_ref` which is critical for virtual refs. Can we do `entry` instead of `contains_key` and then `get`? Single search... and all that. Maybe it gets tricky because of cloning? If we cannot find a workaround let's put a comment in the code.

icechunk/src/session.rs
paraseba: Let's try to avoid this clone. Do you want me to give it a try?

icechunk/src/session.rs
```diff
@@ -840,12 +904,23 @@ impl Session {
...
+
+        if self.splits.iter().any(|(node, our_splits)| {
+            splits
+                .get(node)
+                .is_some_and(|their_splits| !our_splits.compatible_with(their_splits))
+        }) {
+            return Err(SessionErrorKind::IncompatibleSplits.into());
+        }
+
+        self.splits.extend(splits);
```
paraseba: Is a simple `extend` enough for any compatible sets of splits? Shouldn't `our` splits "absorb" theirs? Imagine theirs are smaller...

icechunk/src/session.rs
```diff
@@ -108,6 +109,8 @@ pub enum SessionErrorKind {
     InvalidIndex { coords: ChunkIndices, path: Path },
     #[error("invalid chunk index for splitting manifests: {coords:?}")]
     InvalidIndexForSplitManifests { coords: ChunkIndices },
+    #[error("incompatible manifest splitting config when merging sessions")]
+    IncompatibleSplits,
```
paraseba: Can we provide more information here? Maybe the first two non compatible splits we detected (or all of them)

icechunk/src/session.rs
paraseba: I wonder if for all these cases, instead of using `Option` it's easier to introduce a (constant) match all ManifestExtents: `ManifestExtents::all : ManifestExtents`

icechunk/src/session.rs
```diff
@@ -1493,67 +1576,65 @@ impl<'a> FlushProcess<'a> {
...
+        let mut to = vec![];
+        let chunks = aggregate_extents(&mut from, &mut to, chunks, |ci| &ci.coord);
 
-                self.manifest_refs
-                    .entry(node_id.clone())
-                    .and_modify(|v| v.push(new_ref.clone()))
-                    .or_insert_with(|| vec![new_ref]);
-            }
+        #[allow(clippy::expect_used)]
+        if let Some(new_manifest) = Manifest::from_stream(chunks)
```
paraseba, line 1627: can chunks be empty? iirc from_stream will return None in that case? If it can't add a comment to the code to justify the expect please

icechunk/src/session.rs
```diff
@@ -1576,19 +1676,82 @@ impl<'a> FlushProcess<'a> {
...
+                    .map(|new_ref| refs.insert(extent.clone(), vec![new_ref]));
+            } else {
+                let on_disk_bbox = on_disk_extents
+                    .iter()
+                    .filter_map(|e| e.intersection(extent))
+                    .reduce(|a, b| a.union(&b));
+
+                // split was unmodified in this session. Let's look at the current manifests
+                // and see what we need to do with them
+                for old_ref in manifests.iter() {
```
paraseba: it would help to rename the `manifests` arguments, to something that makes it clear which manifests are these

icechunk/src/session.rs
```diff
@@ -1576,19 +1676,82 @@ impl<'a> FlushProcess<'a> {
...
     ) -> SessionResult<()> {
-        let asset_manager = Arc::clone(&self.asset_manager);
-        let updated_chunks = updated_node_chunks_iterator(
-            asset_manager.as_ref(),
-            self.change_set,
-            self.parent_id,
-            node.clone(),
-        )
-        .await
-        .map_ok(|(_path, chunk_info)| chunk_info);
```
paraseba, line 1589: glorious block of code! very clear

icechunk/src/format/manifest.rs
paraseba: I think it would save us from confusion, yes.

icechunk/src/change_set.rs
paraseba: The `BTreeMap<..., Vec<...>>` you show above is what I was expecting. But this also makes sense:

> Since I'm doing a lookup by ManifestExtents I went with a map...

So whatever is clearer to you.

icechunk/src/session.rs
paraseba: @dcherian maybe add a comment here explaining how the splits get lazily populated

paraseba (commented):
This is looking awesome @dcherian !!!

How do you feel about our benchmark coverage? We'll probably need to introduce splitting in some benchmarks and to create a giant tiny chunk array and write it randomly multiple times?

icechunk/tests/test_gc.rs
dcherian: split size is 10; the script only overwrites the first 10 chunks, so only one manifest needs to be rewritten!

icechunk/src/session.rs
dcherian: `main` has an unwrap. I have modified to raise a new `SessionError::ManifestCreationError(Box<SessionError>)`

icechunk/src/session.rs
dcherian: using `entry` now.

dcherian: using `entry` now.

dcherian: using `entry` now.

icechunk/src/session.rs
dcherian: I'll add a comment.`session.splits` is always complete i.e. it has ALL possible splits. So a simple `extend` is enough `ours` will get overwritten by `theirs` if the same node appears in both but that is fine.

icechunk/src/session.rs
dcherian: I am now reporting the splitting configs for the two sessions. It seems hard to actual report a Path for the node whose splits are incompatible.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -13,18 +15,74 @@
...
+    pytest.mark.filterwarnings(
+        "ignore::zarr.core.dtype.common.UnstableSpecificationWarning"
+    ),
+]
+
+
+import functools
+
+
+def with_frequency(frequency):
```
dcherian, line 43: this was Claude. looks good!

dcherian:
Some benchmarks for a repo with 500_000 chunk refs, written to 50 manifests with 10_000 refs each.

1. Appends are much faster as expected. Each append adds 50_000 chunks refs = 5 new manifest files.
2. we have a slowdown for full rewrites . This is expected since we write each split in serial, this is slow but reduces memory consumption. As an experiment I did a crude async implementation for manifest writes per node, as expected this is a nice speedup.


After discussion with Seba, I have reverted that async write commit, and this PR should be good to go. I'll address #986 in a new PR.

#### Without async writes

```
------------------ benchmark 'refs-write test_write_split_manifest_refs_append': 4 tests ------------------
Name (time in ms)                                                                          Median
-----------------------------------------------------------------------------------------------------------
test_write_split_manifest_refs_append[no-splitting-large-1d] (p/s3_e184e2998)            792.6242 (1.64)
test_write_split_manifest_refs_append[no-splitting-large-1d] (p/s3_main_7e03)            815.8424 (1.69)
test_write_split_manifest_refs_append[split-size-10_000-large-1d] (p/s3_e184e2998)       483.8746 (1.0)
test_write_split_manifest_refs_append[split-size-10_000-large-1d] (p/s3_main_7e03)     1,987.4977 (4.11)
-----------------------------------------------------------------------------------------------------------

---------------- benchmark 'refs-write test_write_split_manifest_refs_full_rewrite': 4 tests ----------------
Name (time in s)                                                                             Median
-------------------------------------------------------------------------------------------------------------
test_write_split_manifest_refs_full_rewrite[no-splitting-large-1d] (p/s3_e184e2998)          1.7160 (1.04)
test_write_split_manifest_refs_full_rewrite[no-splitting-large-1d] (p/s3_main_7e03)          1.6557 (1.0)
test_write_split_manifest_refs_full_rewrite[split-size-10_000-large-1d] (p/s3_e184e2998)     3.9933 (2.41)
test_write_split_manifest_refs_full_rewrite[split-size-10_000-large-1d] (p/s3_main_7e03)     4.2187 (2.55)
-------------------------------------------------------------------------------------------------------------
```

#### With async writes
```
------------------ benchmark 'refs-write test_write_split_manifest_refs_append': 4 tests ------------------
Name (time in ms)                                                                          Median
-----------------------------------------------------------------------------------------------------------
test_write_split_manifest_refs_append[no-splitting-large-1d] (m/s3_HEAD_2a68)            709.7038 (2.97)
test_write_split_manifest_refs_append[no-splitting-large-1d] (m/s3_main_7e03)            764.4788 (3.20)
test_write_split_manifest_refs_append[split-size-10_000-large-1d] (m/s3_HEAD_2a68)       238.7356 (1.0)
test_write_split_manifest_refs_append[split-size-10_000-large-1d] (m/s3_main_7e03)     1,726.8433 (7.23)
-----------------------------------------------------------------------------------------------------------

---------------- benchmark 'refs-write test_write_split_manifest_refs_full_rewrite': 4 tests ----------------
Name (time in s)                                                                             Median
-------------------------------------------------------------------------------------------------------------
test_write_split_manifest_refs_full_rewrite[no-splitting-large-1d] (m/s3_HEAD_2a68)          1.6306 (1.22)
test_write_split_manifest_refs_full_rewrite[no-splitting-large-1d] (m/s3_main_7e03)          1.5534 (1.16)
test_write_split_manifest_refs_full_rewrite[split-size-10_000-large-1d] (m/s3_HEAD_2a68)     1.3361 (1.0)
test_write_split_manifest_refs_full_rewrite[split-size-10_000-large-1d] (m/s3_main_7e03)     3.6363 (2.72)
-------------------------------------------------------------------------------------------------------------
```

paraseba (approved):

PR #965, opened 2025-06-03
Disallow writing with consolidated metadata

### LINKED ISSUE #962: Icechunk should fail early if user tries to write with consolidated metadata
### What happened?

A user set consolidated metadata at the root group by mistake. Icechunk accepted the change. Now new groups are invisible.

### What did you expect to happen?

Failure when consolidated metadata is attempted

### Environment

icechunk 0.2.15

paraseba:
Icechunk provides its own metadata consolidation mechanism.

Closes #962


icechunk/src/store.rs
```diff
@@ -96,6 +96,10 @@ pub enum StoreErrorKind {
         "invalid chunk location, no matching virtual chunk container: `{chunk_location}`"
     )]
     InvalidVirtualChunkContainer { chunk_location: String },
+    #[error(
+        "Icechunk has its own metadata consolidation mechanism. Please disable Zarr's metadata consolidation for writing data on Icechunk. If you are using Xarray, you can pass consolidate=False to to_zarr, or simply use Icechunk's to_icechunk which already disables consolidation."
```
dcherian:
```suggestion
        "Icechunk has its own metadata consolidation mechanism. Please disable Zarr's metadata consolidation for writing data on Icechunk. If you are using Xarray, you can pass `consolidated=False` to `Dataset.to_zarr`, or simply use `icechunk.xarray.to_icechunk` which disables consolidation."
```

dcherian:
My concern here is that this will break existing code & pipelines.

paraseba:
> My concern here is that this will break existing code & pipelines.

@dcherian: But, aren't those pipelines already broken because they are not setting `consolidated=False`?

icechunk/src/store.rs
paraseba: done

paraseba:
We are going to use a different approach

PR #980, opened 2025-06-11
Align main with released tag 0.2.16

paraseba:


Cargo.lock
```diff
@@ -1696,7 +1696,7 @@ dependencies = [
 
 [[package]]
 name = "icechunk"
-version = "0.2.14"
+version = "0.2.16"
```
dcherian, line 1699: is this right?

paraseba: yeah, remember you tried to deploy and successfully the deployed the rust parts but not the python parts? So later changes (downgrading   s3) had to go with a new rust version that I applied in the release branch.

dcherian (approved):

PR #982, opened 2025-06-11
Support changing log levels and setting them in Python

paraseba:
This introduces a new function `icechunk.set_logs_filter(str)` that allows setting and changing the logs Icechunk will produce.

On module import, we still initialize logs to the value of ICECHUNK_LOG environment variable (or error level if not present).

### DIFF
--- a/icechunk-python/src/lib.rs
+++ b/icechunk-python/src/lib.rs
@@ -27,6 +27,7 @@ use conflicts::{
 use errors::{IcechunkError, PyConflictError, PyRebaseFailedError};
 use icechunk::{format::format_constants::SpecVersionBin, initialize_tracing};
 use pyo3::prelude::*;
+use pyo3::types::PyMapping;
 use pyo3::wrap_pyfunction;
 use repository::{PyDiff, PyGCSummary, PyManifestFileInfo, PyRepository, PySnapshotInfo};
 use session::PySession;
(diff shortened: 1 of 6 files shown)


icechunk/src/lib.rs
```diff
@@ -49,24 +49,60 @@ mod private {
 }
 
 #[cfg(feature = "logs")]
-pub fn initialize_tracing() {
+#[allow(clippy::type_complexity)]
+static LOG_FILTER: std::sync::LazyLock<
+    std::sync::Mutex<
```
mpiannucci, line 54: 😍

mpiannucci (approved):

icechunk-python/src/lib.rs
```diff
@@ -69,14 +70,31 @@ fn cli_entrypoint(_py: Python) -> PyResult<()> {
     Ok(())
 }
 
+fn log_filters_from_env(py: Python) -> PyResult<Option<String>> {
+    let os = py.import("os")?;
+    let environ = os.getattr("environ")?;
+    let environ: &Bound<PyMapping> = environ.downcast()?;
+    let value = environ.get_item("ICECHUNK_LOG").ok().and_then(|v| v.extract().ok());
```
dcherian, line 77: hahahah 🤯

PR #985, opened 2025-06-12
Add debug logging when `delete_objects` return Err

dcherian:


paraseba (approved):

dcherian:
I switched to `warn!` to make it prominent and cleaned up the logs.

PR #991, opened 2025-06-13
High concurrency script learns new tricks

paraseba:
* GCS
* not killing the cluster
* coiled workspaces
* different chunk sizes


icechunk-python/examples/high_concurrency.py
```diff
@@ -311,11 +331,13 @@ def report(self) -> None:
 def mk_client(
```
dcherian, line 330: we should probably set a default `keepalive` and allow it to be configurable too.

dcherian (approved):

PR #1002, opened 2025-06-24
Add `rewrite_manifests` EAR-1667

### LINKED ISSUE #986: Allow users to experiment with different splitting configurations
It's be nice to allow users to easily experiment with different splitting configurations. 

I came up with two approaches, though they aren't too different.

-----

### Approach 1: Branches and empty commits

Today I think we can do this (except the 'empty' commit bit):
```python
config1 = RepositoryConfig(...)
# assume this splitting config requires a rewrite of manifests
repo = Repository.open(..., config=config1)
repo.create_branch("split-1, repo.lookup_branch("main"))
session = repo.writable_session("split-1")
# need to allow 'empty commits' with no chunk or metadata changes.
session.commit("test splitting with config ...", empty=True)

# Now again
config2 = RepositoryConfig(...)
# assume this splitting config requires a rewrite of manifests
repo = Repository.open(..., config=config2)
repo.create_branch("split-2", repo.lookup_branch("main"))
session = repo.writable_session("split-2")
# need to allow 'empty commits' with no chunk or metadata changes.
session.commit("test splitting with config ...", empty=True)


# ...
# do some tests


# yay I like this config
repo.update_config(config2)
```
The main issue: the repo config isn't persisted on the branch so the user has to keep track externally of which config was used to create the manifests on a specific branch.

A secondary issue is that we will need to allow 'empty' commits to force a manifest rewrite even without chunk or metadata changes.

### Approach 2: specific `rewrite_manifests` method

We could alternatively allow an in-place rewrite of a snapshot.

```python
repo.rewrite_manifests(snapshot_id, splitting_config=ManifestSplittingConfig(...))
```

- I don't really like the idea of mutating snapshot files, though I guess we do it for GC.
- This doesn't really allow before/after comparisons without an empty commit; i.e. the workflow would be (a) create empty commit which doesn't rewrite manifests (b) explicitly rewrite with a config.

TBH i don't think this approach has much to recommend it over (1).

-----

Regardless, it seems like we need to allow empty commits. The choice is whether that empty commit will rewrite manifests when the current splitting config is incompatible with the splits on disk.

dcherian:
Closes #986

Updated docs here: https://icechunk--1002.org.readthedocs.build/en/1002/icechunk-python/performance/#splitting-manifests 

Questions:
1. Should we allow filtering by a prefix?
2. Shall we automatically add the splitting config to the commit's `metadata`? and make sure it's easy to use for creating a new repo?

Decisions:
1. I punted on (1)
2. I've done (2) just to record the config, but didn't add the ability to easily create a ManifestSplittingConfig from the info in metadata.

### DIFF
--- a/docs/docs/icechunk-python/performance.md
+++ b/docs/docs/icechunk-python/performance.md
@@ -242,9 +242,8 @@ or when reading from a Repository object that was just created.
 At that point, you will want to experiment with different manifest split configurations.
 
 To force Icechunk to rewrite all chunk refs to the current splitting configuration use [`rewrite_manifests`](./reference.md#icechunk.Repository.rewrite_manifests)
---- for the current example this will consolidate to two manifests.
 
-To illustrate, we will use a split size of 3.
+To illustrate, we will use a split size of 3 --- for the current example this will consolidate to two manifests.
 ```python exec="on" session="perf" source="material-block"
 split_config = ManifestSplittingConfig.from_dict(
     {ManifestSplitCondition.AnyArray(): {ManifestSplitDimCondition.Any(): 3}}
@@ -310,7 +309,7 @@ First we persist that configuration to disk
 repo.save_config()
 ```
 
-Now point main to the commit with rewritten manifests
+Now point the `main` branch to the commit with rewritten manifests
 ```python exec="on" session="perf" source="material-block"
 repo.reset_branch("main", repo.lookup_branch("split-experiment-1"))
 ```
--- a/docs/docs/icechunk-python/performance.md
+++ b/docs/docs/icechunk-python/performance.md
@@ -262,10 +262,11 @@ repo = ic.Repository.open(storage, config=repo_config)
 
 We will rewrite the manifests on a different branch
 ```python exec="on" session="perf" source="material-block"
-repo.create_branch("split-experiment-1")
+repo.create_branch("split-experiment-1", repo.lookup_branch("main"))
 snap = repo.rewrite_manifests(
     f"rewrite_manifests with new config", branch="split-experiment-1"
 )
+print(repo.lookup_snapshot(snap).manifests)
 ```
 Now benchmark reads on `main` vs `split-experiment-1`
 ```python exec="on" session="perf" source="material-block"
--- a/docs/docs/icechunk-python/performance.md
+++ b/docs/docs/icechunk-python/performance.md
@@ -16,9 +16,9 @@ Coming Soon.
 ## Splitting manifests
 
 Icechunk stores chunk references in a chunk manifest file stored in `manifests/`.
-For very large arrays (millions of chunks), these files can get quite large.
 By default, Icechunk stores all chunk references in a single manifest file per array.
-Requesting even a single chunk requires downloading the entire manifest.
+For very large arrays (millions of chunks), these files can get quite large.
+Requesting even a single chunk will require downloading the entire manifest.
 In some cases, this can result in a slow time-to-first-byte or large memory usage.
 Similarly, appending a small amount of data to a large array requires
 downloading and rewriting the entire manifest.
@@ -47,7 +47,7 @@ repo_config = ic.RepositoryConfig(
 )
 ```
 
-Then pass the config to `Repository.open` or `Repository.create`
+Then pass the `config` to `Repository.open` or `Repository.create`
 ```python
 repo = ic.Repository.open(..., config=repo_config)
 ```
@@ -206,7 +206,13 @@ This ends up rewriting all refs to two new manifests.
 
 ### Rewriting manifests
 
-To force Icechunk to rewrite all chunk refs to the current splitting configuration use [`rewrite_manifests`](./reference.md#icechunk.Repository.rewrite_manifests) --- for the current example this will consolidate to two manifests.
+Remember, by default Icechunk only writes one manifest per array regardless of size.
+For large enough arrays, you might see a relative performance hit while committing a new update (e.g. an append),
+or when reading from a Repository object that was just created.
+At that point, you will want to experiment with different manifest split configurations.
+
+To force Icechunk to rewrite all chunk refs to the current splitting configuration use [`rewrite_manifests`](./reference.md#icechunk.Repository.rewrite_manifests)
+--- for the current example this will consolidate to two manifests.
 
 To illustrate, we will use a split size of 3.
 ```python exec="on" session="perf" source="material-block"
@@ -219,7 +225,7 @@ repo_config = ic.RepositoryConfig(
 new_repo = ic.Repository.open(storage, config=repo_config)
 
 snap4 = new_repo.rewrite_manifests(
-    f"rewrite_manifests with new config {str(split_config.to_dict())!r}", branch="main"
+    f"rewrite_manifests with new config", branch="main"
 )
 ```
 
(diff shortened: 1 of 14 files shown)


icechunk/src/repository.rs
```diff
@@ -796,6 +798,20 @@ impl Repository {
         Ok(session)
     }
 
+    pub async fn rewrite_manifests(
```
paraseba: Other options for consideration: `optimize_manifests`, `reshard_manifests`

icechunk/src/repository.rs
```diff
@@ -796,6 +798,20 @@ impl Repository {
...
+        &self,
+        branch: &str,
+        message: &str,
+        properties: Option<SnapshotProperties>,
+    ) -> RepositoryResult<SnapshotId> {
+        let mut session = self.writable_session(branch).await?;
+
+        session
```
paraseba: Will this rewrite every array's manifest? would it be useful to let them select a list of arrays? What happens if there is nothing to do (say because the are no new chunks in the new splits) will a commit still happen?

paraseba: Taking the branch is quite strange. Maybe we should only have the session method? Actually, a writable session method?

dcherian: Agree that taking the branch is weird.

I think it would be good to keep this operation (still thinking of it as a "maintenance op") separate from the usual committing workflow. And so it would be nice to disallow all the other things you can do with a writable session.

Some options:
1. Keep it in Repository as now.
2. Add to Session but require that the ChangeSet is empty.

Thoughts?

paraseba: Interesting, we actually have an `ops` module for that, I guess you can put it in there. But bottom line, the important thing is where we add it in the python interface.

docs/docs/icechunk-python/performance.md
```diff
@@ -92,3 +102,105 @@ will result in splitting manifests so that each manifest contains (3 longitude c
...
+
+Why?
+
+Icechunk minimizes how many chunk references are rewritten at each commit (to save time and memory). The previous splitting configuration (split size of 1) results in manifests that are _compatible_ with the current configuration (split size of 5) because the bounding box of every existing manifest `[slice(0, 1), slice(1, 2), ...]` is fully contained in the the bounding boxes implied by the new configuration `[slice(0, 5), slice(5, 10)]`.
+
+To force Icechunk to rewrite all chunk refs to the current splitting configuration use [`rewrite_manifests`](./reference.md#icechunk.Repository.rewrite_manifests) --- for the current example this will consolidate to two manifests.
+```python exec="on" session="perf" source="material-block"
+snap3 = new_repo.rewrite_manifests(
+    f"rewrite_manifests with new config {str(split_config.to_dict())!r}", branch="main"
+)
```
dcherian: Should we automatically insert the splitting config in `metadata` and merge it with any user provided metadata?

icechunk-python/src/repository.rs
```diff
@@ -925,6 +927,28 @@ impl PyRepository {
...
+        branch: &str,
+        metadata: Option<PySnapshotProperties>,
+    ) -> PyResult<String> {
+        // This function calls block_on, so we need to allow other thread python to make progress
+        py.allow_threads(move || {
+            let metadata = metadata.map(|m| m.into());
+            let result =
+                pyo3_async_runtimes::tokio::get_runtime().block_on(async move {
+                    let lock = self.0.read().await;
+                    rewrite_manifests(lock.deref(), branch, message, metadata)
```
dcherian: 😬 not sure about the `deref`

docs/docs/icechunk-python/performance.md
```diff
@@ -92,3 +102,132 @@ will result in splitting manifests so that each manifest contains (3 longitude c
...
+
+```python  exec="on" session="perf" source="material-block"
+snap3 = session.commit("rewrite [3,7)")
+print(repo.lookup_snapshot(snap3).manifests)
+```
+This ends up rewriting all refs to two new manifests.
+
+### Rewriting manifests
+
+To force Icechunk to rewrite all chunk refs to the current splitting configuration use [`rewrite_manifests`](./reference.md#icechunk.Repository.rewrite_manifests) --- for the current example this will consolidate to two manifests.
```
paraseba: Let's have a paragraph on why you would do this. You probably started your dataset with default sharding, now it's huge and slow, how do you make it fast again?

docs/docs/icechunk-python/performance.md
```diff
@@ -92,3 +102,132 @@ will result in splitting manifests so that each manifest contains (3 longitude c
...
+
+To illustrate, we will use a split size of 3.
```
paraseba: I think if we are going to commit automatically for them, we should give an example workflow that includes creating a new branch, exploring a couple of possibilities, then deleting the branch and applying the winning config in main

icechunk-python/python/icechunk/repository.py
```diff
@@ -634,6 +634,31 @@ def garbage_collect(self, delete_object_older_than: datetime.datetime) -> GCSumm
 
         return self._repository.garbage_collect(delete_object_older_than)
 
+    def rewrite_manifests(
+        self, message: str, *, branch: str, metadata: dict[str, Any] | None = None
+    ) -> str:
+        """
+        Rewrite manifests for all arrays and commit to the specified branch.
```
paraseba: I'd say "Start a new writable session on the branch, rewrite manifests for all arrays and then commit"

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -149,12 +114,38 @@ def reopen_with_config(self, data):
...
             inline_chunk_threshold_bytes=0, manifest=ic.ManifestConfig(splitting=sconfig)
         )
         note(f"reopening with splitting config {sconfig=!r}")
         self.repo = Repository.open(self.storage, config=config)
+        if data.draw(st.booleans()):
+            self.repo.save_config()
+        self.store = self.repo.writable_session("main").store
+
+    @precondition(lambda self: not self.store.session.has_uncommitted_changes)
+    @rule(data=st.data())
```
paraseba, line 128: lovely

icechunk/src/session.rs
```diff
@@ -943,6 +943,37 @@ impl Session {
...
+        let nodes = self.list_nodes().await?.collect::<Vec<_>>();
+        for node in nodes.into_iter().flatten() {
+            if let NodeSnapshot {
+                id,
+                path,
+                node_data: NodeData::Array { shape, dimension_names, .. },
+                ..
+            } = node
+            {
+                self.get_splits(&id, &path, &shape, &dimension_names);
```
paraseba, line 968: can you add a comment on what does this line achieve?

icechunk-python/src/repository.rs
paraseba: maybe `&lock` just works

docs/docs/icechunk-python/performance.md
paraseba: oh that's a great idea!

dcherian: Done

paraseba (approved):
Do we need a test to prove ourselves that manifest can also be fused back into fewer ones by calling rewrite_manifests? Or are covered?

dcherian:
> we need a test to prove ourselves that manifest can also be fused back into fewer ones by calling rewrite_manifests? Or are covered?

That's covered. For completeness, I just added one for the case where we split existing multiple manifests in to smaller manifests too.

PR #1008, opened 2025-06-25
Add initial commit id as constant

### LINKED ISSUE #708: Make first snapshot a known id

mpiannucci:
Closes #708 

* Should this be exposed to python somehwere?

### DIFF
--- a/icechunk/src/format/snapshot.rs
+++ b/icechunk/src/format/snapshot.rs
@@ -341,7 +341,8 @@ static ROOT_OPTIONS: VerifierOptions = VerifierOptions {
 impl Snapshot {
     pub const INITIAL_COMMIT_MESSAGE: &'static str = "Repository initialized";
     pub const INITIAL_SNAPSHOT_ID: SnapshotId = SnapshotId::new([
-        0x69, 0x63, 0x65, 0x63, 0x68, 0x75, 0x6E, 0x6B, 0x72, 0x65, 0x70, 0x6F, // "icechunkrepo"
+        0x69, 0x63, 0x65, 0x63, 0x68, 0x75, 0x6E, 0x6B, 0x72, 0x65, 0x70,
+        0x6F, // "icechunkrepo"
     ]);
 
     pub fn from_buffer(buffer: Vec<u8>) -> IcechunkResult<Snapshot> {
--- a/icechunk/src/format/snapshot.rs
+++ b/icechunk/src/format/snapshot.rs
@@ -340,7 +340,9 @@ static ROOT_OPTIONS: VerifierOptions = VerifierOptions {
 
 impl Snapshot {
     pub const INITIAL_COMMIT_MESSAGE: &'static str = "Repository initialized";
-    pub const INITIAL_SNAPSHOT_ID: SnapshotId = SnapshotId::new([0; 12]);
+    pub const INITIAL_SNAPSHOT_ID: SnapshotId = SnapshotId::new([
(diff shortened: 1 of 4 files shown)


icechunk/src/format/snapshot.rs
```diff
@@ -340,6 +340,7 @@ static ROOT_OPTIONS: VerifierOptions = VerifierOptions {
 
 impl Snapshot {
     pub const INITIAL_COMMIT_MESSAGE: &'static str = "Repository initialized";
+    pub const INITIAL_SNAPSHOT_ID: SnapshotId = SnapshotId::new([0; 12]);
```
paraseba: Let's make it something people can use to recognize this is an icechunk repo (we have some other ticket about that). Maybe we can put some bytes that when converted to the serialized id that goes in the object name, spells something similar to icechunk or something

mpiannucci: spells "icechunkrepo" now. Do we need to check for collisions with the random generator?

icechunk/src/format/snapshot.rs
```diff
@@ -340,6 +340,10 @@ static ROOT_OPTIONS: VerifierOptions = VerifierOptions {
 
 impl Snapshot {
     pub const INITIAL_COMMIT_MESSAGE: &'static str = "Repository initialized";
+    pub const INITIAL_SNAPSHOT_ID: SnapshotId = SnapshotId::new([
+        0x69, 0x63, 0x65, 0x63, 0x68, 0x75, 0x6E, 0x6B, 0x72, 0x65, 0x70,
+        0x6F, // "icechunkrepo"
```
paraseba: amazing

paraseba: can we add a python test that verifies a new repo writes an object with name "icechunkrepo"? Just to make sure we don't break anything in between (serializers, formatters, etc)

icechunk/src/format/snapshot.rs
paraseba: no, i don't think we need to check collisions, this is great

icechunk/src/format/snapshot.rs
paraseba: now that I think about it, I think our id serialization doesn't use "i" to avoid confusing it with 1. Now I want to see this test even more. It doesn't use U either... and it's 20 characters, so maybe we have to spell something like

```
1CECHNKREP0R00TSNAP
```

but I welcome more imaginative or fun options

mpiannucci: > and it's 20 characters

Okay this is what i am lost, this is `[u8; 12]` isnt it?

paraseba: Cool cool, let me explain. We want to generate a known file path for the first snapshot, something like:
```
snapshots/1CECHNKREP0R00TSNAP
```

The string that goes to the object name, is that `[u8;12]` id, but encoded with something that is not ASCII or Hex. What we are using is called crockford base 32 https://www.crockford.com/base32.html

We need to find a decoder (there are many onlines) and know what bytes to put in there, so when they get encoded to an object name, they produce the string that we want. 12 bytes will become 20 characters

mpiannucci: got ittttt thank you! ive been away from here for a while

mpiannucci: `1CECHNKREP0F1RSTCMT0`

PR #1010, opened 2025-06-25
Fix typo in Virtual Datasets Docs

jbusecke:
I just went through the example here and I think that no variable called `store` is defined in the steps preceeding this (unless I missed something).


paraseba:
Thank you @jbusecke

PR #1018, opened 2025-06-27
Stateful test for parallel sessions

dcherian:
Looks like it's catching a version of that bug.

```
E           Falsifying example:
E           state = SerialParallelStateMachine()
E           forking with pickle
E           state.fork_pickle()
E           merging forks to base session, merging 1→2→parallel
E           state.merge(two_to_one=False)
E           forking with pickle
E           state.fork_pickle()
E           merging forks to base session, merging 1→2→parallel
E           state.merge(two_to_one=False)
E           Draw 1: int8(1)
E           Adding array:  path='00'  shape=(1, 1, 1)  chunks=(1, 1, 1)
E           state.add_array(array_and_chunks=(array([[[0]]], dtype=int8), (1, 1, 1)), data=data(...), name='00')
E           Draw 2: int8(0)
E           Adding array:  path='0'  shape=(1,)  chunks=(1,)
E           state.add_array(array_and_chunks=(array([0], dtype=int8), (1,)), data=data(...), name='0')
E           forking with pickle
E           state.fork_pickle()
E           Draw 3: '00'
E           Draw 4: '0/0/0'
E           deleting chunk path='00/c/0/0/0'
E           Draw 5: ('parallel', <icechunk.session.Session object at 0x11476e570>)
E           executing on parallel
E           state.delete_chunk(data=data(...))
E           merging forks to base session, merging 1→2→parallel
E           state.merge(two_to_one=False)
E           verifying all arrays
E           state.verify_all_arrays()
E           state.teardown()
```

### DIFF
--- a/icechunk-python/python/icechunk/testing/strategies.py
+++ b/icechunk-python/python/icechunk/testing/strategies.py
@@ -41,3 +41,42 @@ def splitting_configs(
                 key: draw(st.integers(min_value=1, max_value=size + 10))
             }
     return ic.ManifestSplittingConfig.from_dict(config_dict)  # type: ignore[attr-defined, no-any-return]
+
+
+@st.composite
+def chunk_coordinates(draw: st.DrawFn, numblocks: tuple[int, ...]) -> tuple[int, ...]:
+    return draw(
+        st.tuples(*tuple(st.integers(min_value=0, max_value=b - 1) for b in numblocks))
+    )
+
+
+@st.composite
+def chunk_slicers(
+    draw: st.DrawFn, numblocks: tuple[int, ...], chunk_shape: tuple[int, ...]
+) -> tuple[slice, ...]:
+    return tuple(
+        (
+            slice(coord * size, coord + 1 * size)
+            for coord, size in zip(
+                draw(chunk_coordinates(numblocks)), chunk_shape, strict=False
+            )
+        )
+    )
+
+
+@st.composite
+def chunk_paths(draw: st.DrawFn, numblocks: tuple[int, ...]) -> str:
+    blockidx = draw(chunk_coordinates(numblocks))
+    return "/".join(map(str, blockidx))
+
+
+@st.composite
+def chunk_directories(draw: st.DrawFn, numblocks: tuple[int, ...]) -> str:
+    ndim = len(numblocks)
+    blockidx = draw(chunk_coordinates(numblocks))
+    subset_slicer = (
+        slice(draw(st.integers(min_value=0, max_value=ndim)))
+        if draw(st.booleans())
+        else slice(None)
+    )
+    return "/".join(map(str, blockidx[subset_slicer]))
--- a/icechunk-python/tests/test_stateful_parallel_sessions.py
+++ b/icechunk-python/tests/test_stateful_parallel_sessions.py
@@ -1,0 +1,232 @@
+import datetime
+import pickle
+from collections.abc import Callable
+from typing import Any
+
+import hypothesis.extra.numpy as npst
+import numpy as np
+import numpy.testing as npt
+from hypothesis import assume, note, settings
+from hypothesis import strategies as st
+from hypothesis.stateful import RuleBasedStateMachine, precondition, rule
+
+import icechunk as ic
+import icechunk.testing.strategies as icst
+import zarr
+import zarr.testing.strategies as zrst
+from icechunk.distributed import merge_sessions
+from zarr.abc.store import Store
+from zarr.testing.stateful import SyncStoreWrapper
+
+zarr.config.set({"array.write_empty_chunks": True})
+
+
+def simple_dtypes() -> st.SearchStrategy[np.dtype[Any]]:
+    return npst.integer_dtypes(endianness="=") | npst.floating_dtypes(endianness="=")
+
+
+class SerialParallelStateMachine(RuleBasedStateMachine):
+    """
+    This stateful test asserts that two stores :
+    1. one on which all actions are executed in serial
+    2. one on which those same actions may be executed on the parent session,
+       or on forks. Importantly, forks may be created with a 'dirty' state.
+
+    To model this we use the same repo with two branches.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.storage = ic.local_filesystem_storage(
+            f"tmp/icechunk_parallel_stateful/{str(datetime.datetime.now()).split(' ')[-1]}"
+        )
+        self.repo = ic.Repository.create(self.storage)
+        self.repo.create_branch("parallel", self.repo.lookup_branch("main"))
+
+        # TODO: should this just be Zarr memory store instead?
+        #       are version control ops on the serial store useful?
+        self.serial = self.repo.writable_session("main")
+        self.parallel = self.repo.writable_session("parallel")
+
+        self.fork1: ic.Session | None = None
+        self.fork2: ic.Session | None = None
+
+        self.has_changes = False
+        self.all_arrays: set[str] = set()
+
+    def has_forks(self) -> bool:
+        return self.fork1 is not None and self.fork2 is not None
+
+    @precondition(lambda self: not self.has_forks())
+    @rule(
+        data=st.data(),
+        name=zrst.node_names,
+        array_and_chunks=zrst.np_array_and_chunks(
+            arrays=npst.arrays(simple_dtypes(), npst.array_shapes())
+        ),
+    )
+    def add_array(
+        self,
+        data: st.DataObject,
+        name: str,
+        array_and_chunks: tuple[np.ndarray[Any, Any], tuple[int, ...]],
+    ) -> None:
+        array, _ = array_and_chunks
+        # TODO: support size-0 arrays GH392
+        assume(array.size > 0)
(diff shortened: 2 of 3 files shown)


icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,232 @@
...
+        note("committing both sessions")
+        self.serial.commit("foo")
+        self.parallel.commit("foo")
+
+        self.serial = self.repo.writable_session("main")
+        self.parallel = self.repo.writable_session("parallel")
+
+    # @precondition(lambda self: self.has_forks())
+    # @rule(commit_fork1_first=st.booleans())
+    # def commit_on_forks(self, commit_fork1_first: bool):
```
dcherian: This seems hard. I think we might need a different test that exercises `rebase`

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,232 @@
...
+            merge_sessions(self.parallel, self.fork1)
+        else:
+            note("merging forks to base session, merging 1→2→parallel")
+            merge_sessions(self.fork2, self.fork1)
+            merge_sessions(self.parallel, self.fork2)
+
+        self.fork1 = None
+        self.fork2 = None
+
+    @precondition(lambda self: not self.has_forks() and self.has_changes)
```
dcherian: this is basically saying we expect the user to fork, merge, and then commit the "base" `parallel` session

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,232 @@
...
+                store=store,
+                fill_value=fill_value,
+                zarr_format=3,
+                dimension_names=None,
+            )
+        self.all_arrays.add(name)
+
+    @precondition(lambda self: bool(self.all_arrays))
+    @rule(data=st.data())
+    def write_chunk(self, data: st.DataObject) -> None:
```
dcherian, line 96: This is problematic. We could draw the same path on two draws which would model the case where parallel writers write to the same chunk and behaviour is undefined. I guess I can filter on chunk paths not in `self.serial.status().updated_chunks[array]`

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,232 @@
...
+        self.fork2 = None
+
+    @precondition(lambda self: not self.has_forks() and self.has_changes)
+    def commit(self) -> None:
+        note("committing both sessions")
+        self.serial.commit("foo")
+        self.parallel.commit("foo")
+
+        self.serial = self.repo.writable_session("main")
+        self.parallel = self.repo.writable_session("parallel")
```
dcherian, line 212: I should `verify_all_arrays` here because we expect the state on `main` and `parallel` to be identical after committing.

dcherian: `verify_all_arrays` is now an invariant so it gets run after `commit` always

icechunk-python/tests/test_stateful_parallel_sessions.py
dcherian: Now we track state to make sure this doesn't happen. More important given #1815

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,239 @@
...
+
+    @invariant()
+    def verify_all_arrays(self) -> None:
+        """
+        This cannot be an invariant because we may have state on the forks.
+        """
```
ianhi:
```suggestion
```
can be if you have the escape hatch you added

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,239 @@
...
+        note(f"executing on {name}")
+        assert session is not None
+
+        # Track ownership and reject conflicts via assume()
+        if self.has_forks():
+            for coord in chunks:
+                owner = self.chunk_owners.get(coord)
+                if owner is not None and owner != name:
+                    assume(False)
+                self.chunk_owners[coord] = name
```
ianhi: can we push the check of chunk ownership up into the sampling of the chunks to modify in order to avoid this assume?

even just as a filter on `chunk_path = data.draw(icst.chunk_paths(numblocks=arr.cdata_shape))`

ianhi: I think you'd also have to push up the sampling of who is doing the executing, but maybe that could even become a rule on the caller?

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,239 @@
...
+        fork = self.parallel.fork()
+        self.fork1 = pickle.loads(pickle.dumps(fork))
+        self.fork2 = pickle.loads(pickle.dumps(fork))
+
+    @precondition(lambda self: not self.has_forks())
+    @rule()
+    def fork_threads(self) -> None:
+        note("forking with reference (threads)")
+        self.fork1 = self.parallel
+        self.fork2 = self.parallel
```
ianhi, line 182: I don't' see how this does anything? It doesn't actually fork, just creates alias.

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,239 @@
...
+from hypothesis.stateful import RuleBasedStateMachine, invariant, precondition, rule
+
+import icechunk as ic
+import icechunk.testing.strategies as icst
+import zarr
+import zarr.testing.strategies as zrst
+from zarr.abc.store import Store
+from zarr.testing.stateful import SyncStoreWrapper
+
+zarr.config.set({"array.write_empty_chunks": True})
```
ianhi: will this infect other tests?

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,239 @@
...
+            p = zarr.open_array(path=path, store=self.parallel.store)
+            npt.assert_array_equal(s, p)
+
+    def teardown(self) -> None:
+        ic.set_logs_filter(None)
+
+
+SerialParallelStateMachine.TestCase.settings = settings(
+    deadline=None, report_multiple_bugs=False
+)
```
ianhi:
```suggestion
```

handled in conf test now.

icechunk-python/tests/test_stateful_parallel_sessions.py
ianhi: should this be new calls of `self.repo.writable_session("parallel")`?

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        log_filter = "warn,icechunk::storage::object_store=error"
+        ic.set_logs_filter(log_filter)
+
+        self.storage = ic.local_filesystem_storage(
+            f"tmp/icechunk_parallel_stateful/{str(datetime.datetime.now()).split(' ')[-1]}"
+        )
```
ianhi: This is relative to whereever the test gets called from I think? can we just use `tmpdir`

icechunk-python/tests/test_stateful_parallel_sessions.py
dcherian: it models the multithreading case where the same session is broadcast out.

icechunk-python/tests/test_stateful_parallel_sessions.py
dcherian: yeah but forks can be empty so then every rule needs to gain new logic. I'm inclined to leave as-is and improve later if/when we need to

icechunk-python/python/icechunk/testing/strategies.py
```diff
@@ -82,3 +82,30 @@ def repository_configs(
...
+    draw: st.DrawFn, numblocks: tuple[int, ...], chunk_shape: tuple[int, ...]
+) -> tuple[slice, ...]:
+    return tuple(
+        (
+            slice(coord * size, (coord + 1) * size)
+            for coord, size in zip(
+                draw(chunk_coordinates(numblocks)), chunk_shape, strict=False
+            )
+        )
+    )
```
ianhi, line 117: Some comments or docstrings on what things like `numblocks` arae here will for sure help in the future.

also as in tree maybe we should ahve strict here?

icechunk-python/tests/test_stateful_parallel_sessions.py
```diff
@@ -0,0 +1,229 @@
...
+    ) -> None:
+        """
+        Chooses one of self.parallel, self.fork1, or self.fork2
+        as the session on which to make changes using `func`.
+
+        When forks exist, enforces that each (array, chunk_coord) is only
+        modified by a single session to avoid overlapping writes that would
+        cause merge conflicts.
+        """
+        # Track ownership and reject conflicts via assume()
```
ianhi:
```suggestion
```
no longer true

PR #1021, opened 2025-06-30
Implement new `with_read_only` Zarr `Store` method

paraseba:
This new method is not really implementable in Icechunk. If the user created a readonly session there is no real good way to turn it into a writable session, and viceversa.

The implementation we are writing "lies" in this sense. It's pretending to return a new store with the requested readonly-ness, but not really.


dcherian (approved):
Can we add an FAQ entry saying that if you want true read-only create a `readonly_session`

PR #1024, opened 2025-07-01
Address zarr 3.0.9 test failures

mpiannucci:


mpiannucci:
This is the non xarray tests, should fix those i think

paraseba (approved):

PR #1027, opened 2025-07-01
Configure mkdocs site_url so 404s look better on deployed site

maxrjones:
Similar to Icechunk, VirtualiZarr was having issues with 404 pages being rendered poorly on readthedocs. For an Icechunk example, see https://icechunk.io/en/latest/icechunk-python/unknown-page. @chuckwondo pointed me to this elegant solution so I thought I'd propose it for your library as well.


paraseba:
Thank you @maxrjones

PR #1029, opened 2025-07-02
Raise a warning that LocalFileSystem is not safe for concurrent writes

monodeldiablo:


mpiannucci:
Do we wanna warn from rust so all users get it?

icechunk/src/storage/object_store.rs
```diff
@@ -77,6 +77,7 @@ impl ObjectStorage {
     pub async fn new_local_filesystem(
         prefix: &StdPath,
     ) -> Result<ObjectStorage, StorageError> {
+        tracing::warn!("The LocalFileSystem backend is not safe for concurrent writes due to a race condition at commit time. Use with caution, and prefer using object stores for production data.");
```
paraseba: This feel too serious to me. How about this:

The LocalFileSystem storage is not safe for concurrent commits. If more than one thread/process can attempt to commit at the same time, prefer using object stores

paraseba:
I think this is ready to merge @monodeldiablo

PR #1031, opened 2025-07-02
[docs] Manifest preload docs

mpiannucci:


docs/docs/icechunk-python/performance.md
```diff
@@ -318,3 +314,85 @@ Notice that the persisted config is restored when opening a Repository
...
+    preload_if=ic.ManifestPreloadCondition.name_matches("x"),
+    max_total_refs=10000,
+)
+```
+
+This will preload all manifests that match the array name "x" while the number of total chunk references that have been preloaded is less than 10000.
+
+!!! important
+
+    Once you find a preload configuration you like, remember to persist it on-disk using `repo.save_config`. It will still be able to be overridden at runtime for different applications.
```
dcherian:
```suggestion
    Once you find a preload configuration you like, remember to persist it on-disk using `repo.save_config`. The saved config can be overridden at runtime for different applications.
```

dcherian (approved):
❤️

docs/docs/icechunk-python/performance.md
```diff
@@ -318,3 +314,85 @@ Notice that the persisted config is restored when opening a Repository
 ```python exec="on" session="perf" source="material-block"
 print(ic.Repository.open(storage).config.manifest)
 ```
+
+## Preloading manifests
+
+While [manifest splitting](./performance.md#splitting-manifests) is a great way to control the size of manifests, it can be useful to configure the manner in which manifests are loaded. In Icechunk manifests are loaded lazily by default, meaning that when you read a chunk, Icechunk will only load the manifest for that chunk when it is needed to fetch the chunk data. While this is good for memory performance, it can be a bottleneck for read performance when you are reading a lot of chunks.
```
paraseba:
```suggestion
While [manifest splitting](./performance.md#splitting-manifests) is a great way to control the size of manifests, it can be useful to configure the manner in which manifests are loaded. In Icechunk manifests are loaded lazily by default, meaning that when you read a chunk, Icechunk will only load the manifest for that chunk when it is needed to fetch the chunk data. While this is good for memory performance, it can increase the latency to first the first elements of an array. Once a manifest has been loaded, it will usually remain in memory for future chunks in the same manifest.
```

docs/docs/icechunk-python/performance.md
```diff
@@ -318,3 +314,85 @@ Notice that the persisted config is restored when opening a Repository
...
+
+To address this, Icechunk provides a way to preload manifests at the time of opening a Session, loading all manifests that match specific conditions into the cache. This means when it is time to read a chunk, the manifest will already be in the cache and the chunk can be found without having to first load the manifest from storage.
+
+### Configuring preloading
+
+To configure manifest preloading, you can use the `ManifestPreloadConfig` class to specify the `ic.RepositoryConfig.manifest.preload` field.
+
+```python exec="on" session="perf" source="material-block"
+preload_config = ic.ManifestPreloadConfig(
+    preload_if=ic.ManifestPreloadCondition.name_matches("x"),  # preload all manifests that match the regex "x"
```
paraseba: I think name_matches takes a regex, so this may match much more than they want? I don't remember if we do full match or we match anywhere in the name. So, this may need something like `("^x$")`. But I'd have to check how we match it, maybe you already did. It would be nice to add it to this docs.

paraseba (approved):

docs/docs/icechunk-python/performance.md
mpiannucci: You are totally right, I'll refine this

mpiannucci: done

docs/docs/icechunk-python/performance.md
```diff
@@ -318,3 +314,89 @@ Notice that the persisted config is restored when opening a Repository
...
+This will preload all manifests that match the array name "x" while the number of total chunk references that have been preloaded is less than 10000.
+
+!!! important
+
+    Once you find a preload configuration you like, remember to persist it on-disk using `repo.save_config`. The saved config can be overridden at runtime for different applications.
+
+
+#### Default preload configuration
+
+Icechunk has a default `preload_if` configuration that will preload all manifests that match [cf-xarrays coordinate axis regex](https://github.com/xarray-contrib/cf-xarray/blob/1591ff5ea7664a6bdef24055ef75e242cd5bfc8b/cf_xarray/criteria.py#L149-L160).
```
abkfenris, line 420: Is this default condition exposed so that you can build upon it?


```python exec="on" session="perf" source="material-block"
preload_config = ic.ManifestPreloadConfig(
    preload_if=ic.ManifestPreloadCondition.or_conditions(
        ic.ManifestPreloadCondition.DEFAULT_CF_CONDITION,
        ic.ManifestPreloadCondition.path_matches("y"),
    )
)
```

paraseba: I don't think it's exposed today, which sucks and we should change.

mpiannucci:
I should update this with the syntax from @abkfenris PRs

abkfenris:
> I should update this with the syntax from @abkfenris PRs

Maybe as a comment rather than replacing all the usage of the classmethod. 

On [other projects that have similar rule structures](https://docs.dagster.io/guides/automate/declarative-automation/customizing-automation-conditions/), the operators are really nice short hand between just two conditions, but the classmethods can be more explicit once things get more complicated.

Issue #996, opened 2025-06-16
Preload/Splitting conditions: It is confusing that `DimensionName("t")` matches `"time"`


dcherian:
### What happened?

since we are doing regex matches. This was quite unexpected to me and the API doesn't make it clear. 

Perhaps we should have `DimensionNameMatches` (mirroring `PathMatches`) for a regex match; and `DimensionName`  for a literal match (so that `DimensionName("t")` is identical to `DimensionNameMatches("^t$")`).

PR #1038, opened 2025-07-03
Manifest preload magic methods

abkfenris:
Adds magic methods to the `ManifestPreloadCondition` Python class, so that rather than require users to remember the class method for comparison, then can use the `&` and `|` operators.

```py
icechunk.ManifestPreloadCondition.and_conditions(
    [
        icechunk.ManifestPreloadCondition.true(),
        icechunk.ManifestPreloadCondition.name_matches("foo"),
    ]
) ==
icechunk.ManifestPreloadCondition.true() & icechunk.ManifestPreloadCondition.name_matches("foo")
```

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -421,6 +421,12 @@ class ManifestPreloadCondition:
     def false() -> ManifestPreloadCondition:
         """Create a preload condition that never matches any manifests"""
         ...
+    def __and__(other: ManifestPreloadCondition) -> ManifestPreloadCondition:
+        """Create a preload condition that matches if both this condition and `other` match."""
+        ...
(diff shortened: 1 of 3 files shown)


paraseba:
This is lovely @abkfenris , thank you!

abkfenris:
@paraseba I was heckling @mpiannucci for other reasons and things escalated. Now he's conned me into trying to do the same for manifest splitting.

dcherian:
You might as well. 

And while you're there you could also hit https://github.com/earth-mover/icechunk/issues/996 ;)

abkfenris:
I told him I'm not doing all of his homework.

abkfenris:
I tried to stack a branch/PR with manifest splitting. Instead my Git client force pushed the existing commits which has hidden the test failures, and gotten into an angry state itself, so that branch is https://github.com/abkfenris/icechunk/tree/manifest-splitting-magic

paraseba (approved):

PR #1039, opened 2025-07-03
Identify virtual chunk containers by url prefix

paraseba:
This is an API breaking change, new argument to create and open repositories.


icechunk-python/python/icechunk/credentials.py
```diff
@@ -340,15 +340,13 @@ def azure_credentials(
     raise ValueError("Conflicting arguments to azure_credentials function")
 
 
-def containers_credentials(
-    m: Mapping[str, AnyS3Credential] = {}, **kwargs: AnyS3Credential
-) -> dict[str, Credentials.S3]:
+def containers_credentials(m: Mapping[str, AnyS3Credential]) -> dict[str, Credentials.S3]:
```
mpiannucci: Should this be AnyCredential instead of AnyS3Credential

mpiannucci (approved):
I reviewed this as in depth as I could. It's a lot of changes, overall the idea makes sense. Just one question I wasn't sure about

icechunk-python/python/icechunk/credentials.py
paraseba: yes!

paraseba: I'll need to think a bit about what's the right way to fix this. Interestingly, it was already a bug before this PR. I guess a survivor from a time where we only supported S3 virtual chunks

PR #1041, opened 2025-07-04
Add transaction context manager to repo

rabernat:
This is a bit of syntactic sugar that I have always wanted for Icechunk. It will make it much easier to do quick updates to a store.

I plan to use this in a subsequent PR with some doc rewrites.


icechunk-python/python/icechunk/repository.py
```diff
@@ -564,6 +566,62 @@ def writable_session(self, branch: str) -> Session:
         """
         return Session(self._repository.writable_session(branch))
 
+    @contextmanager
+    def transaction(self, branch: str, message: str) -> Iterator["IcechunkStore"]:
```
paraseba: Can't you just say `-> Iterator[IcechunkStore]` here? Do you need the quotes?

icechunk-python/tests/test_store.py
```diff
@@ -95,3 +95,16 @@ async def test_with_readonly() -> None:
...
     reader = store.with_read_only(read_only=True)
     assert reader.read_only
+
+
+async def test_transaction() -> None:
+    repo = parse_repo("memory", "test")
+    with repo.transaction("main", "initialize group") as store:
+        assert not store.read_only
+        root = zarr.group(store=store)
+        root.attrs["foo"] = "bar"
```
paraseba, line 107: let's add  an assertion after the block that verifies the commit happened, and a different test that verifies no commit on exception

paraseba (approved):

icechunk-python/python/icechunk/repository.py
dcherian:
```suggestion
    def transaction(self, branch: str, *, message: str) -> Iterator["IcechunkStore"]:
```

mild preference for making this explicit

icechunk-python/tests/test_store.py
rabernat: done

icechunk-python/python/icechunk/repository.py
```diff
@@ -579,6 +581,31 @@ def writable_session(self, branch: str) -> Session:
         """
         return Session(self._repository.writable_session(branch))
 
+    @contextmanager
+    def transaction(self, branch: str, *, message: str) -> Iterator[IcechunkStore]:
```
paraseba: can you add the `rebase_with` and `rebase_tries` arguments to pass to `commit` ?

rabernat: Done. (Also metadata) But not tested. I trust mypy that it's okay.

paraseba (approved):

PR #1047, opened 2025-07-08
Don't do docstring inspection to solve local docs build error

ianhi:
Consequence is that we won't pick up dynamically generated docstrings but I don't think we have any.

Fixes the build error for me locally.


ianhi:
From the docs: https://mkdocstrings.github.io/python/usage/configuration/general/?h=allow_in#allow_inspection

> If you have compiled modules but also provide stubs for them, you might want to disable the inspection of these modules, because inspection picks up many more members, and sometimes the collected data is inaccurate (depending on the tool that was used to compile the module) or too low-level/technical for API documentation.

so it's definitely not unprecedented.

That said, we should be sure that this is not just papering over some issue that should be fixed for other, non-doc, reasons

rabernat:
Docs LGTM: https://icechunk--1047.org.readthedocs.build/en/1047/icechunk-python/reference/#icechunk.ConflictError

PR #1051, opened 2025-07-09
major WIP: reorganization

rabernat:


paraseba (commented):
It's all very very good! Some comments as I read the rendered documentation, sorry for no making them in place, but I can help to apply them.



The call to action on the landing page is "Get Started". Is it clear that takes you to the documentation? I'd normally think it is taking me to somebody selling me something. Maybe "Learn Icechunk" or something like that?

I see many version dropdowns at the top, as you navigate more and more appear

Is there a way to tabulate the left menu? It would make it easier to understand the sections

> Previous snapshots of a repo remain accessible after new ones have been written.

and while they are being written

> (i.e. serializable) order within branch.

within a branch

> snapshot--a pointer

snapshot --a pointer

> immutable references to snapshot

to a snapshot

> You can then can update

quickstart: let's highlight that they need to start a new session after commit

quickstart should tell them about `transaction`, this avoids the issue with reusing the session after commit

>First we need to create a writable from the

a writable session from the

> session2.rebase(icechunk.ConflictDetector())

let's pass the conflict detector as a keyword argument, so they know what this method takes

> Other types of conflicts are not able to be resolved

let's add that they are not resolved but they are still detected and block the commit. Icechunk always protects your data from unintentional overwrites

I think data expiration section should be after parallel/distributed writing

In expiration section, after "There are 10 snapshots" there is a code block issue

The same FAQ page appears twice in the menu

> The GCS storage config does not yet support anonymous access.

we do now

> and thus configuration updating is potentially unsaf

I think we fixed that

Limitation for filesystem storage: it's unsafe in the presence of concurrent commits

"Migrate to 1.0" feels a bit out of place, maybe it should be in the reference section?

rabernat:
What do you mean by "tabulate the left menu"?

paraseba:
> What do you mean by "tabulate the left menu"?

At least in my browser, the font weight is not enough to distinguish the section titles from the section subtitles in the menu. It would be easier if the contents of each section were a couple of spaces to the right.

:facepalm: I think "nesting" is the word I want, not tabulating 

image

rabernat:
All comments addressed in latest commit. @paraseba can you approve?

PR #1062, opened 2025-07-10
upstream test using pip install --pre

paraseba:


dcherian (approved):
Right, we've run in to this before!

PR #1071, opened 2025-07-11
Fix URL in error message

maxrjones:
Assuming you don't want redirects from pre-v1.0, this just fixes the URL in the error message.


paraseba (approved):
Thank you @maxrjones !!!

maxrjones:
No problem!

BTW it seems silly to open an issue for something so minor, but FYI I've been routinely getting into a pretty weird state of the banner on Firefox after using the back arrow -
Screenshot 2025-07-11 at 5 37 42 PM

paraseba:
@maxrjones we have this problem, yes. I think it works if you go to the `latest` version instead, but last released version (stable) has this documentation bug

PR #1074, opened 2025-07-13
Bring back async methods for python

mpiannucci:
#620 

I want to use icechunk in async context, so looking to bring async methods back to repo and sessions in python. 

I am matching the style of [obstore](https://github.com/developmentseed/obstore/blob/main/obstore/src/get.rs) for this, and created a first pass to get feedback on it. 

The gist is that the simplest possible implementation of this is to duplicate the methods exposed to python with async variants, named with an `async` postfix to make editor searching easier etc.

This also has the benefit of being able to use sync and async versions of functions on the same object.

## Repository

• create_async() - async repository creation
• open_async() - async repository opening
• open_or_create_async() - async repository open or create
• exists_async() - async check if repository exists
• fetch_config_async() - async fetch repository config
• save_config_async() - async save repository config
• reopen_async() - async repository reopen
• expire_snapshots_async() - async snapshot expiration
• rewrite_manifests_async() - async manifest rewriting
• garbage_collect_async() - async garbage collection
• total_chunks_storage_async() - async total chunks storage calculation
• all_virtual_chunk_locations_async() - async get all virtual chunk locations
• create_branch_async() - async branch creation
• list_branches_async() - async list branches
• lookup_branch_async() - async branch lookup
• reset_branch_async() - async branch reset
• delete_branch_async() - async branch deletion
• create_tag_async() - async tag creation
• delete_tag_async() - async tag deletion
• list_tags_async() - async list tags
• lookup_tag_async() - async tag lookup
• lookup_snapshot_async() - async snapshot lookup
• diff_async() - async diff between snapshots

## Session

• readonly_session_async() - async readonly session creation
• writable_session_async() - async writable session creation
• merge_async() - async session merge
• commit_async() - async commit
• rebase_async() - async rebase

## Store

• set_virtual_ref_async() - async set single virtual reference
• set_virtual_refs_async() - async set multiple virtual references

### DIFF
--- a/icechunk-python/tests/test_session.py
+++ b/icechunk-python/tests/test_session.py
@@ -9,7 +9,8 @@ from icechunk import ForkSession, IcechunkError, Repository, local_filesystem_st
 from icechunk.distributed import merge_sessions
 
 
-def test_session_fork() -> None:
+@pytest.mark.parametrize("use_async", [True, False])
+async def test_session_fork(use_async: bool) -> None:
     with tempfile.TemporaryDirectory() as tmpdir:
         repo = Repository.create(local_filesystem_storage(tmpdir))
         session = repo.writable_session("main")
@@ -19,7 +20,10 @@ def test_session_fork() -> None:
         with pytest.raises(ValueError):
             session.fork()
 
-        session.commit("init")
+        if use_async:
+            await session.commit_async("init")
+        else:
+            session.commit("init")
 
         # forking a read-only session
         with pytest.raises(ValueError):
@@ -37,8 +41,12 @@ def test_session_fork() -> None:
         with pytest.warns(UserWarning):
             with pytest.raises(IcechunkError, match="cannot commit"):
                 session.commit("foo")
-        session.merge(fork)
-        session.commit("foo")
+        if use_async:
+            await session.merge_async(fork)
+            await session.commit_async("foo")
+        else:
+            session.merge(fork)
+            session.commit("foo")
 
         session = repo.writable_session("main")
         fork1 = pickle.loads(pickle.dumps(session.fork()))
@@ -56,8 +64,12 @@ def test_session_fork() -> None:
         with pytest.raises(TypeError, match="Received 'Session'"):
             merge_sessions(cast(ForkSession, session), fork1, fork2)
         # this is right
-        session.merge(fork1, fork2)
-        session.commit("all done")
+        if use_async:
+            await session.merge_async(fork1, fork2)
+            await session.commit_async("all done")
+        else:
+            session.merge(fork1, fork2)
+            session.commit("all done")
 
         groups = set(
             name for name, _ in zarr.open_group(session.store, mode="r").groups()
@@ -77,75 +89,6 @@ def test_session_fork() -> None:
         fork2 = pickle.loads(pickle.dumps(fork2))
         session.merge(fork1, fork2)
 
-        groups = set(
-            name for name, _ in zarr.open_group(session.store, mode="r").groups()
-        )
-        assert groups == {"foo", "foo1", "foo2", "foo3", "foo4"}
-
-
-async def test_session_fork_async() -> None:
-    """Test async session forking and merging functionality."""
-    with tempfile.TemporaryDirectory() as tmpdir:
-        repo = await Repository.create_async(local_filesystem_storage(tmpdir))
-        session = await repo.writable_session_async("main")
-        zarr.group(session.store)
-        assert session.has_uncommitted_changes
-
-        with pytest.raises(ValueError):
-            session.fork()
-
-        await session.commit_async("init")
-
-        # forking a read-only session
-        with pytest.raises(ValueError):
-            session.fork()
-        # readonly sessions can be pickled directly
-        with pytest.raises(ValueError, match="can be pickled"):
-            pickle.loads(pickle.dumps(session.fork()))
-        pickle.loads(pickle.dumps(session))
-
-        # Test Session.merge_async
-        session = await repo.writable_session_async("main")
-        fork = pickle.loads(pickle.dumps(session.fork()))
-        zarr.create_group(fork.store, path="/foo")
-        assert not session.has_uncommitted_changes
-        assert fork.has_uncommitted_changes
-        with pytest.warns(UserWarning):
-            with pytest.raises(IcechunkError, match="cannot commit"):
-                await session.commit_async("foo")
-        await session.merge_async(fork)
-        await session.commit_async("foo")
-
-        # Test ForkSession.merge_async with multiple forks
-        session = await repo.writable_session_async("main")
-        fork1 = pickle.loads(pickle.dumps(session.fork()))
-        fork2 = pickle.loads(pickle.dumps(session.fork()))
-        zarr.create_group(fork1.store, path="/foo1")
-        zarr.create_group(fork2.store, path="/foo2")
-
-        with pytest.raises(TypeError, match="Cannot commit a fork"):
-            await fork1.commit_async("foo")
-
-        # Test merging multiple forks using distributed merge
-        with pytest.raises(
-            TypeError, match="can only be merged with another ForkSession"
-        ):
-            await fork1.merge_async(cast(ForkSession, session))
-        await fork1.merge_async(fork2)
-        await session.merge_async(fork1, fork2)
-        await session.commit_async("foo1 and foo2")
-
-        # Test additional fork merging
-        session = await repo.writable_session_async("main")
-        fork3 = pickle.loads(pickle.dumps(session.fork()))
-        fork4 = pickle.loads(pickle.dumps(session.fork()))
-        zarr.create_group(fork3.store, path="/foo3")
-        zarr.create_group(fork4.store, path="/foo4")
-
-        await session.merge_async(fork3, fork4)
-        await session.commit_async("foo3 and foo4")
-
-        # Verify all groups were created
         groups = set(
             name for name, _ in zarr.open_group(session.store, mode="r").groups()
         )


paraseba:
@mpiannucci is there a way to forward the sync methods to the async ones to reduce maintenance cost? Maybe with a macro?

mpiannucci:
@paraseba I can try, but the hard part is that the lifetimes are different for the async/blocking calls so there are some differences in where things need to be cloned and such

paraseba:
If it gets too tricky or fragile I'd be in favor of merging as is, we can explore possibilities in the future.

mpiannucci:
Okay after implementing a macro, i got this error: 
```
macros cannot be used as items in `#[pymethods]` impl blocks
       = note: this was previously accepted and ignored
```

bummer. Will do the simple way

mpiannucci:
A downside of remove thing this feature a while back is that a bunch of the stream processing in `Repository` is not marked as `Send` so its really hard to figure out what needs to change

mpiannucci:
I am not sure of what the best testing method for this is, but for now i duplicated some test in full `test_timetravel` and mixed in async with existing methods `test_gc` to use both. Let me know if there is a different desired approach

icechunk-python/src/repository.rs
```diff
@@ -560,6 +668,19 @@ impl PyRepository {
         })
     }
 
+    #[classmethod]
+    fn from_bytes_async<'py>(
```
paraseba: We shouldn't offer an async version for this method, there is no benefit.

icechunk-python/src/repository.rs
```diff
@@ -572,6 +693,16 @@ impl PyRepository {
         })
     }
 
+    fn as_bytes_async<'py>(&'py self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
```
paraseba: I think we should skip all methods that don't benefit from being async. In several of these the only real await is to lock the repo. If we don't expect the method to be called concurrently, I think it's ok not to offer an async version, which will lower maintenance cost.

I'll flag other methods that IMHO don't need async

icechunk-python/src/repository.rs
```diff
@@ -604,18 +752,63 @@ impl PyRepository {
...
+                .map_err(PyIcechunkStoreError::RepositoryError)?;
+            Ok(())
+        })
+    }
+
     pub fn config(&self) -> PyRepositoryConfig {
         self.0.blocking_read().config().clone().into()
     }
 
+    fn config_async<'py>(&'py self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
```
paraseba: doesn't need async version

icechunk-python/src/repository.rs
```diff
@@ -604,18 +752,63 @@ impl PyRepository {
...
     pub fn storage_settings(&self) -> PyStorageSettings {
         self.0.blocking_read().storage_settings().clone().into()
     }
 
+    fn storage_settings_async<'py>(
+        &'py self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let repository = self.0.clone();
+        pyo3_async_runtimes::tokio::future_into_py::<_, PyStorageSettings>(
```
paraseba: doesn't need async version

icechunk-python/src/repository.rs
```diff
@@ -604,18 +752,63 @@ impl PyRepository {
...
+            },
+        )
+    }
+
     pub fn storage(&self) -> PyStorage {
         PyStorage(Arc::clone(self.0.blocking_read().storage()))
     }
 
+    fn storage_async<'py>(&'py self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
+        let repository = self.0.clone();
```
paraseba: doesn't need async version

icechunk-python/src/repository.rs
```diff
@@ -627,13 +820,40 @@ impl PyRepository {
         })
     }
 
+    fn set_default_commit_metadata_async<'py>(
+        &'py self,
+        py: Python<'py>,
+        metadata: PySnapshotProperties,
+    ) -> PyResult<Bound<'py, PyAny>> {
```
paraseba: doesn't need async version

icechunk-python/src/repository.rs
```diff
@@ -627,13 +820,40 @@ impl PyRepository {
...
         py.allow_threads(move || {
             let metadata = self.0.blocking_read().default_commit_metadata().clone();
             metadata.into()
         })
     }
 
+    fn default_commit_metadata_async<'py>(
+        &'py self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyAny>> {
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -36,6 +36,19 @@ impl PySession {
         })
     }
 
+    #[classmethod]
+    fn from_bytes_async<'py>(
+        _cls: Bound<'_, PyType>,
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -49,30 +62,82 @@ impl PySession {
         })
     }
 
+    pub fn as_bytes_async<'py>(
+        &'py self,
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -49,30 +62,82 @@ impl PySession {
...
     #[getter]
     pub fn read_only(&self, py: Python<'_>) -> bool {
         // This is blocking function, we need to release the Gil
         py.allow_threads(move || self.0.blocking_read().read_only())
     }
 
+    pub fn read_only_async<'py>(
+        &'py self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyAny>> {
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -49,30 +62,82 @@ impl PySession {
...
     pub fn snapshot_id(&self, py: Python<'_>) -> String {
         // This is blocking function, we need to release the Gil
         py.allow_threads(move || self.0.blocking_read().snapshot_id().to_string())
     }
 
+    pub fn snapshot_id_async<'py>(
+        &'py self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let session = self.0.clone();
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -49,30 +62,82 @@ impl PySession {
...
+    }
+
     #[getter]
     pub fn branch(&self, py: Python<'_>) -> Option<String> {
         // This is blocking function, we need to release the Gil
         py.allow_threads(move || self.0.blocking_read().branch().map(|b| b.to_string()))
     }
 
+    pub fn branch_async<'py>(&'py self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
+        let session = self.0.clone();
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -49,30 +62,82 @@ impl PySession {
...
     pub fn has_uncommitted_changes(&self, py: Python<'_>) -> bool {
         // This is blocking function, we need to release the Gil
         py.allow_threads(move || self.0.blocking_read().has_uncommitted_changes())
     }
 
+    pub fn has_uncommitted_changes_async<'py>(
+        &'py self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let session = self.0.clone();
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -86,13 +151,37 @@ impl PySession {
...
         py.allow_threads(move || {
             self.0.blocking_write().discard_changes();
         })
     }
 
+    pub fn discard_changes_async<'py>(
+        &'py self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let session = self.0.clone();
```
paraseba: doesn't need async version

icechunk-python/src/session.rs
```diff
@@ -106,6 +195,19 @@ impl PySession {
         })
     }
 
+    pub fn store_async<'py>(&'py self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
```
paraseba: doesn't need async version

icechunk-python/src/store.rs
```diff
@@ -119,6 +119,23 @@ impl PyStore {
         })
     }
 
+    #[classmethod]
+    fn from_bytes_async<'py>(
```
paraseba: doesn't need async version

icechunk-python/src/store.rs
```diff
@@ -335,6 +369,34 @@ impl PyStore {
         })
     }
 
+    #[allow(clippy::too_many_arguments)]
+    #[pyo3(signature = (key, location, offset, length, checksum = None, validate_container = false))]
```
paraseba: Careful because this method changed silghtly in `main`. The pyo3 signature and the body too

icechunk-python/src/store.rs
```diff
@@ -389,6 +451,59 @@ impl PyStore {
         })
     }
 
+    fn set_virtual_refs_async<'py>(
```
paraseba, line 430: this method changed in main

icechunk-python/tests/test_virtual_ref.py
```diff
@@ -414,3 +414,116 @@ def test_error_on_non_authorized_virtual_chunk_container() -> None:
...
+
+async def test_async_single_virtual_ref() -> None:
+    """Test async API for setting a single virtual reference."""
+    config = RepositoryConfig.default()
+    container = VirtualChunkContainer(
+        "https://earthmover-sample-data.s3.amazonaws.com", http_store()
+    )
+    config.set_virtual_chunk_container(container)
+
+    repo = Repository.open_or_create(
```
paraseba: use the async version

paraseba: let's use async in as many places as possible here

paraseba (commented):
This is almost done, amazing work @mpiannucci !!

Two main comments:
* I think we don't need this many async methods, I flagged the ones I don't think we need. I'd create async versions only for those methods that can benefit from it in practice. This way we keep the maintenance cost at bay.
* Can we make sure we have at least one call in tests for each of the new `_async` methods?

icechunk/src/repository.rs
```diff
@@ -667,24 +699,35 @@ impl Repository {
...
+            match snap.into_iter().next() {
+                Some(snap) => Ok(snap.id),
+                None => Err(RepositoryErrorKind::InvalidAsOfSpec {
+                    branch: branch.clone(),
+                    at: *at,
                 }
+                .into()),
             }
+        } else {
+            Err(RepositoryErrorKind::InvalidVersionInfo(version.clone()).into())
```
paraseba: I think there may be a better way to do this, because if you think about it, this is not an error that can actually happen.

One possibility is instead of having a `From<&VersionInfo> for Option<RefVersionInfo>` how about having:

```rust
impl TryFrom<&VersionInfo> for RefVersionInfo {
   type Error =  (BranchName, DateTime<Utc>)
...
}
```

That way, this function can have a complete pattern match matching on `version.try_into()` and you don't need to introduce `InvalidVersionInfo`, the phantom error.

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1289,25 +1297,63 @@ class PyRepository:
...
     def from_bytes(cls, data: bytes) -> PyRepository: ...
     def as_bytes(self) -> bytes: ...
     @staticmethod
     def fetch_config(storage: Storage) -> RepositoryConfig | None: ...
+    @staticmethod
+    async def fetch_config_async(storage: Storage) -> RepositoryConfig | None: ...
     def save_config(self) -> None: ...
+    async def save_config_async(self) -> None: ...
     def config(self) -> RepositoryConfig: ...
+    def storage_settings(self) -> StorageSettings: ...
```
paraseba, line 1337: oh we missed this one and reopen before? :facepalm:

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1289,25 +1297,63 @@ class PyRepository:
...
+        authorize_virtual_chunk_access: dict[str, AnyCredential | None] | None = None,
+    ) -> PyRepository: ...
+    async def reopen_async(
+        self,
+        *,
+        config: RepositoryConfig | None = None,
+        authorize_virtual_chunk_access: dict[str, AnyCredential | None] | None = None,
+    ) -> PyRepository: ...
     def set_default_commit_metadata(self, metadata: dict[str, Any]) -> None: ...
+    async def set_default_commit_metadata_async(
```
paraseba: I think these don't need async? It's just an in-mem operation right?

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1384,22 +1484,36 @@ class PySession:
     @property
     def has_uncommitted_changes(self) -> bool: ...
     def status(self) -> Diff: ...
+    async def status_async(self) -> Diff: ...
```
paraseba: does this need async?

icechunk-python/src/repository.rs
```diff
@@ -627,13 +764,40 @@ impl PyRepository {
...
+        let repository = self.0.clone();
+        pyo3_async_runtimes::tokio::future_into_py::<_, PySnapshotProperties>(
```
paraseba: do you think this needs to be async?

icechunk/src/repository.rs
mpiannucci: that was my original approach 🤦  shoulda stuck with my gut on it haha

mpiannucci: This is way better now thank you 🙏

mpiannucci:
Okay @paraseba . Duplicated a lot of the tests for now, we can consolidate further possibly at some point but every method should be tried now. 

Also updated the header with a list of methods changed

PR #1078, opened 2025-07-14
Improve performance of GC and stats for repos with multiple refs

paraseba:
This avoids repeating computation (and I/O) when multiple branches/tags repeat parts of the tree (true in almost any repo that has more than one branch/tag)

### DIFF
--- a/icechunk/src/ops/mod.rs
+++ b/icechunk/src/ops/mod.rs
@@ -1,6 +1,8 @@
-use std::{collections::HashSet, future::ready, sync::Arc};
+use std::{collections::HashSet, sync::Arc};
 
+use async_stream::try_stream;
 use futures::{Stream, StreamExt as _, TryStreamExt as _, stream};
+use tokio::pin;
 use tracing::instrument;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


icechunk/src/ops/mod.rs
```diff
@@ -39,21 +39,73 @@ pub async fn pointed_snapshots<'a>(
...
+    use std::collections::{HashMap, HashSet};
+
+    use bytes::Bytes;
+
+    use crate::{
+        Repository, format::Path, new_in_memory_storage, ops::pointed_snapshots,
+    };
+
+    #[tokio::test]
+    async fn test_pointed_snapshots_duplicate() -> Result<(), Box<dyn std::error::Error>>
```
dcherian, line 85: what does "duplicate" mean?

icechunk/src/ops/mod.rs
```diff
@@ -39,21 +39,73 @@ pub async fn pointed_snapshots<'a>(
...
-    Ok(roots
-        .and_then(move |snap_id| {
+    let mut seen: HashSet<SnapshotId> = HashSet::new();
+    let res = try_stream! {
+        let roots = all_roots(storage, storage_settings, extra_roots)
+            .await?
+            .err_into::<RepositoryError>();
+        pin!(roots);
+
+        while let Some(pointed_snap_id) = roots.try_next().await? {
```
dcherian, line 49: How big a drag is this loop? Does it make sense to concurrently iterate over `roots` instead?

EDIT: I guess it's slow since we are pulling new snapshots for `ancestry`

icechunk/src/ops/mod.rs
paraseba: before this change, we were returning duplicate snapshot ids in this iterator, once per ref including the snapshot in its ancestry

icechunk/src/ops/mod.rs
paraseba: yes, there are a few optimizations we need to do here ... it's not totally trivial because we always need to keep track of things we have already seen, so there is mutable state across the threads. But totally doable.

PR #1081, opened 2025-07-15
Parallelize size statistics gathering

paraseba:
This adds concurrency to `repo_chunks_storage`.

I verified manually this improves performance very significantly, but benchmarks should be able to tell us.

### DIFF
--- a/icechunk/src/ops/stats.rs
+++ b/icechunk/src/ops/stats.rs
@@ -1,16 +1,74 @@
-use futures::TryStreamExt as _;
-use std::{collections::HashSet, sync::Arc};
-use tokio::pin;
+use futures::{TryStream, TryStreamExt as _, future::ready, stream};
+use std::{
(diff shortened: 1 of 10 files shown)


icechunk/src/ops/stats.rs
```diff
@@ -28,48 +87,30 @@ pub async fn repo_chunks_storage(
...
+            let manifest_size = manifest_info.size_bytes;
+            manifest_chunks_storage(
+                manifest_info.id,
+                manifest_size,
+                Arc::clone(asset_manager),
+                Arc::clone(seen_chunks),
+            )
+            .await
+        })
+        .try_buffered(process_manifests_concurrently.get() as usize)
```
mpiannucci, line 111: Yesss!! I was hoping it would be structured instead of spawning!

mpiannucci (approved):

PR #1085, opened 2025-07-16
Improve performance of sise stats

paraseba:
This introduces two limits to the concurrency:

* max memory keeping for in flight manifests
* max number of concurrent manifest fetches

With this, we don't need to limit the concurrency to a specific number of manifests, but now we can fetch more manifests if they are small.

Combining these two limits, we get an important performance optimization, particularly for repos with a large number of smallish manifests.

I used this script to test and observe the performance of this:

```python
import arraylake
import time
import icechunk

icechunk.set_logs_filter("icechunk::stream_utils=trace")

token = '...'

client = arraylake.Client(token=token)
repo = client.get_repo("earthmover-public/era5-surface-aws")

start_time = time.time()
repo.total_chunks_storage(None, None)
print(" %s seconds" % (time.time() - start_time))
```

At the time of running this, ERA5 took around 4 seconds to compute.

Testing with real world repos with small manifests shows performance improvements of the order of 30x.


icechunk/src/ops/stats.rs
```diff
@@ -90,27 +76,59 @@ pub async fn repo_chunks_storage(
...
+        })
+        // Now we can buffer a bunch of fetch_manifest operations. Because we are using
+        // StreamLimiter we know memory is not going to blow up
+        .try_buffer_unordered(max_concurrent_manifest_fetches.get() as usize)
+        .map_ok(|(manifest, minfo)| async move {
+            let size = calculate_manifest_storage(manifest, Arc::clone(seen_chunks))?;
+            Ok((size, minfo))
+        })
+        // We do some more buffering to get some concurrency on the processing of the manifest file
+        // TODO: this should actually happen in a CPU bounded worker pool
```
mpiannucci, line 118: If you mean to spawn here I'd rather we don't

icechunk/src/stream_utils.rs
```diff
@@ -0,0 +1,123 @@
...
+/// Note: this could have a much better interface with a bit of extra work:
+/// - We could simple wrap the stream and it would take care of mapping every element
+/// - Elements could be freed automatically using the Drop instance
+#[derive(Debug)]
+pub(crate) struct StreamLimiter<T, F: Fn(&'_ T) -> usize> {
+    get_usage: F,
+    max_usage: usize,
+    state: Mutex<StreamLimiterState>,
+    waiter: Notify,
+    _phantom: PhantomData<T>,
```
mpiannucci, line 54: Why do we need the phantom?

icechunk/src/stream_utils.rs
```diff
@@ -0,0 +1,123 @@
...
+        loop {
+            let mut state = self.state.lock().await;
+            if state.current_usage == 0
+                || state.current_usage + this_item_usage <= self.max_usage
+            {
+                trace!(
+                    item_usage = this_item_usage,
+                    current_usage = state.current_usage,
+                    in_flight = state.in_flight,
+                    "Stats rate limiter: Allowing item"
```
mpiannucci: I'm assuming this will be used for more than stats

icechunk/src/ops/stats.rs
```diff
@@ -90,27 +76,59 @@ pub async fn repo_chunks_storage(
...
+    // we want to fetch many manifests in parallel, but not more than memory allows
+    // for this we use the StreamLimiter using the manifest size in bytes for usage
+    let limiter = &Arc::new(StreamLimiter::new(
+        max_manifest_mem_bytes.get(),
+        |m: &ManifestFileInfo| m.size_bytes as usize,
+    ));
+
+    // The StreamLimiter works by calling limit on every element before they are processed
+    let rate_limited_manifests = unique_manifest_infos
+        .and_then(|m| async move { Ok(limiter.clone().limit(m).await) });
```
mpiannucci, line 99: 😍

mpiannucci (approved):
This is beautiful 👏  a few comments for my own learning

icechunk/src/ops/stats.rs
paraseba: Not exactly, but this can take some time, you are not supposed to do CPU bound shit in a tokio Task, it could block the network stack for example. But, I think this is enough for now, we can iterate.

icechunk/src/stream_utils.rs
paraseba: Rust doesn't like a type declared as `Foo<T>` not to use `T`. `PhantomData` has the mission of making the compiler think I'm using it, and still not take any storage.

icechunk/src/stream_utils.rs
paraseba: good catch, I moved it later, will change.

PR #1090, opened 2025-07-17
Add zarr_transaction that yields a Zarr Store

dcherian:
Repository only knows about Sessions, so Repository.transaction should yield a Session.

I am raising a DeprecationWarning, though we could instead make a breaking change.

### DIFF
--- a/icechunk-python/python/icechunk/repository.py
+++ b/icechunk-python/python/icechunk/repository.py
@@ -1,4 +1,5 @@
 import datetime
+import warnings
 from collections.abc import AsyncIterator, Iterator
 from contextlib import contextmanager
 from typing import Any, Self, cast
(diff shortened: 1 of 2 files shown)


paraseba:
I'm not sure what's best. This is an important method, so we would break a lot of code. Maybe we just live with it and find a new name for the Session yielding transaction?

rabernat:
> Repository only knows about Sessions, so Repository.transaction should yield a Session.

I'm not sure I agree with this. If we switch to yielding a session, what happens if you commit the session inside the transaction? To me this indicates that `transaction` needs to yield something other than a session.

paraseba:
We could play some tricks to not allow you to commit explicitly, or to detect that you did and do nothing on block exit

Issue #1086, opened 2025-07-17
Add static redirects from https://icechunk.io/en/latest/icechunk-python/* to corresponding https://icechunk.io/en/latest/*


jbusecke:
### Problem

I noticed that when I search for `"icechunk virtual"` on google I get this link: https://icechunk.io/en/latest/icechunk-python/virtual/ which results in a 404 error. 

https://icechunk.io/en/latest/virtual/ shows what I am after, and I presume that the prefix was at some point changed globally?

### Suggested Improvement

We could add some static redirects to make these links point to the new structure.

paraseba:
ah good point. Do you happen to know how to do that in readthedocs @jbusecke ? I have no idea.

jbusecke:
[This](https://readthedocs-theme.readthedocs.io/en/latest/user-defined-redirects.html) seems promising!

ianhi:
Partially fixed in #1092. Or rather the original issue should be fixed for latest, and will be fixed for stable with out next release. 

There are a few more docs redirect issues that I will solve in a followup PR

PR #1092, opened 2025-07-17
docs: add redirects to correct for doc reorg

ianhi:
Uses the mkdocs-redirect plugin we already have installed to manage redirects we need from https://github.com/earth-mover/icechunk/pull/1051

This is part 1 of two PRs to fix all redirect issues (https://github.com/earth-mover/icechunk/issues/1086)


ianhi:
Seems to work. https://icechunk--1092.org.readthedocs.build/en/1092/icechunk-python/virtual brings me to the correct place

paraseba (approved):

PR #1094, opened 2025-07-18
Fix Rendering of some code blocks

ianhi:
These blocks were not rendering properly: https://icechunk.io/en/latest/configuration/#example

image

I am a little alarmed that there were no errors thrown by the CI, but this at least means the docs will look good in the short term

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -81,7 +81,8 @@ Icechunk allows repos to contain [virtual chunks](./virtual.md). To allow for re
 
 For example, if we wanted to configure an icechunk repo to be able to contain virtual chunks from an `s3` bucket called `my-s3-bucket` in `us-east-1`, we would do the following:
 
-```python exec="on" session="config" source="material-block"
+```python
+
 config.virtual_chunk_containers = [
     icechunk.VirtualChunkContainer(
         url_prefix="s3://my-s3-bucket/",
@@ -94,7 +95,7 @@ config.virtual_chunk_containers = [
 
 If we also wanted to configure the repo to be able to contain virtual chunks from another `s3` bucket called `my-other-s3-bucket` in `us-west-2`, we would do the following:
 
-```python exec="on" session="config" source="material-block"
+```python
 config.set_virtual_chunk_container(
     icechunk.VirtualChunkContainer(
         url_prefix="s3://my-other-s3-bucket/",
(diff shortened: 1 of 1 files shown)


paraseba:
Do you know if the code in them was failing or something else? Are there alternatives where we don't lose the execution check?

ianhi:
I think these can't really be executed because they are attempting to access S3 buckets. Which the docss build does not have access to, and also may not be real buckets

ianhi:
Although that said, this code seems to be incorrect

ianhi:
I'm getting an error when I run this:

```python
config.virtual_chunk_containers = [
    icechunk.VirtualChunkContainer(
        url_prefix="s3://my-s3-bucket/",
        storage=icechunk.StorageSettings(
            storage=icechunk.s3_storage(bucket="my-s3-bucket", prefix="foo", region="us-east-1"),
        ),
    ),
]
```

`StorageSettings.__new__() got an unexpected keyword argument 'storage'`

I guess we haven't been building using `--strict` which is why the docs build was not errorring, But also whats teh proper way to fix this?

dcherian:
Try
```
config.virtual_chunk_containers = [
    icechunk.VirtualChunkContainer(
        url_prefix="s3://my-s3-bucket/",
        storage==icechunk.s3_storage(bucket="my-s3-bucket", prefix="foo", region="us-east-1"),
    ),
]
````

**Comment by ianhi** [comment]
We can fix this as is. but I got trapped trying to figure out why markdown-exec didn't throw a build error on these (it should have caught this!)

We can probably merge this after I update the code to the correct version and then open an issue about the doc execution not failing

**Comment by ianhi** [comment]
Here is the current docs code, which fails:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "icechunk",
# ]
# ///


import icechunk

config = icechunk.RepositoryConfig.default()


config.set_virtual_chunk_container(
    icechunk.VirtualChunkContainer(
        url_prefix="s3://my-other-s3-bucket/",
        storage=icechunk.StorageSettings(
            storage=icechunk.s3_storage(
                bucket="my-other-s3-bucket", prefix="other-foo", region="us-west-2"
            ),
        ),
    ),
)

```



and i can't @dcherian's suggestion work ebcasue there is no storage objec tto compare to

ianhi:
Fixed with Seba's help

Issue #754, opened 2025-02-19
DOC - Improvement: Making example code executable


ianhi:
# Problem
Code in the docs can have typos, or may work when it was written but becomes out of date with underlying changes. We need a way to regularly run the example code in the docs.

# Solutions?

I think there are two approaches to this. 

1. Transform the docs and include them as part of the test suite

Conceptually just extracting the python and running it, but there is some complexity. That can be managed by somethign like: https://github.com/koaning/mktestdocs

However it does not seem like any of the existing solutions will easily allow us to manage the code blocks where the user chooses between local and s3 storage so that's one potential roadblock.

2. Swap the docs framework to something that natively understands executable documents. e.g. with myst markdown based docs. Then the code can be executed as part of the docs. I personally like this solution as it also lends itself towards allowing a reader to open a jupyterhub with the docs page, but it is python centric and would require redoing the docs configuration.

paraseba:
@jhamman may have opinions on this. He recently migrated us to readthedocs

PR #1099, opened 2025-07-21
Update docs on virtual references to match VirtualiZarr v2.0's updated API

TomNicholas:
We should re-run the example to double-check everything still works as expected.

### DIFF
--- a/icechunk-python/docs/docs/virtual.md
+++ b/icechunk-python/docs/docs/virtual.md
@@ -288,7 +288,7 @@ Here is how we can set the chunk at key `c/0` to point to a file on my local fil
 
 ```python
 config = ic.config.RepositoryConfig.default()
-config.set_virtual_chunk_container(ic.virtual.VirtualChunkContainer("s3://mybucket/my/data/", ic.storage.local_filesystem_store("/path/to/my")))
+config.set_virtual_chunk_container(ic.virtual.VirtualChunkContainer("file:///path/to/my/", ic.storage.local_filesystem_store("/path/to/my")))
 repo = ic.Repository.create(storage, config)
 session = repo.writable_session("main")
 session.store.set_virtual_ref('c/0', 'file:///path/to/my/file.nc', offset=20, length=100)


docs/docs/virtual.md
```diff
@@ -16,13 +16,14 @@ While Icechunk works wonderfully with native chunks managed by Zarr, there is lo
 
 We are going to create a virtual dataset pointing to all of the [OISST](https://www.ncei.noaa.gov/products/optimum-interpolation-sst) data for August 2024. This data is distributed publicly as netCDF files on AWS S3, with one netCDF file containing the Sea Surface Temperature (SST) data for each day of the month. We are going to use `VirtualiZarr` to combine all of these files into a single virtual dataset spanning the entire month, then write that dataset to Icechunk for use in analysis.
 
-Before we get started, we need to install `virtualizarr`, and `icechunk`. We also need to install `fsspec` and `s3fs` for working with data on s3.
+Before we get started, we need to install `virtualizarr` (this notebook uses VirtualiZarr v2.0.0), and `icechunk`. We also need to install `fsspec`, `s3fs`, and `obstore` for working with data on s3.
```
TomNicholas: The only reason we are using `fsspec` and `s3fs` here is literally just to glob the files in the bucket. If we had a [globbing function in `obspec`](https://github.com/developmentseed/obspec/issues/12) then we wouldn't need `fsspec` at all.

docs/docs/virtual.md
```diff
@@ -40,13 +41,27 @@ oisst_files = sorted(['s3://'+f for f in oisst_files])
...
 
 ```python
+from virtualizarr.parsers import HDFParser
+from virtualizarr.registry import ObjectStoreRegistry
+
+bucket = "noaa-cdr-sea-surface-temp-optimum-interpolation-pds/"
+store = S3Store(
+    bucket=bucket,
+    region="us-west-2",
```
TomNicholas: I'm not sure this is the right region - @mpiannucci presumably you know?

mpiannucci: I believe it is, you can double check on the AWS registry tho

ianhi:
Can these docs be autoexecuted? Seems as though they could

TomNicholas:
> Can these docs be autoexecuted? Seems as though they could

Yes should be. But do we have any examples of making docs executable in this repo yet? Looks like this #754 is still open?

docs/docs/virtual.md
TomNicholas: I think it's actually `us-east-1` https://registry.opendata.aws/noaa-cdr-oceanic/

ianhi:
> > Can these docs be autoexecuted? Seems as though they could
>
> Yes should be. But do we have any examples of making docs executable in this repo yet? Looks like this https://github.com/earth-mover/icechunk/issues/754 is still open?


yes, most of them are executed now. For example https://github.com/earth-mover/icechunk/blob/main/docs/docs/xarray.md

essentially it's a bunch of these: 

````python exec="on" session="xarray" source="material-block"`

File: icechunk-python/docs/docs/virtual.md [hunk 3]
```diff
@@ -40,13 +41,31 @@ oisst_files = sorted(['s3://'+f for f in oisst_files])
 #]
 ```
 
-Now that we have the filenames of the data we need, we can create virtual datasets with `VirtualiZarr`. This may take a minute.
+These are netCDF4 files, which are really HDF5 files, so we need to user virtualizarr's `HDFParser`.
+
+We also need to give the parser a way to access our files. We do this by creating an `ObjectStoreRegistry` containing an obstore `S3Store` for that bucket.
+
+```python
```
ianhi, line 52:
```suggestion
```python exec="on" session="virtual" source="material-block"
```

like this, and adding the same on other code blcoks. They share variables and state so long as they have the same "session"

dcherian:
@TomNicholas can this be closed?

PR #1101, opened 2025-07-22
GC: Performance improvement + dry run

paraseba:
A repo with 300k tiny manifests runs in ~ 15-20 minutes. Before this change it took many hours.

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -1350,16 +1350,22 @@ class PyRepository:
         delete_expired_branches: bool = False,
         delete_expired_tags: bool = False,
     ) -> set[str]: ...
-    def garbage_collect(
-        self, delete_object_older_than: datetime.datetime
-    ) -> GCSummary: ...
     def rewrite_manifests(
         self, message: str, *, branch: str, metadata: dict[str, Any] | None = None
     ) -> str: ...
+    def garbage_collect(
+        self,
+        delete_object_older_than: datetime.datetime,
+        dry_run: bool = False,
+        max_snapshots_in_memory: int = 50,
+        max_compressed_manifest_mem_bytes: int = 512 * 1024 * 1024,
+        max_concurrent_manifest_fetches: int = 500,
+    ) -> GCSummary: ...
     def total_chunks_storage(
         self,
-        max_manifest_mem_bytes: int | None = None,
-        max_concurrent_manifest_fetches: int | None = None,
+        max_snapshots_in_memory: int = 50,
+        max_compressed_manifest_mem_bytes: int = 512 * 1024 * 1024,
+        max_concurrent_manifest_fetches: int = 500,
     ) -> int: ...
 
 class PySession:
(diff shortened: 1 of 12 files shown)


icechunk/src/ops/gc.rs
```diff
@@ -153,7 +185,141 @@ pub enum GCError {
...
+                Arc::clone(asset_man).fetch_manifest(&m.id, m.size_bytes).await?;
+            Ok((manifest, m))
+        })
+        // Now we can buffer a bunch of fetch_manifest operations. Because we are using
+        // StreamLimiter we know memory is not going to blow up
+        .try_buffer_unordered(config.max_concurrent_manifest_fetches.get() as usize)
+        .and_then(move |(manifest, minfo)| {
+            let keep_chunks_c = Arc::clone(&keep_chunks_c);
+            async move {
+                task::spawn_blocking(move || {
```
mpiannucci: Is this necessary because there could be millions of chunks and iter would block otherwise?

paraseba: yeah ... a manifest could have 10M chunks or something, this would take a while and in the meantime some HTTP requests in flight could timeout

icechunk/src/repository.rs
```diff
@@ -173,55 +173,49 @@ impl Repository {
...
-                // TODO: we could cache this first snapshot
-                let asset_manager = AssetManager::new_no_cache(
-                    Arc::clone(&storage_c),
-                    storage_settings.clone(),
-                    compression,
-                );
-                // On create we need to create the default branch
-                let new_snapshot = Arc::new(Snapshot::initial()?);
-                asset_manager.write_snapshot(Arc::clone(&new_snapshot)).await?;
+        let create_branch = async move {
```
paraseba, line 176: this is an unrelated improvement, i learned about `join!` macro

icechunk-python/python/icechunk/repository.py
```diff
@@ -673,7 +673,14 @@ def expire_snapshots(
             delete_expired_tags=delete_expired_tags,
         )
 
-    def garbage_collect(self, delete_object_older_than: datetime.datetime) -> GCSummary:
+    def garbage_collect(
+        self,
+        delete_object_older_than: datetime.datetime,
+        dry_run: bool = False,
```
dcherian, line 679: Would be good to make `dry_run` and every kwarg after that kwarg-only

paraseba: good idea, will do

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1350,16 +1350,22 @@ class PyRepository:
...
+        dry_run: bool = False,
+        max_snapshots_in_memory: int = 50,
+        max_compressed_manifest_mem_bytes: int = 512 * 1024 * 1024,
+        max_concurrent_manifest_fetches: int = 500,
+    ) -> GCSummary: ...
     def total_chunks_storage(
         self,
-        max_manifest_mem_bytes: int | None = None,
-        max_concurrent_manifest_fetches: int | None = None,
+        max_snapshots_in_memory: int = 50,
```
dcherian, line 1366: ooh same kwarg comment applies here. Notice that the order of kwargs has changed!

PR #1103, opened 2025-07-22
keyword arguments for gc functions

paraseba:


icechunk-python/python/icechunk/repository.py
```diff
@@ -673,9 +673,42 @@ def expire_snapshots(
             delete_expired_tags=delete_expired_tags,
         )
 
+    def rewrite_manifests(
```
paraseba, line 676: this is just code moved to keep stats and gc together, no changes here

dcherian (approved):

PR #1109, opened 2025-07-23
Fix tagged enum serialiation for GCS and Azure credentials

mpiannucci:
https://github.com/earth-mover/icechunk/discussions/1106

This is necessary because of this: https://github.com/serde-rs/serde/issues/1307

Tested with actual GCS service account file, test not added because we dont have secrets set up for it at this time.

### DIFF
--- a/icechunk/src/config.rs
+++ b/icechunk/src/config.rs
@@ -503,7 +503,7 @@ pub enum S3Credentials {
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
-#[serde(tag = "gcs_static_credential_type")]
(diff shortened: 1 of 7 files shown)


icechunk/src/config.rs
```diff
@@ -503,7 +503,7 @@ pub enum S3Credentials {
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
-#[serde(tag = "gcs_static_credential_type")]
+#[serde(tag = "gcs_static_credential_type", content = "__field0")]
```
paraseba, line 507: could you add a link to the issue in both places?

mpiannucci: done!

PR #1110, opened 2025-07-23
Updates the glad-ingest jupyter notebook to reflect the new session fork and session merge capabilities

jakenotjay:
- Currently the notebook within the guide is out of date, still using the old allow_pickling context manager
- Updates the notebook to use the new behaviour described in [the docs](https://icechunk.io/en/stable/parallel/#cooperative-distributed-writes)
- Note to maintainers: I've not run this code, as I don't have access to coiled (nor do I want to incur the cost 👀 ) so you'll need to run this yourselves to get the output before updating the docs!

### DIFF
--- a/docs/docs/ingestion/glad-ingest.ipynb
+++ b/docs/docs/ingestion/glad-ingest.ipynb
@@ -488,7 +488,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": null,
    "id": "a07f0ef8-da71-480f-b79d-6e6f5564ac58",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


paraseba:
Thank you @jakenotjay !

paraseba (approved):

jakenotjay:
@paraseba I think this set of changes might be incorrect as per [this page](https://icechunk.io/en/latest/xarray/):

> [xarray.Dataset.to_zarr](https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_zarr.html#xarray.Dataset.to_zarr) and [to_icechunk](https://icechunk.io/en/latest/reference/#icechunk.xarray.to_icechunk) are nearly functionally identical.
> 
> In a distributed context, e.g. writes orchestrated with multiprocesssing or a dask.distributed.Client and dask.array, you must use to_icechunk. This will ensure that you can execute a commit that successfully records all remote writes. See [these docs on orchestrating parallel writes](https://icechunk.io/en/latest/parallel/) and [these docs on dask.array with distributed](https://icechunk.io/en/latest/dask/#icechunk-dask-xarray) for more.
> 
> If using to_zarr, remember to set zarr_format=3, consolidated=False. Consolidated metadata is unnecessary (and unsupported) in Icechunk. Icechunk already organizes the dataset metadata in a way that makes it very fast to fetch from storage.

PR #1129, opened 2025-07-28
Slightly more informative errors

dcherian:
Some local changes I made while debugging. Feel free to just fold it in to your PR

### DIFF
--- a/icechunk/src/format/manifest.rs
+++ b/icechunk/src/format/manifest.rs
@@ -215,7 +215,7 @@ pub enum VirtualReferenceErrorKind {
         "no virtual chunk container can handle the chunk location ({0}), edit the repository configuration adding a virtual chunk container for the chunk references, see https://icechunk.io/en/stable/virtual/"
     )]
     NoContainerForUrl(String),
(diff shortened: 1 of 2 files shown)


icechunk/src/format/manifest.rs
```diff
@@ -215,7 +215,7 @@ pub enum VirtualReferenceErrorKind {
         "no virtual chunk container can handle the chunk location ({0}), edit the repository configuration adding a virtual chunk container for the chunk references, see https://icechunk.io/en/stable/virtual/"
     )]
     NoContainerForUrl(String),
-    #[error("error parsing virtual ref URL")]
+    #[error("error parsing virtual ref URL: {0}")]
```
paraseba, line 218: I remember at some point deleting all those {0}. Doesn't `from` get automatically rendered by our error logging system?

paraseba: because we do have printing of the nested errors, is in not working as expected?

dcherian: OK then that is not working as expected

PR #1131, opened 2025-07-28
Validate url on set_virtual_ref[s]

paraseba:
Now we verify the URL can be parsed as an absolute URL and fail immediately if it cannot.

This will have a small impart on `set_virtual_refs` performance.

### DIFF
--- a/icechunk-python/tests/test_virtual_ref.py
+++ b/icechunk-python/tests/test_virtual_ref.py
@@ -414,3 +414,50 @@ def test_error_on_non_authorized_virtual_chunk_container() -> None:
 
     with pytest.raises(IcechunkError, match="file:///foo.*authorize"):
         array[0]
+
+
+def test_cannot_write_invalid_urls() -> None:
+    repo = Repository.create(
+        storage=in_memory_storage(),
+    )
+    session = repo.writable_session("main")
+    store = session.store
+
+    zarr.create_array(store, shape=(5,), chunks=(1,), dtype="i4", compressors=None)
+
+    with pytest.raises(IcechunkError, match="error parsing"):
+        store.set_virtual_ref(
+            "c/0",
+            "",
+            offset=1,
+            length=4,
+        )
+
+    with pytest.raises(IcechunkError, match="error parsing"):
+        store.set_virtual_ref(
+            "c/0",
+            "some/relative/url",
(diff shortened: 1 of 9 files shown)


icechunk-python/tests/test_virtual_ref.py
```diff
@@ -414,3 +414,50 @@ def test_error_on_non_authorized_virtual_chunk_container() -> None:
 
     with pytest.raises(IcechunkError, match="file:///foo.*authorize"):
         array[0]
+
+
+def test_cannot_write_invalid_urls() -> None:
```
dcherian, line 419: The traceback I get running our reproducer is
```
icechunk.IcechunkError:   × error parsing virtual ref URL
  │
  │ context:
  │
  │
  ├─▶ error parsing virtual ref URL
  ╰─▶ relative URL without a base
```

Can we print the url here? Perhaps we're missing a formatting specifier?

paraseba: huh, w t f

paraseba: ah ok ... the unherlying error is printed, but not the url, that's my bad ... fixing

paraseba: can you try again @dcherian ?

paraseba: well, i guess i should put some quotes or something, or the empty string will show nothing

dcherian: in python this would `{url!r}` :)

icechunk/src/format/manifest.rs
```diff
@@ -215,8 +215,12 @@ pub enum VirtualReferenceErrorKind {
         "no virtual chunk container can handle the chunk location ({0}), edit the repository configuration adding a virtual chunk container for the chunk references, see https://icechunk.io/en/stable/virtual/"
     )]
     NoContainerForUrl(String),
-    #[error("error parsing virtual ref URL")]
-    CannotParseUrl(#[from] url::ParseError),
+    #[error("error parsing virtual ref URL: {url}")]
```
dcherian: Perhaps 

```suggestion
    #[error("error parsing virtual ref URL: {url:?}")]
```

PR #1133, opened 2025-07-28
Make `validate_container` default to True in `set_virtual_ref[s]`

paraseba:
Now that virtual chunk containers are mandatory for reads, we should make them at least the default (if not mandatory) for writes.

This gives better UX, users of VZ won't produce repos that are broken for reads, just because they forget to create the containers.

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -1435,13 +1435,14 @@ class PyStore:
         offset: int,
         length: int,
         checksum: str | datetime.datetime | None = None,
-        validate_container: bool = False,
+        validate_container: bool = True,
     ) -> None: ...
     def set_virtual_refs(
         self,
         array_path: str,
         chunks: list[VirtualChunkSpec],
-        validate_containers: bool,
+        *,
+        validate_containers: bool = True,
     ) -> list[tuple[int, ...]] | None: ...
     async def delete(self, key: str) -> None: ...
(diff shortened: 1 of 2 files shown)


dcherian (approved):

maxrjones:
FYI about some downstream impacts of this change - https://github.com/zarr-developers/VirtualiZarr/issues/763#issuecomment-3172245256. Would you be willing to explain the motivation for returning indices rather than raising an error when container validation fails? I'm asking because using the return type to signify a failure made it challenging to isolate why workflows with VZ 2.0.1 and IC 1.1.0 were just returning fill values (silently no manifests were written due to this setting).

paraseba:
@maxrjones I can look at this next week, but If `set_virtual_ref` is not raising an error I'd say it's a bug. The function is typed as returning `None`. If it's a bug we can fix it quickly and release a new version

paraseba:
Ah I see now the problem is with `set_virtual_refs`? I can explain he reasoning for that one not raising, but I'm open to changing or adding different APIs:
- That function is not "atomic", it's there only for performance reasons, so IC can do its thing without having to constantly switch between rust and python as you call the (singular) `set_virtual_ref`.
- If we raised an error when we find a missing container, you could end up with part of your references set, and the rest not set, and you wouldn't know which ones succeeded.
- Getting back the indices of the failed references, allows them to add the container and retry only those missing indices.

If you prefer, we can think about adding another function that raises an error on first failure, leaving things partially done.

Issue #1112, opened 2025-07-23
Make it easier for users to notice they should increase cachnig config


paraseba:
For example, users with manifest larger than 5M refs (the default caching config) will see terrible performance unless they tweak the configuration.

PR #1135, opened 2025-07-29
Warn if caches are too small

paraseba:
We are emiting a warning first time we see a manifest that is more than half the cache size. Same for snapshots that are larger than 1/5 the cache size.

Closes: #1112

### DIFF
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -7,9 +7,9 @@ use serde::{Deserialize, Serialize};
 use std::{
     io::{BufReader, Read},
     ops::Range,
-    sync::Arc,
+    sync::{Arc, atomic::AtomicBool},
 };
-use tracing::{Span, debug, instrument, trace};
+use tracing::{Span, debug, instrument, trace, warn};
 
 use crate::{
     Storage,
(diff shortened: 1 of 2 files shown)


mpiannucci (approved):

icechunk/src/asset_manager.rs
```diff
@@ -184,12 +193,43 @@ impl AssetManager {
...
                 Ok(manifest)
             }
         }
     }
 
+    fn warn_if_manifest_cache_small(&self, manifest: &Manifest) {
+        let manifest_weigth = manifest.len();
+        let capacity = self.num_chunk_refs;
+        // TODO: we may need a config to silence this warning
+        if manifest_weigth as u64 > capacity / 2
```
dcherian: typo:
```suggestion
        if manifest_weight as u64 > capacity / 2
```

And :+1: to the config option. Sucks to have libraries issue warnings you can't turn off.

dcherian (approved):

icechunk/src/asset_manager.rs
paraseba: typo done ...

config option, coming soon I guess?

At least it only logs once :shrug: , and you could silence it if needed via the log config

dcherian: OK good point.

PR #1147, opened 2025-08-01
Better credential refresh

paraseba:
- Start attempting to refresh earlier when using S3 SDK. Credential refresh could take a while.
- Less contending over the credential lock in GCS
- Fix logic in Python refreshable credentials

Context: some users are seeing errors when trying to refresh credentials, but we weren't able to reproduce.


mpiannucci (approved):
Great finds on limiting the GIL

PR #1151, opened 2025-08-04
Design notes towards Icechunk 2.0

paraseba:


design-docs/010-notes-towards-an-IC-2.md
```diff
@@ -0,0 +1,83 @@
...
+good to add a new `moved` operation to our transaction logs and conflict
+detection.
+
+### Hard to know repo spec version
+
+Currently there is no real way to know the spec version of a repo. Best
+chance is going through every branch/tag looking at the spec version in
+the snapshot file
+
+* See [Single Entry Point Object for Refs and Ancestry](./011-ref-and-ancestry-entry-point.md)
```
dcherian, line 83:
```suggestion
* See [Single Entry Point Object for Refs and Ancestry](./011-ref-and-ancestry-entry-point.md)

### Support more complex array updates

Example: backfills, inserts, and upserts. This would require some changes to the manifest to make it efficient, as well as upstream API additions to Zarr.
```

design-docs/011-ref-and-ancestry-entry-point.md
```diff
@@ -0,0 +1,220 @@
...
+    message: str (required);
+    metadata: [MetadataItem] (required);
+}
+
+enum RepoAvailability : ubyte { Online = 0, ReadOnly, Offline }
+
+table RepoStatus {
+    availability: RepoAvailability (required);
+    limited_availability_reason: string;
+    set_at: uint64 (required);
```
dcherian, line 40: Also `set_by: string` perhaps?

design-docs/010-notes-towards-an-IC-2.md
```diff
@@ -0,0 +1,83 @@
+# Notes Towards an Icechunk 2.0
+
+This is intended to be a living document, receiving updates as we come up with
+concrete plans and design documents.
+
+## Current limitations and issues
```
dcherian, line 6: Couple of other things to think about:
1. repos with large numbers of nodes. 
2. supporting multiple arrays in a single manifest.
3. tracking changes to the config file
4. with `Repo.rewrite_manifests` we commit for the user. I don't like this. Do you think we need a "system commit" concept'?

dcherian (approved):

design-docs/011-ref-and-ancestry-entry-point.md
paraseba: I think that can go in the reason maybe? I'm not sure what OS users would put it `set_by`

design-docs/010-notes-towards-an-IC-2.md
paraseba: Love these. Will think and add. Except for 2 that is supported today iiuc, remember that's where we started and we never broke that possibility.

dcherian: You're right. i meant actually implementing that feature.

design-docs/010-notes-towards-an-IC-2.md
```diff
@@ -0,0 +1,93 @@
...
+
+* See [Single Entry Point Object for Refs and Ancestry](./011-ref-and-ancestry-entry-point.md)
+
+### Support more complex array updates
+
+Example: backfills, inserts, and upserts. This would require some
+changes to the manifest to make it efficient, as well as upstream
+API additions to Zarr.
+
+### Is Icechunk efficient in the presence of many thousands of nodes?
```
jleben, line 91: Hello! Just adding my 2 c on this one - I've been thinking about this issue a lot and looking over the recent commits in the repo I found it mentioned here, so seems like a perfect place to comment.

Due to the limitations in updating a single Xarray dataset stored in Icechunk (no inserts, prepends, deletes, as well as limitations in appending), I am considering storing a very large dataset (up to 100 TiB) as many many individual datasets (tens of thousands).

I discovered the issue is that Icechunk stores some data for each Zarr array all in a single snapshot file. Based on some experiments, I think this file could grow to 1+ GiB in size. This file is loaded in entirety upon opening an Icechunk session, and also a new such file is generated for each committed snapshot. Issues:
- If I commit my dataset in many separate snapshots, just the snapshot data will take a lot of space.
- There will be a huge upfront time cost (and memory cost) when opening a session.

The solution I see would go along these lines:

- Distribute this per-array info into multiple files.
- Do not store a copy of all this data with each snapshot, instead store only the changes + references to older snapshot data that together makes a whole snapshot.

Is something like this planned? This design doc seems to still propose a single snapshot file.

PR #1161, opened 2025-08-12
feature: implement zarr's zep8 adapter for icechunk

### LINKED ISSUE #576: URL syntax for icechunk
I'm in the process of implementing support for ZEP 8-style URL syntax in Neuroglancer, and in conjunction with that am planning to also add support for icechunk format.

Here are some examples of existing URLs that are supported or will be supported:

gs://bucket/path/to/array/|zarr3:
gs://bucket/path/to/group/|zarr3:path/to/array/
gs://bucket/path/to/file.zip|zip:path/to/array/|zarr3:
gs://bucket/path/to/file.zip|zip:path/to/nested.zip|zip:path/to/array/|zarr3:
gs://bucket/path/to/ocdbt/|ocdbt:path/to/array/|zarr3:

Note that the URL consists of a "pipeline" of |-separated components, where the first component must be a base kvstore protocol (e.g. gs, s3, http), followed by zero or more kvstore adapter schemes, like zip, followed by a data format scheme, e.g. zarr2, zarr3, precomputed, n5, etc.  There is also format auto-detection, which can add necessary kvstore adapter schemes and the final data format scheme automatically.  For example, if you type just `gs://bucket/path/to/file.zip` then it will first get completed to `gs://bucket/path/to/file.zip|zip:` (based on the content, not the filename) and then if there is a zarr array or group at the root within the zip file it will get further completed to `gs://bucket/path/to/file.zip|zip:|zarr3:`.

Note that whether some part of the path goes before or after the `|zarr:` currently doesn't matter, but if a group storage transformer were used, it would matter.  I'm planning to normalize urls so that the outer kvstore url points to the topmost valid zarr v3 group.  E.g. if we normalize to `gs://bucket/path/to/group/|zarr3:path/to/array/` then that means:
- gs://bucket/path/to/zarr.json does NOT exist, but
- gs://bucket/path/to/group/zarr.json does exist, and
- gs://bucket/path/to/group/path/zarr.json does exist and
- gs://bucket/path/to/group/path/to/zarr.json does exist and
- gs://bucket/path/to/group/path/to/array/zarr.json does exist

With this background out of the way, for icechunk, there are two questions:
1. How to encode the branch / tag / snapshot in the URL
2. Whether to treat icechunk as a key-value store adapter, or as a final data format in place of zarr3.

Some possible options:

gs://bucket/path/to/icechunk_repo/|icechunk:branch.main/path/to/array/|
gs://bucket/path/to/icechunk_repo/|icechunk:branch.main@path/to/array/|
gs://bucket/path/to/icechunk_repo/|icechunk:branch.main|zarr3:path/to/array/|

gs://bucket/path/to/icechunk_repo/refs/branch.main/|icechunk:path/to/array/|
gs://bucket/path/to/icechunk_repo/refs/branch.main/|icechunk:|zarr3:path/to/array/|

Neuroglancer needs a URL syntax to support it at all, unlike e.g. zarr-python, but it would be nice to standardize on a syntax that will also be supported by other tools in the future.

Choosing a URL syntax that includes a final |zarr3: component most closely corresponds to the current zarr-python integration where icechunk just behaves as a key-value store and translates the metadata and chunks back to the standard zarr v3 metadata encoding and key encoding.   I think the right choice, though, depends on how you expect to evolve icechunk in the future.

jhamman:
Combined with https://github.com/zarr-developers/zarr-python/pull/3369, this allows Icechunk stores to be specified with a single string and passed to Zarr-Python and/or Xarray like this:

```python
In [1]: import xarray as xr

In [2]: xr.open_zarr('s3://icechunk-public-data/v1/glad|icechunk:')
```

TODO: 

- [ ] merge upstream PR in Zarr - https://github.com/zarr-developers/zarr-python/pull/3369
- [ ] write docs for this

closes #576

### DIFF
--- a/icechunk-python/python/icechunk/__init__.py
+++ b/icechunk-python/python/icechunk/__init__.py
@@ -223,20 +223,3 @@ ManifestSplittingConfig.from_dict = from_dict  # type: ignore[method-assign]
 ManifestSplittingConfig.to_dict = to_dict  # type: ignore[attr-defined]
 
 initialize_logs()
-
-
-# Auto-register ZEP 8 adapters for zarr URL support
-def _register_zarr_adapters():
-    """Auto-register icechunk adapters when the package is imported."""
-    try:
-        from icechunk.zarr_adapter import IcechunkStoreAdapter, ICStoreAdapter
-        from zarr.registry import register_store_adapter
-
-        register_store_adapter(IcechunkStoreAdapter)
-        register_store_adapter(ICStoreAdapter)
-    except ImportError:
-        # zarr not available or version doesn't support ZEP 8
-        pass
-
-
-_register_zarr_adapters()
--- a/icechunk-python/python/icechunk/zarr_adapter.py
+++ b/icechunk-python/python/icechunk/zarr_adapter.py
@@ -21,7 +21,7 @@ from icechunk import (
     local_filesystem_storage,
     s3_storage,
 )
-from zarr.abc.store_adapter import StoreAdapter  # type: ignore[import-untyped]
+from zarr.abc.store_adapter import StoreAdapter
 
 if TYPE_CHECKING:
     from zarr.abc.store_adapter import URLSegment
@@ -277,7 +277,7 @@ async def _create_icechunk_store(repo: Repository, segment_path: str) -> Icechun
 # =============================================================================
 
 
-class IcechunkStoreAdapter(StoreAdapter):  # type: ignore[misc]
+class IcechunkStoreAdapter(StoreAdapter):
     """Store adapter for Icechunk repositories in ZEP 8 URL chains."""
 
     adapter_name = "icechunk"


icechunk-python/python/icechunk/__init__.py
```diff
@@ -215,3 +223,20 @@ def to_dict(config: ManifestSplittingConfig) -> SplitSizesDict:
...
+        from zarr.registry import register_store_adapter
+
+        register_store_adapter(IcechunkStoreAdapter)
+        register_store_adapter(ICStoreAdapter)
+    except ImportError:
+        # zarr not available or version doesn't support ZEP 8
+        pass
+
+
+_register_zarr_adapters()
```
jhamman, line 242: probably don't need this

jbusecke:
Hey @jhamman this looks really cool and I am very keen to test drive this. I was looking at the code and it seems for now the assumption that credentials are taken from the environment is [hardcoded](https://github.com/earth-mover/icechunk/blob/37508c5cb74731b7c96e465ed91c21eebd86f837/icechunk-python/python/icechunk/zarr_adapter.py#L234-L256)? 

Do you think there will be a way to use the [`storage_options`](https://github.com/zarr-developers/zarr-python/blob/6a546d639e263fe9a5e11740f667cc5db0115da6/src/zarr/api/synchronous.py#L188-L190) kwarg in `zarr.open` to pass other credentials, an in particular virtual container credentials?

jbms:
Note: I have moved the URL pipeline proposal over from the ZEP repo to a separate repository:

https://github.com/jbms/url-pipeline

PR #1166, opened 2025-08-13
Introduce an HTTP concurrency limiter per repo

paraseba:
This PR is in no way final, or perfect. It adds a semaphore to the AssetManager, which will limit concurrency per repo to a certain configurable value (default 256 concurrent requests).

Known ways in which this is not great:
* There is a minor source of extra concurrency: large objects can be further devided in smaller requests by the `Storage` implementation. This is solvable by having the Storage share the same semaphore.
* The semaphore doesn't survive serialization. If the repo is serialized now there are two semaphores and double the max requests. Users should half the permitted max manually.
* Limits are per repo, users having multiple repos open at once will see more requests (this is again solvable by users by lowering the allowed max, but not ideal)

### DIFF
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -9,6 +9,7 @@ use std::{
     ops::Range,
     sync::{Arc, atomic::AtomicBool},
 };
+use tokio::sync::Semaphore;
 use tracing::{Span, debug, instrument, trace, warn};
 
 use crate::{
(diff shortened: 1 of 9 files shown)


rabernat (approved):
Fixes the I/O error: streaming error in my high-throughput test.

icechunk/src/asset_manager.rs
```diff
@@ -338,10 +369,12 @@ impl AssetManager {
             Ok(chunk) => Ok(chunk),
             Err(guard) => {
                 trace!(%chunk_id, ?range, "Downloading chunk");
+                let permit = self.request_semaphore.acquire().await?;
                 let chunk = self
                     .storage
                     .fetch_chunk(&self.storage_settings, chunk_id, range)
                     .await?;
+                drop(permit);
```
dcherian, line 377: nice catch!

dcherian (approved):

paraseba:
Before merging I'd like to add a tiny test to this PR, that verifies the config can be set/get

PR #1170, opened 2025-08-13
New knob for stream stall detection on S3 storage

paraseba:
Affects: `s3_storage`, `s3_store`, `tigris_storage`, `r2_storage`

They all now support a new `network_stream_timeout_seconds` argument with default 20. If set to 0 stream stall detection is deactivated.


rabernat:
Let's add documentation about this option to https://icechunk.io/en/latest/performance/

Specifically, let's explain what users should do if they see the "stream stalled" error. They should either:
- Reduce concurrency (how?) or
- Disable the stalled stream protection completely (with what risks?)

Issue #1172, opened 2025-08-14
Split up top-level python namespace into modules


TomNicholas:
### Problem

The top-level python namespace `icechunk` is really polluted, which makes it harder to navigate the library. There's [~80 things](https://github.com/earth-mover/icechunk/blob/main/icechunk-python/python/icechunk/__init__.py) in it. For example, both the ubiquitous class `Repository` and the extremely niche class `ManifestSplitDimCondition` are in the top-level `icechunk` namespace.

### Solution

It would be better practice to split this up into sub-modules. That would improve the user experience in a few ways:
- auto-complete would be more helpful
- the API docs would be easier to navigate (see also #1095)
- the key classes would be the only ones in the top-level namespace (e.g. `Repo`, `Session`, `IcechunkStore`), making it more immediately obvious how icechunk is structured
- the words "store"/"storage" would become a bit less overloaded (e.g. `icechunk.Storage`, `icechunk.s3_store`, and `icechunk.IcechunkStore` are all currently in the same namespace, but mean totally different things)

# Proposal

We actually already have this splitting by module internally, with modules
- `credentials`
- `dask`
- `distributed`
- `repository`
- `session`
- `storage`
- `store`
- `xarray`

But everything is imported again in the [top-level `__init__`](https://github.com/earth-mover/icechunk/blob/main/icechunk-python/python/icechunk/__init__.py), so e.g. `AzureCredentials` is currently available as both `icechunk.AzureCredentials` and as `icechunk.credentials.AzureCredentials`.

I think we should move things behind dedicated namespaces, starting with the list above. We might also want to add
- `virtual` (for `VirtualChunkContainers` & `VirtualChunkSpec`)
- `config`
- `conflict`
- `manifest` (for the many options related to manifest splitting)
- maybe others.

### Backwards compatibility

We could make this change in a backwards-compatible fashion, by having a deprecation cycle, where we first throw a warning suggesting what you should change the import to. We could leave this up for a really long time. We can also probably get the API docs to display the submodules first and ignore the top-level imports, even whilst they are still available.

a-nu-ragn:
Hi @TomNicholas I'd like to work on this. Is this only specific some people or to anyone? I worked on [kerchunk](https://github.com/fsspec/kerchunk) as a part of GSoC.

paraseba:
[...]

PR #1179, opened 2025-08-18
Design doc for refactoring python API namespaces

TomNicholas:
Towards #1172

### DIFF
--- a/design-docs/012-python-namespaces
+++ b/design-docs/012-python-namespaces
@@ -95,8 +95,6 @@ Existing namespaces, with new members added:
 - `distributed`
   - `distributed.extract_sessions`
   - `distributed.merge_sessions`
-- `garbage`
-  - `garbage.GCSummary`
 - `repository`
   - `repository.Repository`
 - `session`
@@ -134,6 +132,8 @@ New namespaces:
   - `exceptions.ConflictError`
   - `exceptions.IcechunkError`
   - `exceptions.RebaseFailedError`
+- `garbage`
+  - `garbage.GCSummary`
 - `manifests`
   - `manifests.ManifestConfig`
   - `manifests.ManifestFileInfo`


design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+```python
+import icechunk
+storage = icechunk.Storage.s3_storage(bucket="my-bucket", prefix="my-prefix", from_env=True)
+repo = icechunk.Repository.create(storage)
+```
+
+### Top-level `icechunk` module:
+
+- `icechunk.Repository` (also available in `icechunk.repository.Repository`)
+- `icechunk.IcechunkStore` (also available in `icechunk.store.IcechunkStore`)
```
paraseba: Do people need to use `IcechunkStore` directly? Can be a good candidate to hide in a module

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+  - `storage.local_filesystem_storage`
+  - `storage.local_filesystem_store`
+  - `storage.r2_storage`
+  - `storage.s3_storage`
+  - `storage.s3_store`
+  - `storage.tigris_storage`
+- `store`
+  - `store.IcechunkStore`
+- `xarray`
+  - `xarray.to_icechunk`
```
paraseba, line 112: I wonder if `to_icechunk` should be at the top level.

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+storage = icechunk.Storage.s3_storage(bucket="my-bucket", prefix="my-prefix", from_env=True)
+repo = icechunk.Repository.create(storage)
+```
+
+### Top-level `icechunk` module:
+
+- `icechunk.Repository` (also available in `icechunk.repository.Repository`)
+- `icechunk.IcechunkStore` (also available in `icechunk.store.IcechunkStore`)
+- `icechunk.spec_version`
+- `icechunk.print_debug_info`
```
paraseba, line 70: `set_logs_filter` should probably be at the top level

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+- `conflict`
+  - `conflict.BasicConflictSolver`
+  - `conflict.Conflict`
+  - `conflict.ConflictDetector`
+  - `conflict.ConflictSolver`
+  - `conflict.ConflictType`
+  - `conflict.VersionSelection`
+- `exceptions`
+  - `exceptions.ConflictError`
+  - `exceptions.IcechunkError`
```
paraseba, line 133: maybe `IcechunkError` at the top level

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+  - `exceptions.RebaseFailedError`
+- `logging`
+  - `logging.initialize_logs`,
+  - `logging.set_logs_filter`,
+- `manifests`
```
paraseba, line 137: All of these should probably go in the `config` module

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+  - `snapshots.SnapshotInfo`
+  - `snapshots.Diff`
+- `virtual`
+  - `virtual.VirtualChunkContainer`
+  - `virtual.VirtualChunkSpec`
+
+## Outstanding questions
+
+- It's generally a bit confusing that there are multiple ways to make most classes, e.g. `local_filesystem_store` and `Storage.new_local_filesystem`. Do we really need both?
+  - If we only kept `Storage` we could get rid of a _lot_ of functions in favour of methods...
```
paraseba, line 155: Not against hiding

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+  - Alternatively should base classes that aren't really meant to be called by the user (e.g. `Storage`) be public API at all?
```
paraseba, line 156: `Storage` can still be useful for typing

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+- What's the difference between `storage.local_filesystem_storage` and `storage.local_filesystem_store`?
+  - Similarly how can we better disambiguate `storage.s3_storage` and `storage.s3_store`?
```
paraseba, line 158: we probably need better names. `Storage` is what you pass to create/open repos. `s3_store` is a way to indicate the target of a `VirtualChunkContainer`

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+- Is there a similar question for credentials?
```
paraseba, line 159: Credentials are currently a mess

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+- Should cloud-provider-specific classes be further grouped together within `.storage`/`.credentials`?
+- Should all configs live together?
+  - Should `StorageSettings` really be a `StorageConfig`? Should it live with configs or in `storage`?
+  - Similarly should `S3Options` really be `S3Config`?
+- Should summary dataclasses such as `Diff` and `GCSummary` live in the top-level?
+- Is `distributed` meant to be a public module at all? Why is it's API defined in a different place? Should it be merged with `dask`?
+- Union types like `AnyCredential` are re-exported as top-level API for importing, but it seems like they are only useful for internal type-hinting purposes?
```
paraseba, line 166: Only for typing, but not sure if they are needed publicly or only privately

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+
+## Implementation steps (including deprecation cycle)
+
+We need to implement this in multiple steps, including a deprecation cycle which maintains backwards-compatibility until the cycle is complete.
+
+1. **Decide on public API structure (i.e. review then merge this design doc)**
+
+2. **Move everything into desired locations whilst keeping all top-level imports**
+
+We should add tests that both import paths work.
```
paraseba, line 176: Yes please, particularly we need to add tests that verify we are not breaking user code

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+3. **Change how docs render to display using fine-grained namespaces instead of single top-level namespace**
+
+We can probably configure the API docs to display the submodules first and ignore the top-level imports, even whilst the top-level imports are still available.
+
+4. **Add deprecation warning pointing users to specific fine-grained import**
+
+User experience should look like this:
+
+```python
+# works, but emits a warning
```
paraseba, line 187: I think we should use Icechunk 2 as the moment to make the change and start warning people about deprecation.

design-docs/012-python-namespaces
TomNicholas: I also wondered this. Maybe useful for type checking? I also think this could be hidden in `icechunk.store.IcechunkStore`.

design-docs/012-python-namespaces
TomNicholas: We might want to leave open the possibility of adding API for serializing non-xarray things `to_icechunk` though...

design-docs/012-python-namespaces
TomNicholas: i.e. you would support getting rid of all the functions-as-constructors in favour of the methods-as-constructors?

design-docs/012-python-namespaces
TomNicholas: In user code though? Also we could separately have a `typing` module...

design-docs/012-python-namespaces
TomNicholas: Okay yeah that's really confusing

design-docs/012-python-namespaces
TomNicholas: If they are only needed for typing, and you don't expect users to make great use of them for typing, I think they should be private.

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,216 @@
...
+- Modules with only one member are fine if they add clarity.
+- Adding an extra intermediate namespace doesn't make things much harder for users to find. In other words this:
+
+```python
+import icechunk
+storage = icechunk.s3_storage(bucket="my-bucket", prefix="my-prefix", from_env=True)
+repo = icechunk.Repository.create(storage)
+```
+
+is not significantly harder than this:
```
TomNicholas:
```suggestion
is not significantly easier than this:
```

design-docs/012-python-namespaces
paraseba: yeah, let's move it. Even typechecking would be rare, because you can use Zarr's store

design-docs/012-python-namespaces
paraseba: Storage instances get passed around quite a bit, because they are what you need to open/create a repo. So, in user code, it's common to have functions that take and return Storage instances.

Having a separate typing module sounds interesting.

design-docs/012-python-namespaces
dcherian: I agree. I expect ConflictError  to be used more regularly so it should be top-level

design-docs/012-python-namespaces
TomNicholas: Just noting that this is a whole separate deprecation cycle

design-docs/012-python-namespaces
TomNicholas: Removed in https://github.com/earth-mover/icechunk/pull/1179/commits/db4671cf39c2a2ba92df33152df409059ebc6daf.

Note that this makes it clearer that the only zarr-related part of icechunk is that one class...

design-docs/012-python-namespaces
TomNicholas: I think either all exceptions should be top-level or none of them

design-docs/012-python-namespaces
TomNicholas: > Storage instances get passed around quite a bit, because they are what you need to open/create a repo. So, in user code, it's common to have functions that take and return Storage instances.

:+1:

Sounds like we're leaning towards keeping `Storage` and getting rid of the functions-as-constructors, which would mean we can still use `Storage` for typing.

design-docs/012-python-namespaces
TomNicholas: You mean the manifest splitting classes? They also seem like they should be grouped together though... They could actually go even deeper, i.e. `icechunk.config.manifests.ManifestConfig`

design-docs/012-python-namespaces
TomNicholas: Should `initialize_logs` also go in the top-level?

design-docs/012-python-namespaces
TomNicholas: Tracked separately in #1182

design-docs/012-python-namespaces
TomNicholas: See #1182

design-docs/012-python-namespaces
TomNicholas: I sketched an alternative proposal in https://github.com/earth-mover/icechunk/issues/1182

design-docs/012-python-namespaces
TomNicholas: These deserve their own separate issue. The class hierarchy is complicated so it's not immediately obvious to me how to simplify the API.

TomNicholas:
@paraseba there are still a few unresolved question in this design doc, when you have a moment.

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,213 @@
...
+  - `credentials.s3_refreshable_credentials`
+  - `credentials.s3_static_credentials`
+- `dask`
+  - `dask.computing_meta`
+  - `dask.store_dask`
+- `distributed`
+  - `distributed.extract_sessions`
+  - `distributed.merge_sessions`
+- `garbage`
+  - `garbage.GCSummary`
```
a-nu-ragn: There is no such module **garbage** in icechunk!!

design-docs/012-python-namespaces
```diff
@@ -0,0 +1,213 @@
...
+- `icechunk.spec_version`
+- `icechunk.print_debug_info`
+- `icechunk.__version__`
+
+### Sub-modules
+
+Existing namespaces, with new members added:
+
+- `credentials`
+  - `credentials.Credentials`
```
a-nu-ragn, line 78: As per this doc, the classes and functions from the  existing sub-modules would be exported from the module itself using `__all__` variable instead of the top-level `icechunk` package?

design-docs/012-python-namespaces
TomNicholas: Oh yes you're right. Let's make one then. I'll edit this design doc to move `garbage` into another section.

design-docs/012-python-namespaces
TomNicholas: For now, we actually want both: exported from the module itself, and then re-imported from the sub-module into the top-level `__init__.py`. That way your initial PR doesn't change the user experience, but does define a clearer underlying structure of sub-modules. So your current PR is for step 2 as described at the end of this doc.

a-nu-ragn: So for now only add the new namespaces to `icechunk`?

TomNicholas: We currently have this:

```python
# __init__.py

from icechunk._icechunk_python import Credentials

__all__ = [
    "Credentials",
]
```

```python
# icechunk.icechunk_python.pyi

class Credentials
    ...
```

Which only allows you to import via `from icechunk import Credentials`.


What we want is this:

```python
# __init__.py

from icechunk.credentials import Credentials

__all__ = [
    "Credentials",
]
```

```python
# icechunk.credentials.py

from icechunk._icechunk_python import Credentials

__all__ = [
    "Credentials",
]
```

```python
# icechunk_python.pyi

class Credentials
    ...
```

This would enable importing via `from icechunk.credentials import Credentials` (i.e. the end state we want) whilst for now still allowing importing via `from icechunk import Credentials` (i.e. backwards compatibility), and requires no changes to the part of the python code that interacts with rust (i.e. separates API from implementation).

PR #1186, opened 2025-08-20
Update virtual dataset documentation

laurengulland:
Draft PR with some small docs updates as we're following the virtual datasets documentation and finding/correcting issues!

[icechunk's contributing guide here](https://icechunk.io/en/latest/contributing/)

Edit: just realized some of this overlaps with https://github.com/earth-mover/icechunk/pull/1099, but a couple small things are new, so keeping this open for the time being -- if we decide to take this out of draft mode once we're doing going through all of this we'll make sure it's deduplicated with that PR (and possibly just even make this a PR against that branch)!

cc @mosegontar


docs/docs/virtual.md
```diff
@@ -38,20 +39,31 @@ oisst_files = sorted(['s3://'+f for f in oisst_files])
...
 
 ```python
+from obstore.store import from_url
+from virtualizarr.parsers import HDFParser
+from virtualizarr.registry import ObjectStoreRegistry
 
-virtual_datasets =[
-    open_virtual_dataset(url, indexes={})
+store = from_url(bucket_name, region=bucket_region, skip_signature=True)
```
laurengulland, line 55: todo: I forgot to copy the bucket region, but it's also `us-1-east` for their example oisst data (see https://registry.opendata.aws/noaa-cdr-oceanic/ for confirmation)

paraseba:
Thank you @laurengulland

TomNicholas:
Thanks @laurengulland ! I finally merged #1099, and incorporated your suggestions there.

Issue #1181, opened 2025-08-20
deserialization error when using distributed dask and virtualizarr


alxthm:
Hi,

I am trying to process a large dataset (NASA NEX GDDP CMIP6, 36TB of `.nc` files) in order to compute some climate indices.

Icechunk combined with virtualizarr seems a great way of doing this (thanks a lot for the work you are doing in open-source btw, it's exciting to see all these new capabilities !). In principle I could build virtual references with `vz.open_virtual_dataset`, save them in an icechunk store, and work with xarray and dask to manipulate the data, without having to copy and transform the 36TB of original data to a "proper" zarr store.

The 1st step (creating the virtual datatree / dataset and saving it to an icechunk store) works well. I am also able to load and manipulate the data from the store (plot or compute indicators on slices of the data).

I am encountering issues when trying to do distributed computation on the data with dask: as soon as I instantiate a `distributed.Client(cluster)` (even with a LocalCluster), dask fails when trying to deserialize the computation graph.

```
lib/python3.12/site-packages/icechunk/store.py:88, in __setstate__()
     87 store_repr = state["_store"]
---> 88 state["_store"] = PyStore.from_bytes(store_repr)
     89 self.__dict__ = state

ValueError: Failed to deserialize store from bytes: deserialization error

...

The above exception was the direct cause of the following exception:

...
RuntimeError: Error during deserialization of the task graph. This frequently
occurs if the Scheduler and Client have different environments.
For more information, see
https://docs.dask.org/en/stable/deployment-considerations.html#consistent-software-environments
```

I am suspecting this is linked to the virtualizarr-based icechunk store, since I was able to reproduce your [Icechunk + Dask + Xarray example](https://icechunk.io/en/latest/dask/#icechunk-dask) without issues.

I made a [minimal reproduction example notebook](https://gist.github.com/alxthm/90f6a76d458eb64be4e56f95106f18ad).

Any idea why this error arises ? Apologies in advance if I misconfigured or misunderstood something !

Thanks

paraseba:
Thanks for reporting this @alxthm . It looks like a bug with Icechunk Session [de]serialization. Let me look into it later today and I'll get back to you.

paraseba:
[This](https://github.com/3Hren/msgpack-rust/issues/86) is the upstream issue. We'll need to find a way around it

paraseba:
[...]

PR #1187, opened 2025-08-20
Fix config serialization

paraseba:
We were failing to deserialize in certain situations when there are virtual chunk containers.

This change breaks compatibility for users who wrote config with virtual chunk containers to their repos. 

Those users will need to edit the config.yaml file adding 

```
name: null
```

to their virtual chunk containers.

Closes: #1181

### DIFF
--- a/icechunk/src/strategies.rs
+++ b/icechunk/src/strategies.rs
@@ -1,30 +1,38 @@
+#![allow(clippy::unwrap_used, clippy::expect_used, clippy::panic)]
 use std::collections::HashMap;
-use std::num::NonZeroU64;
-use std::ops::Range;
+use std::num::{NonZeroU16, NonZeroU64};
+use std::ops::{Bound, Range};
+use std::path::PathBuf;
 
 use prop::string::string_regex;
 use proptest::prelude::*;
 use proptest::{collection::vec, option, strategy::Strategy};
 
-use crate::Repository;
+use crate::config::{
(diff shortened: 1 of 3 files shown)


icechunk/src/strategies.rs
```diff
@@ -109,29 +114,265 @@ pub fn manifest_extents(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
...
+        s3_options().prop_map(S3Compatible),
+        s3_options().prop_map(Tigris),
+        any::<HashMap<String, String>>().prop_map(Gcs),
+        any::<HashMap<String, String>>().prop_map(Http),
+        any::<HashMap<String, String>>().prop_map(Azure),
+    ]
+    .boxed()
+}
+
+pub fn bound<T>(inner: impl Strategy<Value = T>) -> impl Strategy<Value = Bound<T>>
```
dcherian, line 205: nice!

paraseba:
> LGTM!
> 
> For some extra coverage at the python end we could add a pickle/unpickle of repo config here: https://github.com/earth-mover/icechunk/blob/main/icechunk-python/tests/test_zarr/test_stateful.py, though that only exercises the splitting config.

Love it, let me add.

dcherian:
> Those users will need to edit the config.yaml file adding

are we able to catch the error from serde and raise a nicer error for the user?

PR #1189, opened 2025-08-21
Added new namespaces for icechunk

### LINKED ISSUE #1172: Split up top-level python namespace into modules

Anu-Ra-g:
This PR is part of for the fix #1172. I've added new namespaces in this PR as per this doc: #1179


.github/workflows/rust-ci.yaml
TomNicholas, line 1: None of the changes to this file should be in this PR - somethings gone wrong with your usage of git here 😕

icechunk-python/python/icechunk/config.py
```diff
@@ -0,0 +1,9 @@
+from icechunk._icechunk_python import (
+    CachingConfig,
+    CompressionConfig,
+    CompressionAlgorithm,
+    ObjectStoreConfig,
+    RepositoryConfig,
+    initialize_logs,
+    set_logs_filter
+)
```
TomNicholas: You should also add the names to python's `__all__` variable in each file.

TomNicholas (commented):
Thanks for starting work on this!

We should alter the top-level imports in `__init__.py` to import from these new namespaces.

.github/workflows/rust-ci.yaml
a-nu-ragn: I mentioned this issue [here](https://github.com/earth-mover/icechunk/issues/1172#issuecomment-3209303282). I made the PR to the `icechunk2` branch and not the `main` branch like @paraseba  asked.

a-nu-ragn: I tried to raise the PR again with a fresh new fork, but the issue was still there.

a-nu-ragn:
The usage of `pyi` file was a little new for me. The types mentioned there are for the Rust code?

TomNicholas:
> The types mentioned there are for the Rust code?

Yes I believe they are auto-generated from the rust code.

.github/workflows/rust-ci.yaml
a-nu-ragn: @TomNicholas Should I raise PR on the main branch then?

TomNicholas: No, definitely this should be against the icechunk2 branch. The question is how @paraseba's commit ended up there too.

a-nu-ragn: image

I tried to raise a rough PR but still the same issue. This is new for me on Github.

paraseba:
I bet @Anu-Ra-g started with a checkout of `main` instead of `icechunk2`. I just merged `main` into the branch. You should try rebasing your fork @Anu-Ra-g.

Re: .pyi file, unfortunately, we have to maintain this file manually, which is a pain

a-nu-ragn:
> I bet @Anu-Ra-g started with a checkout of `main` instead of `icechunk2`. I just merged `main` into the branch. You should try rebasing your fork @Anu-Ra-g.
> 
> Re: .pyi file, unfortunately, we have to maintain this file manually, which is a pain

Yes, I actually branched out from `main` and then raised a PR against `icechunk2` :sweat_smile: . I fixed it.

a-nu-ragn:
@TomNicholas I've the `__all__` variable for existing namespaces. Can you please check?

icechunk-python/python/icechunk/credentials.py
```diff
@@ -399,3 +399,21 @@ def containers_credentials(
         else:
             raise ValueError(f"Unknown credential type {type(cred)}")
     return res
+
+
+__all__ = [
```
TomNicholas: `__all__` conventionally should go at the top of the file, just underneath the imports.

a-nu-ragn: Done

a-nu-ragn:
@TomNicholas Can you please review and suggest the updates?

a-nu-ragn:
@TomNicholas Can you help me out with this CI check? Its failing and I can't tell why?

paraseba:
Don't worry about the Rust failures @Anu-Ra-g . Those are expected because you don't have permission to run with secrets. But the python lint issues are real

a-nu-ragn:
@paraseba python checks are part of the `pre-commit` config?

paraseba:
> @paraseba python checks are part of the `pre-commit` config?

If you install `just`, you can run `just pre-commit-python` locally.

a-nu-ragn:
`pre-commit-config` is at the root level. `just` should be installed as part of the virtual environment or by default as part of my system? Why `just` is not part of the `pyproject.toml` file? I actually confused as contributing guide doesn't specify these !!

TomNicholas:
Sorry for forgetting about this @Anu-Ra-g !

The branch deletion here is because of some work @paraseba is doing on Icechunk 2.0. Are you still interested in contributing here?

a-nu-ragn:
Sorry @TomNicholas I got busy and totally forgot about this. This branch is deleted, right? Yes, I'd like to continue with this. Should I make a new branch?

TomNicholas:
> Yes, I'd like to continue with this.

Great!

> This branch is deleted, right?

I believe it was deleted in favour of `main`. You probably either want to rebase your PR branch onto `main` or `cherry-pick` your commits into a new branch.

PR #1203, opened 2025-09-02
QOL: add a warnign if you slightly misspell icechunk for logs

### LINKED ISSUE #1202: Catch small spelling errors when setting logs
### Problem
It's easy to mispell `icehunk=trace` (in fact typing this issue I literally did mispell it there...) `icechunk=trace`

### Proposed Solution
We should catch mispellings close to this and warn.

ianhi:
Close #1202

Compare name in the logs_filter using levenstein distance 
Example output:
```
Warning: Did you mean 'icechunk' instead of 'icehunk' in log filter?
```

but it also gets colorcodes:

so it will look like this:

image

### DIFF
--- a/icechunk-python/Cargo.toml
+++ b/icechunk-python/Cargo.toml
@@ -43,6 +43,7 @@ clap = { version = "4.5", features = ["derive"], optional = true }
 rand = "0.9.2"
 rmp-serde = "1.3.0"
 pyo3-bytes = "0.2.0"
+strsim = "0.11"
 
 [features]
 cli = ["clap", "icechunk/cli"]
--- a/icechunk-python/src/lib.rs
+++ b/icechunk-python/src/lib.rs
@@ -9,6 +9,8 @@ mod streams;
 
 use std::env;
 
+use strsim;
+
 use config::{
     PyAzureCredentials, PyAzureStaticCredentials, PyCachingConfig,
     PyCompressionAlgorithm, PyCompressionConfig, PyCredentials, PyGcsBearerCredential,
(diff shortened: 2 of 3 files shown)


ianhi:
draft until i can get the tests fixed

icechunk-python/src/repository.rs
```diff
@@ -591,10 +591,7 @@ impl PyRepository {
     }
 
     #[staticmethod]
-    fn exists_async<'py>(
-        py: Python<'py>,
-        storage: PyStorage,
```
ianhi, line 596: changes in the file only because pre-commit was bitterly complaining if i didn't also change this. No functionality changes here as far as I can tell

ianhi:
Ok the only failing test is now unrelated. I also had to update a dependency to fix the cargo deny failures.

Interestingly those failures were also related to ansi sequences from user input being a security risk. But I think that as done now this is ok

ianhi:
Turns out we can just santize user input using a crate we already have and keep ourselves safe.

icechunk-python/src/repository.rs
paraseba: you may have a too old/too new Rust version

icechunk-python/Cargo.toml
```diff
@@ -43,6 +43,7 @@ clap = { version = "4.5", features = ["derive"], optional = true }
 rand = "0.9.2"
 rmp-serde = "1.3.0"
 pyo3-bytes = "0.2.0"
+strsim = "0.11"
```
paraseba, line 46: so we no longer need this?

icechunk-python/src/lib.rs
```diff
@@ -88,10 +88,61 @@ fn initialize_logs(py: Python) -> PyResult<()> {
     Ok(())
 }
 
+fn extract_module_names(filter: &str) -> Vec<&str> {
+    filter
+        .split(',')
+        .map(|part| part.trim())
+        .filter_map(|part| {
```
paraseba, line 94: this code is a bit weird ... there is no much point in calling map followed by filter_map followed by filter, it could all be combined in a single filter_map

icechunk-python/src/lib.rs
```diff
@@ -88,10 +88,61 @@ fn initialize_logs(py: Python) -> PyResult<()> {
     Ok(())
 }
 
+fn extract_module_names(filter: &str) -> Vec<&str> {
```
paraseba: this should return an iterator ... something like `impl Iterator<Item=&str>`, by skipping the `collect` call

icechunk-python/src/lib.rs
```diff
@@ -88,10 +88,61 @@ fn initialize_logs(py: Python) -> PyResult<()> {
...
+            } else {
+                None
+            }
+        })
+        .filter(|module| !module.is_empty())
+        .collect()
+}
+
+fn is_likely_icechunk_misspelling(module: &str) -> bool {
+    const GENERIC_LOG_LEVELS: &[&str] = &["trace", "debug", "info", "warn", "error"];
```
paraseba: I feel it would be much better to only analyze the segments that have indeed a module name, instead of matching against log levels. That way we don't have to think about corner cases, like modules named `error` (pretty common)

icechunk-python/src/lib.rs
```diff
@@ -88,10 +88,61 @@ fn initialize_logs(py: Python) -> PyResult<()> {
...
+    for module in modules {
+        if is_likely_icechunk_misspelling(module) {
+            let misspelled = module.split("::").next().unwrap_or(module);
+            // Sanitize the user input to prevent ANSI injection attacks
+            let sanitized =
+                misspelled.chars().filter(|c| !c.is_control()).collect::<String>();
+
+            eprintln!(
+                "\x1b[93m\x1b[1mWarning\x1b[0m: Did you mean '\x1b[92m\x1b[1micechunk\x1b[0m' instead of '\x1b[91m{}\x1b[0m' in log filter?",
+                sanitized
```
paraseba, line 113: we should only warn once

icechunk-python/src/lib.rs
```diff
@@ -88,10 +88,61 @@ fn initialize_logs(py: Python) -> PyResult<()> {
...
+
+    for module in modules {
+        if is_likely_icechunk_misspelling(module) {
+            let misspelled = module.split("::").next().unwrap_or(module);
+            // Sanitize the user input to prevent ANSI injection attacks
+            let sanitized =
+                misspelled.chars().filter(|c| !c.is_control()).collect::<String>();
+
+            eprintln!(
+                "\x1b[93m\x1b[1mWarning\x1b[0m: Did you mean '\x1b[92m\x1b[1micechunk\x1b[0m' instead of '\x1b[91m{}\x1b[0m' in log filter?",
```
paraseba: Let's not do colors. Otherwise, we need to respect the NO_COLOR variable, verify terminal type, etc

icechunk-python/Cargo.toml
ianhi: > Turns out we can just santize user input using a crate we already have and keep ourselves safe.


I think you're refereing to the above. That was about sanitizing ANSI control characters. The `strstim` is for the distance metric. WHich I suppose we could re-write instead as its no too complicated.

icechunk-python/src/lib.rs
ianhi: done

icechunk-python/src/lib.rs
ianhi: my rust noobiness + claudes confidence tricked me here. refactored into a single filter_map

paraseba (approved):

PR #1207, opened 2025-09-05
Split up just pre-commit to massively speed up pre-commit

ianhi:
Prior to this PR the `just` pre-commit step was very slow (45 seconds or more) which makes it very likely that contributors (and specifically me) will just skip pre-commit checks. In this PR I:

1.  add a new `just` command with only the faster steps for use in pre-commit
2. Filter to only run when the relevant files change (before even changing just the python parts of icechunk-python triggered rust checks)
3. Keep using the old (complete checks) in CI

This does check less, but I think overall it's better because the alternative with the full pre-commit checks is that people (again, me) don't run them at all. Which is worse.

## Speedup
`time pre-commit run --all-files`

**This PR:** `3.76s user 0.89s system 107% cpu 4.325 total`
**main:** `125.57s user 10.14s system 358% cpu 37.851 total`

### DIFF
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -116,7 +116,7 @@ jobs:
           TIGRIS_SECRET_ACCESS_KEY: ${{ secrets.TIGRIS_SECRET_ACCESS_KEY }}
 
         run: |
-          just pre-commit
+          just pre-commit-ci
 
       - name: Run unit tests only
         if: matrix.os != 'ubuntu-latest' && matrix.os != 'ubuntu-24.04-arm'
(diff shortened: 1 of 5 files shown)


.github/workflows/rust-ci.yaml
```diff
@@ -116,7 +116,7 @@ jobs:
           TIGRIS_SECRET_ACCESS_KEY: ${{ secrets.TIGRIS_SECRET_ACCESS_KEY }}
 
         run: |
-          just pre-commit
+          just pre-commit-ci
```
paraseba, line 119: I think you need to make the same change in other workflow files

ianhi: Nice catch. I've added it to the other file where it was used. the other uses of `just` in workflow files should be unchanged i think

paraseba (approved):

dcherian (approved):

PR #1209, opened 2025-09-06
Add Python reprs (implemented in python)

TomNicholas:
With this we get

```python
In [1]: import icechunk
   ...: 
   ...: storage = icechunk.in_memory_storage()
   ...: repo = icechunk.Repository.create(storage)

In [2]: repo.readonly_session("main")
Out[2]: 
<icechunk.Session>
read_only: True
snapshot_id: 1CECHNKREP0F1RSTCMT0

In [3]: repo.writable_session("main")
Out[3]: 
<icechunk.Session>
read_only: False
snapshot_id: 1CECHNKREP0F1RSTCMT0
branch: main
has_uncommitted_changes: False
```

For compound classes I want to have them make nested repr calls, i.e. `Repository.__repr__` includes `RepositoryConfig.__repr__` and `Storage.__repr__`, passed as "attributes". This way we should be able to have nice nested reprs for all classes be implemented simply and consistently.

Then once that works we can use the same approach but for HTML reprs.

### DIFF
--- a/icechunk-python/python/icechunk/display.py
+++ b/icechunk-python/python/icechunk/display.py
@@ -1,0 +1,35 @@
+def dataclass_repr(
+    obj: object,
+    # make this default to inspecting the cls?
+    cls_name: str,
+    attributes: list[str] = None,
+    # TODO optional indent
+) -> str:
+    """
+    Dynamically create a repr for this dataclass-like object.
+
+    Parameters
+    ----------
+    obj : object
+        Object for which to make a repr.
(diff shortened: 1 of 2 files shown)


TomNicholas:
> But the reason why I'd prefer to write it in Rust is because of all the tests we would need if we do it in python. If we change the name read_only to readonly things break, in Rust we would get a compilation error.

That's fair. Closing this in favour of the rust implementation in #1217.

PR #1217, opened 2025-09-10
Add Python reprs (implemented in rust)

### LINKED ISSUE #1208: Nicer python reprs (string and HTML)

```
```

These could be much nicer.

Thoughts?

TomNicholas:
Would close #1208, supercedes #1209.

To-do:
- [x] A non-executable repr
- [x] An executable repr
- [ ] A nested repr
- [ ] (Bonus) a nested HTML repr

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -45,14 +45,14 @@ run-all-examples:
 
 # fast pre-commit - format and lint only
 pre-commit-fast:
-  just format "--check"
+  just format
   just lint "-p icechunk -p icechunk-python"
 
 # medium pre-commit - includes compilation checks (~2-3 minutes)
 pre-commit $RUSTFLAGS="-D warnings -W unreachable-pub -W bare-trait-objects":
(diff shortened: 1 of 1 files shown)


icechunk/src/display.rs
```diff
@@ -0,0 +1,26 @@
+pub fn dataclass_repr(class_name: &str, attributes: &[(&str, &str)]) -> String {
```
TomNicholas: I used a slice here because it seemed to be a simple way to pass an _ordered_ list of str-str pairs. I first tried a `HashMap` but that doesn't preserve order which led to my reprs looking different every time!

Apparently there is an [`IndexMap`](https://docs.rs/indexmap/latest/indexmap/) thing that preserves insertion order, which sounds neat, but I didn't use that because it would have added a dependency.

TomNicholas:
superceded by #1944

PR #1218, opened 2025-09-10
Apply formatting suggestions automatically

TomNicholas:
This changes the rust formatter from merely suggesting formatting changes to always applying them. It's not very helpful to have them suggested but then have to manually apply them.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -45,14 +45,14 @@ run-all-examples:
 
 # fast pre-commit - format and lint only
 pre-commit-fast:
-  just format "--check"
+  just format
   just lint "-p icechunk -p icechunk-python"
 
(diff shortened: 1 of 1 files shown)


paraseba (approved):

TomNicholas:
Some CI tests are failing, which I don't think are related to my change? But apparently I don't have the rights to force the merge regardless.

ianhi:
@TomNicholas If you open this same PR from a branch on this repo, rather than from a fork they will pass. It's a complciation with hiding secrets from external PRs. I have a fix in the works. But for now i think it's ok for us to bypass right @paraseba ? I have the rights to make that happen if need be

paraseba:
It's weird that you don't get the secrets applied @TomNicholas . I thought people in the GH org would :shrug:

PR #1222, opened 2025-09-11
Split CI workflows for better parallelization and separation of concerns [EAR-2135]

ianhi:
## General Organization

For both the rust and python CI i have extracted the code quality checks (cargo fmt, pre-commit, etc...) into separate workflows. And the main CI pathways just run compilations and tests.

- Use `uv` as much as possible when interacting with python
- use `sccache` on the maturin build step
- The two python steps now share a single `build-wheel` step

- Fixed  MinIO/Azurite health checks in both python-check and rust-ci 
    - previously they always waited the whole time becuase they were hitting the wrong url.
    - This is 1 minute 30seconds saving for the rust-ci

## Potential future improvement
Right now the maturin build doesn't benefit at all from the fact that we are already building the icechunk rust library separately. It feels like this should be possible, but I couldn't figure it out

# Current status

I think this is basically ready. It seems to significantly speed things up. For example the python tests go from 8 minutes on a PR from before this to ~5 min on this PR. The rust ci checks are also sped up.

My one worry is that the tests that normally fail on PR due to credentials are not failing here. But I think that might be because I gained some icechunk status on github.

After merging this we will need to edit the names of the required workflows for merging. That's the cause of this

image

it's waiting for a workflow that no longer exists. We will just update that to be the "tests" or "xarray-tests" jobs

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Run clippy linting
         run: |
-          just lint -p icechunk -p icechunk-python
+          just lint --workspace
 
       - name: Run doc tests
         run: |
--- a/Justfile
+++ b/Justfile
@@ -46,22 +46,22 @@ run-all-examples:
 # fast pre-commit - format and lint only
 pre-commit-fast:
   just format
-  just lint "-p icechunk -p icechunk-python"
+  just lint "--workspace"
 
 # medium pre-commit - includes compilation checks (~2-3 minutes)
 pre-commit $RUSTFLAGS="-D warnings -W unreachable-pub -W bare-trait-objects":
   just compile-tests "--locked"
   just build
   just format
-  just lint "-p icechunk -p icechunk-python"
+  just lint "--workspace"
   just check-deps
 
 # full pre-commit for CI - runs all checks including tests
 pre-commit-ci $RUSTFLAGS="-D warnings -W unreachable-pub -W bare-trait-objects":
   just compile-tests "--locked"
   just build
-  just format
-  just lint "-p icechunk -p icechunk-python"
+  just format "--check"
+  just lint "--workspace"
   just doctest
   just test
   just run-all-examples


ianhi:
Not sure what's up with this double compilation in the build wheels:

```
   Compiling icechunk-python v1.1.5 (/home/runner/work/icechunk/icechunk/icechunk-python)
    Finished `release` profile [optimized] target(s) in 1m 49s
📦 Built wheel for CPython 3.11 to dist/icechunk-1.1.5-cp311-cp311-manylinux_2_39_x86_64.whl
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/runner/work/icechunk/icechunk/icechunk-python/Cargo.toml
workspace: /home/runner/work/icechunk/icechunk/Cargo.toml
   Compiling pyo3-build-config v0.24.2
   Compiling pyo3-ffi v0.24.2
   Compiling pyo3-macros-backend v0.24.2
   Compiling pyo3 v0.24.2
   Compiling pyo3-macros v0.24.2
   Compiling pyo3-async-runtimes v0.24.0
   Compiling pyo3-bytes v0.2.0
   Compiling icechunk-python v1.1.5 (/home/runner/work/icechunk/icechunk/icechunk-python)
```

ianhi:
[`55ebe1a` (#1222)](https://github.com/earth-mover/icechunk/pull/1222/commits/55ebe1afa0c0b6ae43c1e74d352b8fd21b4868af)

Seems to be good for a 30second time save every run. and is also a fully human brain generated commit :)

.github/workflows/code-quality.yaml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          key: ${{ env.RUST_CHANNEL }}
+
+      - name: Run formatting checks
+        run: |
+          just format --check
+
+      - name: Run clippy linting
+        run: |
+          just lint -p icechunk -p icechunk-python
```
paraseba: we should add -p icechunk-macros everywhere too. In fact, I think there is `--workspace` to do this more maintainable

ianhi: Nice call!

ianhi:
This is ready now! I will merge in the morning when I have more brainpower. So I can immediately follow up and approriately modify the branch protection rules

paraseba:
> This is ready now! I will merge in the morning when I have more brainpower. So I can immediately follow up and approriately modify the branch protection rules

perfect, lmk if you need help merging because of that pending status

ianhi:
image
 I seem to be ok!

PR #1224, opened 2025-09-11
[DRAFT] Enable sccache for faster Rust compilation in maturin builds

ianhi:
As it's own PR to keep things simpler to figure out.

Added sccache: true to all PyO3/maturin-action build steps across:
- python-ci.yaml (linux, musllinux, windows, macos, sdist jobs)
- python-check.yaml (both build steps)
- python-upstream.yaml (both build steps)

This will significantly reduce build times by caching Rust compilation artifacts.

🤖 Generated with [Claude Code](https://claude.ai/code)


dcherian:
This doesn't really help if we don't save/restore cache across runs

ianhi:
> This doesn't really help if we don't save/restore cache across runs

I think it does? Where do you see that it doesn't?

ianhi:
Main benefite incorporated into #1222

PR #1225, opened 2025-09-11
Optimize rust ci containers

ianhi:


ianhi:
Better solution in #1222

PR #1241, opened 2025-09-22
update typing of partial writes method to follow zarr

ianhi:
result of https://github.com/zarr-developers/zarr-python/pull/3413 to get mypy passing


TomNicholas:
My bad for apparently merging something unlinted here, that's causing CI to complain

PR #1242, opened 2025-09-22
improve nightly workflow open issues for mypy

ianhi:


.github/workflows/python-upstream.yaml
```diff
@@ -75,13 +75,46 @@ jobs:
...
         shell: bash
         working-directory: icechunk-python
         run: |
           set -e
           python3 -m venv .venv
           source .venv/bin/activate
-          mypy python
+          mypy python 2>&1 | tee mypy-output.log || true
+
+      - name: Create mypy failure issue
```
paraseba, line 89: can this be done for other types of failures too?

ianhi: We already do this for pytest failures. i think every other type of fialure we might have should never make it into main.

Also i'm going to test that this actually works on this PR so there may be some noise of opened issues

PR #1245, opened 2025-09-22
Speedup `list_prefix` and `getsize_prefix` when prefix is not empty

paraseba:
These two operations were very inefficient in repositories with a large number of arrays. This PR implements pre-filtering of chunks, so we only need to open manifests for the arrays that match the prefix.

Also in this PR: limit to 50 the number of preloaded array manifests. Making this number configurable is an exercise left to a future PR.

### DIFF
--- a/icechunk/src/storage/mod.rs
+++ b/icechunk/src/storage/mod.rs
@@ -903,7 +903,7 @@ pub fn new_tigris_storage(
 ) -> StorageResult<Arc<dyn Storage>> {
     let config = S3Options {
         endpoint_url: Some(
-            config.endpoint_url.unwrap_or("https://fly.storage.tigris.dev".to_string()),
+            config.endpoint_url.unwrap_or("https://t3.storage.dev".to_string()),
         ),
         ..config
     };


icechunk/src/store.rs
```diff
@@ -770,19 +768,24 @@ impl Store {
...
-                    Ok((path, chunk)) => {
-                        let chunk_key = Key::Chunk { node_path: path, coords: chunk.coord }.to_string();
-                        if is_prefix_match(&chunk_key, prefix) {
-                            yield chunk_key;
+            let session = Arc::clone(&self.session).read_owned().await;
+            for node in session.list_nodes().await? {
+                let node = node?;
+                if node.node_type() == NodeType::Array &&
+                    // FIXME: utf8 handling
+                    // skip the initial / in the path
```
dcherian, line 797: do we know that the node.path always starts with `/`?

icechunk/src/store.rs
```diff
@@ -770,19 +768,24 @@ impl Store {
...
+                    is_prefix_match(&node.path.to_string()[1..], prefix) {
+                        for await maybe_path_chunk in session.all_node_chunks(&node).await {
```
dcherian, line 799: could `try_for_each` here?

icechunk/src/store.rs
```diff
@@ -770,19 +768,24 @@ impl Store {
...
+                            match maybe_path_chunk {
+                                Ok((path, chunk)) => {
+                                    let chunk_key = Key::Chunk { node_path: path, coords: chunk.coord }.to_string();
+                                    if is_prefix_match(&chunk_key, prefix) {
```
dcherian: is this check needed given the prefiltering on node?

icechunk/src/store.rs
paraseba: Yes, Path constructor checks that and there is a `PathError::NotAbsolute`

icechunk/src/store.rs
paraseba: it's tricky because this is a stream generating other stream (see `for await`) I'm not sure if the macro can handle it. But I could try.

icechunk/src/store.rs
paraseba: your question makes me think that this may not be good enough, like, if the prefix goes beyond the array, none of this is going to work, because it won't be a prefix for the array.

As in `list_prefix('my-group/my-array/c/0')`.

I long time ago I asked about not allowing this kind of thing, but .... we should think how we want to treat it. iirc we said it was OK no be "undefined" but it's still ugly.

I kept the extra case thinking about this case of extended perfixes, but without thinking that they won't match the node

icechunk/src/store.rs
dcherian: ah i was wondering if that was the issue.

icechunk/src/store.rs
dcherian: Oh right. I think that kind of use-case is better served by `chunk_coordinates` anyway, which could expose some filtering logic. It might be OK to simply disallow a prefix that is not a node path.

rabernat: Zarr should not issue list_prefix or getsize_prefix calls for non-nodal paths.

paraseba:
@rabernat @dcherian I changed things so users get a explicit failure if they try a prefix that is not a node

icechunk/src/store.rs
```diff
@@ -770,19 +771,40 @@ impl Store {
...
+                    prefix: prefix.to_owned(),
+                    message: "Cannot convert to a path".to_string(),
+                }
+            })?;
+
+            if path != Path::root() {
+                let _ = session.get_node(&path).await.map_err(|_| {
+                    StoreErrorKind::BadKeyPrefix {
+                        prefix: prefix.to_owned(),
+                        message: "Only prefixes pointing to a group or array are allowed".to_string(),
```
dcherian, line 788: Could mention `chunk_coordinates` API here to redirect the user

PR #1248, opened 2025-09-23
ForkSession: deepcopy underlying Session objects.

dcherian:
We've been implicitly replying on the fact that users are using pickle/unpickle. By doing it explicitly we make ForkSession objects more usable, and useful.


icechunk-python/python/icechunk/session.py
```diff
@@ -387,7 +387,7 @@ def fork(self) -> "ForkSession":
                 "You should not need to fork a read-only session. Read-only sessions can be pickled and transmitted directly."
             )
         self._allow_changes = True
-        return ForkSession(self._session)
+        return ForkSession(PySession.from_bytes(self._session.as_bytes()))
```
paraseba, line 394: maybe just add a comment about why

PR #1253, opened 2025-09-24
Fix (ignore) mypy errors generated from supporting divergent xarray imports

ianhi:
closes: https://github.com/earth-mover/icechunk/issues/1262


ianhi:
The failure is the upstream xarary backend tests which were failing before. this fixes the mypy typing bot that was harrassing us with issues.

PR #1254, opened 2025-09-24
simplify dask check

ianhi:
Inspired by some pyright warnings I was getting about directly using both dask.base and importlib.util we were apparently relying on not perfectly defined import behavior https://discuss.python.org/t/python3-11-importlib-no-longer-exposes-util/25641/23

image
image

this simplifies this a bit and removes the warnings that were driving me nuts, and i think is slightly more secure against weird importing importlib errors.

### DIFF
--- a/icechunk-python/python/icechunk/xarray.py
+++ b/icechunk-python/python/icechunk/xarray.py
@@ -1,4 +1,3 @@
-import importlib
 from collections.abc import Hashable, Mapping, MutableMapping
 from dataclasses import dataclass, field
 from typing import Any, Literal, overload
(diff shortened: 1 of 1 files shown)


paraseba (approved):

dcherian:
IDK: https://docs.python.org/3/library/importlib.html#checking-if-a-module-can-be-imported

icechunk-python/python/icechunk/xarray.py
```diff
@@ -1,4 +1,3 @@
-import importlib
 from collections.abc import Hashable, Mapping, MutableMapping
```
dcherian, line 1: Would jkust explicitly importing `importlib.util` fix the errors you see?

ianhi: yes!

PR #1257, opened 2025-09-25
Speedup list_dir and list_prefix

paraseba:
This avoids going through every node each time `list_dir` is called. This is important because Zarr calls `list_dir` recursively to build `tree`.

In flat repositories with ~ 60k nodes, we verified ~15x performance improvement for `Group.tree`.


dcherian (approved):
wow!

PR #1259, opened 2025-09-26
Safer `reset_branch`

paraseba:
An optional third argument was added to `reset_branch`:

```python
def reset_branch(self, branch: str, snapshot_id: str, *, from_snapshot_id: str | None = None) -> None:
```

If `from_snapshot_id` is passed, the operation raises an exception and leaves the branch untouched if it's not currently pointing to that snapshot id.


mpiannucci (approved):
This is great

PR #1267, opened 2025-10-01
try to fix xarray-backend-test

ianhi:


ianhi:
@paraseba i think this did fix the original error. but possibly revealed some new, possibly real errors:

```
 FAILED ../../xarray::TestIcechunkRegionAuto::test_dataset_to_zarr_align_chunks_true[3-tmp_path] - TypeError: to_icechunk() got an unexpected keyword argument 'align_chunks'
FAILED ../../xarray::TestIcechunkRegionAuto::test_dataset_to_zarr_align_chunks_true[3-ZipStore] - TypeError: to_icechunk() got an unexpected keyword argument 'align_chunks'
FAILED ../../xarray::TestIcechunkRegionAuto::test_dataset_to_zarr_align_chunks_true[3-Dict] - TypeError: to_icechunk() got an unexpected keyword argument 'align_chunks'
```
Maybe one for @TomNicholas

ianhi:
Those errors are the result of a new parameter coming from xarray that we don't handle https://github.com/pydata/xarray/pull/10516

I've pushed a zeroth order guess at how to fix but please take a close look at what I did before any merging.

ianhi:
@paraseba I've updated the CI a bit, and fixed the tmp store error. 

- After that there was an error with the `align_chunks` which I think I fixed by just passing it through (https://github.com/earth-mover/icechunk/pull/1267/commits/58328eceb48bda62a75819aab1500761ed5ad371)
- Then that uses dask which exposes some errors from icechunk coming from here: https://github.com/earth-mover/icechunk/blob/f377a54dcb2ba07e1f918f19ecf677351c125abe/icechunk-python/python/icechunk/xarray.py#L294-L304 because dask is being invoked when `align_chunks=True`

maybe we should just throw away align_chunks? or ignore this test? Not totally sure. But this is the only failure in both upstream and non-upstream xarray tests.


```

=========================== short test summary info ============================
FAILED ../../xarray::TestIcechunkRegionAuto::test_dataset_to_zarr_align_chunks_true[3-tmp_path] - ValueError: Calling `to_icechunk` is not allowed on a Session with uncommitted changes. Please commit first.
FAILED ../../xarray::TestIcechunkRegionAuto::test_dataset_to_zarr_align_chunks_true[3-ZipStore] - ValueError: Calling `to_icechunk` is not allowed on a Session with uncommitted changes. Please commit first.
FAILED ../../xarray::TestIcechunkRegionAuto::test_dataset_to_zarr_align_chunks_true[3-Dict] - ValueError: Calling `to_icechunk` is not allowed on a Session with uncommitted changes. Please commit first.
```

.github/workflows/python-check.yaml
```diff
@@ -74,16 +73,18 @@ jobs:
...
         working-directory: icechunk-python
         run: |
           set -e
-          # Create virtual environment and install the built wheel
           uv venv
-          uv pip install icechunk['test'] --find-links dist
+          source .venv/bin/activate
+          python --version
+          WHEEL=$(ls dist/*-cp311-*.whl)
+          uv pip install "${WHEEL}[test]"
```
ianhi, line 87: Somewhat horrifyingly this was previously not installing the source icechunk. I'm a little alarmed I didn't catch that earlier.

I have confirmed that it does install the local build. This was not the cause of the failures, but a separate also important thing I discovered in the process.

PR #1275, opened 2025-10-02
Create mechanism for checking docstring consistency with xarray

ianhi:
A `uv run`abble script that compares the docstring of `to_icechunk` against xarray's `to_zarr` and highlights the differences. 

To account for known differences (e.g. we don't include compute=False text) store a hash of known differences and don't fail on them.

some examples of output:

image
image

image


ianhi:
Intentionally pushed a bad known hash and indeed it failed: https://github.com/earth-mover/icechunk/actions/runs/18201352291/job/51820814621

PR #1286, opened 2025-10-06
Add `.authorized_virtual_container_prefixes` property

### LINKED ISSUE #1285: ENH: Expose a repo's `.authorized_virtual_containers` in python
### Problem

There is currently no way to discover which of a repo's virtual chunk containers have been authorized from python.

### Proposed Solution

There is already an `authorized_virtual_containers` attribute on the `Repository` struct in rust

https://github.com/earth-mover/icechunk/blob/cbd285b8a478ecd9bbd341b3618bd3e866a655f1/icechunk/src/repository.rs#L167

So we just need to expose this.

The only question is whether or not we want the new `.authorized_virtual_containers` python property to return the credentials too. For my use case I don't need that - the url_prefixes would be fine. What do you think @paraseba ?

TomNicholas:
Closes #1285


icechunk/src/repository.rs
```diff
@@ -455,6 +455,10 @@ impl Repository {
         &self.asset_manager
     }
 
+    pub fn authorized_virtual_containers(&self) -> &HashMap<String, Option<Credentials>> {
```
mpiannucci: To decrease surface area for now, do yuou want to just make this match the pyuthon impl and just return prefixes for now?

mpiannucci (commented):
one note of suggestion

mpiannucci (approved):

Issue #1298, opened 2025-10-14
ZarrStore.open_group got an unexpected keyword argument 'align_chunks'


Melissari1997:
### Problem

I'm using an environment composed by:

- xarray version 2025.1.2
- icechunk version 1.1.9

In icechunk version 1.1.9, `ZarrStore.open_group` method is used with the `align_chunks` parameter: https://github.com/earth-mover/icechunk/blob/v1.1.9/icechunk-python/python/icechunk/xarray.py#L128

But in xarray version 2025.1.2, this parameter is not present: https://github.com/pydata/xarray/blob/v2025.01.2/xarray/backends/zarr.py#L676

And it's present in version 2025.6.0: https://github.com/pydata/xarray/blob/v2025.06.0/xarray/backends/zarr.py#L639

So in this environment, I get this error:

```powershell
TypeError: ZarrStore.open_group() got an unexpected keyword argument 'align_chunks'
``` 

### Suggested Improvement
Update the xarray version or manage the retrocompatibility with xarray versions < 2025.6.0

ianhi:
Thanks for the report @Melissari1997 it seems I introduced that when adding compatibility for this change in xarray: https://github.com/pydata/xarray/pull/10516

One thing to note if you are using this parameter is that `align_chunks` did not actually function properly

> Update the xarray version or manage the retrocompatibility with xarray versions < 2025.6.0

This is definitely the right fix, we need to support more than 3 months of xarray. I'll make a PR today. In the meantime I believe that 1.1.6 is the most recent icechunk verssion that will work for you.

That said, can I ask what's keeping you on those older versions? Is it pandas compatibility witht geopandas?

Melissari1997:
Thank @ianhi!

I'm using rioxarray, but I think that I can update the xarray version :)

ianhi:
I hope to release this tomorrow

ianhi:
@Melissari1997 I beat my expectations. This is now released in version 1.1.10

Thanks again for the report, please let me know if you have further problems with this release

Melissari1997:
Great news @ianhi ! Thanks for the effort, I'll try it as soon as possible!

PR #1299, opened 2025-10-14
Fix `align_chunks` xarray backcompat + add min version tests

ianhi:
Fixes: https://github.com/earth-mover/icechunk/issues/1298

Initially opening without the fix to check that the new min version tests introduced here would have caught this.

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -239,3 +239,31 @@ jobs:
           # pass xarray's pyproject.toml so that pytest can find the `flaky` fixture
           source .venv/bin/activate
           python -m pytest -c=../../xarray/pyproject.toml -W ignore --override-ini="strict_markers=false" tests/run_xarray_backends_tests.py
+
+  check-xarray-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          path: "icechunk"
+
+      - uses: actions/checkout@v5
+        with:
+          repository: "pydata/xarray"
+          path: "xarray"
+          fetch-depth: 0
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Check xarray documentation consistency
+        shell: bash
+        working-directory: icechunk/icechunk-python
+        env:
+          XARRAY_DIR: ../../xarray
+        run: |
+          set -e
+          uv run scripts/check_xarray_docs_sync.py
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -125,6 +125,35 @@ python -m pytest -xvs tests/run_xarray_backends_tests.py::TestIcechunkStoreFiles
   --override-ini="addopts="
 ```
 
+#### Checking Xarray Documentation Consistency
+
+Icechunk's `to_icechunk` function shares several parameters with Xarray's `to_zarr` function. To ensure documentation stays in sync, use the documentation checker script.
+
+From the `icechunk-python` directory:
+
+```bash
+# Set XARRAY_DIR to point to your local Xarray clone
+export XARRAY_DIR=~/Documents/dev/xarray
+
+# Run the documentation consistency check
+uv run scripts/check_xarray_docs_sync.py
+```
+
+The script will display a side-by-side comparison of any documentation differences, with missing text highlighted in red.
+
+**Known Differences**: Some differences are acceptable (e.g., Sphinx formatting like `:py:func:` doesn't work in mkdocs). These are tracked in `scripts/known-xarray-doc-diffs.json`. Known differences are displayed but don't cause the check to fail.
+
+**Updating Known Differences**: After making intentional documentation changes, update the known diffs file:
+
+```bash
+# Mark current diffs as known (creates/updates scripts/known-xarray-doc-diffs.json)
+uv run scripts/check_xarray_docs_sync.py --update-known-diffs
+
+# Edit scripts/known-xarray-doc-diffs.json to add reasons for each difference
+```
+
+**CI Integration**: The script returns exit code 0 if only known differences exist, allowing CI to pass while still displaying diffs for review.
+
 ### Rust Development Workflow
 
 #### Prerequisites
--- a/icechunk-python/python/icechunk/xarray.py
+++ b/icechunk-python/python/icechunk/xarray.py
@@ -1,6 +1,6 @@
-import importlib
 from collections.abc import Hashable, Mapping, MutableMapping
 from dataclasses import dataclass, field
+from importlib.util import find_spec
 from typing import Any, Literal, overload
 
 import numpy as np
@@ -22,7 +22,7 @@ ZarrWriteModes = Literal["w", "w-", "a", "a-", "r+", "r"]
 
 
 try:
-    has_dask = importlib.util.find_spec("dask") is not None
+    has_dask = find_spec("dask") is not None
 except ImportError:
     has_dask = False
 
@@ -45,9 +45,9 @@ else:
 
 def is_dask_collection(x: Any) -> bool:
     if has_dask:
-        import dask
+        from dask.base import is_dask_collection
 
-        return dask.base.is_dask_collection(x)
+        return is_dask_collection(x)
     else:
         return False
 
--- a/icechunk-python/scripts/check_xarray_docs_sync.py
+++ b/icechunk-python/scripts/check_xarray_docs_sync.py
@@ -1,0 +1,671 @@
+#!/usr/bin/env -S uv run
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "rich>=13.0.0",
+# ]
+# ///
+"""
+Check that parameter documentation in icechunk.xarray.to_icechunk is consistent
+with xarray.Dataset.to_zarr.
+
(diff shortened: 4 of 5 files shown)


ianhi:
```
    
>       self.xarray_store = ZarrStore.open_group(
            store=self.store,
            group=group,
            mode=concrete_mode,
            zarr_format=3,
            append_dim=append_dim,
            write_region=region,
            safe_chunks=self.safe_chunks,
            align_chunks=self.align_chunks,
            synchronizer=None,
            consolidated=False,
            consolidate_on_close=False,
            zarr_version=None,
        )
E       TypeError: ZarrStore.open_group() got an unexpected keyword argument 'align_chunks'
```

nice! it caught it

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -375,6 +376,10 @@ def check_list_prefix_from_root(self) -> None:
         )
 
 
+@pytest.mark.skipif(
```
dcherian: this can't be right; it's been running so far!

icechunk-python/tests/test_zarr/test_properties.py
```diff
@@ -2,10 +2,17 @@
...
+from packaging.version import Version
 
 from icechunk import IcechunkStore, Repository, in_memory_storage
 
 pytest.importorskip("hypothesis")
+import zarr
+
+pytestmark = pytest.mark.skipif(
+    Version(zarr.__version__) < Version("3.1.0"),
+    reason="BytesCodec Issuess with less than 3.1.0",
```
dcherian: do we just need to skip when `dtype.kind == "S"`? If so we can use `hypothesis.strategies.assume` in the body of the test. It'd be nice to be much narrower here

.github/workflows/python-check.yaml
```diff
@@ -161,6 +197,22 @@ jobs:
           WHEEL=$(ls dist/*-cp311-*.whl)
           uv pip install "${WHEEL}[test]"
           uv pip install pytest-mypy-plugins
+          # Install specific xarray version based on matrix
+          if [ "${{ matrix.xarray-version.version }}" = "latest" ]; then
+            echo "Using xarray from git checkout (latest)"
```
dcherian: let's have this be the latest released versions. Otherwise our PRs will start failing when there's some Xarray/Zarr incompatibility on their main branches. we already run nightly upstream tests.

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: this was a bit confusing. but i saw very consistent failures on the min version tests

ianhi: agree that this seems wrong though

icechunk-python/tests/test_zarr/test_properties.py
ianhi: will look into it

.github/workflows/python-check.yaml
ianhi: good call. Although the nightly tests don't actually test the backends :(

Also I'm somewhat shocked that this never failed in the past. it's always been checking out the latest git version, not the most recent release.

I'll look into grabbing the xarray verssion and using that to checkout, but it may get a bit complicated.

ianhi:
ok this is ready now!

The pending tests:

 
image


are that way because the github settings require those tests, but we no longer have tests by those names (because of teh version info in the test name) so I can update that post merge

.github/workflows/python-check.yaml
```diff
@@ -47,13 +49,25 @@ jobs:
...
     runs-on: ubuntu-latest
     needs: [build-wheels]
+    strategy:
+      matrix:
+        deps-version:
+          - name: minimum
+            xarray: "2025.01.2"
+            dask: "2024.11.0"
+            distributed: "2024.11.0"
+            zarr: "3.0.8"
```
dcherian, line 59: is there a way to use these in pyproject.toml and have uv resolve them?

ianhi: If you do `uv sync --resolution lowest-direct` then it will resolve to these. but i could not get that to work with installing from the built wheel as we do in the action. I can have another go at it

.github/workflows/python-check.yaml
```diff
@@ -114,11 +137,22 @@ jobs:
...
     runs-on: ubuntu-latest
     needs: [build-wheels]
+    strategy:
+      matrix:
+        xarray-version:
+          - name: minimum
+            version: "2025.01.2"
+            checkout-ref: "v2025.01.2"
+            zarr: "3.0.8"
+          - name: latest
```
dcherian: we do this nightly; I don't think we need it on every PR. Let's just do latest released version.

For the main branch tests, let's use this workflow: https://github.com/earth-mover/icechunk/blob/5a453d6009cadf227d9adffea51b820bc691ee15/.github/workflows/python-upstream.yaml#L203-L207

ianhi: this just means latest released, not the upstream main version. Is that the concern? I could update the wording here

.github/workflows/python-check.yaml
ianhi: Ok I found the issue. Becauses we aren't doing uv sync on the local package here we can't becuase we use pip (which I think is the right way to test this)

https://github.com/pypa/pip/issues/8085

.github/workflows/python-check.yaml
dcherian: oh i see, the "checkout-ref" is confusing; but it's obnly sued to checkout a version of the backend tests. correct?

ianhi: yes exactly. I'll try renaming it to make it more obvious

.github/workflows/python-check.yaml
dcherian: So is there a solution, or shall we merge and move on?

ianhi: Left for lunch and got distracted with a few napari things. I might be able to make this work with a uv compile step, then pip install. but it will be more complicated

.github/workflows/python-check.yaml
ianhi: I was actually able to fully remove it, because now I checkout the tag that matches the installed version.

.github/workflows/python-check.yaml
ianhi: Ok. This **is** posssible using something like

`uv pip compile --resolution lowest pyproject.toml`

however it will require adding min version requirements all over the place because the min resolution affects all thesse other deps. So on balance i think the current direct encoded approach is better. I'd like to leave as is and revist if this becomes problem.
 
```
⠙ Resolving dependencies...                                                                      warning: The direct dependency `boto3` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `coverage` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `mypy` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `object-store-python` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `pytest` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `pytest-cov` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `pytest-asyncio` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `pytest-xdist` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `ruff` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `hypothesis` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `pandas-stubs` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `boto3-stubs` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `termcolor` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `pooch` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `netcdf4` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
```

dcherian:
@jhamman or @paraseba you'll need to change the "required" check list for us to merge this

dcherian (approved):

.github/workflows/python-check.yaml
dcherian: sounds good

ianhi:
> @jhamman or @paraseba you'll need to change the "required" check list for us to merge this

I have this power :) I will update those required checks now

Issue #1194, opened 2025-08-25
Get a list of all the files referenced by a virtual icechunk store


jsignell:
### Problem

In a virtual icechunk store I don't think there is currently a way to access a list of all the files that store the chunks. 

### Proposed Solution

https://github.com/earth-mover/icechunk/issues/104 has some references to the idea of a `get_chunk_ref` which would be a good general solution to this. 

### Additional context

I also considered whether virtualizarr should be able to read virtual icechunk stores back in (https://github.com/zarr-developers/VirtualiZarr/issues/786), but I don't think this is possible until this kind of functionality is added within icechunk.

rabernat:
Hi @jsignell! 👋  Great to see you around here!

Curious what's your use case for this?

paraseba:
@jsignell closest thing we have today is `Session.all_virtual_chunk_locations`. But that applies only to a single snapshot. You would need to call repeatedly over all snapshots in the repo, adding them to a set.

jsignell:
The current use case is to use the virtual icechunk store as the sole source of truth for generating the STAC item-level metadata per-legacy file. Not sure if this will work or not, but there is sometimes a requirement that the legacy files each be represented by their own item in STAC and we were scheming about whether everything we need to make that item is contained in the virtual store.

jsignell:
So I guess what I really would need is not just the chunk location but also the dimension range of each chunk.

paraseba:
@jsignell what do you mean exactly by "dimension range of each chunk"? Would it be enough if we give you the chunk index in the chunk grid?

jsignell:
> Would it be enough if we give you the chunk index in the chunk grid?

That would be enough as long as I can also represent the chunk grid in dimension (time, lat, lon) space.

paraseba:
> That would be enough as long as I can also represent the chunk grid in dimension (time, lat, lon) space.

Can you get that from the Zarr array metadata?

jsignell:
Yeah I think basically if there were a version of `all_virtual_chunk_locations` that specified which array the chunk is part of and what the chunk index is then that would be great. I was imagining that the response could have the same shape as the chunk layout. So if you have an array called `air_temp` with 2 chunks in the 0th dim and 3 chunks in the 1st dim then you could have the response be like:

```
{
  air_temp: [
    [location, location, location], 
    [location, location, location],
  ]
}
```

paraseba:
[...]

Issue #104, opened 2024-09-25
Reading virtual references back out into VirtualiZarr Manifests


TomNicholas:
I would like to be able to read virtual references back out from an icechunk store into VirtualiZarr `ManifestArray` objects.

Note this issue is the icechunk equivalent of https://github.com/zarr-developers/VirtualiZarr/issues/118, which is about reading kerchunk references into `ManifestArray` objects.

The main use case is appending new data to an existing store (see https://github.com/zarr-developers/VirtualiZarr/issues/21#issuecomment-2114001053), so that when some new data arrives (e.g. a new grib file with today's weather data), I can add an updated snapshot just with something like:

```python
import virtualizarr as vz

# avoids re-extracting all the metadata from all the past grib files, so should be quick
existing_vds = vz.open_virtual_dataset(icechunkstore, reader='icechunk')

new_vds = vz.open_virtual_dataset('todays_weather.grib', reader='grib')

updated_vds = xr.concat([existing_vds, new_vds], dim='time')

# commit new snapshot that includes today's data
# requires https://github.com/earth-mover/icechunk/issues/103
updated_vds.virtualize.to_icechunk(icechunkstore)
icechunkstore.commit('<todays-date>')
```

In order to implement that Icechunk reader for virtualizarr I would need some API for getting all virtual (and non-virtual) references for a snapshot back out of the Icechunk store, ideally as a `vz.ManifestArray` or something I can cheaply coerce to one (see [`ChunkManifest.from_arrays()`](https://github.com/zarr-developers/VirtualiZarr/blob/47a5e8702e44f71bb355bcba0ff6214fe6d09d83/virtualizarr/manifests/manifest.py#L142)).

Writing the updated references as a new snapshot also requires https://github.com/earth-mover/icechunk/issues/103.

(I guess the `.virtualize.to_icechunk` method might also need to know to do `array.resize` in this example... (see the Append example in [this notebook](https://github.com/earth-mover/icechunk/blob/main/icechunk-python/notebooks/demo-dummy-data.ipynb).)

Running that above snippet as a cron job / event-driven serverless function should go a long way towards making ingestion of regularly-updated data archives easier. (cc @mpiannucci)

This feature might also be useful to allow using icechunk as a serialization format during large serverless reductions (xref https://github.com/zarr-developers/VirtualiZarr/issues/123).

cc @paraseba

abarciauskas-bgse:
[...]

PR #1307, opened 2025-10-22
Add `get_virtual_ref` and `all_virtual_refs` functionality

mpiannucci:
#104 
#1194

### DIFF
--- a/icechunk-python/python/icechunk/session.py
+++ b/icechunk-python/python/icechunk/session.py
@@ -1,6 +1,5 @@
 import contextlib
 import warnings
-from datetime import datetime
 from collections.abc import AsyncIterator, Callable, Generator, Iterable, Sequence
 from typing import Any, NoReturn, Self
 


TomNicholas:
I think this needs more thought - see https://github.com/earth-mover/icechunk/issues/104#issuecomment-3453976262.

TomNicholas:
Sorry @mpiannucci - despite the fact that this is not enough for me to solve the thorny problem of #104, adding these methods would still be useful. I should have approved a while ago.

icechunk/src/session.rs
```diff
@@ -1086,6 +1086,27 @@ impl Session {
         Ok(stream)
     }
 
+    #[instrument(skip(self))]
+    pub async fn all_virtual_refs(
```
TomNicholas: Is there a reason why this is on the `session` but `get_virtual_refs`, `set_virtual_ref`, and `set_virtual_refs` are all on the `store`?

TomNicholas: moved everything to `Session` in https://github.com/earth-mover/icechunk/pull/1307/commits/44d9b9bcee9f3046b9a5a59e070152f44533ae95

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags=["-D", "warnings", "-W", "unreachable-pub",  "-W", "bare-trait-objects"]
```
paraseba, line 2: I don't think we want this here?

PR #1353, opened 2025-11-03
New method Repository.manifest_files to get the snap manifests

paraseba:
This replaces the previous API of `SnapshotInfo.manifests` and it's the only API breaking change in Icechunk 2. We don't expect the old API to be heavily used.

### DIFF
--- a/docs/docs/performance.md
+++ b/docs/docs/performance.md
@@ -280,10 +280,10 @@ And commit
 snap = session.commit("Add 5 chunks")
 ```
 
-Use [`repo.lookup_snapshot`](./reference.md#icechunk.Repository.lookup_snapshot) to examine the manifests associated with a Snapshot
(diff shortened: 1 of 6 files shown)


mpiannucci (approved):

docs/docs/performance.md
```diff
@@ -280,10 +280,10 @@ And commit
 snap = session.commit("Add 5 chunks")
 ```
 
-Use [`repo.lookup_snapshot`](./reference.md#icechunk.Repository.lookup_snapshot) to examine the manifests associated with a Snapshot
+Use [`repo.manifest_files`](./reference.md#icechunk.Repository.manifest_files) to examine the manifests associated with a Snapshot
```
dcherian, line 283: should we use a verb instead? `list_manifest_files`?

paraseba: too late, but I like your name better. Creating a ticket

PR #1357, opened 2025-11-03
CI: update for support branches

ianhi:
CI runs on PRs against support branch


ianhi:
> Shouldn't we run both on support and main?

yes, will need two PRs. but i got caught up in the one to main trying to fix the nightly wheels to be based off both branches and stalled out. I'm planning on cherry-picking this commit over to main as well

PR #1359, opened 2025-11-04
Ian/v1 ci main

ianhi:
Update Main branch to support:

1. Same cherry picked commit https://github.com/earth-mover/icechunk/pull/1357
2. Nightly wheels build for both support v1 and main branch + optionally manual build of either/both
3. update main version to 2.0.0-alpha.0 (normalizes to pep compliant 2.0.0.a0 for wheels)

after this we will need to tag 2.0.0-alpha.0 to make the CI work properly

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -43,6 +43,7 @@ Create / activate a virtual environment:
 
     ```bash
     uv sync
+    source .venv/bin/activate
     ```
 
 Install `maturin`:
@@ -217,6 +218,8 @@ This builds the site to `docs/.site` directory.
 
 #### Prerequisites
 
+You need to have already created and activated a virtual environment ([see above](#python-development-workflow)), because the full rust build will also compile the python bindings.
+
 Install the `just` command runner (used for build tasks and pre-commit hooks):
 
 ```bash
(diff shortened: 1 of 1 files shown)


.github/workflows/python-ci.yaml
```diff
@@ -18,6 +18,14 @@ on:
         required: true
         default: true
         type: boolean
+      branch:
+        description: 'Branch to build/release (for manual builds only)'
+        type: choice
+        options:
+          - main
+          - support/v1.x
+          - both
```
paraseba: I think I don't like having `both` here. It sounds risky, non atomic, etc. Am I wrong?

.github/workflows/python-ci.yaml
```diff
@@ -18,6 +18,14 @@ on:
...
+        default: main
```
paraseba: Should the default be support until we release IC2?

paraseba (approved):

.github/workflows/python-ci.yaml
ianhi: yea i guess since this is only manual builds no harm in removing th eboth. Hard to imagine a use case for that byeond "i want to publish nightly wheels for both. But you can just run it twice

.github/workflows/python-ci.yaml
ianhi: ah yes it should

PR #1362, opened 2025-11-05
Add some missing steps for first-time rust contributors

TomNicholas:


docs/docs/contributing.md
```diff
@@ -228,6 +231,8 @@ Or using other package managers:
 - **macOS**: `brew install just`
 - **Ubuntu**: `snap install --edge --classic just`
 
+Ensure you have navigated to the root directory of the cloned repo (i.e. not the `icechunk-python` subdirectory).
```
ianhi, line 234: I thought this is system level, not per project. I was able to run this just from my home dir. Is this instruction for a different part?

TomNicholas: You were able to run what from your home dir?

IIUC to run any of the `cargo` commands you need to be in the correct directory, so cargo knows which rust crate you are attempting to compile.

PR #1373, opened 2025-11-06
Expose merge method in python in IC V1

TomNicholas:
Same as #1372 but for IC1

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - 'support/**'
 
 concurrency:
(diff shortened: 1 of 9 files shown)


paraseba (approved):

TomNicholas:
mypy failed, but for a very unrelated-looking reason

ianhi:
Agreed unrelated to the changes. i think we can just mypy ignore it as it will still be available: 
image

TomNicholas:
Okay I'm going to merge this then try to fix that if it fails on ~~`main`~~ `support/v1.x`.

TomNicholas:
Should be fixed by #1374

PR #1374, opened 2025-11-06
Ignore mypy complaint about `dask.__version__`

TomNicholas:
Fix the mypy error seen in #1373, by ignoring it.


ianhi:
can you cherry pick this and bring to main as well

PR #1379, opened 2025-11-06
New HTTP Storage instance

paraseba:
New method:

```python
def http_storage(base_url: str, opts: dict[str, str] | None = None) -> Storage:
```

Credential handling is not implemented yet.

The Storage instance is read only.


dcherian (approved):
As discussed would be good to add docs at some point

PR #1386, opened 2025-11-08
Introduce `redirect_storage`

paraseba:
This new Storage instance executes an http GET on a URL passed. It expects the response to be a redirect to the URL of a destination target dataset. After the destination is resolved, it behaves as a storage appropiate for the target url.

- It follows up to 10 redirects until one matches one of the handled protocols
- Currently it handles the following protocols:
    - s3
    - gs or gcs
    - http+icechunk, http+ic, https+icechunk, or https+ic
    - r2
    - tigris
- The redirect for s3 must include a region as query argument, example: `Location: s3://foo/bar/baz?region=us-east-1`
- The redirect for r2 must include an account as query argument, example: `Location: r2://foo/bar/baz?account_id=12345`
- Currently the target Storage instances are created using anonymous credentials
- Currently no auth is supported in the redirection service

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2456,7 +2456,7 @@ checksum = "a948666b637a0f465e8564c73e89d4dde00d72d4d473cc972f390fc3dcee7d9c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.108",
+ "syn 2.0.110",
 ]
 
 [[package]]


paraseba:
This is missing documentation, left for a followup PR.

paraseba:
Probably the best way to understand how to use this is to look at the new python test, after reading the commit message.

PR #1390, opened 2025-11-11
builds: fix RTD explicitly add patchelf + update macos runners

ianhi:
hopefully fix RTD build errors


ianhi:
This seems to have fixed the RTD issue. the failure is the macos-13 build issue

PR #1391, opened 2025-11-11
DOC: update overview page with more concrete examples

ianhi:
I moved the old overview page into the new "concepts" page and re-wrote the overview to focus on three concrete improvements on top of zarr. I also tried to to structure it so that someone who knows what an ACID transaction is will still immediately be able to see what is going on, but for those who don't there is an easy path to learning what that is and why it is important.


docs/docs/concepts.md
```diff
@@ -0,0 +1,225 @@
...
+group["pressure"][:, 6] = pressure_forecast     # Written immediately to disk
+# ... computer crashes here ...
+group["humidity"][:, 6] = humidity_forecast     # Never gets written!
+```
+
+Now your dataset has temperature and pressure for hour 6, but no humidity or wind speed. Anyone reading this data will get inconsistent results, and there's no automatic way to detect or fix this problem.
+
+Transactions solve this by making all the writes atomic—they happen together or not at all. This is the foundation that makes Icechunk reliable enough to use as a database.
+
+### ACID Transactions
```
ianhi, line 116: This is ACID transactions section is pretyt much claude vibing out explanations because I coudln't find a single great source to link to. But fairly confident it did a good job, but someone who understands this better than me should hopefully check

docs/docs/concepts.md
```diff
@@ -0,0 +1,225 @@
...
+```python
+# Reader sees a consistent snapshot, unaffected by concurrent writers
+session = repo.readonly_session("main")
+data = zarr.open_group(session.store)["temperature"][:]
+# This data won't change even if someone commits updates right now
+```
+
+**Durability** - "Commits are permanent"
+
+Once your transaction is committed, it's saved permanently. Even if your computer crashes immediately after committing, your data is safe. The commit won't be lost or partially applied.
```
rabernat, line 151:
```suggestion
Once your transaction is committed, it's saved permanently. Even if your computer crashes immediately after committing, your data is safe. The commit won't be lost or partially applied.
Icechunk's durability is inherited from the durability of the underlying storage medium, such as S3.
```

docs/docs/concepts.md
```diff
@@ -0,0 +1,225 @@
...
+    # All these operations happen together as one transaction
+    group["temperature"][:, 6] = temp_forecast
+    group["pressure"][:, 6] = pressure_forecast
+    group["humidity"][:, 6] = humidity_forecast
+    group["wind_speed"][:, 6] = wind_forecast
+
+# Transaction commits here - all changes become visible at once
+```
+
+While the transaction is running (inside the `with` block), these changes exist in a kind of "scratch space" that only you can see. Other users reading the data see the previous committed version. When the transaction completes successfully:
```
rabernat:
```suggestion
While the transaction is running (inside the `with` block), these changes exist in a kind of "scratch space" that only you can see. (The scratch space in is the remote storage, i.e. S3, not your local computer.)
Other users reading the data see the previous committed version. When the transaction completes successfully:
```

docs/docs/overview.md
```diff
@@ -3,128 +3,177 @@ title: Overview
...
+    # Reader process (concurrent with writer)
+    session = repo.readonly_session("main")
+    forecast = zarr.open_group(session.store)
+    # Always sees a complete, consistent snapshot
+    # Either all old values OR all new values, never mixed
+    ```
+
+There are other consistency issues with Zarr that you can read about in more detail in the [Multi-Player Mode](https://earthmover.io/blog/multi-player-mode-why-teams-that-use-zarr-need-icechunk) blog post.
+
+In contrast, **Icechunk** is, by design, always consistent. It implements ACID transactions, which ensures you will never run into these issues.
```
rabernat: "ACID transactions" should be a link to "concepts"

rabernat (approved):
This is excellent. Sorry it took me so long to review.

PR #1395, opened 2025-11-13
Speed up `ops_log` by storing up to 100 changes per repo file

paraseba:
This makes `ops_log` require I/O only once every 100 changes.

Other improvement is that we store the update time for every change, that can now be returned in `ops_log`

TODO in future PR: make the repo file store a configurable number of changes instead of hardcoding 100.

Closes: #1404

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,23 +26,14 @@ body:
     attributes:
       label: Minimal Complete Verifiable Example
       description: |
-        Minimal, self-contained copy-pastable example that demonstrates the issue.         This will be automatically formatted into code, so no need for markdown backticks.
+        Minimal, self-contained copy-pastable example that demonstrates the issue.
+
+        Consider listing additional or specific dependencies in [inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#example)
+        so that calling `uv run issue.py` shows the issue when copied into `issue.py`.
+
+        This will be automatically formatted into code, so no need for markdown backticks.
       render: Python
-
-  - type: checkboxes
-    id: mvce-checkboxes
-    attributes:
-      label: MVCE confirmation
-      description: |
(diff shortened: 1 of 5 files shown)


mpiannucci (approved):

mpiannucci:
Sorry my review took so long! Wanted to fully read it

paraseba:
> Sorry my review took so long! Wanted to fully read it

No problem at all. Thank you @mpiannucci !!

icechunk/src/format/repo_info.rs
```diff
@@ -247,6 +253,91 @@ impl RepoInfo {
...
+    fn mk_latest_updates<
+        'bldr,
+        'a,
+        I: IntoIterator<Item = IcechunkResult<(UpdateType, DateTime<Utc>, Option<&'a str>)>>,
+    >(
+        builder: &mut flatbuffers::FlatBufferBuilder<'bldr>,
+        update: UpdateInfo<I>,
+        backup_path: Option<&'a str>,
+    ) -> IcechunkResult<(
+        WIPOffset<
```
dcherian, line 266: 😵‍💫

icechunk/src/format/repo_info.rs
```diff
@@ -247,6 +253,91 @@ impl RepoInfo {
...
+        );
+
+        let new_updates: Box<dyn Iterator<Item = _>> =
+            if let Some(last_update) = last_update {
+                Box::new(
+                    [Ok((update.update_type, update.update_time, None)), last_update]
+                        .into_iter(),
+                )
+            } else {
+                Box::new([Ok((update.update_type, update.update_time, None))].into_iter())
```
dcherian, line 299: Just curious, why `Box` and not  `[Ok((update.update_type, update.update_time, None)), None]`?

paraseba: Rust doesn't like they types, arrays have fixed size, and iterator types are different.

dcherian: Right I mean that we could `[Result<_, E>, Option<_>]` so the array is the same size in both branches; 

In other words, why not `[Ok((update.update_type, update.update_time, None)), last_update].into_iter()` and handle the Option in the body.

Issue #1400, opened 2025-11-13
Linux Wheel Generations Broken


ianhi:
https://github.com/earth-mover/icechunk/pull/1386 added a requirement for openssl which is now causing the release action to fail:

https://github.com/earth-mover/icechunk/actions/runs/19286028078/job/55147013938

```
  Could not find openssl via pkg-config:

  pkg-config exited with status code 1
  > PKG_CONFIG_PATH=/usr/local/lib/pkgconfig PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags openssl

  The system library `openssl` required by crate `openssl-sys` was not found.
  The file `openssl.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
  PKG_CONFIG_PATH contains the following:
      - /usr/local/lib/pkgconfig
```

I think this is because the `PyO3/Maturin` action step we use here:

https://github.com/earth-mover/icechunk/blob/3be837a72439971b86d8d9184b7ad3cb4d5a209d/.github/workflows/python-ci.yaml#L170-L176

uses the `manylinux` image which does not have openssl in it

https://github.com/rust-openssl/rust-openssl/issues/1985#issuecomment-1612208107

> manylinux containers do not have an openssl installed in them, because they provide only the bare set of libraries that can be assumed on all linux installations (e.g., an old glibc). To resolve this, you need to bring your own openssl and statically link it (and therefore, update it whenever openssl does a security release).

Also from https://peps.python.org/pep-0513/

>  In particular, because it lacks a stable ABI, OpenSSL cannot be included in the manylinux1 profile.

And from the manylinux repo (https://github.com/pypa/manylinux/issues/1195#issuecomment-932721624)

> If OpenSSL is a direct dependency of your build, IMHO, you should rebuild it from sources as a static library (see [pyca/cryptography](https://github.com/pyca/cryptography), [scikit-build/cmake-python-distributions](https://github.com/scikit-build/cmake-python-distributions) wheel build for example).

## Possible Solution

Adding the `openssl` crate as vendored may be the solution here: https://docs.rs/openssl/latest/openssl/#vendored

PR #1402, opened 2025-11-13
use rustls-tls instead of native-tls -> openssl

ianhi:
openssl not available in the manylinux docker. so instead depend on rust-tls 
see: https://github.com/seanmonstar/reqwest/issues/495

closes: #1400

### DIFF
--- a/icechunk/Cargo.toml
+++ b/icechunk/Cargo.toml
@@ -73,6 +73,10 @@ flexbuffers = "25.9.23"
 # https://github.com/earth-mover/icechunk/issues/1400
 reqwest = { version = "0.12.24", default-features = false, features = [
   "rustls-tls",
+  # defaults we rely on. Add explicitly because we set default-features to false
+  "charset",
+  "http2",
+  "system-proxy",
 ] }
 [dev-dependencies]
 fs_extra = "1.3.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -624,7 +624,7 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
- "windows-link",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -788,7 +788,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-link",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -1174,6 +1174,15 @@ name = "encode_unicode"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
+
+[[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+dependencies = [
+ "cfg-if",
+]
 
 [[package]]
 name = "env_filter"
@@ -1748,9 +1757,11 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "socket2 0.6.1",
+ "system-configuration",
  "tokio",
  "tower-service",
  "tracing",
+ "windows-registry",
 ]
 
 [[package]]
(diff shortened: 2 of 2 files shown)


deny.toml
```diff
@@ -25,6 +25,7 @@ allow = [
   "CC0-1.0",
   "Unicode-3.0",
   "Zlib",
+  "CDLA-Permissive-2.0",
```
ianhi, line 28: required for https://github.com/rustls/webpki-roots#license which we pick up for getting the certs for rustls-tls

.github/workflows/python-check.yaml
```diff
@@ -98,7 +97,8 @@ jobs:
           uv venv
           source .venv/bin/activate
           python --version
-          WHEEL=$(ls dist/*-cp311-*.whl)
+          PYTHON_TAG="cp$(echo ${{ env.PYTHON_VERSION }} | tr -d '.')"
+          WHEEL=$(ls dist/*-${PYTHON_TAG}-*.whl)
```
ianhi, line 101: this is not directly necessary but good future proofing for when we update the python version

.github/workflows/python-check.yaml
```diff
@@ -36,10 +36,9 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           working-directory: icechunk-python
-          args: --release --out dist --find-interpreter
+          args: --release --out dist -i python${{ env.PYTHON_VERSION }}
           sccache: true
-        env:
-          CARGO_TERM_QUIET: true
+          manylinux: auto
```
ianhi, line 41: this matches the way we build in the release workflow (using manylinux) but constrains to only build for the PYTHON_VERSION version of python to speed things up

icechunk/Cargo.toml
```diff
@@ -67,8 +67,13 @@ dirs = { version = "6.0.0", optional = true }
 assert_fs = { version = "1.1.3", optional = true }
 flatbuffers = "25.2.10"
 flexbuffers = "25.9.23"
-reqwest = "0.12.24"
 
+# reqwest's default-features enables default-tls which ultimately leads to an openssl dependencies
+# which we are not including as it is not included in manylinux wheels. Instead depend on rustls-tls only
```
ianhi, line 72:
```suggestion
# which we are not including as it is not included in the manylinux image. Instead depend on rustls-tls only
```

icechunk/Cargo.toml
```diff
@@ -67,8 +67,13 @@ dirs = { version = "6.0.0", optional = true }
 assert_fs = { version = "1.1.3", optional = true }
 flatbuffers = "25.2.10"
 flexbuffers = "25.9.23"
-reqwest = "0.12.24"
 
+# reqwest's default-features enables default-tls which ultimately leads to an openssl dependencies
+# which we are not including as it is not included in the manylinux iamge. Instead depend on rustls-tls only
```
ianhi:
```suggestion
# which we are not including as it is not included in the manylinux image. Instead depend on rustls-tls only
```

icechunk/Cargo.toml
```diff
@@ -67,8 +67,13 @@ dirs = { version = "6.0.0", optional = true }
 assert_fs = { version = "1.1.3", optional = true }
 flatbuffers = "25.2.10"
 flexbuffers = "25.9.23"
-reqwest = "0.12.24"
 
+# reqwest's default-features enables default-tls which ultimately leads to an openssl dependencies
+# which we are not including as it is not included in the manylinux image. Instead depend on rustls-tls only
+# https://github.com/earth-mover/icechunk/issues/1400
+reqwest = { version = "0.12.24", default-features = false, features = [
```
paraseba, line 74: actually, now that i look, i think we need to enable a couple more features:
- charset
- http2
- system-proxy

paraseba (changes requested):

icechunk/Cargo.toml
ianhi: Should we open an issue for adding a test taht would have caught these missing ones?

ianhi: Actually @paraseba, all of those are enabled by default: https://docs.rs/reqwest/latest/reqwest/#optional-features

so we don't need to explcitly include them

paraseba: but you are saying (as you should) `default-features = false`

paraseba: > Should we open an issue for adding a test taht would have caught these missing ones?

We can open an issue, but it's significant work to make tests for this kind of thing

ianhi: oooh whoops i didn't connect that that would remove all the defaults

ianhi: ok added them + a comment with why

paraseba (approved):

Issue #1220, opened 2025-09-11
IC2 `total_chunks_storage` doesn't work in IC1 repos


paraseba:
<img width="522" height="237" alt="Image" src="https://github.com/user-attachments/assets/c1683916-cf0b-48a3-b7bf-038151f55df5" />

PR #1407, opened 2025-11-17
Fix `Repository.total_chunks_storage` for V1 repos.

paraseba:
TODO in future PR: actually allow the creation of V1 repos. Currently the API is added, but the parameter is ignored.

Closes: #1220

### DIFF
--- a/icechunk-python/python/icechunk/repository.py
+++ b/icechunk-python/python/icechunk/repository.py
@@ -33,6 +33,7 @@ class Repository:
         storage: Storage,
         config: RepositoryConfig | None = None,
         authorize_virtual_chunk_access: dict[str, AnyCredential | None] | None = None,
+        spec_version: int | None = None,
     ) -> Self:
         """
         Create a new Icechunk repository.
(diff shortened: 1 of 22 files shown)


mpiannucci (approved):

icechunk-python/python/icechunk/repository.py
```diff
@@ -55,6 +56,9 @@ def create(
             environment, or anonymous credentials will be used if the container allows it.
             As a security measure, Icechunk will block access to virtual chunks if the
             container is not authorized using this argument.
+        spec_version : int, optional
+            Use this version of the spec for the new repository. If not passed, the latest version
+            of the spec that was available before the library version release will be used.
```
dcherian, line 61: can we add a small helper `ic.default_spec_version()`  or `ic.latest_spec_version()` so this is easy to query for a user.

EDIT: Actually `ic.supported_spec_versions()` that listed out the `SpecVersionBin` enum with an indicator of which is default would be nice.

paraseba: good idea, adding as a ticket

PR #1417, opened 2025-11-19
More write operations on V1 repos

paraseba:
In this PR we add support for the following operations on spec vesion 1 repos:

- branch creation
- branch delete
- branch reset
- tag creation
- tag delete


dcherian:
Updating `test_stateful_repo_ops.py` to test both V1 and V2 would give us some confidence here. I'm happy to do this if you'd like.

paraseba:
> Updating `test_stateful_repo_ops.py` to test both V1 and V2 would give us some confidence here. I'm happy to do this if you'd like.

Oh yeah @dcherian ... I plan to do that, and we will need to do much more. But it won't work yet. Soon.

PR #1418, opened 2025-11-19
`Session.flush` on V1 repos

paraseba:
Also in this PR, expand test_timetravel to test all spec versions.

Still missing: expiration and GC on V1 repos


icechunk-python/src/repository.rs
```diff
@@ -1903,15 +1903,14 @@ impl PyRepository {
...
-                        &lock,
-                        branch,
-                        message,
-                        metadata,
-                        icechunk::session::CommitMethod::Amend,
-                    )
-                    .await
-                    .map_err(PyIcechunkStoreError::ManifestOpsError)
+                    let commit_method = if lock.spec_version() > SpecVersionBin::V1dot0 {
+                        icechunk::session::CommitMethod::Amend
```
mpiannucci, line 1907: So in IC2 rewriting manifest is an amend, not a new commit?

paraseba: I thought that would be better ... but ... very open to change this. Maybe we should keep it as commit?

mpiannucci: All good just calling it out as a change! I think it makes sense as a default

dcherian: the reason we went with new commit initially was to allow users to switch between versions and measure perf differences. I guess now a user *has* to create a new branch, call rewrite manifests and compare between branches.

icechunk-python/tests/test_timetravel.py
```diff
@@ -279,9 +289,14 @@ async def test_branch_reset() -> None:
     ) is None
 
 
-async def test_tag_delete() -> None:
+@pytest.mark.parametrize(
```
dcherian, line 292: might be good to use a fixture for this.

paraseba: good idea! will change

icechunk-python/src/repository.rs
paraseba: yeah ... it may be better to keep as commit ... not sure. I'll create a ticket to think about it

icechunk/src/session.rs
```diff
@@ -1137,6 +1137,36 @@ impl Session {
         self._commit(message, properties, false, CommitMethod::Amend).await
     }
 
+    async fn flush_v2(&mut self, new_snap: Arc<Snapshot>) -> SessionResult<()> {
+        let update_type =
```
dcherian, line 1141: a comment here would be helpful.

IIUC:
1. we flush and create the snapshot file before this function is called. This is "detached" since "repo_info" hasn't been updated yet.
2. In this function we update the repo info file, which could fail and the snapshot would remain dangling.

So `UpdateType::NewDetachedSnapshotUpdate` means we are updating to add the detached snapshot  to ancestry list.

EDIT: I find that enum variant confusingly named. It suggests the detached snapshot is being updated, rather than being inserted in to repo info.

paraseba: yes, you UC. All variants in that enum end with `Update` ... i don't know why ... it sounds stupid now that you mention it

paraseba: Ah, maybe because this goes all the way to the flatbuffers, where you need to put the names in the global namespace

PR #1422, opened 2025-11-19
GC and expiration for V1 repos

paraseba:
Also in this PR: a pytest fixture to test across spec versions as suggested by @dcherian.

TODO in other PR: make the stateful test work for V1. It currently assumes the IC2 behavior that differs slightly, for example, by making expired snapshots not findable.


icechunk/src/ops/expiration_v1.rs
```diff
@@ -0,0 +1,231 @@
+use std::{collections::HashSet, sync::Arc};
```
paraseba, line 1: This is the expiration code in the support branch, with minor modifications to adapt it to new APIs

Issue #1011, opened 2025-06-25
Support requester_pays access for virtual chunk credentials


jbusecke:
### Problem

I am wondering if it would be possible (or if I simply missed an existing way) to enable requester-pays buckets as sources for virtual chunks in icechunk. 

I am currently trying to build a virtual dataset for NASA data that can be accessed on a requester pays bucket. 

### Proposed Solution

I would like to be able to enable requester pays access to the virtualized chunks, maybe along the lines of a keyword argument like  `icechunk.s3_credentials(..., requester_pays=True)`

### Additional context

Note: I am exploring the option to implement [refreshable credentials](https://icechunk.io/en/latest/icechunk-python/storage/#__tabbed_1_4) but was still curious about this option.

jbusecke:
Additionally I guess this would be a somewhat risky way of 'shipping' virtual data to folks, since they might open the virtual dataset and not realize that they are paying to use the data when they load it...

paraseba:
Yes this should be possible @jbusecke . As you mention, one tricky part is how to make it safe, forcing the user to explicitly acknowledge they are paying.

PR #1425, opened 2025-11-20
Enable a requester pays flag for use with S3 compatible stores.

sharkinsspatial:
This adds requester pays support for S3 compatible stores and also utilizes the flag for virtual chunk fetch configuration.  Closes https://github.com/earth-mover/icechunk/issues/1011.

### DIFF
--- a/icechunk-python/python/icechunk/storage.py
+++ b/icechunk-python/python/icechunk/storage.py
@@ -195,7 +195,6 @@ def s3_object_store_storage(
     anonymous: bool | None = None,
     from_env: bool | None = None,
     force_path_style: bool = False,
-    requester_pays: bool = False,
 ) -> Storage:
     credentials = s3_credentials(
         access_key_id=access_key_id,
@@ -211,7 +210,6 @@ def s3_object_store_storage(
         endpoint_url=endpoint_url,
         allow_http=allow_http,
         force_path_style=force_path_style,
-        requester_pays=requester_pays,
     )
     return Storage.new_s3_object_store(
         config=options,


paraseba:
amazing @sharkinsspatial !!! Will review today

paraseba:
Don't worry about the remaining test errors @sharkinsspatial , they are just our CI system not giving you access to the secrets necessary to run those integration tests.

icechunk-python/python/icechunk/storage.py
```diff
@@ -189,6 +195,7 @@ def s3_object_store_storage(
     anonymous: bool | None = None,
     from_env: bool | None = None,
     force_path_style: bool = False,
+    requester_pays: bool = False,
```
paraseba: @sharkinsspatial we should either not include the option here, or also implement requester_pays in `object_store.rs`. Otherwise people will pass `True` here and see no effect.

icechunk-python/python/icechunk/storage.py
```diff
@@ -230,6 +238,7 @@ def tigris_storage(
     get_credentials: Callable[[], S3StaticCredentials] | None = None,
     scatter_initial_credentials: bool = False,
     network_stream_timeout_seconds: int = 60,
+    requester_pays: bool = False,
```
paraseba: I don't think tigris supports requester pays, let's remove the option and  hardcode `False`

icechunk-python/python/icechunk/storage.py
```diff
@@ -315,6 +327,7 @@ def r2_storage(
...
```
paraseba: I don't think R2 supports requester pays, let's remove the option

paraseba:
@ianhi any chance we can port your recent fixes to the support branch to fix the python issues in this PR? I think it's otherwise ready to merge

ianhi:
They have been backported. @sharkinsspatial rebasing on the latest changes to the target branch should fix the remaining errors. I don't seem to have the power to merge main via the github UI here.

paraseba:
Thank you @sharkinsspatial !

Issue #1355, opened 2025-11-03
Rename `Repository.manifest_files` -> `Repository.list_manifest_files`


paraseba:

paraseba:
I'm not sure if this was completed or not.

li-em:
Done in https://github.com/earth-mover/icechunk/pull/1428

Issue #1405, opened 2025-11-15
Documentation on IceChunk 2


abhiaagarwal:
### Problem

Hi there! I've been evaluating ways to store tensor data efficiently, and I stumbled upon IceChunk. I'm interested in building around it, but I noticed there's an IceChunk 2 in progress. Since I'm just prototyping for now, I'd be happy to build around IceChunk 2, but I'm trying to figure out what the changes are / how stable it is. There appears to be no documentation about this.

### Suggested Improvement

Document the changes in IceChunk 2

paraseba:
Hello @abhiaagarwal. Glad do know you have discovered Icechunk! Icechunk 2 will be an evolution of the library and on-disk format. It brings several very interesting new features and some performance improvements. Unfortunately there is no documentation yet, but you can get some understanding of the biggest parts by looking at the [Icechunk 2](https://github.com/orgs/earth-mover/projects/11/views/1) project on github. In particular, you can look at the `Done` items for a peak into most of the new features.

My recommendation would be for you to start on Icechunk 1, the latest version of the spec, using the latest release of the Icechunk library. Here are the reasons why:

- Icechunk 2 library will be API compatible, so you won't have to change your code to later migrate to Icechunk 2.
- Icechunk 2 library will be able to read and write to Icechunk 1 repositories.
- There will be a very easy and cheap mechanism to upgrade your repositories to the Icechunk 2 spec if you want to use the new features.
- Icechunk 2 won't be ready for several weeks, there is no need to wait if you can use Icechunk 1 today.

Please, let us know if you have any questions about Icechunk 1 or 2, and feel free to join the Slack community by clicking the slack button on https://icechunk.io/

abhiaagarwal:
Thank you for the long response! Looking forward to IceChunk 2, though your comment makes me feel confident about building around IceChunk 1!

PR #1428, opened 2025-11-20
repository: add `list_` prefix to manifest_files() method

evaporei:
Addresses: https://github.com/earth-mover/icechunk/issues/1355

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -112,11 +112,18 @@ However, because it sits _below_ Zarr in the stack, it's not immediately clear h
 No, we do not recommend implementing Icechunk independently of the existing Rust library.
 There are two reasons for this:
 
-1. The spec has not yet been stabilized and is still evolving rapidly.
+1. The 1.0 version of the spec is stable. We are continuing to make improvements and will publish new versions of the spec as needed, accompanied with working reference implementation.
 1. It's probably much easier to bind to the Rust library from your language of choice,
    rather than re-implement from scratch.
 
 We welcome contributions from folks interested in developing Icechunk bindings for other languages!
+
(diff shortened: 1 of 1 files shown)


ianhi:
Thanks for the PR @evaporei !

Note for @paraseba. In https://github.com/earth-mover/icechunk/issues/1405#issuecomment-3536753071 you wrote

> Icechunk 2 library will be API compatible, so you won't have to change your code to later migrate to Icechunk 2.


this, and any other renames, will break that promise unless we include the old method name as an alias, possibly with an admonition to swap to the new name.

evaporei:
I can change so both are available, and `list_` is the one used in the docs.

paraseba:
This is the one breaking change we are allowing ourselves to make. Reason: this method doesn't exist in IC1, in IC1 there was a field in the SnapshoInfo struct to retrieve the manifests. This is not something we expect users to use on their pipelines, so we decided it was OK to introduce this minor breaking change.

ianhi:
> Reason: this method doesn't exist in IC1,

oops I missed that! I guess that makes it not really a breaking change?

> This is not something we expect users to use on their pipelines, so we decided it was OK to introduce this minor breaking change.

Sounds good to me. But to be thorough, we should include a line about this in the changelog.

> This is the one breaking change we are allowing ourselves to make

As a little treat :)

paraseba:
> But to be thorough, we should include a line about this in the changelog.

Yeah, I'll create a ticket so we don't forget

paraseba:
@evaporei I merged the main branch into this, the only relevant error is about code formatting. Could you please run `ruff` on the code? After that we can merge this PR.

Issue #1427, opened 2025-11-20
No solution when resolving dependencies


evaporei:
### What happened?

I followed the development instructions in: https://icechunk.io/en/latest/contributing/#python-development-workflow and for some reason I get this error in `uv run pytest`:

>   × No solution found when resolving dependencies for split (markers: python_full_version >= '3.13'):
  ╰─▶ Because there is no version of dask==2025.11.0 and distributed==2025.11.1.dev10+gd8e3fd201 depends on dask==2025.11.0, we can
      conclude that distributed==2025.11.1.dev10+gd8e3fd201 cannot be used.
      And because only the following versions of distributed are available:
          distributed<2024.11.0
          distributed==2025.11.1.dev10+gd8e3fd201
      we can conclude that distributed>=2024.11.0 cannot be used.
      And because icechunk[docs] depends on distributed>=2024.11.0 and your project requires icechunk[docs], we can conclude that your
      project's requirements are unsatisfiable.

Is this expected to happen?

I've been able to fix it by moving both to their latest `pypi` packages (both version: `2025.11.0`), instead of GitHub repo: https://github.com/evaporei/icechunk/commit/058d26a692da6c45bd15d13343d6ca9506cc4c93

If that's the correct fix, I can open a PR no problem 🙂 

### What did you expect to happen?

The python package to resolve dependencies properly.

### Minimal Complete Verifiable Example

```Python
Just run `uv run pytest` under latest main (098945d).
```

### MVCE confirmation

- [ ] Minimal example — the example is as focused as reasonably possible to demonstrate the underlying issue in icechunk.
- [ ] Complete example — the example is self-contained, including all data and the text of any traceback.
- [ ] Verifiable example — the example copy & pastes into an IPython prompt, returning the result.
- [ ] New issue — a search of GitHub Issues suggests this is not a duplicate.
- [ ] Recent environment — the issue occurs with the latest version of icechunk and its dependencies.

### Relevant log output

```Python

```

### Environment

Not applicable.

ianhi:
Thanks for the report @evaporei

This has been dogging our upstream tests for a little bit:

https://github.com/earth-mover/icechunk/actions/runs/19520479260/job/55882612787

and I wasn't seeing it locally due to an old uv lock file.

I suspect, though haven't confirmed, that it's due dask and distributed switching how they do git based versioning the most recent releases: https://github.com/dask/dask/pull/12133

The quick solution to get moving is to comment out these lines from the `upstream` group.

[...]

PR #1429, opened 2025-11-21
simplify upstream dependecies to avoid resolution errors

ianhi:
Solution for https://github.com/earth-mover/icechunk/issues/1427

This will temporarily leave us not getting upstream dask. But that will be fixed once:

https://github.com/dask/dask/pull/12168 and https://github.com/dask/dask/issues/12166 come through

### DIFF
--- a/.github/setup-failure-template.md
+++ b/.github/setup-failure-template.md
@@ -1,0 +1,22 @@
+# Nightly Upstream Dependency Installation Failed
+
+**Workflow:** {{WORKFLOW}}
+**Run:** [{{RUN_ID}}]({{RUN_URL}})
+**Date:** {{DATE}}
+
+## Issue
+
+The upstream dependency installation step failed during the nightly CI run.
+
+This likely indicates an issue with:
+- The nightly wheels repository
+- Dependency resolution conflicts
+- Version incompatibilities between upstream packages
+
+## Setup Output
+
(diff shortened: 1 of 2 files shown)


ianhi:
I took this a little bit further and moved everythign from optional-depencies to PEP-753 dependency groups: https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups

All of our optional dependencies were actually just dev dependencies, which is an explicit use case for these dependency groups.

these groups have been supported by `uv` for a long time, and by `pip` since april 2025 so i think we are ok to use them. You can eitehr specify them with `--group` or with the same bracket syntax as for extras. `-e ".[dev]"`

ianhi:
@paraseba can we fix these unrelated mypy errors, which i think are from the new zarr release, in a separate PR?

paraseba:
Yes @ianhi as long as we can work on it very soon, because it's going to block PR merging. Let me know if you want me to merge this one

ianhi:
ok great thanks! As soon as everythign else passes lets merge. We unfortunately missed the upstream mypy checks catching this before the zarr release because they were failing due to the dask resolution errors. Which was a failure mode that did not make the bot open an issue. the last commit here should fix that, resulting in an issue being opened/updated if we fail on the install step.

PR #1433, opened 2025-11-21
fix typing errors

ianhi:
after https://github.com/zarr-developers/zarr-python/pull/3304 we need to be specific about if we have V2 or V3 arrays. So i've constrained the typing to be V3 only

@paraseba quickly, as requested :)


ianhi:
I thought it would be harder! That's why i left it out. alas it was easy once I did the barest minimum of reading release notes

ianhi:
it was hard in the end :(

PR #1442, opened 2025-11-27
Backport: fix typing errors

paraseba:
This change: https://github.com/earth-mover/icechunk/commit/dcbd54368c509ba93041cb0da34fcbc1ba962632

after https://github.com/zarr-developers/zarr-python/pull/3304 we need to be specific about if we have V2 or V3 arrays

### DIFF
--- a/icechunk-python/python/icechunk/dask.py
+++ b/icechunk-python/python/icechunk/dask.py
@@ -1,6 +1,6 @@
 import functools
 from collections.abc import Callable, Mapping
-from typing import Any, ParamSpec, TypeAlias, TypeVar
+from typing import TYPE_CHECKING, Any, ParamSpec, TypeAlias, TypeVar
 
 import numpy as np
 from packaging.version import Version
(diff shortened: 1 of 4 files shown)


ianhi:
Did a strict backport not work? I think that some of the extra things in the original were important to avoid the real failures here.

paraseba:
@ianhi I don't think I missed anything from the original commit

icechunk-python/python/icechunk/dask.py
```diff
@@ -57,7 +60,7 @@ def _assert_correct_dask_version() -> None:
 def store_dask(
     *,
     sources: list[Array],
-    targets: list[zarr.Array],
+    targets: list[zarr.Array[ArrayV3Metadata]],
```
ianhi, line 63: in #1433 this, and others, are wrapped in quotes to avoid the import failures

paraseba: that's weird ... i didn't do this manually. Anyway, I can try this, thank you!

ianhi: linter going crazy?

paraseba:
@ianhi did it better

Issue #180, opened 2024-10-10
Enable mypy everywhere


paraseba:
In particular, we need to mypy the tests directory, which is the main source of API usage in the codebase.

agrigoriev:
Should it be closed?

paraseba:
If I remember correctly we still need to mypy the benchmarks directory

PR #1443, opened 2025-11-28
mypy tests

evaporei:
Solves #180 

Technically there's still the `benchmarks`, `integration_tests` and `examples` folders, but this PR got too big for that.

Most of the `type: ignored` or `Any` can be solved with improvements in type annotations in `zarr` and other dependencies (eg: `hypothesis`).

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -40,7 +40,7 @@ cd icechunk-python
     uv run pytest
 
     # Run type checking
-    uv run mypy python
+    uv run mypy python tests
 
     # Run linting
     uv run ruff check python
--- a/icechunk-python/README.md
+++ b/icechunk-python/README.md
@@ -19,7 +19,7 @@ source .venv/bin/activate
 uv run pytest
 
 # Run type checking
-uv run mypy python
+uv run mypy python tests
 
 # Run linting
 uv run ruff check python


paraseba:
omg this is amazing @evaporei !! Thank you! I'll review tomorrow in detail.

icechunk-python/tests/test_dask.py
```diff
@@ -124,5 +124,8 @@ def test_fork_session_deep_copies(scheduler, any_spec_version: int | None) -> No
             session.merge(*forks)
             session.commit("yaya writes succeeded")
 
-            actual = xr.open_dataset(repo.readonly_session("main").store, engine="zarr")
+            actual = xr.open_dataset(
+                repo.readonly_session("main").store,  # type: ignore [arg-type]
```
paraseba, line 128: I wonder what the issue is here, do we need to fix xarray typing?

icechunk-python/tests/test_timetravel.py
```diff
@@ -428,15 +428,16 @@ async def test_timetravel_async(using_flush: bool, any_spec_version: int | None)
 
     session.discard_changes()
     assert not session.has_uncommitted_changes
-    assert air_temp[200, 6] == 54
+    # I don't understand why I need to ignore here
+    assert air_temp[200, 6] == 54  # type: ignore [unreachable]
```
paraseba, line 432: so weird

.github/workflows/code-quality.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
         shell: bash
         working-directory: icechunk-python
         run: |
-          uvx --with mypy --with zarr --with hypothesis --with xarray --with dask mypy python
+          uv run mypy python tests
```
ianhi: I know that `uvx`  is far less ergonomic, but it's used here to avoid this workflow building the icechunk library, which unfortunately `uv run` forces :(

I'd love to find a better solution that uses our define dependencies in pyproject.toml but doesn't spend a bunch of time building icechunk

evaporei: I imagined that was the case, now Code Quality Checks is the slowest part of CI 😢 

9 min vs 1, 2 or 4
スクリーンショット 2025-11-28 13 22 49

+5 new minutes to install it
スクリーンショット 2025-11-28 13 22 28

evaporei: I can try to make a `uvx` equivalent of the above to see if we get any significant performance improvements.

Although whenever a new dependency gets added to `pyproject.toml`, this script will need to be updated too 😕

paraseba: Let's use uvx, and add a comment in the code explaining why we need it

.github/workflows/code-quality.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
         shell: bash
         working-directory: icechunk-python
         run: |
-          uvx --with mypy --with zarr --with hypothesis --with xarray --with dask mypy python
+          uv run --no-project mypy python tests
```
ianhi: ❤️ if this works. very slick. the power of actually reading the docs :)

evaporei: It didn't I've updated the commit message with the CI error, I'm trying a second one lol

evaporei: Third time is the charm, `--all-groups` worked 🎉  

Although it says it has built `icechunk-python` 🤔 it seems to be faster than usual: 2m 42s 🕙 

```
 ...
 Downloaded mypy
 Downloaded mkdocs-material
      Built icechunk @ file:///home/runner/work/icechunk/icechunk/icechunk-python
Installed 262 packages in 537ms
Success: no issues found in 53 source files
```

evaporei: I found out about `--no-build` too, but that didn't work.

```
$ uv run --all-groups --no-build mypy python tests
error: Distribution `icechunk @ editable+.` can't be installed because it is marked as `--no-build` but has no binary distribution
```

It seems `--all-groups` is the best we can get, it's at the same speed of the slowest CI job (Rust Ubuntu). Last build from `main`:

```
Rust CI / Rust CI (ubuntu-24.04-arm) (push) Successful in 5m
```

ianhi: 🤔 I don't really see how installing more deps speeds this up. maybe we got lucky with some caching? But happy to run with this for now and remove later if it becomes an issue.

> It seems --all-groups is the best we can get, it's at the same speed of the slowest CI job (Rust Ubuntu). Last build from main:

it's actually slightly faster than the slowest! the python checks are broken into a build and then a test step that installs from the wheel (to mimic real life usage) so they look faster than they actually are.

paraseba:
@ianhi do we have a plan to deal with the remaining mypy errors?

icechunk-python/tests/test_concurrency.py
```diff
@@ -17,15 +18,17 @@
 
 
 async def write_to_store(
-    array: zarr.Array, x: int, y: int, barrier: asyncio.Barrier
+    array: zarr.Array[Any], x: int, y: int, barrier: asyncio.Barrier
```
ianhi: I think if we wrap these in quotes we might be able to escape the errors we are seeing in the min deps test cases:
```
 tests/test_concurrency.py:21: in <module>
    array: zarr.Array[Any], x: int, y: int, barrier: asyncio.Barrier
           ^^^^^^^^^^^^^^^
E   TypeError: type 'Array' is not subscriptable
```

evaporei: You were correct, the minimum test CI job is working now 🙂

evaporei:
@paraseba @ianhi I've done a rebase, everything should be good now. I just don't understand why the Rust jobs are failing with:

> Error: ICError { kind: StorageError(S3PutObjectError(ConstructionFailure(ConstructionFailure { source: InterceptorError { kind: ReadBeforeExecution, interceptor_name: Some("PutObjectEndpointParamsInterceptor"), source: Some(BuildError { kind: MissingField { field: "bucket", details: "A required field was not set" } }) } }))), context: SpanTrace [] }

.github/workflows/code-quality.yaml
evaporei: You were right on this as well, I just done a rebase and that step went up to 7 min.

paraseba:
Don't worry about those @evaporei, it's just that you don't have the permissions needed to execute those integration tests, they need secrets to different object stores.

This looks great!

paraseba:
Thank you @evaporei !!!

icechunk-python/tests/test_timetravel.py
evaporei: I saw some related issues in safe-assert, but I don't see how the same would be happening here: https://github.com/wemake-services/safe-assert/pull/131/files

PR #1452, opened 2025-12-01
Document proxy support

TomNicholas:


docs/docs/configuration.md
```diff
@@ -289,3 +289,21 @@ repo = icechunk.Repository.open(
...
+export HTTPS_PROXY=http://proxy.corp.com:8080
+export HTTP_PROXY=http://proxy.corp.com:8080
+```
+
+No code changes are required. Icechunk checks proxy sources in this order:
+1. `HTTPS_PROXY` / `https_proxy` - highest priority
+2. `HTTP_PROXY` / `http_proxy`
+3. `ALL_PROXY` / `all_proxy`
+
+Note that you must set the environment variable at the OS level, not via the python interpreter.
```
paraseba, line 313: should we mention that currently this only works for S3-compatible storage instances?

Issue #1463, opened 2025-12-09
S3 virtual references fail when object keys contain spaces due to URL encoding


elyall:
### What happened?

When using icechunk with S3 virtual references, objects with spaces in their keys fail because icechunk URL-encodes the location, while S3 object keys use literal spaces.

For instance for a virtual reference at "s3://bucket/abc def/image 001.tif", icechunk tries to read "s3://bucket/abc%20def/image%20001.tif" which doesn't exist.

### What did you expect to happen?

I expected the s3 path to not be url encoded and the data to be loaded.

### Minimal Complete Verifiable Example

```Python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "icechunk",
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
#   "virtualizarr",
#   "virtual-tiff",
#   "xarray",
#   "obstore",
# ]
# ///
#   "icechunk@git+https://github.com/earth-mover/icechunk.git@main#subdirectory=icechunk-python", # <- fails to install
#
# This script automatically imports the development branch of icechunk to check for issues.
# Please delete this header if you have _not_ tested this script with `uv run`!

import icechunk
import xarray as xr
from obstore.store import S3Store
from virtual_tiff import VirtualTIFF
from virtualizarr import open_virtual_dataset
from virtualizarr.registry import ObjectStoreRegistry

icechunk.print_debug_info()

file_url = "s3://icechunk-test-s4rads2/path/abc def/image 001.tif"
bucket_url = "s3://icechunk-test-s4rads2/"
REGION = "us-west-2"

# create virtual dataset of file on s3
s3_store = S3Store(
    bucket=bucket_url.lstrip("s3://").rstrip("/"),
    region=REGION,
    skip_signature=True,
)
registry = ObjectStoreRegistry({bucket_url: s3_store})
parser, variable = VirtualTIFF(ifd=0), "0"
vds = open_virtual_dataset(
    url=str(file_url),
    registry=registry,
    parser=parser,
    # loadable_variables=[variable],
)
# validate virtual dataset
for chunk_key, entry in list(vds[variable].data.manifest.dict().items()):
    print(f"Chunk {chunk_key}:")
    print(f"  Path: {entry['path']}")
    print(f"  Offset: {entry['offset']}")
    print(f"  Length: {entry['length']}")
    print(f"  Path starts with bucket_url? {entry['path'].startswith(bucket_url)}")


# save virtual dataset to icechunk
[...]

PR #1473, opened 2025-12-10
Fix icechunk dependency URL in bug report template

TomNicholas:
Addresses @elyall's comment in https://github.com/earth-mover/icechunk/issues/1463#issuecomment-3638375276 by updating the icechunk dependency URL in the bug report template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -37,7 +37,7 @@ body:
         # /// script
         # requires-python = ">=3.11"
         # dependencies = [
-        #   "icechunk",
+        #   "icechunk<2",
         #   "zarr",
         # ]
         #


ianhi:
Can we just point to the nightly wheel instead? I think that would be a better user experience. With something like this:
https://docs.astral.sh/uv/concepts/projects/dependencies/#index

it may end up being uv specific and fall back to pypi for other script runners. but that's fine with me for the convenience of not needing to compile icechunk to make a bug report.

Same for zarr I guess, though that's less important

ianhi:
This is the magic set of settings:

```python
        # /// script
        # requires-python = ">=3.11"
        # dependencies = [
        #   "icechunk",
        #   "zarr",
        # ]
        #
        # [[tool.uv.index]]
        #
        # [tool.uv.sources]
        # zarr = { index = "scientific-python-nightly-wheels" }
        #
        # [tool.uv]
        # prerelease = "allow"
        # ///

```

which does grab icechunk 2.0.0a. So we might want to constrain icechunk to be <2 for now

TomNicholas:
Looks like you finished this @ianhi ? (Thank you!) I can't self-approve, so if you want to approve it then either of us can merge it. (Or maybe @paraseba can just give me god privileges)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -37,9 +37,20 @@ body:
         # /// script
         # requires-python = ">=3.11"
         # dependencies = [
-        #   "icechunk@git+https://github.com/earthmover/icechunk.git@main",
-        #   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
+        #   "icechunk",
```
ianhi:
```suggestion
        #   "icechunk<2",
```
Whoops inforgot this

Issue #1481, opened 2025-12-11
Icechunk 1 repo opened and committed with icechunk 2 alpha can no longer be opened by icechunk 1


mpiannucci:
### What happened?

* Created a repo with icechunk 1
* Opened, wrote data, and committed with icechunk 2 alpha
* tried to open with icechunk 1 again and got this error:

```
{
	"name": "IcechunkError",
	"message": "  x Icechunk cannot read from repository written with a more modern version\n  | \n  | context:\n  |    0: icechunk::asset_manager::fetch_snapshot\n  |            with snapshot_id=WNJ88HTCBBR74NMBJ63G\n  |              at icechunk/src/asset_manager.rs:286\n  |    1: icechunk::asset_manager::snapshot_ancestry\n  |            with snapshot_id=WNJ88HTCBBR74NMBJ63G\n  |              at icechunk/src/asset_manager.rs:402\n  |    2: icechunk::asset_manager::snapshot_info_ancestry\n  |            with snapshot_id=WNJ88HTCBBR74NMBJ63G\n  |              at icechunk/src/asset_manager.rs:386\n  | \n  `-> Icechunk cannot read from repository written with a more modern version\n",
[...]

PR #1482, opened 2025-12-12
Fix spec version for new snapshots on V1 repos.

paraseba:
Also fix metadata serialization for V1 repos.

Fixes: #1481


icechunk/src/format/repo_info.rs
```diff
@@ -263,7 +268,7 @@ impl RepoInfo {
             branches: Some(branches),
             deleted_tags: Some(deleted_tags),
             snapshots: Some(snapshots),
-            spec_version: SpecVersionBin::current() as u8,
+            spec_version: spec_version as u8,
```
dcherian, line 271: lol i should've spotted that yesterday!

dcherian (approved):

PR #1483, opened 2025-12-12
Extend storage stats calculation to include virtual and inline chunks

### LINKED ISSUE #1478: Add a way to compute total size of all virtual chunks
There is currently no way to compute the total size of all chunks (virtual and native) in a repo, because the existing method [repo.total_chunks_storage](https://icechunk.io/en/latest/reference/#icechunk.Repository.total_chunks_storage_async) excludes virtual (and inlined) chunks. We should add a way to calculate this.

Not sure what the API should be though... We want an easy way to calculate total storage for virtual + native or just native. But it's an expensive operation so we don't want to call two versions of `repo.total_chunks_storage` concurrently - they are doing exactly the same work traversing the manifests.

Returning both numbers from one method could be done efficiently, but that's a breaking change to the return type of the existing API.

TomNicholas:
Closes #1478.

Compiles (at least the core rust crate does), but I haven't tested it yet.

EDIT: Ready

### DIFF
--- a/.github/workflows/dependency-check.yaml
+++ b/.github/workflows/dependency-check.yaml
@@ -22,7 +22,7 @@ env:
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10
   RUST_CHANNEL: '1.91.0'
-  CARGO_DENY_VERSION: '0.18.4'
+  CARGO_DENY_VERSION: '0.18.9'
 
 jobs:
   check-dependencies:
--- a/.github/workflows/publish-rust-library.yml
+++ b/.github/workflows/publish-rust-library.yml
@@ -9,7 +9,7 @@ env:
   RUSTUP_MAX_RETRIES: 10
   # update RUST_CHANNEL and CARGO_DENY in sync
   RUST_CHANNEL: '1.91.0'
-  CARGO_DENY_VERSION: '0.18.4'
+  CARGO_DENY_VERSION: '0.18.9'
 
 on:
   workflow_dispatch:
--- a/.github/workflows/python-ci.yaml
+++ b/.github/workflows/python-ci.yaml
@@ -352,7 +352,7 @@ jobs:
           merge-multiple: true
 
       - name: Upload nightly wheels
-        uses: scientific-python/upload-nightly-action@b36e8c0c10dbcfd2e05bf95f17ef8c14fd708dbf
+        uses: scientific-python/upload-nightly-action@5748273c71e2d8d3a61f3a11a16421c8954f9ecf
         with:
           artifacts_path: dist
           anaconda_nightly_upload_token: ${{ secrets.ANACONDA_ORG_UPLOAD_TOKEN }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1288,9 +1288,9 @@ checksum = "3a3076410a55c90011c298b04d0cfa770b00fa04e1e3c97d3f6c9de105a03844"
 
 [[package]]
 name = "flatbuffers"
-version = "25.9.23"
+version = "25.12.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09b6620799e7340ebd9968d2e0708eb82cf1971e9a16821e2091b6d6e475eed5"
+checksum = "35f6839d7b3b98adde531effaf34f0c2badc6f4735d26fe74709d8e513a96ef3"
 dependencies = [
  "bitflags 2.10.0",
  "rustc_version",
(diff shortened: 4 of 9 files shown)


icechunk/src/ops/stats.rs
```diff
@@ -10,37 +11,138 @@ use tracing::trace;
...
+                    insert_and_increment_size_if_new(
+                        &seen_virtual_checksums,
+                        checksum.clone(),
+                        virtual_ref.length,
+                        &mut virtual_bytes,
+                    )?;
+                } else {
+                    // No checksum: deduplicate by (location, offset, length)
+                    let virtual_identifier = (
+                        virtual_ref.location.clone(),
```
TomNicholas: FYI I think it's possible to remove these clones (by passing a reference and then only actually cloning if needed), but it comes at the cost of an additional if. Maybe worth it though to minimize memory usage?

icechunk-python/src/stats.rs
```diff
@@ -0,0 +1,66 @@
...
+impl PyChunkStorageStats {
+    /// Total bytes excluding virtual chunks.
+    ///
+    /// This represents the approximate size of all objects stored in the
+    /// icechunk repository itself (native chunks plus inline chunks).
+    /// Virtual chunks are not included since they reference external data.
+    ///
+    /// Returns:
+    ///     int: The sum of native_bytes and inlined_bytes
+    pub fn non_virtual_bytes(&self) -> u64 {
```
paraseba: nitpicking but, I wonder if this is the best approach. Should we instead create this with a `ChunkStorageStats` and delegate all methods and properties?

icechunk/src/ops/stats.rs
```diff
@@ -10,37 +11,138 @@ use tracing::trace;
...
+                insert_and_increment_size_if_new(
+                    &seen_native_chunks,
+                    chunk_ref.id,
+                    chunk_ref.length,
+                    &mut native_bytes,
+                )?;
+            }
+            Ok(ChunkPayload::Virtual(virtual_ref)) => {
+                // Deduplicate by checksum if available, otherwise by (location, offset, length)
+                if let Some(checksum) = &virtual_ref.checksum {
```
paraseba: I'm very unsure about the viability of using checksum for uniqueness.  For example, for timestamps as checksum we are using second resolution. Other example, some object stores may use very low entropy etags.

TomNicholas: If it is only a ballpark estimate I think that's okay for virtual chunks... 

OTOH we could always use the (url, offset, length) identifier? That might actually be more unique?

icechunk-python/src/stats.rs
TomNicholas: You mean like delegate the addition of two numbers to a wrapped class?

Or inherit the methods instead (or whatever rust's equivalent is)?

paraseba: Sorry, the former. Having this class wrap over the pure rust one, and delegate all its methods. I'm ok with it one way or another.

icechunk/src/ops/stats.rs
paraseba: I think always using (url, offset, length) is the safer approach

TomNicholas: done in https://github.com/earth-mover/icechunk/pull/1483/commits/68589c6e627c4771bdcd9836ac56cc8c8d794977

icechunk-python/src/stats.rs
TomNicholas: done in https://github.com/earth-mover/icechunk/pull/1483/commits/03fbac5026f6f75735e8b55b1094db868f971121, using composition

icechunk/src/ops/stats.rs
TomNicholas: I just left a todo for this. It might be worthwhile to run it on a really big virtual repo before merging just to see what the memory usage is like.

TomNicholas:
I added some tests, including of virtual chunk deduplication.

icechunk-python/src/repository.rs
```diff
@@ -2215,59 +2216,62 @@ impl PyRepository {
         })
     }
 
-    pub(crate) fn total_chunks_storage(
```
TomNicholas, line 2218: I didn't bother trying to do a deprecation cycle in the rust crate.

icechunk-python/python/icechunk/repository.py
```diff
@@ -1642,11 +1719,20 @@ async def total_chunks_storage_async(
...
-        return await self._repository.total_chunks_storage_async(
+        warnings.warn(
+            "The ``total_chunks_storage_async`` method has been deprecated in favour of the ``chunk_storage_stats_async`` method. " \
+            "The new method is superior, as it actually calculates storage size occupied by inlined and virtual chunks in addition to native chunks. " \
+            "You can still access just the total native bytes: to keep your existing behaviour using API that will not be removed in a future version, " \
+            "please replace your existing ``.total_chunks_storage_async(**kwargs)`` method call with ``.chunk_storage_stats_async(**same_kwargs).native_bytes``.",
+            DeprecationWarning,
+        )
+        
+        stats = await self._repository.chunk_storage_stats(
```
paraseba: I think we need to call the `_async` version here?

icechunk-python/src/repository.rs
```diff
@@ -2215,59 +2216,62 @@ impl PyRepository {
...
+    pub(crate) fn chunk_storage_stats(
```
paraseba, line 2219: I think you are missing to mirror this change in the `.pyi` file.

paraseba (approved):

Issue #1487, opened 2025-12-15
Make S3Storage get_client pub


d0rianb:
Is there any particular reason for this method to be private  : 
https://github.com/earth-mover/icechunk/blob/4ce8e672c61821272d8301252ad18dc843579f02/icechunk/src/storage/s3.rs#L246

It would be nice to be able to use it to retrieve the s3 client from the storage, to benefit from the init_or_get single initalization

paraseba:
I'm OK with this. Would you like to provide a PR @d0rianb ?

paraseba:
Thanks for the PR @d0rianb, it has been merged. If you need this feature in Icechunk 1.x, we will have to port the change to the `support/v1.x` branch.

Thank you.

d0rianb:
Thanks, v2 will be be good for me

PR #1489, opened 2025-12-16
Made get_client and get_path_str public (#1487)

d0rianb:
Expose the two methods, so the cached s3 clients can be reused.


paraseba:
This fixes #1487

paraseba (approved):

PR #1490, opened 2025-12-17
added support for __str__, __repr__ and _repr_html_

### LINKED ISSUE #1208: Nicer python reprs (string and HTML)

jayendra13:
Improves Python repr for core classes

- Add __repr__, __str__, _repr_html_ to Repository, Session, IcechunkStore
- Add display helpers in icechunk/src/display.rs

  Fixes #1208

### DIFF
--- a/icechunk-python/src/repository.rs
+++ b/icechunk-python/src/repository.rs
@@ -229,7 +229,7 @@ impl From<SnapshotInfo> for PySnapshotInfo {
 impl PyManifestFileInfo {
     pub(crate) fn __repr__(&self) -> String {
         dataclass_repr(
-            "ManifestFileInfo",
+            "icechunk.ManifestFileInfo",
             &[
                 ("id", format!("\"{}\"", self.id)),
                 ("size_bytes", self.size_bytes.to_string()),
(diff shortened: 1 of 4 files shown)


paraseba:
Wow this is looking great @jayendra13 !!

paraseba:
Don't worry about the Rust failures, that's because you don't have permissions to run tests that require secrets. But you do need to run `ruff` to fix the code quality test.

Also, it would be great to have some minimal testing

jayendra13:
Yes, tests are the next things, I am going to work on, created a draft PR to know weather the efforts are in correct directions or not and I am covering the necessary classes or not. Would love to hear any feedbacks.
Let me run the ruff as well.

jayendra13:
Amended the commit to handle the rust linting failures.

paraseba:
This looks great @jayendra13 !! Feel free to flag as ready for review. We'll have to wait a few days until @TomNicholas has some time to review, he has more context on this topic than I do. But we'll probably merge soon.

Thank you!

jayendra13:
Will work on test cases here soon, if we want them in the same PR.

paraseba:
Same PR is preferred @jayendra13 . We don't need to be very thorough for this, just a basic test or two. Thanks again!

jayendra13:
Added simple regex and string matching based python test cases for three classes 
- Repository
- Store
- Session

with only InMemory storage use-case.

paraseba:
rebasing will get rid of the dependency check error

jayendra13:
Rebased.

ianhi:
Hi @jayendra13 I'd like to pick up reviewing here. This is a super useful feature that we're excited to include. Are you. able to rebases once more? Alternatively I'm happy to handle the rebase on main so we can move. forward here.

jayendra13:
Done

PR #1493, opened 2025-12-18
Fix anonymous parameter not passed to S3Options in storage functions

### LINKED ISSUE #1183: anonymous=True option appears to be ignored in s3_storage configuration
Hi everyone,

I think there's a small bug in how anonymous=True is handled.

Here's what I ran:

```python 
storage = icechunk.s3_storage(
    bucket='pythia',
    prefix='radar/ams2025/CASSM.zarr',
    endpoint_url='https://js2.jetstream-cloud.org:8001',
    anonymous=True,
    region='us-east-1',
    force_path_style=True
)

print(storage)
```
But the output shows:

```python 
S3Storage(bucket=pythia, prefix=radar/ams2025/CASSM.zarr, config=S3Options(region=us-east-1, endpoint_url=https://js2.jetstream-cloud.org:8001, anonymous=false, allow_http=false, force_path_style=true))
```

So `anonymous=False` is being set, even though I explicitly passed `anonymous=True`.

Let me know if you need anything else.

aladinor:
- [x] Closes #1183 

The anonymous parameter was accepted by s3_storage, s3_store, s3_object_store_storage, tigris_storage, and r2_storage functions but not passed to S3Options, causing __repr__ to show anonymous=false even when anonymous=True was specified.

Now it is showing `anonymous=True` when expected.

### DIFF
--- a/icechunk-python/python/icechunk/storage.py
+++ b/icechunk-python/python/icechunk/storage.py
@@ -103,7 +103,7 @@ def s3_store(
         force_path_style=force_path_style,
         network_stream_timeout_seconds=network_stream_timeout_seconds,
         requester_pays=requester_pays,
-        anonymous=anonymous or False,
+        anonymous=anonymous,
     )
     return (
         ObjectStoreConfig.S3Compatible(options)


icechunk-python/python/icechunk/storage.py
```diff
@@ -103,6 +103,7 @@ def s3_store(
         force_path_style=force_path_style,
         network_stream_timeout_seconds=network_stream_timeout_seconds,
         requester_pays=requester_pays,
+        anonymous=anonymous or False,
```
paraseba: I think you can just say `anonymous=anonymous`? Is there any advantage to adding `or False` if the variable is of type `bool`?

aladinor: Thanks @paraseba for your comment. The test will fail as it does not accept `None`. Passing None directly to S3Options() causes:  `TypeError: argument 'anonymous': 'NoneType' object cannot be cast as 'bool'`

paraseba: Ah I see, but that's only in the functions that take `anonymous: bool | None`, but some of these take `anonymous: bool`. Let's pass anonymous only in the second type, that shouldn't fail

paraseba:
```suggestion
        anonymous=anonymous,
```

paraseba:
Thank you @aladinor

PR #1498, opened 2025-12-19
Alternative way to do manifest split without caching the splits

paraseba:


icechunk/src/session.rs
```diff
@@ -5205,6 +5140,101 @@ mod tests {
...
+    // 2. Splits are cached for the array with extent [0..5)
+    // 3. Session A writes chunk [3]
+    // 4. Meanwhile, another session resizes array to [20, 1] and writes chunk [10]
+    // 5. Session A rebases to the new snapshot
+    // 6. BUG: Session A's splits are NOT updated (still [0..5))
+    // 7. Session A commits - during flush, verified_node_chunk_iterator filters
+    //    parent chunks by extent, dropping chunk [10] because it's outside [0..5)
+    // 8. Result: chunk [10] is lost
+    //
+    // The fix requires updating Session.splits during rebase to match the new
```
```suggestion
    // With IC1, doing this correctly requires updating Session.splits during rebase to match the new
```

icechunk/src/session.rs
```diff
@@ -2346,9 +2267,16 @@ async fn do_flush(
 
     // Now we need to go through all the new arrays, and generate manifests for them
 
-    for (node_path, node_id) in flush_data.change_set.new_arrays() {
+    for (node_path, node_id, array_data) in flush_data.change_set.new_arrays() {
+        // FIXME: do we need to cache this?
```
dcherian: don't think so since we iterate over each node once and query get_split_sizes once.

icechunk/src/session.rs
```diff
@@ -2096,29 +1985,50 @@ impl<'a> FlushProcess<'a> {
         existing_manifests: Vec<ManifestRef>,
         old_snapshot: &Snapshot,
         rewrite_manifests: bool,
+        splits: &ManifestSplits,
     ) -> SessionResult<()> {
-        #[allow(clippy::expect_used)]
-        let splits =
-            self.splits.get(&node.id).expect("splits should exist for this node.");
+        let init: HashMap<ManifestExtents, Vec<ChunkInfo>> = Default::default();
+        // FIXME: this duplicates chunk storage for the array
```
dcherian: yeah IIUC this loads *all* on-disk manifests to memory instead of simply rewriting the ones that need to be rewritten. if so, that should not be needed. We just need to load the existing manifest for a split in line 2030

icechunk/src/session.rs
```diff
@@ -2096,29 +1985,50 @@ impl<'a> FlushProcess<'a> {
...
+
         let mut refs =
             HashMap::<ManifestExtents, Vec<ManifestRef>>::with_capacity(splits.len());
 
         let on_disk_extents =
             existing_manifests.iter().map(|m| m.extents.clone()).collect::<Vec<_>>();
 
-        let modified_splits = self
+        //let modified_splits = classified_chunks.keys().collect::<HashSet<_>>();
+        // FIXME: this is another pass through all chunks
```
dcherian: but this is a pass over all chunks _in the changeset_, so that's fine. IIRC our goal was to have flush scale with size of Changeset, not size of array

icechunk/src/session.rs
```diff
@@ -2062,16 +1950,17 @@ impl<'a> FlushProcess<'a> {
...
-            self.splits.get(node_id).expect("getting split for node unexpectedly failed");
-
         for extent in splits.iter() {
-            if self.change_set.array_manifest(node_id, extent).is_some() {
+            if self.change_set.array_manifest(node_id).is_some() {
                 let chunks = stream::iter(
                     self.change_set
-                        .new_array_chunk_iterator(node_id, node_path, extent.clone())
+                        .new_array_chunk_iterator(node_id, node_path)
+                        // FIXME: do we need to optimize this so we don't need multiple passes over all chunks calling
```
dcherian, line 1997: we should just do the groupby into hashmaps once outside, and pass the appropriate iterator in?

icechunk/src/session.rs
paraseba: ups you're right, this is wrong in that way

icechunk/src/session.rs
dcherian: So basically reconstruct the `HashMap<ManifestExtents, ChunkTable>` that we used to store on the ChangeSet earlier. Then I think the rest of the logic will be easy to adapt.

icechunk/src/session.rs
```diff
@@ -2132,88 +2009,178 @@ impl<'a> FlushProcess<'a> {
...
+        //         * create a new manifest with all chunks in `extent` from all
+        //           `intersecting_manifests`, overriding chunks with those coming from the change
+        //           set, add the new manifest to the list of refs
+        //     * else (no changes in this session to `extent`)
+        //         * for each intersecting manifest:
+        //             * if it's fully contained in the extent, we can reuse it, just add it to th
+        //             elist of refs
+        //             * else create a new manifest filtering out the chunks that are outside of
+        //             the extent
+        let updated_chunks_by_extent: HashMap<ManifestExtents, ChunkTable> = self
```
dcherian, line 2103: For perf we could fold in the `aggregate_extents` detection with this pass over the chunks.

paraseba: aggregate_extents is lazy, accumulating as the stream is navigated to construct the manifest. I don't think it would make any difference?

dcherian: nope makes no sense since this is only _updated_ chunks

icechunk/src/session.rs
```diff
@@ -2132,88 +2009,178 @@ impl<'a> FlushProcess<'a> {
...
+            if let Some(payload) = maybe_payload.as_ref() {
+                all_chunks_vec.push(Ok(ChunkInfo {
+                    node: node_id.clone(),
+                    coord: idx.clone(),
+                    payload: payload.clone(),
+                }));
+            }
+        }
+
+        self.write_manifest_from_iterator(stream::iter(all_chunks_vec).err_into()).await
```
dcherian, line 2060: could stream here instead of realizing a Vec. Is this where you had lifetime issues?

paraseba: Yeah, basically, I'd have to replace the mutable for loops by a bunch of `then`, `filter_map` and others. I tried and failed. I think what is to blame is `Manifest.iter()` which is one of these weird things that take self as an Arc.

We should revisit if we find problems once we go into benchmarking.

icechunk/src/session.rs
```diff
@@ -5561,6 +5552,101 @@ mod tests {
...
+    // 5. Session A rebases to the new snapshot
+    // 6. BUG: Session A's splits are NOT updated (still [0..5))
+    // 7. Session A commits - during flush, verified_node_chunk_iterator filters
+    //    parent chunks by extent, dropping chunk [10] because it's outside [0..5)
+    // 8. Result: chunk [10] is lost
+    //
+    // With IC1, doing this correctly requires updating Session.splits during rebase to match the new
+    // parent snapshot's array shapes.
+    async fn test_rebase_over_resize() -> Result<(), Box<dyn Error>> {
+        struct YoloSolver;
```
li-em, line 5572: 😹

PR #1512, opened 2025-12-23
Clean up stateful tests.

dcherian:
I upstreamed a bunch of these "transitions" a while back.


dcherian:
Looks like we'll need to bump minimum zarr version to 3.0.9 for https://github.com/zarr-developers/zarr-python/pull/3130

ianhi:
can you remove even more given that https://github.com/zarr-developers/zarr-python/pull/3161 was merged?

ianhi (approved):
One question re whether mroe can be removed. other 👍 from me

PR #1514, opened 2025-12-23
Allow reindex_array on V1 repos

dcherian:


paraseba (approved):

malosellier:
Hello, does this PR prefigure the support for this feature in the next v1 (v1.1.17?) of Icechunk? Or will it be available only in version 2?
Anyway, thank you for the effort!

paraseba:
It's still undecided if we'll support this operation in Icechunk 1 @malosellier

PR #1515, opened 2025-12-23
initial version compat tests

ianhi:
This has a lot TODOs still in it. like using hypothesis and actually writing more tests. But opening to get feedback on teh general structure @paraseba @dcherian 

and to see if it runs properly in CI

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -619,7 +619,8 @@ class VersionControlStateMachine(RuleBasedStateMachine):
 
     # TODO: v1 has bugs in expire_snapshots, only test for v2
     # https://github.com/earth-mover/icechunk/issues/1520
-    @precondition(lambda self: bool(self.model.commits) and self.model.spec_version >= 2)
+    # https://github.com/earth-mover/icechunk/issues/1534
+    @precondition(lambda self: bool(self.model.commits))
     @rule(
         data=st.data(),
         delta=st.timedeltas(
@@ -670,6 +671,20 @@ class VersionControlStateMachine(RuleBasedStateMachine):
 
         assert self.initial_snapshot.id not in actual
         assert actual == expected.expired_snapshots, (actual, expected)
+
+        # Check that expired snapshots are actually removed from ancestry
+        # TODO: this should have failed earlier on the model step. the model to
+        # need to fix the model here as well
+        remaining_snapshot_ids = set()
+        for branch in branches_after:
+            for snap in self.repo.ancestry(branch=branch):
+                remaining_snapshot_ids.add(snap.id)
+        expired_but_remaining = actual & remaining_snapshot_ids
+        note(expired_but_remaining)
+        assert (
+            not expired_but_remaining
+        ), f"Snapshots marked as expired but still in ancestry: {expired_but_remaining}"
+
         assert (
             actual_deleted_branches == expected.deleted_branches
         ), f"deleted branches mismatch: actual={actual_deleted_branches}, expected={expected.deleted_branches}"


dcherian:
Should we just install `icechunk_v1` in all CI envs instead of a subpackage?

ianhi:
> Should we just install icechunk_v1 in all CI envs instead of a subpackage?

This was lower friction to start. but i guess i don't really see an issue with your proposal. then i can just wrap this into the existing stateful tests

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -301,14 +303,12 @@ def __init__(self) -> None:
...
 
     @initialize(data=st.data(), target=branches)
     def initialize(self, data: st.DataObject) -> str:
-        # FIXME: currently this test is IC2 only
-        spec_version = data.draw(
-            st.one_of(st.integers(min_value=2, max_value=2), st.none())
-        )
-        self.repo = Repository.create(in_memory_storage(), spec_version=spec_version)
+        self.storage = in_memory_storage()
+        self.repo = Repository.create(self.storage, spec_version=1)
```
dcherian, line 372: let's draw `spec_version` too, so we sometimes start from a V2 repo. i'd use `sampled_from([1, 2])` or `sampled_from([1, 1, 2])` so that V1 is prioritized and we preferentially exercise the upgrade path

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -348,6 +349,31 @@ def set_doc(self, path: str, value: Buffer) -> None:
             with pytest.raises(IcechunkError, match="read-only store"):
                 self.sync_store.set(path, value)
 
+    @rule()
+    @precondition(lambda self: len(self.model.commits) > 5)
```
dcherian: let's instead explicitly check the spec version

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -348,6 +349,31 @@ def set_doc(self, path: str, value: Buffer) -> None:
...
+    def upgrade_spec_version(self):
+        icechunk.upgrade_icechunk_repository(self.repo)
```
dcherian, line 418: alternatively catch the error here if expected.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -362,6 +388,63 @@ def commit(self, message: str) -> str:
         self.commit_times.append(snapinfo.written_at)
         return commit_id
 
+    @rule(message=st.text(max_size=MAX_TEXT_SIZE), target=commits)
+    @precondition(lambda self: self.model.changes_made)
```
dcherian: add a spec_version precondition here

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -348,6 +349,31 @@ def set_doc(self, path: str, value: Buffer) -> None:
             with pytest.raises(IcechunkError, match="read-only store"):
                 self.sync_store.set(path, value)
 
+    @rule()
+    @precondition(lambda self: len(self.model.commits) > 5)
+    def upgrade_spec_version(self):
```
dcherian, line 415: let's add this step to `test_zarr/test_stateful.py`

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -362,6 +388,63 @@ def commit(self, message: str) -> str:
...
+        Amend requires spec_version >= 2. For spec_version=1, it raises an error.
+        """
+        branch = self.session.branch
+        assert branch is not None
+        note(
+            f"Amending commit on branch {branch!r} (spec_version={self.model.spec_version})"
+        )
+
+        # Amend is only supported on spec_version >= 2
+        if self.model.spec_version == 1:
```
dcherian: prefer just making a precondition to avoid this; and use a unit test for this error (its trivial)

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -301,14 +303,12 @@ def __init__(self) -> None:
         note("----------")
         self.model = Model()
```
dcherian, line 366: it's be nice to parameterize this over `Model` and an icechunk V1 repo (created with icechunk_v1).

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -114,12 +134,11 @@ def __init__(self, **kwargs: Any) -> None:
...
         self.ondisk_snaps: dict[str, CommitModel] = {}
         self.tags: dict[str, TagModel] = {}
-        # TODO: This is only tracking the HEAD,
-        # Should we model the branch as an ordered list of commits?
-        self.branches: dict[str, str] = {}
+        self.branches: dict[str, BranchModel] = {}
 
         # a tag once created, can never be recreated even after expiration
         self.created_tags: set[str] = set()
+        self.spec_version: Literal[1, 2] | None = None
```
dcherian, line 141: Conceptually this model should not need to care about `spec_version`. It's an abstract simplified model of an icechunk repo.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -170,15 +224,24 @@ def checkout_commit(self, ref: str) -> None:
 
     def create_branch(self, name: str, commit: str) -> None:
         assert commit in self.commits
-        self.branches[name] = commit
+        self.branches[name] = BranchModel(commits=[commit])
```
dcherian, line 227: Probably doesn't matter but the Branch should inherit the ancestry of `commit` too, right? If so, let's add a comment saying that detail doesn't matter for the test as written.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -101,6 +102,25 @@ class TagModel:
     # message: str | None
 
 
+@dataclass
+class BranchModel:
```
dcherian, line 106: nice! though isn't this basically `list` with an extra `head` property?

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -157,7 +181,37 @@ def commit(self, snap: SnapshotInfo) -> None:
...
-        self.branches[self.branch] = ref
+        self.branches[self.branch].append(ref)
+
+    def amend(self, branch: str, snap: SnapshotInfo) -> None:
+        """Amend the HEAD commit on the given branch."""
+        old_head = self.branches[branch].head
+        old_commit = self.commits[old_head]
+        parent_id = old_commit.parent_id
+
+        # Only remove old commit if no other branch references it in their history
```
dcherian, line 192: nice!

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -348,6 +410,37 @@ def set_doc(self, path: str, value: Buffer) -> None:
             with pytest.raises(IcechunkError, match="read-only store"):
                 self.sync_store.set(path, value)
 
+    @rule()
+    @precondition(lambda self: self.model.spec_version == 1)
```
dcherian, line 414: could just query `self.repo` since you have access to it.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -348,6 +410,37 @@ def set_doc(self, path: str, value: Buffer) -> None:
...
+        # that should be covered in unit tests
+        icechunk.upgrade_icechunk_repository(self.repo)
+        # TODO: remove the reopen after https://github.com/earth-mover/icechunk/issues/1521
+        self.reopen_repository()
+
+    @rule()
+    def reopen_repository(self) -> None:
+        """Reopen the repository from storage to get fresh state.
+
+        This discards any uncommitted changes.
```
dcherian, line 426: My first instinct is that the existing `discard_changes` rule exercises the necessary logic. Please add a comment if your intent is to test something else.

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: yes. at first i had more. Then simplified but still thought that in the future we would add more. should i go back to just list? Being able to say `.head` is pretty nice

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: My understanding is that this is a statement of fact, not of intent. although. looking again i think I need to have the model match that

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: now that im thinking more. im questioning if branches need their history. since commits all have a parent i think we can in fact just track the heads

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: turns out you don't need this at all. commit parents have all the info you need for this logic

icechunk-python/tests/test_stateful_repo_ops.py
dcherian: Seems to me like the minimal possible abstract model is the goal here. Reduces chances of bugs in the model

PR #1525, opened 2025-12-25
Add roundtrip serialization property tests for types in change sets file

ebarylko:
## Summary

There are property tests in `icechunk/src/change_set.rs` which check that the composition of serializing and deserializing a `ChangeSet` is equivalent to the identity function.

### icechunk/src/change_set.rs
* Added arbitraries for the `ChangeSet` type
* Added a roundtrip serialization test for the `ChangeSet` type

### icechunk/src/format/manifest.rs
* Changed the `ManifestExtents` strategy used for some of the tests

### icechunk/src/strategies.rs
* Added new strategies used for constructing the `ChangeSet` type in `icechunk/src/change_set.rs`

### icechunk/src/storage/s3.rs
* Upgraded the behaviour version for the aws config

### DIFF
--- a/icechunk/src/strategies.rs
+++ b/icechunk/src/strategies.rs
@@ -119,7 +119,11 @@ pub fn shapes_and_dims(max_ndim: Option<usize>) -> impl Strategy<Value = ShapeDi
         })
 }
 
-pub fn manifest_extents2(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
+// Generates a subset of possible manifest extents where the width of each extent is
+// in [1, 999]
+pub fn limited_width_manifest_extents(
+    ndim: usize,
+) -> impl Strategy<Value = ManifestExtents> {
     (vec(0u32..1000u32, ndim), vec(1u32..1000u32, ndim)).prop_map(|(start, delta)| {
         let stop = std::iter::zip(start.iter(), delta.iter())
             .map(|(s, d)| s + d)
@@ -128,6 +132,8 @@ pub fn manifest_extents2(ndim: usize) -> impl Strategy<Value = ManifestExtents> 
     })
 }
 
+// Generates possible manifest extents where the width of each extent is
+// in [1, 4_294_967_295]
 pub fn manifest_extents(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
     vec(
         any::<Range<u32>>()
@@ -599,11 +605,6 @@ fn chunk_payload() -> impl Strategy<Value = ChunkPayload> {
         chunk_ref().prop_map(Ref)
     ]
 }
-
-// pub fn chunk_indices2() -> impl Strategy<Value = ChunkIndices> {
-//     (any::<u8>().prop_map(usize::from), any::<Range<u32>>())
-//         .prop_flat_map(|(dim, data)| chunk_indices(dim, data))
-// }
 
 pub fn large_chunk_indices(dim: usize) -> impl Strategy<Value = ChunkIndices> {
     any::<Range<u32>>().prop_flat_map(move |data| chunk_indices(dim, data))
--- a/icechunk/src/format/manifest.rs
+++ b/icechunk/src/format/manifest.rs
@@ -637,7 +637,7 @@ static ROOT_OPTIONS: VerifierOptions = VerifierOptions {
 mod tests {
     use super::*;
     use crate::strategies::{
-        ShapeDim, manifest_extents, manifest_extents2, shapes_and_dims,
+        ShapeDim, limited_width_manifest_extents, manifest_extents, shapes_and_dims,
     };
     use icechunk_macros;
     use itertools::{all, multizip};
@@ -685,7 +685,7 @@ mod tests {
 
     #[proptest]
     fn test_property_extents_widths(
-        #[strategy(manifest_extents2(4))] extent1: ManifestExtents,
+        #[strategy(limited_width_manifest_extents(4))] extent1: ManifestExtents,
         #[strategy(vec(0..100, 4))] delta_left: Vec<i32>,
         #[strategy(vec(0..100, 4))] delta_right: Vec<i32>,
     ) {


icechunk/src/strategies.rs
```diff
@@ -456,3 +478,141 @@ pub fn azure_credentials() -> BoxedStrategy<AzureCredentials> {
...
+    prop_oneof![
+        bytes().prop_map(Inline),
+        virtual_chunk_ref().prop_map(Virtual),
+        chunk_ref().prop_map(Ref)
+    ]
+}
+
+type SplitManifest = BTreeMap<ChunkIndices, Option<ChunkPayload>>;
+
+pub fn chunk_indices2() -> impl Strategy<Value = ChunkIndices> {
```
paraseba: why not using the existing strategy?

icechunk/src/strategies.rs
```diff
@@ -456,3 +478,141 @@ pub fn azure_credentials() -> BoxedStrategy<AzureCredentials> {
...
+    (any::<u8>().prop_map(usize::from), any::<Range<u32>>())
+        .prop_flat_map(|(dim, data)| chunk_indices(dim, data))
+}
+
+pub fn split_manifest() -> impl Strategy<Value = SplitManifest> {
+    btree_map(chunk_indices2(), option::of(chunk_payload()), 3..10)
```
paraseba: An issue here is that all chunk indices in a single split manifest should have the same number of dimensions

icechunk/src/strategies.rs
```diff
@@ -108,7 +119,7 @@ pub fn shapes_and_dims(max_ndim: Option<usize>) -> impl Strategy<Value = ShapeDi
         })
 }
 
-pub fn manifest_extents(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
+pub fn manifest_extents2(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
```
paraseba: Let's delete this function if we believe the new one is superior

icechunk/src/change_set.rs
```diff
@@ -965,4 +966,40 @@ mod tests {
...
+
+    use crate::strategies::{
+        array_data, bytes, chunk_indices2, gen_move, manifest_extents, node_id, path,
+        split_manifest,
+    };
+    use proptest::collection::{btree_map, hash_map, hash_set, vec};
+    use proptest::prelude::*;
+
+    prop_compose! {
+        fn edit_changes()(num_of_dims in any::<u16>().prop_map(usize::from))(
```
paraseba: to make it more realistic let's keep number of dimensions in [1, 20]

icechunk/src/strategies.rs
ebarylko: I see. I will change this so all generated chunk indices are of the same size.

icechunk/src/strategies.rs
ebarylko: I thought it would be an easier way to use the existing `chunk_indices` function by having the new function generate the inputs used for the old function.

icechunk/src/change_set.rs
ebarylko: I see. I will shrink the range of values for `num_of_dims`.

icechunk/src/strategies.rs
ebarylko: Although the new function is simpler, it does not work when used in the `test_property_extents_widths` test in `manifest.rs`. 

I still have not figured out how to replicate the property that the  test expects in the new function. That is why I left both versions of the function

ebarylko: Do you happen to know why the delta values fall in [0, 99] for the `test_property_extents_width` test?

paraseba: Feel free to modify that test as needed. There is no profound reason for the values selected there.

icechunk/src/strategies.rs
```diff
@@ -456,3 +478,147 @@ pub fn azure_credentials() -> BoxedStrategy<AzureCredentials> {
...
+    prop_oneof![
+        bytes().prop_map(Inline),
+        virtual_chunk_ref().prop_map(Virtual),
+        chunk_ref().prop_map(Ref)
+    ]
+}
+
+type SplitManifest = BTreeMap<ChunkIndices, Option<ChunkPayload>>;
+
+// pub fn chunk_indices2() -> impl Strategy<Value = ChunkIndices> {
```
paraseba: let's delete this

icechunk/src/strategies.rs
```diff
@@ -456,3 +478,147 @@ pub fn azure_credentials() -> BoxedStrategy<AzureCredentials> {
...
+//     (any::<u8>().prop_map(usize::from), any::<Range<u32>>())
+//         .prop_flat_map(|(dim, data)| chunk_indices(dim, data))
+// }
+
+pub fn chunk_indices2(dim: usize) -> impl Strategy<Value = ChunkIndices> {
```
paraseba: maybe a name like `large_chunk_indices`

icechunk/src/strategies.rs
ebarylko: I see. Thank you for clarifying that.

icechunk/src/strategies.rs
ebarylko: Ok. I will do that.

icechunk/src/strategies.rs
ebarylko: That sounds good to me. I will apply that change.

icechunk/src/strategies.rs
```diff
@@ -117,8 +128,19 @@ pub fn manifest_extents(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
     })
 }
 
+pub fn manifest_extents(ndim: usize) -> impl Strategy<Value = ManifestExtents> {
```
paraseba, line 137: Can we unify the two `manifest_extents` functions? Otherwise let's use better names and document the different between the two

icechunk/src/strategies.rs
paraseba: Let's delete this please @ebarylko

icechunk/src/strategies.rs
ebarylko: At the moment, I have not found a way to do so. I will change the names and add a docstring for each function.

icechunk/src/strategies.rs
ebarylko: I am sorry. I did not realize I forgot to delete the commented code

ebarylko: Just deleted it.

paraseba (approved):

Issue #1534, opened 2025-12-29
spec v1 can expire tags pointing at head preserved snapshots - unlike v2


ianhi:
### What happened?

If you `expire_snapshots(..., delete_expired_tags=True)` on a spec 1 repo then tags pointing at snapshots that are preserved get deleted. I think this is because those snapshots are incorrectly returned as expired. 

```
(icechunk) ➜  icechunk-python git:(ian/v1-v2-regression-tests) ✗ uv run tests/test_tag_deletion_clearer.py

======================================================================
BEFORE EXPIRATION
======================================================================
Repo spec_version: 1

All snapshots in ancestry:
  DVY5A1SCJ3EPKTR0F14G: 'head commit' @ 2025-12-29 21:44:51.104297+00:00
  XZV89ZCW23PEEDKJGHGG: 'middle commit' @ 2025-12-29 21:44:50.596733+00:00
  1CECHNKREP0F1RSTCMT0: 'Repository initialized' @ 2025-12-29 21:44:50.090960+00:00

Branches:
  'main' HEAD: DVY5A1SCJ3EPKTR0F14G

Tags:
  'tag-head' → DVY5A1SCJ3EPKTR0F14G
  'tag-middle' → XZV89ZCW23PEEDKJGHGG
  'tag-initial' → 1CECHNKREP0F1RSTCMT0
----------------------------------------------------------------------

======================================================================
AFTER EXPIRATION
======================================================================
Expired snapshots: {'XZV89ZCW23PEEDKJGHGG', 'DVY5A1SCJ3EPKTR0F14G'}

All snapshots in ancestry:
  DVY5A1SCJ3EPKTR0F14G: 'head commit' @ 2025-12-29 21:44:51.104297+00:00
  1CECHNKREP0F1RSTCMT0: 'Repository initialized' @ 2025-12-29 21:44:50.090960+00:00

Branches:
  'main' HEAD: DVY5A1SCJ3EPKTR0F14G

Tags:
  (none)
----------------------------------------------------------------------

======================================================================
VALIDATION
======================================================================
✓ Middle commit expired: True
✓ HEAD unchanged: True
❌ 'tag-initial' preserved: False
✓ 'tag-middle' deleted: True
❌ 'tag-head' preserved: False
❌ No overlap (expired ∩ remaining): False
  ⚠️  Snapshots marked as expired but still in ancestry: {'DVY5A1SCJ3EPKTR0F14G'}
```

### What did you expect to happen?

[...]

Issue #1520, opened 2025-12-24
Difference in v1 vs v2 branch expiration


ianhi:
### What happened?

icechunk v2 spec version 2 has different behavior for expiring branches compared to all other combinations.

When both feature and main point to the same commit (which is old enough to be expired) then the v2/2 combo will not expire the feature branch, but other versions will

```
=== Same tip (both point to commit2) ===
v1 python: ['feature', 'main'] -> expired 2 snapshots -> remaining branches: ['main']
v2 spec=1: ['feature', 'main'] -> expired 2 snapshots -> remaining branches: ['main']
v2 spec=2: ['feature', 'main'] -> expired 1 snapshots -> remaining branches: ['feature', 'main']

=== Separate tips (feature->commit1, main->commit2) ===
v1 python: ['feature', 'main'] -> expired 2 snapshots -> remaining branches: ['main']
v2 spec=1: ['feature', 'main'] -> expired 2 snapshots -> remaining branches: ['main']
v2 spec=2: ['feature', 'main'] -> expired 1 snapshots -> remaining branches: ['main']
```

### What did you expect to happen?

I guess consistency in expiring feature, or none of the other combos expire.

### Minimal Complete Verifiable Example

```Python
#!/usr/bin/env python3
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "icechunk>=2.0.0.dev0",
#     "icechunk_v1",
#     "zarr>=3",
# ]
#
# [tool.uv]
# extra-index-url = [
#     "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple",
#     "http://127.0.0.1:8123/simple/",
# ]
# index-strategy = "unsafe-best-match"
# ///

"""
Requires the wheel-rename proxy server running:
    uvx --from "wheel-rename[server] @ git+https://github.com/earth-mover/rename-wheel" \
        wheel-rename serve \
        -u https://pypi.anaconda.org/scientific-python-nightly-wheels/simple \
        -r "icechunk=icechunk_v1:<2" \
        --port 8123
"""

import datetime
import icechunk
import icechunk_v1
import zarr

future = datetime.datetime.now(datetime.timezone.utc) + datetime.timedelta(days=1)


def test_expire(name, repo, separate_tips=False):
    session = repo.writable_session("main")
    root = zarr.open_group(session.store, mode="a")
    root.create_array("arr", shape=(2,), dtype="i4")
    commit1 = session.commit("first")

    if separate_tips:
[...]

Issue #1513, opened 2025-12-23
Add new IC2 actions to stateful test suite


dcherian:
For https://github.com/earth-mover/icechunk/blob/main/icechunk-python/tests/test_stateful_repo_ops.py

- [ ] add a V1 -> V2 migration step
- [ ] add `amend`

For https://github.com/earth-mover/icechunk/blob/main/icechunk-python/tests/test_zarr/test_stateful.py
- [ ] add `shift_array`
- [ ] add `insert`

PR #1537, opened 2025-12-30
Add amend to stateful_repo_ops

ianhi:
This required checking if a commit was anywhere in any branches history so I also had to update to storing a full commit history for each branch. I made a branchModel object to make it a bit nicer to do `branch.head` instead of grabbing the last item in a list.

I also added some more checks to the expiry that would have caught issues #1520  and #1534 

This addresses most, but not all of @dcherian review comments from #1515 but I think this is a nice standalone change. I can work on deeper paramaterization of the model and spec separately if thats ok (ref https://github.com/earth-mover/icechunk/pull/1515#discussion_r2651434510)

These tests currently fail due to #1534 and #1520

This is the first half of https://github.com/earth-mover/icechunk/issues/1513

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -118,6 +118,9 @@ class Model:
         self.HEAD: None | str = None
         self.branch: None | str = None
 
+        # we maintain a list of `commits` == snapshots in repo object file
+        # and `ondisk_snaps` = commits + expired_snaphsots
+        # expired snapshots are removed from the repo object file and can be garbage collected
         self.commits: dict[str, CommitModel] = {}
         self.ondisk_snaps: dict[str, CommitModel] = {}
         self.tags: dict[str, TagModel] = {}
@@ -169,41 +172,10 @@ class Model:
         assert self.branch is not None
         self.branch_heads[self.branch] = ref
 
-    def amend(self, branch: str, snap: SnapshotInfo) -> None:
-        """Amend the HEAD commit on the given branch."""
-        old_head = self.branch_heads[branch]
-        old_commit = self.commits[old_head]
-        parent_id = old_commit.parent_id
-
-        # Only remove old commit if no other branch or tag references it
-        # Check if it is another branch's HEAD
-        is_other_branch_head = any(
-            old_head == other_model
-            for other_branch, other_model in self.branch_heads.items()
-            if other_branch != branch
-        )
-        # Check if it is in another branches' history by searching if it is the parent
-        # of any other commits. This is a valid check because we can only amend the HEAD
-        # commit, so if it is the parent of a commit, that commit is on a different branch
-        is_parent = any(old_head == commit.parent_id for commit in self.commits.values())
-        # check if the commit has been tagged
-        has_tag = any(tag.commit_id == old_head for tag in self.tags.values())
-        # if none of the above then delete the commit
-        if not is_other_branch_head and not is_parent and not has_tag:
-            self.commits.pop(old_head)
-
-        # Add new commit with same parent
-        new_commit_id = snap.id
-        self.commits[new_commit_id] = CommitModel.from_snapshot_and_store(
-            snap, copy.deepcopy(self.store)
-        )
-        self.commits[new_commit_id].parent_id = parent_id
-
-        # Update branch HEAD
-        self.branch_heads[branch] = new_commit_id
-
-        self.HEAD = new_commit_id
-        self.changes_made = False
+    def amend(self, snap: SnapshotInfo) -> None:
+        """Amend the HEAD commit."""
+        # this is simpe because we aren't modeling the branch as a list of commits
+        self.commit(snap)
 
     def checkout_commit(self, ref: str) -> None:
         assert str(ref) in self.commits
@@ -276,6 +248,7 @@ class Model:
 
         for id in expired_snaps:
             # notice we don't delete from self.ondisk_snaps, those can still be deleted by GC
+            # however we do pop from `commits` since that is a list of unexpired snaps
             self.commits.pop(id, None)
 
         for c in self.commits.values():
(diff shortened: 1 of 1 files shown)


icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -300,15 +341,13 @@ def __init__(self) -> None:
...
 
     @initialize(data=st.data(), target=branches)
     def initialize(self, data: st.DataObject) -> str:
-        # FIXME: currently this test is IC2 only
-        spec_version = data.draw(
-            st.one_of(st.integers(min_value=2, max_value=2), st.none())
-        )
-        self.repo = Repository.create(in_memory_storage(), spec_version=spec_version)
+        self.storage = in_memory_storage()
+        spec_version = cast(Literal[1, 2], data.draw(st.sampled_from([1, 2])))
```
dcherian: if you stick this in `initialize`, the hypothesis logs will be nicer I think. Same with that draw on `v3_array_metadata` below.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -519,31 +614,66 @@ def expire_snapshots(
...
+        note(f"actual_deleted_branches: {actual_deleted_branches}")
+        note(f"actual_deleted_tags: {actual_deleted_tags}")
+
         assert self.initial_snapshot.id not in actual
         assert actual == expected.expired_snapshots, (actual, expected)
 
-        for branch in expected.deleted_branches:
+        # Check that expired snapshots are actually removed from ancestry
+        # TODO: this should have failed earlier on the model step. the model to
+        # need to fix the model here as well
```
dcherian: not sure what this means

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -321,6 +360,7 @@ def initialize(self, data: st.DataObject) -> str:
         self.model.HEAD = HEAD
         self.model.create_branch(DEFAULT_BRANCH, HEAD)
         self.model.checkout_branch(DEFAULT_BRANCH)
+        self.model.spec_version = spec_version
```
dcherian: should not be needed

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: making sure people know this comment is by a human by making it incoherent :)

I think I mean that the incorrect branches should never be expired becuase that's all downstream of expiring the correct snapshots. The bug I was seeing was that we were overeager in expriing snapshots, which was then leading to overeager branch/tag expiration. This comment is overspecific to that bug. I'll remove it. Not really sure what I meant with the last sentence.

ianhi: I have removed this. I think i meant that I was surprised this wasn't catching the excessive expiration for v1 spec, delete_expired tags true case the model is not doing quite the right thing in these lines:

```python
        expired_snaps = set()
        branch_pointees = set(self.branch_heads.values())
        tag_pointees = set(map(operator.attrgetter("commit_id"), self.tags.values()))
        for snap in self.commits.values():
            if (
                snap.written_at < older_than
                and snap.parent_id is not None
                and (delete_expired_tags or snap.id not in tag_pointees)
                and (
                    (delete_expired_branches and self.branch_heads["main"] != snap.id)
                    or snap.id not in branch_pointees
                )
            ):
                expired_snaps.add(snap.id)

```
but now im not 100% sure what the right expiration behavior is. If there is a branch that has the commit as it's head. should it not be expired? 

of if there is a expired snapshot that is the head, but there is other history for that branch what should happen?

ianhi:
I think this is ready now. I've temporarily had it skip the condition that causes expire_snapshots to fail which will be fixed by https://github.com/earth-mover/icechunk/issues/1534 and https://github.com/earth-mover/icechunk/issues/1520 neither of which I managed to get to today

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -169,41 +172,10 @@ def commit(self, snap: SnapshotInfo) -> None:
...
-        )
-        # Check if it is in another branches' history by searching if it is the parent
-        # of any other commits. This is a valid check because we can only amend the HEAD
-        # commit, so if it is the parent of a commit, that commit is on a different branch
-        is_parent = any(old_head == commit.parent_id for commit in self.commits.values())
-        # check if the commit has been tagged
-        has_tag = any(tag.commit_id == old_head for tag in self.tags.values())
-        # if none of the above then delete the commit
-        if not is_other_branch_head and not is_parent and not has_tag:
-            self.commits.pop(old_head)
```
dcherian, line 193: this pop was wrong; amending keep the snapshot ID in the repo object file, so we shouldn't pop from commits.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -169,41 +172,10 @@ def commit(self, snap: SnapshotInfo) -> None:
...
-
-        # Update branch HEAD
-        self.branch_heads[branch] = new_commit_id
-
-        self.HEAD = new_commit_id
-        self.changes_made = False
+    def amend(self, snap: SnapshotInfo) -> None:
+        """Amend the HEAD commit."""
+        # this is simpe because we aren't modeling the branch as a list of commits
+        self.commit(snap)
```
dcherian, line 178: very simple now since we only model branch heads

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -169,41 +172,10 @@ def commit(self, snap: SnapshotInfo) -> None:
...
-        # if none of the above then delete the commit
-        if not is_other_branch_head and not is_parent and not has_tag:
-            self.commits.pop(old_head)
-
-        # Add new commit with same parent
-        new_commit_id = snap.id
-        self.commits[new_commit_id] = CommitModel.from_snapshot_and_store(
-            snap, copy.deepcopy(self.store)
-        )
-        self.commits[new_commit_id].parent_id = parent_id
```
dcherian, line 200: the error here is that we forgot to update `self.ondisk_snaps` I've chosen to just call `self.commit` to avoid such errors in the future.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -714,6 +689,15 @@ def check_commit(self, commit: str) -> None:
         assert actual.written_at == expected.written_at
 
     @invariant()
+    def checks(self) -> None:
```
dcherian, line 692: making the output less verbose.

dcherian (approved):

PR #1538, opened 2025-12-30
fail with nice error when trying amend the very first commit

ianhi:
There was logic for this, but a different, less clear error was being hit first.

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "icechunk>=2.0.0.dev0",
#     "zarr>=3",
# ]
#
# [tool.uv]
# extra-index-url = [
#     "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple",
# ]
# index-strategy = "unsafe-best-match"
# ///

import zarr
from icechunk import Repository, in_memory_storage

storage = in_memory_storage()
repo = Repository.create(storage, spec_version=2)

root = zarr.group(session.store)
root.create_array("data", shape=(10,), dtype="i4")

# Try to amend the initial commit
amended_commit = session.amend("Amending initial commit")

```

would fail with a confusing error:

```
➜  icechunk-python git:(ian/fix-amend-initial-commit) ✗ uv run tests/test_amend_initial_commit.py
  File "/Users/ian/Documents/dev/icechunk/icechunk-python/tests/test_amend_initial_commit.py", line 35, in <module>
    amended_commit = session.amend("Amending initial commit")
  File "/Users/ian/.cache/uv/environments-v2/test-amend-initial-commit-a826cda45e443de5/lib/python3.14/site-packages/icechunk/session.py", line 377, in amend
    return self._session.amend(message, metadata)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
icechunk.IcechunkError:   × session error: object not found
  │
  │ context:
  │    0: icechunk::storage::object_store::get_object_range
  │            with settings=Settings { concurrency: Some(ConcurrencySettings { max_concurrent_requests_for_object: Some(5), ideal_concurrent_request_size: Some(1) }), retries: Some(RetriesSettings
  │ { max_tries: Some(1), initial_backoff_ms: Some(0), max_backoff_ms: Some(0) }), unsafe_use_conditional_update: None, unsafe_use_conditional_create: None, unsafe_use_metadata: None, storage_class:
  │ None, metadata_storage_class: None, chunks_storage_class: None, minimum_size_for_multipart_upload: None } path="transactions/1CECHNKREP0F1RSTCMT0" range=None
  │              at icechunk/src/storage/object_store.rs:379
  │    1: icechunk::asset_manager::fetch_transaction_log
  │            with transaction_id=1CECHNKREP0F1RSTCMT0
  │              at icechunk/src/asset_manager.rs:447
  │    2: icechunk::session::_commit
  │            with Amending initial commit rewrite_manifests=false commit_method=Amend
  │              at icechunk/src/session.rs:1257
  │    3: icechunk::session::amend
  │            with Amending initial commit
  │              at icechunk/src/session.rs:1134
  │
```

After this PR the error becoems:

```
➜  icechunk-python git:(ian/fix-amend-initial-commit) ✗ uv run python tests/test_amend_initial_commit.py
  File "/Users/ian/Documents/dev/icechunk/icechunk-python/tests/test_amend_initial_commit.py", line 26, in <module>
    amended_commit = session.amend("Amending initial commit")
  File "/Users/ian/Documents/dev/icechunk/icechunk-python/python/icechunk/session.py", line 377, in amend
    return self._session.amend(message, metadata)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
icechunk.IcechunkError:   × session error: the first commit in the repository cannot be an amend, create a new commit instead
  │
  │ context:
  │    0: icechunk::session::amend
  │            with Amending initial commit
  │              at icechunk/src/session.rs:1134
  │

```

**Rust coding stuff**
I tried with

```
       let current_snapshot_info =
           self.asset_manager.fetch_snapshot_info(self.snapshot_id()).await?;
```

instead of what I have here. but then that hits a similar issue where it doesn't work. which is maybe an issue of it's own right?


paraseba:
@ianhi that error you mention at the end with fetch snap info sounds scary. Could you investigate in a separate PR?

icechunk/src/session.rs
```diff
@@ -1140,6 +1140,13 @@ impl Session {
         // Icechunk 1 doesn't support amend
         self.asset_manager.fail_unless_spec_at_least(SpecVersionBin::V2dot0)?;
 
+        // Cannot amend the initial commit (which has no parent and no transaction log)
```
paraseba: It's weird, because the existing code is actively trying to generate this error message. There is some bug, but can we find where the existing code is failing and fix it instead of adding another check?

ianhi: You mean i should leave the original error throwing and handle this further down in the calls?

ianhi: @paraseba should the initial commit have a transaction log? the error i was seeing was coming from this get_transaction_log when old_snapshot_id is the first commit id:

https://github.com/earth-mover/icechunk/blob/0d09a31316e3350735f137ca2fd0f4ab9e1ee755/icechunk/src/session.rs#L2501-L2504

ianhi:
ok i have moved the catch to be near the other commit pre-validation checks (i.e. allow_empty checks) and added a is_initial to snapshotID to make it easy to check.

ianhi:
alternatively i think that the initial snapshot could have an empty transaction log (nice because then consistent with all other commits) and we could catch the error a bit later

ianhi:
I think this is now waiting on https://github.com/earth-mover/icechunk/pull/1623 then i will rebase and figure out what the way to do this is.

ianhi:
I looked at this again. and even with the changes of transaction log for the intial commit i think this is the correct way where we are catching it. and the changes I thought I was waiting on the other PR in order to remove I seem to have already removed.

PR #1539, opened 2025-12-30
Have default contributing instructions do less intensive rebuilds

ianhi:
Following https://quanttype.net/posts/2025-09-12-uv-and-maturin.html

gives me a significantly more enjoyable experience

multiple minutes to ~10 seconds to recompile on my laptop.

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -206,13 +206,14 @@ The documentation is built with [MkDocs](https://www.mkdocs.org/) using [Materia
     sudo dnf install cairo-devel
     ```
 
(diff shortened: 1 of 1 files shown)


ianhi:
For posterity. @samantha-earthmover asked a good question about if this will mess with release builds. We decided the answer was no becuase the release happens with the maturin github action step which does not use uv so should not touch this part of pyproject.toml

https://github.com/earth-mover/icechunk/blob/cf32bf1c1dcb775abe84231bfb15d9e53c08baf7/.github/workflows/python-ci.yaml#L316-L322

PR #1547, opened 2026-01-09
Identify chunks as virtual/native

### LINKED ISSUE #1469: Feature: Identify chunks as virtual/native
This is a summary of a discussion on the Earthmover Slack

### Problem

The core issue is that several users would like to have a mechanism to identify if any given individual chunk is native or virtual. 

Some workarounds (cc @sharkinsspatial @abarciauskas-bgse) have been proposed in the meantime, but are not ideal: 

```python
repo = icechunk.Repository.open(icechunk_storage, config, authorize_virtual_chunk_access=credentials)

root = zarr.open(session.store, mode='r')

# Get the array
time_arr = root['/time']

# Check the array's attributes
print(time_arr[0:10])

# Get the array
base_arr = root['/base']

# Check the array's attributes
print(base_arr[0:10])
```
This works for both arrays, but when one removes `authorize_virtual_chunk_access=credentials` it will fail for virtual chunks with: `virtual chunk in this repository resolves to the url prefix gs://ismip6/, to be able to fetch the chunk you need to authorize the virtual chunk container when you open/create the repository`

### Proposed Solution

@TomNicholas and @paraseba pointed out that this check is trivial in rust, but the mechanism needs to be exposed to the python API. I created this issue in order to track progress on this and have the discussion outside the slack.

li-em:
Closes #1469

Expose both sync and async versions of the new `chunk_type` method.

### DIFF
--- a/icechunk-python/tests/test_session.py
+++ b/icechunk-python/tests/test_session.py
@@ -114,7 +114,7 @@ async def test_session_fork(use_async: bool, any_spec_version: int | None) -> No
     [(10_000, ChunkType.INLINE), (1, ChunkType.NATIVE)],
     ids=["inline", "native"],
 )
-def test_chunk_type(
+async def test_chunk_type(
     inline_threshold: int, chunk_type: ChunkType, any_spec_version: int | None
 ) -> None:
     config = RepositoryConfig.default()
@@ -156,51 +156,6 @@ def test_chunk_type(
     assert session.chunk_type("/air_temp", [0, 0]) == ChunkType.VIRTUAL
     assert session.chunk_type("/air_temp", [0, 2]) == chunk_type
     assert session.chunk_type("/air_temp", [0, 3]) == ChunkType.UNINITIALIZED
-
-
-@pytest.mark.parametrize(
-    "inline_threshold,chunk_type",
-    [(10_000, ChunkType.INLINE), (1, ChunkType.NATIVE)],
-    ids=["inline", "native"],
-)
-async def test_chunk_type_async(
-    inline_threshold: int, chunk_type: ChunkType, any_spec_version: int | None
-) -> None:
-    config = RepositoryConfig.default()
-    config.inline_chunk_threshold_bytes = inline_threshold
-    store_config = local_filesystem_store("/foo")
-    container = VirtualChunkContainer("file:///foo/", store_config)
-    config.set_virtual_chunk_container(container)
-
-    repo = await Repository.create_async(
-        storage=in_memory_storage(),
-        config=config,
-        spec_version=any_spec_version,
-    )
-
-    session = await repo.writable_session_async("main")
-    store = session.store
-    group = zarr.group(store=store, overwrite=True)
-    air_temp = group.create_array("air_temp", shape=(1, 4), chunks=(1, 1), dtype="i4")
-
-    # set index [0, 0] to be a virtual chunk
-    # note: we can't ACCESS it, since the file `foo` is never instantiated
-    store.set_virtual_refs(
-        array_path="/air_temp",
-        validate_containers=False,
-        chunks=[
-            VirtualChunkSpec(
-                index=[0, 0],
-                location="file:///foo",
-                offset=0,
-                length=1_000,
-            ),
-        ],
-    )
-
-    # This forces the chunk to be initialized, either to inline or native chunks
-    air_temp[0, 2] = 42
-    assert air_temp[0, 2] == 42
 
     assert await session.chunk_type_async("/air_temp", [0, 0]) == ChunkType.VIRTUAL
     assert await session.chunk_type_async("/air_temp", [0, 2]) == chunk_type


icechunk-python/src/session.rs
```diff
@@ -318,6 +327,34 @@ impl PySession {
         Ok(PyAsyncGenerator::new(prepared_list))
     }
```
paraseba, line 329: Let's also add `chunk_type_async`, you'll see we follow this pattern of offering the two options using an `_async` suffix.

icechunk-python/src/lib.rs
```diff
@@ -166,6 +166,7 @@ fn _icechunk_python(py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<PyRepository>()?;
     m.add_class::<PyRepositoryConfig>()?;
     m.add_class::<PySession>()?;
+    m.add_class::<ChunkType>()?;
```
paraseba, line 169: You'll need to add the `ChunkType` class in the `.pyi` file.

paraseba: and also in the python `__init__` file. You'll see we add all types there (for now)

icechunk-python/python/icechunk/session.py
```diff
@@ -206,6 +206,19 @@ async def chunk_coordinates(
             for coord in batch:
                 yield tuple(coord)
 
+    def chunk_type(
```
paraseba, line 210: Se my comment below, let's add also `chunk_type_async`

icechunk-python/tests/test_session.py
```diff
@@ -96,3 +104,22 @@ async def test_session_fork(use_async: bool, any_spec_version: int | None) -> No
...
+    )
+
+    session = repo.writable_session("main")
+    group = zarr.group(store=session.store, overwrite=True)
+    air_temp = group.create_array("air_temp", shape=(9, 9), chunks=(3, 3), dtype="i4")
+
+    air_temp[:, :] = 42
+    assert air_temp[0, 0] == 42
+
+    assert session.chunk_type("/air_temp", [0, 0]) == ChunkType.INLINE
```
paraseba: You'll need to find some examples of virtual chunks, so you can test all possibilities

icechunk-python/src/lib.rs
TomNicholas: It would be nice to better document this part of the development process somewhere.

paraseba: yes! I believe you would get, at least, mypy errors if you forget, but we need to document it.

icechunk-python/python/icechunk/session.py
li-em: I moved the impl into an `_inner` function, and call it in both sync and async versions. It avoids repeating the impl, but ended up at the end of the file because it needs to be out of the `[pymethods]` macro (since it is not exposed to Python).

Any preference on keeping or removing the `_inner_` function?

icechunk-python/src/lib.rs
li-em: Will add a bit on the dev docs (maybe link this PR as an example on how to expose something all the way from Rust -> PyO3 -> Python?)

icechunk-python/src/session.rs
li-em: done!

icechunk-python/python/icechunk/session.py
paraseba: love the inner function

icechunk-python/tests/test_session.py
li-em: Added virtual chunks, and parametrized the test for inline vs native, since the only way to control inlining is with the `config.inline_chunk_threshold_bytes` config

icechunk-python/python/icechunk/session.py
```diff
@@ -206,6 +207,52 @@ async def chunk_coordinates(
...
+            The path to the array inside the Zarr store. Example: "/groupA/groupB/outputs/my-array".
+        chunk_coordinates: Sequence[int]
+            A sequence of integers (list or tuple) used to locate the chunk. Example: [0, 1, 5].
+
+        Returns
+        -------
+        ChunkType
+            One of the supported chunk types.
+        """
+        ...
```
TomNicholas:
```suggestion
```

icechunk-python/python/icechunk/session.py
```diff
@@ -206,6 +207,52 @@ async def chunk_coordinates(
...
```
TomNicholas:
```suggestion
```

icechunk-python/tests/test_session.py
```diff
@@ -96,3 +107,101 @@ async def test_session_fork(use_async: bool, any_spec_version: int | None) -> No
...
+    container = VirtualChunkContainer("file:///foo/", store_config)
+    config.set_virtual_chunk_container(container)
+
+    repo = await Repository.create_async(
+        storage=in_memory_storage(),
+        config=config,
+        spec_version=any_spec_version,
+    )
+
+    session = await repo.writable_session_async("main")
```
paraseba: TBH I think I'd just write a single test calling both `chunk_type` and `chunk_type_async`. But I don't oppose to this, it's just more code to maintain

Issue #831, opened 2025-03-11
CLI: Fetch and export metadata files in human readable form


paraseba:
* snapshots,
* manifests,
* transaction logs

Issue #826, opened 2025-03-11
Extend the functionality of the CLI


paraseba:

jleben:
Hi, I can work on this!

Here's a set of commands I propose that could be implemented quickly - let me know what you think please.

List ancestry of a ref (branch, tag or snapshot ID) in a human-friendly format: just snapshot ID, date and message (replacing the current `snapshot list` (alternative to the `-n` option, maybe I can implement interactive scrolling):

```
history <repo> <ref> -n <max count>
```

List branches, create branch from any ref (branch, tag or snapshot ID), and delete branch:

```
branch list <repo>
branch create <repo> <branch name> <from ref>
branch delete <repo> <branch name>
```

Similarly for tags:

```
tag list <repo>
tag create <repo> <tag name> <ref>
tag delete <repo> <tag name>
```

I have some ongoing work here: https://github.com/jleben/icechunk/tree/cli

Should this be done against the main branch? Seems like that one is already targeting v2. Could this possibly be released as a 1.x?

paraseba:
Hey @jleben . Sounds great! We are limiting new features to the `main` branch where we are developing Icechunk 2.0. But, if you need this in 1.x, we can think about porting it later.

Another nice thing would be a way to invoke `inspect_snapshot` and a similar `inspect_manifest` (which I don't think exists yet). This would be very useful to understand the structure of a  repo.

Let us know if you need anything!

PR #1554, opened 2026-01-13
CLI improvements

jleben:
This PR expands the CLI with the following commands:

```
branch list <repo>
branch create <repo> <branch name>
branch delete <repo> <branch name>
tag list <repo>
tag create <repo> <tag name> [<ref>]
tag delete <repo> <tag name>
history <repo> [<ref>]
diff <repo> <from ref> <to ref>
inspect <repo> [<ref>]
```

I think the purpose of these commands is pretty self-explanatory.

The actual output format of these commands was chosen according to fairly subjective criteria for utility + completeness + compactness. I will welcome any suggestions for change.

There is a beginning of a unit test there, nothing more in terms of automated testing. I am finding myself short of time for this lately, so I thought I'd pass it off to someone else at this point.

Related to: #826 #827 #831

### DIFF
--- a/icechunk/src/cli/interface.rs
+++ b/icechunk/src/cli/interface.rs
@@ -42,8 +42,8 @@ enum Command {
     Config(ConfigCommand),
     #[command(subcommand, about = "Manage repositories")]
     Repo(RepoCommand),
-    #[clap(name = "history", about = "Show history of a branch, tag or snapshot")]
-    History(HistoryArgs),
+    #[clap(name = "ancestry", about = "Show ancestry of a branch, tag or snapshot")]
+    Ancestry(AncestryArgs),
     #[clap(name = "inspect", about = "Show snapshot details")]
     Inspect(InspectArgs),
     #[command(subcommand, about = "Manage branches")]
@@ -61,12 +61,12 @@ enum RepoCommand {
 }
 
 #[derive(Debug, Args)]
-struct HistoryArgs {
+struct AncestryArgs {
     #[arg(name = "alias", help = "Alias of the repository in the config")]
     repo: RepositoryAlias,
     #[arg(
         name = "reference",
-        help = "ID of snapshot to show history for"
+        help = "ID of snapshot to show ancestry for"
     )]
     reference: String,
     #[arg(short = 'n', default_value_t = 10, help = "Number of snapshots to list")]
@@ -468,8 +468,8 @@ fn show_snapshot(mut writer: impl std::io::Write, snapshot: SnapshotInfo, with_m
     Ok(())
 }
 
-async fn history(
-    args: &HistoryArgs,
+async fn ancestry(
+    args: &AncestryArgs,
     config: &CliConfig,
     mut writer: impl std::io::Write,
 ) -> Result<()> {
@@ -849,8 +849,8 @@ pub async fn run_cli(args: IcechunkCLI) -> Result<()> {
         Command::Repo(RepoCommand::Create(init_cmd)) => {
             repo_create(&init_cmd, &config).await
         }
-        Command::History(history_args) => {
-            history(&history_args, &config, stdout()).await?;
+        Command::Ancestry(ancestry_args) => {
+            ancestry(&ancestry_args, &config, stdout()).await?;
             Ok(())
         }
         Command::Inspect(inspect_args) => {
@@ -967,7 +967,7 @@ mod tests {
     }
 
     #[tokio_test]
-    async fn test_history() {
+    async fn test_ancestry() {
         let temp = assert_fs::TempDir::new().unwrap();
         let path = temp.path().to_path_buf();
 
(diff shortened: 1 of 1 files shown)


paraseba:
This is awesome! We'll review in detail. Thank you @jleben !

icechunk/src/cli/interface.rs
```diff
@@ -203,6 +310,54 @@ async fn get_storage(
...
+        .await
+        .context(format!("Failed to open repository {:?}", repo_alias))?;
+
+    Ok(repository)
+}
+
+async fn resolve_ref_by_name(
+    repository: &Repository,
+    reference: &String,
+) -> Result<SnapshotId> {
```
paraseba: This is useful  logic, and in fact I think we have something similar somewhere. But for now I'd prefer to keep the CLI as logic-free as possible. I'd recommend we only take a snapshot_id as reference, which is what the python API supports today. Then we can add this logic to the main rust code and extend support in the python library and CLI

icechunk/src/cli/interface.rs
```diff
@@ -219,32 +374,336 @@ async fn repo_create(init_cmd: &CreateCommand, config: &CliConfig) -> Result<()>
...
+    println!(
+        "✅ Deleted tag {:?} in repository {:?}",
+        args.tag, args.repo
+    );
+
+    Ok(())
+}
+
+fn show_snapshot(mut writer: impl std::io::Write, snapshot: SnapshotInfo, with_meta: bool) -> Result<()> {
+    writeln!(writer, "Snapshot: {}", snapshot.id)?;
```
paraseba, line 457: we have an `inspect.rs` module. Can we use that instead?

icechunk/src/cli/interface.rs
```diff
@@ -219,32 +374,336 @@ async fn repo_create(init_cmd: &CreateCommand, config: &CliConfig) -> Result<()>
...
+
+async fn history(
+    args: &HistoryArgs,
+    config: &CliConfig,
+    mut writer: impl std::io::Write,
+) -> Result<()> {
+    let repository = open_repository(&args.repo, config).await?;
+    let snapshot = resolve_ref_by_name(&repository, &args.reference).await?;
+    let ancestry = repository.ancestry(&VersionInfo::SnapshotId(snapshot)).await?;
+    let snapshots: Vec<_> = ancestry.take(args.n).collect().await;
```
paraseba, line 479: I don't think you need to `collect` here

icechunk/src/cli/interface.rs
```diff
@@ -219,32 +374,336 @@ async fn repo_create(init_cmd: &CreateCommand, config: &CliConfig) -> Result<()>
...
+                )
+                .join(", ")
+        },
+        None => {
+            values.join(", ")
+        }
+    }
+}
+
+async fn inspect(
```
paraseba, line 506: I'd love if we instead reused and extended the logic in `inspect.rs`

icechunk/src/cli/interface.rs
```diff
@@ -219,32 +374,336 @@ async fn repo_create(init_cmd: &CreateCommand, config: &CliConfig) -> Result<()>
...
+        // TODO: Human-friendly size
+        writeln!(writer, "    size (bytes): {}", manifest_info.size_bytes)?;
+    }
+
 
     Ok(())
 }
 
+async fn diff(
+    args: &DiffArgs,
```
paraseba, line 577: It would be great if we could DRY things up, some of this logic is present in the python bindings module too: repository.rs, see PyDiff. But maybe this is OK for an initial version

paraseba (commented):
This looks great! Made a few small comments

icechunk/src/cli/interface.rs
jleben: Done.

jleben: I've also renamed the `history` command to `ancestry` in the spirit of conforming to the Python API.

icechunk/src/cli/interface.rs
jleben: @paraseba Could you explain what the benefit of using `inspect.rs` would be? Note that this function prints individual snapshot info in the context of my `ancestry` command. It (intentionally) does not print all the information about a snapshot. If I was to use `inspect_snapshot` from `inspect.rs` for example, it would do more work per snapshot than here. What exactly did you have in mind?

jleben: @paraseba I also wonder what exactly is the intention behind `NodeSnapshotInspect` - it seems like it contains mostly the same info as `NodeSnapshot`, but pruned down, simplified somewhat. The level of such a simplification is always going to be specific to some prupose, so I doubt the value of it for multiple purposes. Currently, it seems it's only used in Python's `inspect_snapshot`. However, a human-friendly display in the CLI may have other needs than the python API. Why try to squeeze a common abstraction between both end-uses and `NodeSnapshot`?

DahnJ:
> I am finding myself short of time for this lately, so I thought I'd pass it off to someone else at this point.

@jleben happy to take this PR over :) interested in having these features in the CLI. Let me know if that sounds good to you, would probably do some work on it later this upcoming week.

jleben:
@DahnJ Sure, I'm happy to pass the torch, and thank you for picking this up!

PR #1559, opened 2026-01-14
Update changelog for 1.1.16 release and fix anonymous S3 access to public buckets bug

li-em:
Two PRs backported in #1557, plus #1533

There was also an one-line fix to a bug with anonymous access to public S3 buckets, added here instead of opening a new PR to the support/v1.x branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1805,7 +1805,7 @@ dependencies = [
 
 [[package]]
 name = "icechunk-python"
-version = "1.1.15"
+version = "1.1.16"
 dependencies = [
  "async-stream",
  "async-trait",


Changelog.python.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Python Icechunk Library 1.1.16
+
+### Features
+
+- New method for returning a chunk type.
```
paraseba: let's include the methods names, something like `Session.chunk_type` and `Session.chunk_type_async`.

icechunk-python/Cargo.toml
```diff
@@ -21,7 +21,7 @@ crate-type = ["cdylib"]
 bytes = "1.11.0"
 chrono = { version = "0.4.42" }
 futures = "0.3.31"
-icechunk = { path = "../icechunk", version = "0.3.18", features = ["logs"] }
+icechunk = { path = "../icechunk", version = "0.3.19", features = ["logs"] }
```
paraseba, line 24: You'll also need to change line 3 of this file, that's what ends up an the version number for the python package

paraseba (approved):

Issue #1558, opened 2026-01-14
Commiting 72M virtual chunks at once results in panic


vladidobro:
### What happened?

Hi,

I am virtualizing some gribs, and my icechunk panicked during the commit phase.
I believe the problem is not with my data, because when the commits are done in batches, there is no panic.
There are 72M virtual refs, I can share a complete reproducible example if needed, but it would include a ~700MB duckdb database.

It only happens when I run it on the big sample, and it takes ~20min to write all the chunks before the erroneous commit.

I have tried both with VirtualiZarr and manual `IcechunkStore.set_virtual_ref`, and both panic when run on the full sample.

It may only be a problem on macos, I have not tried it on linux

### What did you expect to happen?

No panic

### Minimal Complete Verifiable Example

```Python
import obstore.auth.planetary_computer
from itertools import batched
import obstore
import xarray as xr
import gribberish.zarr
import dask.array as da
import numpy as np
from datetime import datetime, timedelta
from zarr.core.dtype import parse_data_type
from zarr.core.metadata.v3 import ArrayV3Metadata
import sparse
import virtualizarr
import virtualizarr.manifests
import duckdb
import asyncio
import polars as pl
from tqdm import tqdm
from polars import col as c
from typer import Typer
import icechunk
import icechunk.storage
import shutil

config = icechunk.RepositoryConfig.default()
config.set_virtual_chunk_container(
    icechunk.VirtualChunkContainer('s3://ecmwf-forecasts/', icechunk.s3_store(region='eu-central-1', anonymous=True)),
)
config.set_virtual_chunk_container(
    icechunk.VirtualChunkContainer('az://ecmwf/', icechunk.storage.azure_store(account='ai4edataeuwest')),
)
config.set_virtual_chunk_container(
    icechunk.VirtualChunkContainer('gcs://ecmwf-open-data/', icechunk.gcs_store()),
)
planetary_cred = obstore.auth.planetary_computer.PlanetaryComputerCredentialProvider(
    account_name='ai4edataeuwest',
    container_name='ecmwf',
)
creds = icechunk.containers_credentials({
        's3://ecmwf-forecasts': icechunk.s3_anonymous_credentials(),
        'az://ecmwf': icechunk.azure_credentials(sas_token=planetary_cred()['sas_token']),
        'gcs://ecmwf-open-data': None,
})
[...]

PR #1568, opened 2026-01-20
Add warning for large number of chunks in a single commit

li-em:
As part of debugging #1558, add an early warning when the number of chunks in a single commit is reaching the limits of the current IC implementation.

Also add an example for generating millions of chunk_refs more easily in the future.

### DIFF
--- a/icechunk/src/change_set.rs
+++ b/icechunk/src/change_set.rs
@@ -19,8 +19,12 @@ use crate::{
     session::{SessionErrorKind, SessionResult, find_coord},
 };
 
-// Constant derived from docs in https://icechunk.io/en/stable/performance/#splitting-manifests
-const NUM_CHUNKS_LIMIT: u64 = 1_000_000;
+// We have limitations on how many chunks we can save on a single commit.
+// Mostly due to flatbuffers not supporting 64-bit offsets,
+// but also because we can't do transaction log splitting (like we can with manifests).
+// For now we suggest smaller commits as a solution.
+// See discussion in https://github.com/earth-mover/icechunk/issues/1558 for more details
+const NUM_CHUNKS_LIMIT: u64 = 50_000_000;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ArrayData {
@@ -481,7 +485,7 @@ impl ChangeSet {
 
         if edits.num_chunks > NUM_CHUNKS_LIMIT && !edits.excessive_num_chunks_warned {
             warn!(
-                "There are more than {NUM_CHUNKS_LIMIT} chunk references being loaded into this commit. This is close to the maximum number of chunk modifications Icechunk supports in a single commit, we recommend to split into smaller commits and/or setting up manifest splitting for this array. More info at https://icechunk.io/en/stable/performance/#splitting-manifests",
+                "There are more than {NUM_CHUNKS_LIMIT} chunk references being loaded into this commit. This is close to the maximum number of chunk modifications Icechunk supports in a single commit, we recommend to split into smaller commits."
             );
 
             edits.excessive_num_chunks_warned = true;


icechunk/examples/limits_chunk_refs.rs
```diff
@@ -0,0 +1,134 @@
+use std::path::PathBuf;
```
paraseba, line 1: I think we run all examples in GHA, can you check and skip this one that takes a long time and a bunch of memory?

icechunk/src/change_set.rs
```diff
@@ -27,13 +31,22 @@ pub struct ArrayData {
...
 pub struct EditChanges {
     new_groups: HashMap<Path, (NodeId, Bytes)>,
     new_arrays: HashMap<Path, (NodeId, ArrayData)>,
     updated_arrays: HashMap<NodeId, ArrayData>,
     updated_groups: HashMap<NodeId, Bytes>,
     set_chunks: BTreeMap<NodeId, HashMap<ManifestExtents, SplitManifest>>,
+
+    // Number of chunks added to this change set
+    #[serde(skip)]
+    num_chunks: u64,
```
paraseba, line 47: why skipping these from serialization?

icechunk/src/change_set.rs
```diff
@@ -461,6 +476,19 @@ impl ChangeSet {
...
             .or_default()
             .insert(coord, data);
+
+        if old.is_none() {
+            edits.num_chunks += 1;
+        }
+
+        if edits.num_chunks > NUM_CHUNKS_LIMIT && !edits.excessive_num_chunks_warned {
+            warn!(
+                "There are more than {NUM_CHUNKS_LIMIT} chunk references being loaded into this commit. This is close to the maximum number of chunk modifications Icechunk supports in a single commit, we recommend to split into smaller commits and/or setting up manifest splitting for this array. More info at https://icechunk.io/en/stable/performance/#splitting-manifests",
```
paraseba: I think the advice of splitting manifest is not correct, they are still going to hit the tx log limit. I'd just recommend smaller commits

icechunk/src/change_set.rs
```diff
@@ -42,6 +55,21 @@ pub struct EditChanges {
     deleted_arrays: HashSet<(Path, NodeId)>,
 }
 
+impl PartialEq for EditChanges {
```
paraseba: i don' tthink we need this if we don't serde-skip them?

icechunk/src/change_set.rs
```diff
@@ -18,6 +19,9 @@ use crate::{
     session::{SessionErrorKind, SessionResult, find_coord},
 };
 
+// Constant derived from docs in https://icechunk.io/en/stable/performance/#splitting-manifests
+const NUM_CHUNKS_LIMIT: u64 = 1_000_000;
```
paraseba: why so low? Shouldn't this be something like 50M? We should only warn when something is close to breaking, not when the should be splitting manifests for performance

icechunk/examples/limits_chunk_refs.rs
li-em: This one takes 15s to run with 1M chunks (the default if you don't pass arguments in the CLI). Will filter it out in the `Justfile`

icechunk/src/change_set.rs
li-em: I saw a proptest failure, but it seems it was related to test taking too long to run (generate valid data) in linux arm64:
https://github.com/earth-mover/icechunk/actions/runs/21300402081/job/61316446036?pr=1568#step:8:179

icechunk/src/change_set.rs
li-em: I suspected this was what triggered the proptest error, will remove

icechunk/examples/limits_chunk_refs.rs
paraseba: Feel free to make it smaller by default and run it then. As long as it doesn't take more than ~ 1sec

icechunk/src/change_set.rs
paraseba: I think that's unrelated, we got some new proptest strategies lately that may have some bugs

icechunk/src/change_set.rs
li-em: Bumped to 50M, and add a small explanation and pointing to https://github.com/earth-mover/icechunk/issues/1558 for longer discussion

icechunk/src/change_set.rs
li-em: Fixed, only mentioning smaller commits now

PR #1570, opened 2026-01-22
Add flatbuffer schematics

dcherian:
I found these useful. In general, they *seem* right but I don't know enough to completely review it myself.


icechunk/src/format/flatbuffers/README.md
```diff
@@ -0,0 +1,408 @@
...
+    +-- <snapshot-id>       # Transaction log (TransactionLog flatbuffer)
+    +-- ...
+```
+
+## File Header Format (8 bytes)
+
+All flatbuffer files (except chunks) have an 8-byte header:
+
+```
++---+---+---+---+---+---+---+---+
```
paraseba: Something fishy because we use 12 magic bytes:
```
    pub const ICECHUNK_FORMAT_MAGIC_BYTES: &[u8] = "ICE🧊CHUNK".as_bytes();
```

paraseba: There is also the IC implementation name i believe.

PR #1572, opened 2026-01-23
Retry on 408

dcherian:
I have reviewed this code and the aws-smithy code. IIUC `.retry_classifier` calls `.push_retry_classifier`, so this is additive. We add the 408 status code for retries; we see this error code with Tigris.

Relevant code: https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs#L124

Co-Authored-By: Claude <noreply@anthropic.com>


icechunk/src/storage/s3.rs
```diff
@@ -192,6 +193,12 @@ pub async fn mk_client(
 
     s3_builder = s3_builder.identity_cache(id_cache);
 
+    // Add retry classifier for HTTP 408 (Request Timeout)
+    // The default HttpStatusCodeClassifier only retries on 500, 502, 503, 504
```
paraseba, line 197: are we sure this is additive?

paraseba: Can you include a comment in the code, maybe with a link, that shows that is additive?

dcherian: Yes i had the same worry. I've navigated through the call chain and am convinced it's fine.

We are eventually calling `push`
https://github.com/smithy-lang/smithy-rs/blob/cfcc39cf4b5bea665bba684b64bfca2b89e4bc73/rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs#L755

on this `Vec<SharedRetryClassifier>`
https://github.com/smithy-lang/smithy-rs/blob/cfcc39cf4b5bea665bba684b64bfca2b89e4bc73/rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs#L370

I've added a comment

PR #1576, opened 2026-01-26
Add `store.set_virtual_refs_arr()`

### LINKED ISSUE #1574: Method to allow setting array of virtual refs in a zero-copy way
Currently VirtualiZarr uses icechunk's `store.set_virtual_refs` method when writing virtual references. But this implementation is super slow on both the VZ and I think on the IC side too.

The problem is that while both libraries store collections of virtual references in a pretty efficient manner (virtualizarr uses numpy arrays internally, and IC presumably does something sensible in rust), the only way to pass them across right now is as a list of python objects, one element per chunk.

This is wildly inefficient - we're creating millions upon millions of python objects just to immediately unpack them again. All we really need to do here is pass some kind of array between the two libraries.

Previously I suggested using numpy like VZ uses, but @paraseba understandably doesn't want icechunk-python to take on the heavy dependency of numpy.

Instead, what about if we use arrow? It seems to be the more lightweight and targeted dependency for solving exactly this problem. The only tricky part is how to represent the multidimensional nature of a zarr `ManifestArray` using 1D Arrow arrays. I think this proposal would work though:

Add a new method to Icechunk `store.set_virtual_refs_arr`, with API:

```python
if TYPE_CHECKING:
    import pyarrow as pa


class IcechunkStore:
    def set_virtual_refs_arr(
        array_path: str,
        chunk_grid_shape: tuple[int, ...],
        locations: pa.StringArray,
        offsets: pa.IntegerArray,
        lengths: pa.IntegerArray,
        *,
        etag_checksum: pa.StringArray | None = None,
        arr_offset: tuple[int, ...] | None = None,
    ) -> None
        """
        Set an array of virtual references at once in a zero-copy fashion.

        Requires PyArrow to be installed.

        Parameters
        ----------
        array_path
            The path to the array inside the Zarr store. Example: "/groupA/groupB/outputs/my-array"
        chunk_grid_shape
            Shape of the grid of chunks the pyarrow arrays represent. The product of the elements of this tuple must equal the size of each of the pyarrow arrays.
        locations
        offsets
        lengths
        etag_checksum, optional
        arr_offset, optional
            Offset in chunk grid space (like for `.shift_array`)
        """
```

This would allow me to do zero-copy conversion from the numpy arrays inside `vz.ManifestArray`, then presumably IC can still work with these arrow arrays efficiently on the rust side too. This should massively slash both memory usage and execution time of writing large numbers of virtual refs.

The only trick here is that by passing `chunk_grid_shape` (and assuming that all refs form a contiguous rectilinear block), I can get away with representing ND arrays as 1D arrays. Then I can just construct the pyarrow arrays on the virtualizarr side via `pa.array(ManifestArray.chunkmanifest._locations.ravel())` etc. This way we don't have to explicitly pass the N-dimensional chunk indices (which could be tricky in 1D array form).

The `append_dim` case would be supported by the `arr_offset` kwarg.

I don't need a `validate_containers` kwarg as (a) I prefer to do it on the virtualizarr side, and (b) 
I think I can use `pyarrow.compute.startswith()` to do this check very efficiently on the virtualizarr side. 

PyArrow would become an optional dependency of icechunk-python, and an optional dependency of VirtualiZarr (bundled along with the `virtualizarr[icechunk]` dependency group). `arrow-rs` would become a dependency of the icechunk rust crate, if it isn't already.

I've never used arrow before so maybe there are other tricks/problems I'm missing here. (Dictionary encoding?)

TomNicholas:
Attempt to close #1574. Doesn't seem to decrease memory usage, but allows https://github.com/zarr-developers/VirtualiZarr/pull/861, which slashes write time by ~6x.

### DIFF
--- a/icechunk/src/store.rs
+++ b/icechunk/src/store.rs
@@ -517,7 +517,7 @@ impl Store {
             .filter(|&i| !locations.is_null(i))
             .map(|i| {
                 let indices = flat_to_nd_indices(i, chunk_grid_shape, arr_offset);
-                let location = VirtualChunkLocation::from_absolute_path(locations.value(i))
+                let location = VirtualChunkLocation::from_url(locations.value(i))
                     .map_err(|e| StoreErrorKind::Other(e.to_string()))?;
                 let vref = VirtualChunkRef {
                     location,
--- a/icechunk/src/change_set.rs
+++ b/icechunk/src/change_set.rs
@@ -463,7 +463,6 @@ impl ChangeSet {
         &mut self,
         node_id: NodeId,
         chunks: I,
-        splits: &ManifestSplits,
     ) -> SessionResult<()>
     where
         I: IntoIterator<Item = (ChunkIndices, Option<ChunkPayload>)>,
@@ -473,21 +472,10 @@ impl ChangeSet {
         let node_chunks = edits
             .set_chunks
             .entry(node_id)
-            .or_insert_with(|| {
-                HashMap::<
-                    ManifestExtents,
-                    BTreeMap<ChunkIndices, Option<ChunkPayload>>,
-                >::with_capacity(splits.len())
-            });
+            .or_default();
 
         for (coord, data) in chunks {
-            #[allow(clippy::expect_used)]
-            let extent = splits.find(&coord).expect("logic bug. Trying to set chunk ref but can't find the appropriate split manifest.");
-
-            let old = node_chunks
-                .entry(extent.clone())
-                .or_default()
-                .insert(coord, data);
+            let old = node_chunks.insert(coord, data);
 
             if old.is_none() {
                 edits.num_chunks += 1;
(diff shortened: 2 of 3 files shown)


icechunk/src/store.rs
```diff
@@ -414,29 +414,116 @@ impl Store {
...
+                    offset: offsets.value(i),
+                    length: lengths.value(i),
+                    checksum: checksum.clone(),
+                };
+                Ok((ChunkIndices(indices), vref))
+            })
+            .collect::<Result<Vec<_>, StoreError>>()?;
+
+        // Delegate to existing method
+        self.set_virtual_refs(array_path, validate_container, refs_iter).await
```
li-em, line 533: If we put all this validation in `icechunk-python`, we don't need to depend on arrow in the rust crate, do we? It would end up with just this call to `.set_virtual_refs()`

TomNicholas: I guess so... I don't actually need to pass arrow all the way down to the core IC rust crate, I just need to pass arrow to `icechunk-python` and have some way to pass refs in bulk from there to the core IC rust crate.

TomNicholas:
We decided to try passing numpy array buffers instead of arrow arrays, that way we can avoid the dependency on arrow entirely.

TomNicholas:
Closed as this was implemented in #2049

Issue #1580, opened 2026-01-27
Add snapshot id to NewCommitUpdate


paraseba:

mpiannucci:
We should also add the repo org and name

PR #1581, opened 2026-01-27
Add snapshot id to NewCommitUpdate repo change

paraseba:
This is an Icechunk 2 on-disk format breaking change.

Closes: #1580

### DIFF
--- a/icechunk-python/src/repository.rs
+++ b/icechunk-python/src/repository.rs
@@ -610,7 +610,7 @@ impl PyBranchResetUpdate {
 impl PyNewCommitUpdate {
     fn __repr__(&self) -> PyResult<String> {
         Ok(format!(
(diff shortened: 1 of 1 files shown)


icechunk/flatbuffers/repo.fbs
```diff
@@ -55,6 +55,7 @@ table BranchResetUpdate {
 }
 table NewCommitUpdate {
   branch: string (required);
+  snap_id: ObjectId12 (required);
```
li-em: This will touch almost every change in this PR, but... is `snapshot_id` too long? :sweat_smile:

paraseba: we have snap_id in several places. We should change them all at once in a separate PR

PR #1583, opened 2026-01-27
Add long form documentaiton explaining rust architecture

ianhi:
opening early to review myself via github and rendered docs

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -20,7 +20,7 @@
 1. Go to the [`Python CI and library release` github action](https://github.com/earth-mover/icechunk/actions/workflows/python-ci.yaml).
 1. Check `Make a PyPI release`.
 1. Uncheck `Use git-based version for nightly builds`.
-1. Under `Branch to build/release (for manual builds only)` select the branch you want to release (`main` for a normal release, `support/v1.x` for a V1 support release).
+1. Under `Branch to build/release (for manual builds only)` select the branch you want to release (`main` for a normal release, `support/v1.x` for a V1 support release). Where it asks which branch workflow you want to use the workflow from just use the one from `main`.
 1. Click the green `Run workflow` button to launch.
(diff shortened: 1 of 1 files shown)


docs/docs/rust/architecture.md
```diff
@@ -0,0 +1,241 @@
...
+    Store-->>User: Bytes
+```
+-->
+
+Key points:
+
+- **ChangeSet is checked first** for uncommitted writes in the current session
+- **Inline chunks** return data directly from memory (no storage fetch)
+- **Ref chunks** (uncommitted but already written) require a storage fetch
+- **Committed chunks** are looked up via the base snapshot's manifests, with caching
```
paraseba, line 100: delete "with caching" (it's not enabled by default)

docs/docs/rust/architecture.md
```diff
@@ -0,0 +1,241 @@
...
+    end
+    Store-->>User: Bytes
+```
+-->
+
+Key points:
+
+- **ChangeSet is checked first** for uncommitted writes in the current session
+- **Inline chunks** return data directly from memory (no storage fetch)
+- **Ref chunks** (uncommitted but already written) require a storage fetch
```
paraseba, line 99: We call them `Native chunks`

docs/docs/rust/architecture.md
```diff
@@ -0,0 +1,241 @@
...
+| Component | Owns/Contains | Used By |
+|-----------|---------------|---------|
+| `Repository` | Config, storage ref | External callers, `Store` |
+| `Session` | `ChangeSet`, snapshot ref, `AssetManager` | `Store`, `Repository` |
+| `ChangeSet` | In-memory modifications | `Session` |
+| `Store` | `Session` (wrapped) | Python bindings |
+| `AssetManager` | Caches, `Storage` ref | `Session` |
+| `Storage` | Backend connection | `AssetManager` |
+
+## Thread Safety
```
paraseba, line 219: I'd delete this section, it can be obtained from the rust documentation and it's a bit confusing.

paraseba: What would be more useful is to know what object contains a reference to other objects, and if they use internal mutability or not. Does the session point to the repo? Does the store hold a `Mutex` to the Session? etc

PR #1599, opened 2026-02-02
Bump min zarr to 3.1.0

dcherian:
Zarr 3.1.0 launched the new dtypes.

 I think this is fine for a min bound.


dcherian:
This bump necessitatd a bump in min Xarray to support Zarr 3.1.0. I bumped dask too to a version from a year ago

PR #1601, opened 2026-02-02
Add two actor stateful tests

dcherian:
Toward #1321


icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -414,6 +415,43 @@ def check_list_prefix_from_root(self) -> None:
...
+    @rule()
+    def upgrade_spec_version(self) -> None:
+        pass
+
+    @rule(data=st.data())
+    def reopen_with_new_actor(self, data: st.DataObject) -> None:
+        # We use the Zarr's memory store as the model,
+        # Since we cannot `reset_branch` on the model; we must commit here.
+        if self.store.session.has_uncommitted_changes:
+            self.commit_with_check(data)
```
ianhi, line 384: Can we maintain a commited model. and an uncommited model? to allow these two repos to have truly diverged state?

ianhi (approved):
seems to overall make sense and be an improvement. espcially for enabling v1 vs v2.

How much more difficult would it be to have the methods randomly select between which actor is calling it instead of this strategy of reopening to swap actors?

icechunk-python/tests/test_zarr/test_stateful.py
dcherian: I'm not sure what we'd gain? Commits will conflict...

PR #1602, opened 2026-02-02
FEAT: expose session type as an enum

ianhi:
helpful for testing and likely helpful for some edgecases for users

This was previously available in IC1 via `read_only` which was sufficient to differentiate between readonly and writable. but with re-arrange you either need a new boolean or to expose this enum.

One other thing I noticed when doing this is that readonly is determined by whether or the session has an associated branch instead of atracking it's state internally. This seems mostly fine but I found it a bit strange that:

gives:

```python
import icechunk as ic

# Create an in-memory repo
storage = ic.in_memory_storage()
repo = ic.Repository.create(storage)

# Create a readonly session from the same branch
readonly_from_branch = repo.readonly_session(branch="main")
print(f"  branch: {readonly_from_branch.branch}")

```

```
  branch: None
```

even though this readonly session is clearly associated with main.

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -53,10 +53,10 @@ jobs:
       matrix:
         deps-version:
           - name: minimum
-            xarray: "2025.01.2"
-            dask: "2024.11.0"
-            distributed: "2024.11.0"
-            zarr: "3.0.8"
+            xarray: "2025.07.1"
+            dask: "2025.2.0"
+            distributed: "2025.2.0"
+            zarr: "3.1.0"
           - name: latest
             xarray: "latest-release"
             dask: "latest-release"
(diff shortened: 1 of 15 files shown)


dcherian:
```
python/icechunk/session.py:87: error: "PySession" has no attribute "mode"  [attr-defined]
python/icechunk/__init__.py:5: error: Module "icechunk._icechunk_python" has no attribute "SessionMode"  [attr-defined]
```

mypy failure is real

li-em:
> ```
> python/icechunk/session.py:87: error: "PySession" has no attribute "mode"  [attr-defined]
> python/icechunk/__init__.py:5: error: Module "icechunk._icechunk_python" has no attribute "SessionMode"  [attr-defined]
> ```
> 
> mypy failure is real

sadly the enum need to be redeclared in `icechunk-python/python/icechunk/_icechunk_python.pyi`, [example here](https://github.com/earth-mover/icechunk/pull/1547/changes#diff-911ee317e7a7771354968cd643a488a106432be747d731255a4d262bd00d4de6R1841-R1860)

ianhi:
I also thought about making new classes for each so that users could do type hints properly. but that seemed far more invasive. was that the rigth call?

ianhi:
huh. auto merge ignored the failing code quality check 😠

PR #1603, opened 2026-02-03
Support empty commits

### LINKED ISSUE #1595: Allow empty commits
These are useful in certain contexts. We need to allow them to happen, and have `diff` and everything work. But at the same time we don't want users to produce empty commits by mistake. So this may require a flag to commit, `allow_empty = True` or something.

mpiannucci:
Closes #1595

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -4362,10 +4362,7 @@ mod tests {
             .map_ok(|si| si.id)
             .try_collect()
             .await?;
-        assert_eq!(ancestry.len(), 3);
-        assert_eq!(ancestry[0], snap2);
-        assert_eq!(ancestry[1], snap1);
-        assert_eq!(ancestry[2], Snapshot::INITIAL_SNAPSHOT_ID);
+        assert_eq!(ancestry, vec![snap2, snap1, Snapshot::INITIAL_SNAPSHOT_ID]);
 
         Ok(())
     }


icechunk/src/session.rs
```diff
@@ -4288,6 +4284,28 @@ mod tests {
...
+        let snap2_info = repo.lookup_snapshot(&snap2).await?;
+        assert_eq!(snap2_info.parent_id, Some(snap1.clone()));
+
+        let diff = repo
+            .diff(
+                &VersionInfo::SnapshotId(snap1.clone()),
+                &VersionInfo::SnapshotId(snap2.clone()),
+            )
+            .await?;
+        assert!(diff.is_empty());
```
dcherian, line 4356: One other assertion to add here is that number of manifests didn't change (we'll need to write an array with multiple manifests using a splitting config). Can you/Claude add an empty commit to this test please: https://github.com/earth-mover/icechunk/blob/787626c398294392c0a0397ffc9424f242c7c12b/icechunk/src/session.rs#L3008

icechunk/src/session.rs
```diff
@@ -3118,6 +3123,26 @@ mod tests {
...
 
+        let session =
+            repo.readonly_session(&VersionInfo::SnapshotId(updated_snapshot)).await?;
+        let updated_manifest_count = match session.get_array(&array_path).await?.node_data
+        {
+            NodeData::Array { manifests, .. } => manifests.len(),
+            NodeData::Group => panic!("expected array at /array"),
+        };
+        assert_eq!(updated_manifest_count, initial_manifest_count);
+
```
```suggestion
        assert_manifest_count(repo.asset_manager(), initial_manifest_count)

```

icechunk/src/session.rs
```diff
@@ -3118,6 +3123,26 @@ mod tests {
...
+        let mut session = repo.writable_session("main").await?;
+        let empty_snapshot = session.commit("empty commit", None).await?;
+        let session =
+            repo.readonly_session(&VersionInfo::SnapshotId(empty_snapshot)).await?;
+        let empty_manifest_count = match session.get_array(&array_path).await?.node_data {
+            NodeData::Array { manifests, .. } => manifests.len(),
+            NodeData::Group => panic!("expected array at /array"),
+        };
+        assert_eq!(empty_manifest_count, initial_manifest_count);
+
```
```suggestion
         assert_manifest_count(repo.asset_manager(), initial_manifest_count)

```

mpiannucci:
thanks @dcherian did not see that new assert, beautiful

dcherian (approved):
LGTM. One minor quibble is that we could hard-code `initial_manifest_count`

dcherian:
shoot that helper is in repository.rs for some reason; perhaps move the test there?

dcherian:
Right, this is missing the bit where a user needs to opt-in with `allow_empty=True`. 

Otherwise users can fork a session, make writes to the ForkSession, and then commit the empty base session without merging, resulting in surprised pikachu face.

mpiannucci:
My baddd. i completely missed that bit 🤦 I was hoping to avoid touching python bindings lol

mpiannucci:
Okay so in order to preserve API compatibility, i added a new `commit_with_options` function. Open to how this could be better though

dcherian:
works for me but we should get Seba's approval on the new `commit_with_options`

icechunk/src/session.rs
```diff
@@ -4288,6 +4328,35 @@ mod tests {
...
+        let result = session.commit("an empty commit", None).await;
+        assert!(matches!(
+            result,
+            Err(SessionError { kind: SessionErrorKind::NoChangesToCommit, .. })
+        ));
+
+        let mut session = repo.writable_session("main").await?;
+        let snap2 = session.commit_with_options("an empty commit", None, true).await?;
+        let snap2_info = repo.lookup_snapshot(&snap2).await?;
+        assert_eq!(snap2_info.parent_id, Some(snap1.clone()));
```
paraseba, line 4348: can we assert the `ancestry`?

paraseba (approved):

PR #1606, opened 2026-02-03
TEST: Add stateful move

ianhi:
Builds on top of https://github.com/earth-mover/icechunk/pull/1602 to allow using preconditions appropriately. 

Beyond adding the move functionality I made two important changes:

1. check_list_dir

now will check the entire tree instead of a random directory
is used as an invariant

I think this is pretty low cost so wasn't worried about calling it more frequently.

2. added a rule to all inherited methods

Because we can't write to a store during a rearrange session we have to preclude several of the inherited methods from running during a re-arrange session. as far as I can tell this requires overriding them to add our pre-condtion. and also retyping the pre-conditions from the original method.

part of https://github.com/earth-mover/icechunk/issues/1513

### DIFF
--- a/icechunk-python/tests/test_zarr/test_stateful.py
+++ b/icechunk-python/tests/test_zarr/test_stateful.py
@@ -294,7 +294,6 @@ class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
         assume(array.size > 0)
         super().add_array(data, name, array_and_chunks)
 
-    @rule(data=st.data())
     def _compare_list_dir(
         self, model: ModelStore, store: ic.IcechunkStore, paths: set[str]
     ) -> None:


icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -165,11 +237,92 @@ def upgrade_spec_version(self) -> None:
...
+            if path.startswith(source + "/"):
+                # child path: "foo/child" -> "bar/child"
+                return dest + path[len(source) :]
+            # unrelated path: "other" -> "other"
+            return path
+
+        self.all_arrays = {remap(p) for p in self.all_arrays}
+        self.all_groups = {remap(p) for p in self.all_groups}
+
+    def move_in_model(self, source: str, dest: str) -> None:
```
dcherian: the pattern is to put model actions as a method on model s oyou just call `self.model.move`

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -165,11 +237,92 @@ def upgrade_spec_version(self) -> None:
         self.repo = Repository.open(self.storage)
         self.store = self.repo.writable_session("main").store
 
+    # NOTE:
+    # Bundles (https://hypothesis.readthedocs.io/en/latest/stateful.html#rule-based-state-machines)
```
dcherian: Bundles have other limitations (https://github.com/HypothesisWorks/hypothesis/issues/3944). I always start with bundles and end up removing them

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -179,24 +332,180 @@ def add_array(
...
 
-    @precondition(lambda self: bool(self.all_groups))
+    @rule(name=node_names, data=st.data())
+    @precondition(lambda self: self.store.session.mode == SessionMode.WRITABLE)
+    def add_group(self, name: str, data: st.DataObject) -> None:
+        super().add_group(name, data)
+
     @rule(data=st.data())
-    def check_list_dir(self, data: st.DataObject) -> None:
+    @precondition(lambda self: bool(self.all_arrays))
```
dcherian: This is a BIG wart.

How about in `move` we create rearrange session; execute an arbitrary series of moves; and then either commit or discard.

Rearrange sessions only allow move so the only sequence of ops is a bunch of moves and a commit or discard.

ianhi: doesn't that mess with the hypothesis sampling mechanism a bit? in the sense that moves are now decoupled from rules?
Overall agree thouhg. I was pretty unhappy with this.

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: The problem is is that model is just a MemoryStore and is instantiated in the super class. So i guess we could monkeypatch it on, or do some trickery with a subclass

icechunk-python/tests/test_zarr/test_stateful.py
dcherian: there will be one move rule, which will sample an arbitrary number of source/destination pairs. It should still work oK and be much less ugly.

icechunk-python/tests/test_zarr/test_stateful.py
dcherian: oh right. I forgot that; 🤦🏾‍♂️ just add a comment

ianhi: nevermind this wasn't as hard as I thought.

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: feels like I almost want a statemachine to be able to spawn a sub-statemachine as one of its steps. then that machine runs until completion.

ianhi:
Ok. move is now self contained. to support multiple moves within and to allow potentially discarding a commit instead of committing it. To support this I implemented a way to clone the model and make changes on that copy. which is hten adopted as the model if we commit, and otherwise thrown away.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -165,6 +261,84 @@ def upgrade_spec_version(self) -> None:
...
+                    # Destination must not already exist: foo -> bar (when bar exists)
+                    and dest not in existing_nodes
+                ):
+                    # Found a valid move - break out of discovery loop
+                    break
+            else:
+                # we shouldn't be reaching here based on our precondition
+                # this is for safety
+                raise AssertionError("Could not find valid move after 100 attempts")
+
```
dcherian:
```suggestion
                # we shouldn't be reaching here based on our precondition
                # this is for safety
                return
```

we needn't bork the test run for this

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -165,6 +261,84 @@ def upgrade_spec_version(self) -> None:
...
+                    else p
+                    for p in s
+                }
+                for s in (pending_arrays, pending_groups)
+            ]
+
+            # Verify store matches pending model after each move
+            # failing due to https://github.com/earth-mover/icechunk/issues/1562#issuecomment-3755544352
+            # TODO: uncomment when that is fixed
+            # self._compare_list_dir(
```
dcherian, line 265: can we not call `self.check_list_dir()`?

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -165,6 +199,84 @@ def upgrade_spec_version(self) -> None:
...
+                source_name = source.split("/")[-1]
+                new_name = (
+                    source_name if data.draw(st.booleans()) else data.draw(node_names)
+                )
+                dest_parent = data.draw(st.sampled_from(possible_parents))
+                dest = f"{dest_parent}/{new_name}".lstrip("/")
+                # Only move to existing groups (non-existent parent loses data)
+                if (
+                    # Can't move into itself: foo -> foo/bar
+                    dest_parent != source
```
dcherian: this and the following condition could be a filter on `st.sampled_from(possible_parents)`?

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: not here. because this is the intermediate step where I haven't modifed the base model yet, in case we end up discarding.

adds a bit of complexity, but i thought it was nice to have the option to discard.

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: yes. refactored to be much nicer

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: with the new changes i don't think this is necssary. pretty sure we will always find a valid move

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -181,22 +292,36 @@ def add_array(
...
-            "c"
-        }:
-            # Consider .list_dir("path/to/array") for an array with a single chunk.
-            # The MemoryStore model will return `"c", "zarr.json"` only if the chunk exists
-            # If that chunk was deleted, then `"c"` is not returned.
-            # LocalStore will not have this behaviour :/
-            # In Icechunk, we always return the `c` so ignore this inconsistency.
-            assert model_ls == store_ls, (model_ls, store_ls)
+    @precondition(
+        # Parent filters to nested groups (containing "/"), so we need at least one
```
dcherian: didn't understand this comment.

ianhi: parent as in the super class. I think this is basically the solution to this comment: https://github.com/zarr-developers/zarr-python/blob/b712f96d60da811e737cf23569e1ccc64281b7a5/src/zarr/testing/stateful.py#L343

but i should have removed this when i removed all my over superclass overides

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -181,22 +292,36 @@ def add_array(
...
+        lambda self: any("/" in g for g in self.all_groups)
+    )
+    def delete_group_using_del(self, data: st.DataObject) -> None:
+        super().delete_group_using_del(data)
```
ianhi:
```suggestion
```

PR #1608, opened 2026-02-03
Update dask + forking error message

dcherian:
People keep asking about this.

Basically

```
ds = xr.open_dataset(session.store, group="foo", engine="zarr", consolidated=False, chunks={})
to_icechunk(ds, session, group="bar")
```

will fail in a distributed context. The solution is to either pass `readonly_session` to open_dataset or fork before open_dataset. 

I can't think of any possible workarounds here, except going to back to unpickling to a read-only store. @paraseba do we want to do the latter?


paraseba:
Let's try this @dcherian . Let's see if the rate of questions drops.

paraseba (approved):
this is a good candidate to backport

PR #1609, opened 2026-02-03
Don't panic when writing manifest fails.

### LINKED ISSUE #1586: Rust panic on S3 write timeout during concurrent commits poisons shared Repository state
## Description

When multiple threads commit concurrently to the same `Repository` object and one thread hits an S3 PUT timeout during `write_manifest`, icechunk panics instead of returning an error. The panic corrupts shared state, causing all subsequent operations on the same `Repository` (across all threads) to fail with `dispatch failure` / `get_ref` errors.

## Reproduction

- 8 Python threads, each writing zarr groups to the same `Repository` via `writable_session("main")` + `session.commit()` with `BasicConflictSolver` and rebase
- S3 backend (`us-west-2`), `max_concurrent_requests = 10`
- High commit throughput (~100+ commits in a few minutes)
- After ~200 successful commits, one thread hits an S3 HTTP connect timeout (3.1s)

## Panic trace

```
thread '<unnamed>' panicked at icechunk/src/session.rs:1779:79:
logic bug. for a new node, we must always write the manifest: ICError {
  kind: RepositoryError(StorageError(S3PutObjectError(DispatchFailure(DispatchFailure {
    source: ConnectorError {
      kind: Timeout,
      source: hyper_util::client::legacy::Error(Connect, HttpTimeoutError {
        kind: "HTTP connect", duration: 3.1s
      }),
      connection: Unknown
    }
  })))),
  context: SpanTrace [
    { target: "icechunk::storage::s3", name: "write_manifest", fields: "id=DZ88Y8A9N3HC7Q7ZTC00" },
    { target: "icechunk::asset_manager", name: "write_manifest" },
    { target: "icechunk::session", name: "_commit" },
    { target: "icechunk::session", name: "commit" }
  ]
}
```

## Cascading failure

After the panic, every other thread immediately fails:
```
error getting object from object store dispatch failure
  context: icechunk::storage::s3::get_ref with ref_key="branch.main"
```

This continues for all remaining writes — the `Repository` object is permanently broken.

## Expected behavior

1. An S3 timeout during `write_manifest` should return a `StorageError` to Python, not panic. The assertion at `session.rs:1779` assumes `write_manifest` always succeeds, but transient network failures are expected under high concurrency.
2. A single thread's S3 failure should not corrupt the shared connection pool / internal state used by other threads.

## Workaround

From Python, we catch `BaseException` (since `pyo3_runtime.PanicException` doesn't inherit from `Exception`) and retry with a fresh session and exponential backoff. This handles the panic for the affected thread, but does not prevent the cascading failure across other threads sharing the same `Repository`.

## Environment

- icechunk Python package (latest from PyPI)
- Python 3.12
- S3 storage, `us-west-2`
- 8 concurrent threads via `ThreadPoolExecutor`

dcherian:
This was my mistake I believe :( . It isn't a logic bug and writing manifest files can fail for valid reasons.

Closes #1586

cc @espg

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -102,7 +102,9 @@ pub enum SessionErrorKind {
     NotAGroup { node: Box<NodeSnapshot>, message: String },
     #[error("node already exists at `{node:?}`: {message}")]
     AlreadyExists { node: Box<NodeSnapshot>, message: String },
-    #[error("cannot commit, no changes made to the session")]
+    #[error(
+        "cannot commit, no changes made to the session (use `allow_empty=true` to commit anyway)"
+    )]
     NoChangesToCommit,
     #[error("invalid snapshot timestamp ordering. parent: `{parent}`, child: `{child}` ")]
     InvalidSnapshotTimestampOrdering { parent: DateTime<Utc>, child: DateTime<Utc> },
--- a/icechunk-python/tests/test_session.py
+++ b/icechunk-python/tests/test_session.py
@@ -190,8 +190,5 @@ def test_session_mode() -> None:
     # after commit, session becomes readonly
     writable = repo.writable_session("main")
     assert writable.mode == SessionMode.WRITABLE
-    # opening a group to make a change is necessary
-    # until https://github.com/earth-mover/icechunk/issues/1595 fixed
-    zarr.group(writable.store)
-    writable.commit("test")
-    assert writable.mode == SessionMode.READONLY
+    writable.commit("test", allow_empty=True)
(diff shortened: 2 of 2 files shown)


li-em (approved):

icechunk/src/session.rs
```diff
@@ -2087,9 +2087,7 @@ impl<'a> FlushProcess<'a> {
                         .map(Ok),
                 );
                 #[allow(clippy::expect_used)]
```
li-em: Can also remove the clippy allow here

PR #1611, opened 2026-02-03
Improve empty commit error message to suggest allow_empty

ianhi:
When a commit fails because no changes were made, the error message now suggests using `allow_empty=true` to commit anyway.

I got caught by this and was briefly confused until I went and read the PR.


mpiannucci:
thanks ian!!

Issue #327, opened 2024-10-25
Better packaging


dcherian:
- [ ] minimize rust dependencies, add feature flags
- [ ] move test strategies to its own feature flag or crate

PR #1614, opened 2026-02-04
Move test deps to dev-deps

mpiannucci:


dcherian:
It would be good to prompt Claude into moving those to a test feature: https://github.com/earth-mover/icechunk/issues/327.

I tried a long time ago and failed but almost certainly due to user error: https://github.com/frozenlib/test-strategy/issues/14

mpiannucci:
Whats the benefit of that over just `dev`? thats effectively what this does

dcherian:
Yeah perhaps that's true and we had bad expectations.

PR #1618, opened 2026-02-04
feat: add `Repository.fetch_spec_version` to Python API

### LINKED ISSUE #1480: Expose function to get repo spec version without opening the repo in the Python API

lindseynield:
Makes the private `repo_spec_version` method public (renamed to `fetch_spec_version`) in the core Rust `Repository` struct, then exposes it through the Python bindings as `Repository.fetch_spec_version()` and `Repository.fetch_spec_version_async()` static methods. 

This allows checking a repository's spec version without the overhead of fully opening it. Closes #1480

### DIFF
--- a/icechunk-python/python/icechunk/repository.py
+++ b/icechunk-python/python/icechunk/repository.py
@@ -354,7 +354,7 @@ class Repository:
         Fetch the spec version of a repository without fully opening it.
 
         This is useful for checking the repository format version before opening,
-        for example to determine if a migration is needed.
+        for example to know what version of the library is needed to open it.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


icechunk/src/repository.rs
```diff
@@ -431,11 +431,11 @@ impl Repository {
     pub async fn exists(
         storage: Arc<dyn Storage + Send + Sync>,
     ) -> RepositoryResult<bool> {
-        Ok(Self::repo_spec_version(storage).await?.is_some())
+        Ok(Self::fetch_spec_version(storage).await?.is_some())
     }
 
     #[instrument(skip_all)]
-    async fn repo_spec_version(
+    pub async fn fetch_spec_version(
```
lindseynield, line 438: Just want to double check that making this function public isn't going to cause problems

paraseba: all good

Issue #1619, opened 2026-02-04
`shift` Different behavior for negative and positive shifts


ianhi:
`shift_array` has different behavior depending on if you pass a negative or positive shift.

Negative shift always works and can delete chunks.

Positive shift will fail - requiring you to resize first. Might be nice if it auto resized for you

I don't have a strong feeling on what is the best behavior, but I do think that they should be the same behavior, possibly with an argument controlling the behavior.

```python

import numpy as np

import icechunk as ic
import zarr

# Negative shift - silently drops out-of-bounds chunks
repo = ic.Repository.create(ic.in_memory_storage())
session = repo.writable_session("main")
arr = zarr.create(store=session.store, path="arr", shape=(10,), chunks=(2,), dtype="i4")
arr[:] = np.arange(10)
print("before:      ", arr[:])
session.shift_array("/arr", (-2,))
print("shift(-2):   ", arr[:])

# Positive shift - must resize first to make room
repo = ic.Repository.create(ic.in_memory_storage())
session = repo.writable_session("main")
arr = zarr.create(store=session.store, path="arr", shape=(10,), chunks=(2,), dtype="i4")
arr[:] = np.arange(10)
print("before:      ", arr[:])
arr.resize((14,))  # add room for 2 more chunks
print("resized:     ", arr[:])
session.shift_array("/arr", (2,))
print("shift(2):    ", arr[:])

# Positive shift without resize - errors
repo = ic.Repository.create(ic.in_memory_storage())
session = repo.writable_session("main")
arr = zarr.create(store=session.store, path="arr", shape=(10,), chunks=(2,), dtype="i4")
arr[:] = np.arange(10)
print("before:      ", arr[:])
session.shift_array("/arr", (2,))  # errors
```

```
before:       [0 1 2 3 4 5 6 7 8 9]
shift(-2):    [4 5 6 7 8 9 6 7 8 9]

before:       [0 1 2 3 4 5 6 7 8 9]
resized:      [0 1 2 3 4 5 6 7 8 9 0 0 0 0]
shift(2):     [0 1 2 3 0 1 2 3 4 5 6 7 8 9]

before:       [0 1 2 3 4 5 6 7 8 9]
Traceback (most recent call last):
  File "/Users/ian/Documents/dev/icechunk/icechunk-python/shift_asymmetry.py", line 34, in <module>
    session.shift_array("/arr", (2,))  # errors
[...]

PR #1621, opened 2026-02-05
FEAT/FIX: more consistent move behavior + options on reindex

### LINKED ISSUE #1513: Add new IC2 actions to stateful test suite

ianhi:
Closes #1513 

This PR makes (to my eye) the behavior of shift_array and reindex more consistent and explicit about user intent. 

**discard asym**
Prior:

asymmetric behavior of chunk deletion vs error for negative and positive shfits: https://github.com/earth-mover/icechunk/issues/1619

now: symmetric discard for both negative and positive

**shift_array return**

since shift array works in chunk space it might be convenient for the user get back the shifts in array index space. I added a return to shift array for this.

## Docs

move, shift_array, and reindex have docs pages now!

shift/reindex: https://icechunk--1621.org.readthedocs.build/en/1621/moving-chunks/
move: https://icechunk--1621.org.readthedocs.build/en/1621/moving-nodes/

### DIFF
--- a/icechunk-python/docs/docs/moving-chunks.md
+++ b/icechunk-python/docs/docs/moving-chunks.md
@@ -30,7 +30,7 @@ shift_array("/arr", (-2,)) # → shifts by 2 chunks = 4 elements
 
 Why chunks instead of elements? Because these are **metadata-only operations**. Shifting by partial chunks would require splitting and rewriting chunk data.
 
-## shift_array { #shift_array }
+## `shift_array` { #shift_array }
 
 The [`shift_array`][icechunk.Session.shift_array] method moves all chunks by a fixed offset per dimension (negative to shift toward index 0, positive toward higher indices). Out-of-bounds chunks are discarded, and vacated positions retain stale data.
 
@@ -132,7 +132,7 @@ Computing the index shift in element space is straightforward: multiply each chu
 
 This pattern works identically whether your array is 1 KB or 1 PB, and whether it's on local disk or cloud object storage—the shift is always instant with zero data transfer.
 
-## reindex_array { #reindex_array }
+## `reindex_array` { #reindex_array }
 
 For transformations that [`shift_array`][icechunk.Session.shift_array] can't express, [`reindex_array`][icechunk.Session.reindex_array] gives you complete control. You provide a function that maps each chunk's old position to its new position.
 


ianhi:
I still have the stateful tests to fill out with this new behavior (and i need to look more careful at the claude generated unit tests) but I wanted to open now to get a review on the behavior changes before I put that effort in. I know @dcherian was. committed to reviewing this, but for the behavior @paraseba would be great to hear from you.

ianhi:
after talking wiht @dcherian :

1. no more ability to delete vacated chunks - too slow!
2. keep shift array discard behaivor on 
3. move the wrap behavior to a new function roll instead of splitting on arguments


test reindiex by using hypothesis property test (equivalent to vindex with a random array) using permutation

icechunk-python/python/icechunk/session.py
```diff
@@ -170,9 +170,21 @@ def config(self) -> RepositoryConfig:
         return self._session.config
 
     def move(self, from_path: str, to_path: str) -> None:
+        """Move or rename a node (array or group) in the hierarchy.
+
+        This is a metadata-only operation—no data is copied. Requires a rearrange session.
```
paraseba: Oh we didn't have any docs? Thank you!. Can we show them here how to start a rearrange session?

icechunk-python/python/icechunk/session.py
```diff
@@ -191,10 +203,73 @@ def reindex_array(
...
+        Chunks that shift out of bounds are discarded. Vacated positions retain
+        stale chunk references — the caller typically writes new data there.
+
+        Parameters
+        ----------
+        array_path : str
+            The path to the array to shift.
+        chunk_offset : Iterable[int]
+            The number of chunks to shift by in each dimension. Positive values
+            shift right/down, negative values shift left/up.
```
paraseba, line 237: I don't think right/left and down/up mean much. Maybe we explain with a formula? Saying that this offset will be added to each chunk coordinate, so chunk at coordinate `x` moves to `x + chunk_offset` and give an example in 3-D

icechunk-python/python/icechunk/session.py
```diff
@@ -191,10 +203,73 @@ def reindex_array(
...
+
+        Returns
+        -------
+        tuple[int, ...]
```
paraseba: Like this idea for the return type. The documentation may need an example.

icechunk-python/python/icechunk/session.py
```diff
@@ -191,10 +203,73 @@ def reindex_array(
...
+        array.resize(), then use shift_array.
+        """
+        return tuple(self._session.shift_array(array_path, list(chunk_offset)))
+
+    def roll_array(
+        self,
+        array_path: str,
+        chunk_offset: Iterable[int],
+    ) -> tuple[int, ...]:
+        """Roll (circular shift) all chunks in an array by the given chunk offset.
```
paraseba: Is this a common use case? Can you give me examples?

ianhi: Admittedly I can't think of one, but that would've also been true for shift-array. It seems to me like the other natural move operation that users will  eventually want, so nice to not ask them to re-implement. its also the behavior I would've expected naively. 

https://github.com/earth-mover/icechunk/issues/1619#issuecomment-3849450508

icechunk-python/python/icechunk/session.py
ianhi: in the docstring or in the main docs?

jbusecke:
Reminder that I stole the `--livereload` part of the docs into #1640

ianhi:
@paraseba is there anything more here? I think i addressed all of deepak's feedback

icechunk-python/python/icechunk/session.py
ianhi: I removed this behavior. we can always easily bring it back later if we want

icechunk-python/src/session.rs
```diff
@@ -226,19 +226,23 @@ impl PySession {
         })
     }
 
-    pub fn shift_array(&mut self, array_path: String, offset: Vec<i64>) -> PyResult<()> {
+    pub fn shift_array(
+        &mut self,
+        array_path: String,
+        chunk_offset: Vec<i64>,
+    ) -> PyResult<Vec<i64>> {
```
dcherian: I don't see why we should change the return type. Also this will add complexity with other chunk grids. IMO its  best to avoid.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -337,9 +413,41 @@ def _compare_list_dir(
                     f"list_dir mismatch for {path=}: {model_ls=} != {store_ls=}"
                 )
 
-    @invariant()
-    def check_list_dir(self) -> None:
-        self._compare_list_dir(self.model, self.store, self.all_groups | self.all_arrays)
+    @precondition(lambda self: bool(self.all_groups))
+    @rule(data=st.data())
```
dcherian: why did you switch this out from an invariant? It's been a very effective one!

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -337,9 +413,41 @@ def _compare_list_dir(
...
+        exist. MemoryStore only returns 'c' if chunks are present. We ignore
+        this specific difference.
+        """
+        path = self.draw_directory(data)
+        note(f"list_dir for {path=!r}")
+        self._compare_list_dir(self.model, self.store, {path})
+
+    @rule(data=st.data())
+    @precondition(lambda self: bool(self.all_arrays))
+    def check_array_data(self, data: st.DataObject) -> None:
```
dcherian: This already exists in the super class no? I know it's a bit confusing... but it's always good to check the super class first :) It's pretty featureful

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -302,6 +335,49 @@ def valid_name(
...
+                arr_model.shape[i] + max(0, offset[i]) * chunks[i]
+                for i in range(len(chunks))
+            )
+            note(f"resizing array '{array_path}' from {arr_model.shape} to {new_shape}")
+            arr_model.resize(new_shape)
+            arr_store.resize(new_shape)
+            num_chunks = tuple(
+                (new_shape[i] + chunks[i] - 1) // chunks[i] for i in range(len(chunks))
+            )
+
```
dcherian: can we reread cdata_shape instead to avoid the possibility of subtle bugs?

icechunk-python/tests/test_shift.py
```diff
@@ -49,19 +52,50 @@ async def test_shift_using_shift_by_offset() -> None:
...
+    """Test shifting right - vacated positions retain stale data."""
     repo = ic.Repository.create(
         storage=ic.in_memory_storage(),
     )
     session = repo.writable_session("main")
     root = zarr.group(store=session.store, overwrite=True)
+    array = root.create_array(
+        "array", shape=(50,), chunks=(2,), dtype="i4", fill_value=-1
+    )
+    array[:] = np.arange(50)
```
dcherian: FYI `create_array` has a data arg you can use

icechunk-python/tests/test_shift.py
```diff
@@ -71,19 +105,132 @@ async def test_resize_and_shift_right() -> None:
...
+    # 4x4 array with 2x2 chunks = 2x2 chunks grid
+    array = root.create_array(
+        "array", shape=(4, 4), chunks=(2, 2), dtype="i4", fill_value=-1
+    )
+    array[:] = np.arange(16).reshape(4, 4)
+    session.commit("create array")
+
+    session = repo.writable_session("main")
+    root = zarr.group(store=session.store, overwrite=False)
+    # Shift by (1, 0) - shift down by one chunk, keep horizontal
```
dcherian: "shift down" = imaging bias!

icechunk/src/session.rs
```diff
@@ -648,13 +657,52 @@ impl Session {
         Ok(())
     }
 
+    /// Shift all chunks in an array by the given chunk offset.
+    ///
+    /// Out-of-bounds chunks are discarded. Vacated source positions retain stale references.
```
dcherian: Let's recommend resizing to preserve these chunks

icechunk/src/session.rs
```diff
@@ -648,13 +657,52 @@ impl Session {
...
+        chunk_offset: &[i64], // FIXME: overflow
+    ) -> SessionResult<Vec<i64>> {
+        let node = self.get_array(array_path).await?;
+        let (num_chunks, chunk_sizes): (Vec<u32>, Vec<u64>) = match &node.node_data {
+            NodeData::Array { shape, .. } => (
+                shape.num_chunks().collect(),
+                shape.iter().map(|dim| dim.chunk_length()).collect(),
+            ),
+            _ => unreachable!("get_array returned non-array"),
+        };
```
dcherian, line 655: does this whole change get reverted if we go back to the old return type?

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: I think i did that because I found it was taking a large amount of time. but can restore.

icechunk-python/docs/docs/moving-chunks.md
```diff
@@ -0,0 +1,216 @@
+# Moving Chunks
+
```
paraseba, line 2: awesome job on this docs!!!

icechunk-python/python/icechunk/session.py
paraseba: I think it's useful here, because it's a constraint. This method con only be called on a special type of session.

icechunk-python/src/session.rs
paraseba: Good point. Let's not get in trouble in the future for a mere convenience

icechunk/src/session.rs
paraseba: yeah, what's the deal here?

ianhi: yes it all goes

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: ah good call

icechunk/src/session.rs
ianhi: well not. all. because i effectively inlined shift_by i dont think it needs to be a separate function

icechunk-python/docs/docs/moving-chunks.md
```diff
@@ -0,0 +1,220 @@
...
+```python
+# With chunk_size=2:
+shift_array("/arr", (-1,)) #  → shifts by 1 chunk = 2 elements
+shift_array("/arr", (-2,)) # → shifts by 2 chunks = 4 elements
+```
+
+Why chunks instead of elements? Because these are **metadata-only operations**. Shifting by partial chunks would require splitting and rewriting chunk data.
+
+## shift_array { #shift_array }
```
dcherian:
```suggestion
## `shift_array` { #shift_array }
```

icechunk-python/docs/docs/moving-chunks.md
```diff
@@ -0,0 +1,220 @@
...
+arr[element_shift[0]:] = todays_readings
+
+session.commit(f"Updated sensor data for {today}")
+```
+
+Computing the index shift in element space is straightforward: multiply each chunk offset by the corresponding chunk size. This tells you exactly where to write new data.
+
+This pattern works identically whether your array is 1 KB or 1 PB, and whether it's on local disk or cloud object storage—the shift is always instant with zero data transfer.
+
+## reindex_array { #reindex_array }
```
dcherian:
```suggestion
## `reindex_array` { #reindex_array }
```

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -33,9 +32,41 @@
...
 
 class ModelStore(MemoryStore):
-    """MemoryStore with move and copy methods for testing."""
+    """MemoryStore with move, copy, and shift_array methods for testing."""
 
-    spec_version: int
+    async def shift_array(
+        self,
+        array_path: str,
+        offset: tuple[int, ...],
```
dcherian, line 41:
```suggestion
        *, chunk_offset: tuple[int, ...],
```

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -33,9 +32,41 @@
...
+                chunk_data[indices] = data
+
+        # Write chunks at new positions (skip out-of-bounds)
+        for old_idx, data in chunk_data.items():
+            new_idx = tuple(idx + off for idx, off in zip(old_idx, offset, strict=True))
+            if any(
+                idx < 0 or idx >= nchunks
+                for idx, nchunks in zip(new_idx, num_chunks, strict=True)
+            ):
+                continue  # Out of bounds - discard
```
dcherian: minor: I would just do this when setting `chunk_data`

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -294,8 +303,8 @@ def valid_name(
             #     pending_model, session.store, pending_arrays | pending_groups
             # )
 
-        # Prefer commit (75%) over discard (25%)
-        if data.draw(st.sampled_from([True, True, True, False])):
+        # Prefer commit (75%) over discard (25%) to exercise more code paths
```
dcherian: Let's avoid pointless comments like `3:1` odds :P

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -306,6 +315,41 @@ def valid_name(
             note("discarding moves")
             self.store = self.repo.writable_session("main").store
 
+    @rule(data=st.data())
+    @precondition(lambda self: self.repo.spec_version >= 2)
+    @precondition(lambda self: bool(self.all_arrays))
+    def shift_array(self, data: st.DataObject) -> None:
```
dcherian, line 346: could be fun to add a `reindex_array` with arbitrary permutations and subsetting of chunks (does that even work; for e.g. the function maps chunk index 3 to -1); not necessary to merge this PR.

icechunk-python/tests/test_shift.py
```diff
@@ -32,58 +32,199 @@ def reindex(idx: Iterable[int]) -> Iterable[int] | None:
...
+    session.shift_array("/array", (1, 0))
+    array = cast("zarr.Array[Any]", root["array"])
+
+    # First 2 rows retain stale data (original rows 0-1)
+    np.testing.assert_equal(array[0:2, :], np.arange(8).reshape(2, 4))
+    # Rows 2-3 should have original rows 0-1 (shifted from chunk 0 to chunk 1)
+    np.testing.assert_equal(array[2:4, :], np.arange(8).reshape(2, 4))
+
+
+async def test_shift_3d_mixed_offset() -> None:
```
dcherian: You could make this a property test and use `@example` to test the specific values used in this test. Not necessary to merge this PR.

icechunk/src/session.rs
```diff
@@ -607,8 +638,9 @@ impl Session {
                 if shape.valid_chunk_coord(&new_chunk_index) {
                     change_set.set_chunk_ref(
                         node.id.clone(),
-                        new_chunk_index,
+                        new_chunk_index.clone(),
```
```suggestion
                        new_chunk_index,
```

Issue #1617, opened 2026-02-04
Should initial commit have an empty transaction log?


ianhi:
Empty commits currently have an empty transaction log. only the initial commit does not have a transaction log. Maybe it should have an empty one instead? 

xref: https://github.com/earth-mover/icechunk/pull/1538

paraseba:
I like it

ianhi:
I had a go at this but ran into the following:

Does this create a version compatibility issue at all? with diffs, and writing a a v1 format repo, then read by an old client, could this cause issues?

PR #1623, opened 2026-02-05
Give initial snapshot an empty transcation log

ianhi:
For consistency make sure the initial commit has an empty transaction log (just like later empty commits)

closes: https://github.com/earth-mover/icechunk/issues/1617


icechunk/src/repository.rs
```diff
@@ -1446,7 +1458,8 @@ impl Repository {
         let fut: FuturesOrdered<_> = all_snaps
             .iter()
             .filter_map(|snap_info| {
-                if snap_info.is_initial() {
+                // v1 repos don't have transaction logs for initial snapshots
+                if self.spec_version == SpecVersionBin::V1dot0 && snap_info.is_initial() {
```
ianhi, line 1518: not entirely sure if this is necessary, or even good? at first i had fully removed it and justs used a map. but then i got worried about breakage when reading a 1.0 repo without thesse transcations available

icechunk/src/session.rs
```diff
@@ -4391,6 +4391,29 @@ mod tests {
...
 
+    /// Test that the initial snapshot has an empty transaction log that can be fetched.
+    #[tokio_test]
+    async fn test_initial_snapshot_has_empty_transaction_log()
+    -> Result<(), Box<dyn Error>> {
+        let repo = create_memory_store_repository().await;
+        let asset_manager = repo.asset_manager();
+
+        let initial_snap_id = repo.lookup_branch("main").await?;
+        let tx_log = asset_manager.fetch_transaction_log(&initial_snap_id).await?;
```
paraseba, line 4418: Let's test a `diff` against the initial commit

icechunk/src/repository.rs
```diff
@@ -251,6 +252,17 @@ impl Repository {
             asset_manager_c.write_snapshot(Arc::clone(&new_snapshot)).await?;
 
             if spec_version >= SpecVersionBin::V2dot0 {
```
paraseba, line 257: Can we do the writing of the tx log and the snapshot concurrently? LMK if it's too much work.

ianhi: ahh as in only one await so that they async and write at the same time? I'll be i can figure that out

paraseba: There are examples, see `Repository::open` for example. You `spawn` and then `try_join!` or similar.

icechunk/src/session.rs
paraseba: One more thing to test: We create a ic1 repo using the ic2 lib, then read it using the ic1 lib

icechunk/src/repository.rs
```diff
@@ -248,21 +249,38 @@ impl Repository {
...
-                let _ = asset_manager_c.create_repo_info(Arc::clone(&repo_info)).await?;
+
+                // Write snapshot, transaction log, and repo info concurrently
+                // since they target independent storage keys
+                let write_tx = asset_manager_c.write_transaction_log(
+                    Snapshot::INITIAL_SNAPSHOT_ID,
+                    Arc::new(empty_tx_log),
+                );
+                let write_repo = asset_manager_c.create_repo_info(Arc::clone(&repo_info));
+                try_join!(write_snap, write_tx, write_repo)?;
```
ianhi: @paraseba is this how? I nkow comment seems a little silly. but its the kind of comment that would have really helped me if i came across this for the first time.

icechunk/src/repository.rs
```diff
@@ -248,21 +249,38 @@ impl Repository {
...
+                        storage_c.as_ref(),
+                        settings_ref,
+                        Ref::DEFAULT_BRANCH,
+                        new_snapshot.id().clone(),
+                        None,
+                    )
+                    .await
+                    .map_err(RepositoryError::from)
+                };
+                try_join!(write_snap, write_branch)?;
```
paraseba: This introduces a race condition. The branch could be written before the snapshot, rendering the IC repository invalid. Usually, the time interval will be very brief, but, the machine or the request to write the snap could fail, and you end up with a repo that is indefinitely valid.

Here we need to do the operations sequentially.

icechunk/src/repository.rs
paraseba: This introduces a race condition. The repo object is the entry point to the whole repository on disk format. It could happen that the repo object is written before the snapshot, rendering the IC repository invalid. Usually, the time interval will be very brief, but, the machine or the request to write the snap could fail, and you end up with a repo that is indefinitely valid.

The writes of the snap and tx can be done concurrently, but only after both succeed we can write the repo object. In this way, if the process is interrupted, the repository is never created, you only have some garbage in the bucket.

paraseba (changes requested):

icechunk/src/session.rs
ianhi: Can we do this in rust? Or you mean as a one-off I test it?

icechunk/src/repository.rs
ianhi: yikes.  I totally missed this. this async stuff is scary.  I've split what I had into two separate `try_join`

I looked at repo open and it uses `tokio:spawn`. do i need that here?  i.e. since this is IO dominated (i think) its k to be on the same thread? So is there an advantage of spawn over this (seemignly simpler) approach?

icechunk/src/repository.rs
ianhi: should be fixed by https://github.com/earth-mover/icechunk/pull/1623/commits/786c2f56a4efe767a060a7c99cb7f57eeef54abd

icechunk/src/session.rs
ianhi: I did test this one off and it worked

icechunk/src/repository.rs
paraseba: So, in IC1 the "consistency point" is the write of the branch. So, we need to first write the snap and only when that finishes successfully we can write the branch.

icechunk/src/repository.rs
```diff
@@ -248,21 +249,42 @@ impl Repository {
...
+                    Snapshot::INITIAL_SNAPSHOT_ID,
+                    Arc::new(empty_tx_log),
+                );
+                try_join!(write_snap, write_tx)?;
+
+                // Only write the repo info after both succeed, since the repo
+                // object is the entry point that makes the repository valid.
+                // Writing it last ensures we never create a repo pointing to
+                // missing snapshot/tx data.
+                asset_manager_c.create_repo_info(Arc::clone(&repo_info)).await?;
```
paraseba, line 277: Great! thank you.

icechunk/src/session.rs
paraseba: Sorry, a one off is fine.

icechunk/src/repository.rs
ianhi: ah so for IC1 we just cant have concurrency here i guess?

ianhi:
@paraseba this is now hopefully race condition free

paraseba (approved):

paraseba:
It looks great, thank you!

PR #1625, opened 2026-02-05
Tweak build configs for Rust

li-em:
- Consolidate profile config in the workspace Cargo.toml
- Enabling optimizations for deps in dev profile (for tests)
- Create a new "ci" profile for running in GHA, with a small amount of optimization but targeting faster compile times.
- Configure tests to use `cargo nextest`, since it can run tests in parallel
- Add instructions to contributor docs on how to set up `cargo-nextest`, and also the `mold` linker (which is much faster than default, but only available in Linux)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,10 @@ sign-commit = true
 sign-tag = true
 push = false
 
+# Recommended configs for maximizing runtime speed
+# https://nnethercote.github.io/perf-book/build-configuration.html#maximizing-runtime-speed
+# This impacts compile time (since it becomes serial instead of parallel)
+# but can lead to 20-30% faster runtime
 [profile.release]
 lto = "thin"
 codegen-units = 1
@@ -31,13 +35,20 @@ opt-level = 1
 debug = "line-tables-only" # allow file:line info in backtraces
 
 # Enable higher optimizations for dependencies, but not our code
+# Setting it to "2" because gets most optimizations, without
+# the compile time impact of "3"
 [profile.dev.package."*"]
 opt-level = 2
 
+# The CI profile optimizes for reduced compile times, since they
+# dominate runtime in github actions, and cache between runs
+# doesn't always work well.
 [profile.ci]
 inherits = "dev"
 opt-level = 0
 debug = 0
 
+# Dependencies tend to cache better between runs,
+# enable a small bit of optimization in this case.
 [profile.ci.package."*"]
 opt-level = 1


.github/workflows/rust-ci.yaml
```diff
@@ -76,6 +76,9 @@ jobs:
         if: matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-24.04-arm'
         run: sudo snap install --edge --classic just
 
+      - name: Install cargo nextest
+        uses: taiki-e/install-action@nextest
```
paraseba, line 79: I've been using it forever, it's great.

paraseba: But you need to call it explicitly no? I use it as `cargo nextest run --workspace --no-fail-fast`

Cargo.toml
```diff
@@ -21,3 +21,11 @@ push = false
 [profile.perf]
 inherits = "release"
 debug = 1
+
+# Enable a small amount of optimization in debug mode
+[profile.dev]
+opt-level = 1
```
paraseba, line 34: Can you explain what is the advantage here and in the next change?. I want to make sure we are not imposing preferences on others, but this may be useful generally.

What I do in my `~/.cargo/config.toml` is the following, and it's quite helpful
 
```toml
[profile.dev]
debug = 0
strip = "debuginfo"
```

Justfile
```diff
@@ -3,18 +3,18 @@ alias pre := pre-commit
 
 # run all tests
 test *args='':
-  cargo test --all --all-targets {{args}}
+  cargo nextest run --all --all-targets {{args}}
```
paraseba: `--all` is deprecated, it's `--workspace` these days.

icechunk-python/Cargo.toml
```diff
@@ -49,7 +49,3 @@ default = ["cli"]
 
 [lints]
 workspace = true
-
```
paraseba, line 52: If only I remembered why I added this

.github/workflows/rust-ci.yaml
li-em: I changed the `Justfile` to call it instead of `cargo test`, as you noted below

Cargo.toml
li-em: Running `cargo clean && cargo nextest run --workspace --all-targets`:
The time for running tests is a bit misleading, there seems to be variability due to proptests.

```
[profile.dev]
opt-level = 1
[profile.dev.package."*"]
opt-level = 3
```
3m40s to compile and run tests, 7s for running tests


```
[profile.dev]
opt-level = 0
[profile.dev.package."*"]
opt-level = 3
```
3m04s to compile and run tests, 13s for running tests

```
[profile.dev]
opt-level = 0
```
2m15s to compile and run tests, 36s for running tests

---

`opt-level=3` for deps is overkill for CI (unless there is a previous cache), but it helps a lot locally since the deps are only compiled once and iterating in icechunk doesn´t trigger recompiles.

icechunk-python/Cargo.toml
dcherian: i added it for profiling with `samply`.

dcherian:
Shall we merge this?

Cargo.toml
```diff
@@ -17,7 +17,27 @@ sign-commit = true
 sign-tag = true
 push = false
 
+[profile.release]
+lto = "thin"
+codegen-units = 1
```
paraseba, line 26: @li-em can you add a code comment with the explanation? this being the release section we should understand what this does.

paraseba:
> Shall we merge this?

I only want to understand the rationale for the change in the release section of the cargo file

Cargo.toml
li-em: This comes from https://github.com/Kobzol/cargo-wizard and the rationale is described in https://nnethercote.github.io/perf-book/build-configuration.html#maximizing-runtime-speed

paraseba (approved):

PR #1626, opened 2026-02-06
Add plan for decoupling storage backends and enabling WASM support

mpiannucci:
Describes a 5-step plan to decouple built-in storage backends from core
types so the icechunk crate can compile for wasm32-wasip1-threads. The
key insight is that replacing the closed ObjectStoreConfig enum with a
typetag-serializable VirtualStoreConfig trait makes feature-gating
backends trivial and lets Rust users bring their own Storage implementations.

https://claude.ai/code/session_01NSNyLnvbaEnQ78tJUCSLEh

### DIFF
--- a/docs/plans/wasm-storage-decoupling.md
+++ b/docs/plans/wasm-storage-decoupling.md
@@ -6,6 +6,19 @@ Enable the core `icechunk` crate to compile for `wasm32-wasip1-threads` by decou
 built-in storage backends from core types. This also lets Rust users bring their own
 `Storage` implementations without pulling in heavy cloud SDK dependencies.
 
+## Delivery Priority
+
+This plan is executed in two phases:
+
+- **Phase 1 (first pass, current priority)**: achieve WASM compilation and clean module
+  boundaries, even if virtual chunks are unsupported on WASM.
+- **Phase 2 (follow-up)**: complete the full virtual-chunk serialization transition and
+  compatibility behavior.
+
+For Phase 1, virtual chunks can be explicitly disabled on WASM (compile-time gating and
+clear runtime errors for unsupported operations) while we keep the architecture aligned
+with the eventual trait/typetag model.
+
 ## Current Coupling
 
 The core crate unconditionally depends on the AWS SDK, cloud-specific `object_store`
@@ -129,6 +142,21 @@ This separation means:
 - No awkward `make_storage()` method that some impls don't need
 
 ## Steps
+
+### Phase 1 (WASM compile-first)
+
+1. Implement Step 1 (`StorageErrorKind` decoupling).
+2. Perform the structural reorganization needed to remove unconditional backend coupling
+   from core modules and make feature gating effective.
+3. Ensure WASM builds without cloud backends and with virtual chunks unsupported (by
+   design in this phase), returning explicit unsupported/backend-unavailable errors.
+4. Keep Python user-facing API unchanged.
+
+### Phase 2 (Virtual chunk transition follow-up)
+
+Complete the full `ObjectStoreConfig` enum→trait migration and serde compatibility
+strategy described below. This includes spec-version-driven serialization policy,
+legacy/native compatibility handling, and typetag-backed representations.
 
 ### Step 1: Genericize `StorageErrorKind`
 
@@ -488,6 +516,10 @@ reorganizations with no impact on public API behavior. Step 4 adds Cargo feature
 
 `ObjectStoreConfig` is serialized through three paths. Understanding these is critical
 for backward compatibility.
+
+Note: this section is the **Phase 2** target behavior. In **Phase 1**, WASM can ship
+with virtual chunks unsupported and therefore does not need to implement this migration
+yet.
 
 ### Path 1: `config.yaml` (YAML) — persistent, per-repository
 


docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+S3UploadPartError(..),
+S3CompleteMultipartUploadError(..),
+S3CopyObjectError(..),
+S3HeadObjectError(..),
+S3ListObjectError(..),
+S3DeleteObjectError(..),
+S3StreamError(..),
+
+// AFTER
+S3Error(String),
```
paraseba: Weird, why not `S3Error(Box<dyn std::error::Error>)`?

We need to check if we are not capturing the specific errors anywhere. Easy to try.

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+that returns an error. Backends that support virtual chunks override it.
+
+```rust
+#[async_trait]
+#[typetag::serde(tag = "object_store_type")]
+pub trait ObjectStoreConfig: fmt::Debug + Send + Sync {
+    /// Create a Storage backend from this configuration.
+    async fn make_storage(
+        &self,
+        credentials: Option<&Credentials>,
```
paraseba, line 204: Then there is a step needed before this one to create a typetagged `Credentials` type?

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+    ) -> StorageResult<Arc<dyn Storage>>;
+
+    /// Validate that the provided credentials match this store type.
+    fn validate_credentials(&self, cred: Option<&Credentials>) -> Result<(), String>;
+
+    /// Validate that this config is appropriate for the given URL scheme.
+    fn validate_url(&self, url: &Url) -> Result<(), String>;
```
paraseba, line 213: This I don't like. It's too much responsibility for this type. But I guess we can find a way to move it somewhere more concrete

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Whether the fetcher cache key needs a bucket/host component.
+    fn is_bucket_constrained(&self) -> bool { false }
```
paraseba: we need a better name for this one, and probably a better design?

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Construct a ChunkFetcher for virtual chunk access.
+    /// Default returns an error; backends that support virtual chunks override this.
+    async fn mk_fetcher(
```
paraseba, line 231: too much responsibility. Probably this should be more similar to:
```rust
    async fn fetch_part(
        &self,
        chunk_location: &Url,
        range: Range<ChunkOffset>,
        checksum: Option<&Checksum>,
    ) -> Result<Box<dyn Buf + Unpin + Send>, VirtualReferenceError>;
```

Ideally removing the url from there, constraining to a prefix, changing the error to not talk about virtual

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+        _credentials: Option<&Credentials>,
+        _settings: &storage::Settings,
+        _chunk_location: &Url,
+    ) -> Result<Arc<dyn ChunkFetcher>, VirtualReferenceError> {
+        Err(VirtualReferenceErrorKind::VirtualChunksNotSupported.into())
+    }
+}
+```
+
+Each backend implements this trait. Backends that support virtual chunks override
```
paraseba: In general, I don't want backends to know about virtual chunks if possible

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+```rust
+async fn mk_fetcher_for(&self, cont: &VirtualChunkContainer, chunk_location: &Url)
+    -> Result<Arc<dyn ChunkFetcher>, VirtualReferenceError>
+{
+    let creds = self.credentials.get(&cont.url_prefix).and_then(|c| c.as_ref());
+    cont.store.mk_fetcher(creds, &self.settings, chunk_location).await
+}
+```
+
+**The factory functions in `storage/mod.rs`** (`new_s3_storage`, `new_gcs_storage`, etc.)
```
paraseba: This I don't get. Somebody needs to know about the concrete implementations to be able to construct the  right one? Or are we going to go one by one asking "are you good for this url?". In that case, how do we register them?

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+remain as convenience constructors but now live in their backend modules. They can also
+be expressed as thin wrappers around `ObjectStoreConfig::make_storage()`.
+
+**`RedirectStorage::mk_storage()`** constructs backend-specific config structs and calls
+`make_storage()` instead of calling factory functions directly.
+
+**`Credentials` enum**: Stays as-is for now. It is a small enum at the `Repository` API
+boundary with no heavy dependencies. Each `ObjectStoreConfig` impl matches on the
+variant it expects. This can be further decoupled later.
```
paraseba: Doesn't this break the magic?

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+
+**Serde migration**: The old `ObjectStoreConfig` used serde's built-in enum tagging
+(`#[serde(rename_all = "snake_case")]`). The new approach uses typetag's tag field
+(`"object_store_type": "s3"`). A migration adapter can deserialize the old format and
```
paraseba: ObjectStoreConfig goes into serialized yaml files for config. We need to be extremely careful not to break existing repos.

paraseba: Probably we also need to think about people accessing the same repo with pre-this-change and post-this-change versions of Icechunk, so the config can probably not be serialized in a new format?

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+
+After this step `config.rs` retains only backend-agnostic types: `RepositoryConfig`,
+`CachingConfig`, `CompressionConfig`, `ManifestConfig`, `Credentials`.
+
+Each backend module becomes self-contained: its `ObjectStoreConfig` impl, its
+`Storage` impl, its config types, and its fetcher.
+
+### Step 4: Feature-gate the backends
+
+Now trivial because each backend is an island.
```
paraseba, line 456: Oh claude you always say the right things...

docs/plans/wasm-storage-decoupling.md
```diff
@@ -0,0 +1,339 @@
...
+| `lib.rs` | | | Update exports | |
+| `Cargo.toml` | | | | Add features, make deps optional |
+
+Each step is independently shippable and testable. Steps 1-3 are internal
+reorganizations with no impact on public API behavior. Step 4 adds Cargo features with
+`default` enabling all, so existing users are unaffected.
+
+## Notes
+
+- `typetag` supports wasm via the `inventory` crate's `__wasm_call_ctors` mechanism.
```
paraseba: we should test this before paying more costs

docs/plans/wasm-storage-decoupling.md
mpiannucci: Agree with this. Will think about this. This is the crux of most of your review

docs/plans/wasm-storage-decoupling.md
mpiannucci: I think it does, this smells, but it doesnt break wasm because there is not actual dependent types in them, its all our custom defined types.

docs/plans/wasm-storage-decoupling.md
mpiannucci: Yes i agree. The goal for this is to: 
* Not break API compat in python
* Not break ic1 serialization of storage. This is tough

docs/plans/wasm-storage-decoupling.md
mpiannucci: I researched this actually and fed this to it. Its documented in the crate: https://docs.rs/typetag/latest/typetag/#so-many-questions

mpiannucci: tested it and it works with wasm

mpiannucci:
Okay i got myself twisted here: 

* Using icechunk from rust, you manually call the storage factory functions right now, there is no magic from object store config to storage backend
* Virtual chunks require the object storage config 

With this, the storage is already decoupled in the rust layer which is what we need except for some leakage in the vcirtual chunks implementation and some error types. So we really just need to abstract away the usage of the config and the storage for the virtual use case

mpiannucci:
The biggest issue here is virtual chunk config being persisted in config.yaml and perserving compatability for reading with old icechunk versions.

mpiannucci:
Current thinking: Allow deserializing from non type tagged version of the vonfig always. When writing, only allow writing the config (Updating it) with the new version, if using a v1 repo we reject updating the config?

mpiannucci:
Basically: one way migration is easy for this, so its a social problem of how we fit that in with everything.

docs/plans/wasm-storage-decoupling.md
mpiannucci: yes same as now, that is treated as either binding or user space thing to choose

PR #1628, opened 2026-02-07
`wasm` build of icechunk core

mpiannucci:


.github/workflows/rust-ci.yaml
```diff
@@ -146,3 +146,61 @@ jobs:
...
+          CC_wasm32_wasi: clang
+          CXX_wasm32_wasi: clang++
+          AR_wasm32_wasi: llvm-ar
+          WASI_SYSROOT: /usr
+          CFLAGS_wasm32_wasip1_threads: --sysroot=/usr -isystem /usr/include/wasm32-wasi
+          CXXFLAGS_wasm32_wasip1_threads: --sysroot=/usr -isystem /usr/include/wasm32-wasi
+          CFLAGS_wasm32_wasi: --sysroot=/usr -isystem /usr/include/wasm32-wasi
+          CXXFLAGS_wasm32_wasi: --sysroot=/usr -isystem /usr/include/wasm32-wasi
+        run: |
+          cargo check -p icechunk --no-default-features --target wasm32-wasip1-threads
```
li-em: I tried out
```suggestion
          cargo build -p icechunk --no-default-features --target wasm32-wasip1-threads
```
and the build time is not much different, and would also exercise the linker. Maybe a bit more futureproof than just `check`?

icechunk/Cargo.toml
```diff
@@ -99,6 +106,20 @@ warp = { version = "0.4.2", features = ["server"] }
...
+default = ["s3", "gcs", "azure", "http-store", "redirect", "local-store"]
+s3 = [
+  "dep:aws-sdk-s3",
+  "dep:aws-config",
+  "dep:aws-credential-types",
+  "dep:aws-smithy-runtime",
+  "dep:aws-smithy-types-convert",
+  "object_store/aws",
+]
+gcs = ["object_store/gcp"]
```
paraseba: but aren't all these + aws already included with object_store? we are not making the optional I think

icechunk/src/storage/mod.rs
```diff
@@ -77,29 +75,25 @@ pub enum StorageErrorKind {
     #[error("bad object store prefix {0:?}")]
     BadPrefix(OsString),
     #[error("error getting object from object store {0}")]
-    S3GetObjectError(#[from] Box<SdkError<GetObjectError, HttpResponse>>),
+    S3GetObjectError(#[source] Box<dyn std::error::Error + Send + Sync>),
```
paraseba, line 81: At this point, is it even worth it having all these different cases? Or a generic `S3Error`? Are we ever catching any of these in a `match`?

paraseba (commented):
@mpiannucci here is my request before merging this: convince me that we we can still open a repository that has configuration stored on-disk, in all situations, for example with the configuration including virtual chunk containers. I think test-can-read-old kind of covers this, but I'm not sure if we have enough coverage off all cases.

I guess we are also saying, to open a repo that has virtual chunks to X, you need to include feature X. It would be better if we could ignore the extra config and open the repo, and only fail if a virtual chunk is fetched. But it may be hard to do that. We should think about all the edge cases.

paraseba:
> @mpiannucci here is my request before merging this: convince me that we we can still open a repository that has configuration stored on-disk, in all situations, for example with the configuration including virtual chunk containers. I think test-can-read-old kind of covers this, but I'm not sure if we have enough coverage off all cases.
> 
> I guess we are also saying, to open a repo that has virtual chunks to X, you need to include feature X. It would be better if we could ignore the extra config and open the repo, and only fail if a virtual chunk is fetched. But it may be hard to do that. We should think about all the edge cases.

maybe we can create some new tests to guarantee this

icechunk/Cargo.toml
mpiannucci: no, because we specify no default features from object store. Ill verify tho

icechunk/src/storage/mod.rs
mpiannucci: The aws errors dont always print out pretty but this at least gives easy matches to switch on for the user

mpiannucci:
Okay so essentially we need tests that show basically this flow:

* User installs icechunk with local filestore enabled but no other cloud systems
* User opens a local repo that has virtual chunks in s3, can open it
* User tries to read virtual chunks from s3 and we throw an appropriate error

Is this the gist of what you are looking for? It might be s3/http etc instead of local filesystem.

icechunk/src/config.rs
```diff
@@ -773,4 +662,103 @@ mod tests {
         );
         assert!(!vccs.contains_key("s3://bucket1/"), "Should not contain bucket1");
     }
+
+    #[icechunk_macros::test]
+    fn test_yaml_config_deserialization() {
```
mpiannucci, line 737: @paraseba lert me know what you think of this test!

paraseba: I like it. What is the trick for this to test? Running it without the default features? How do we achieve that?

mpiannucci: You can use the --features flag when running the test with cargo to specify feature combinations. I ran with none and only s3 and a few other combos locally and it worked. 

I'm not sure if we should single this out in ci, by default it will test with all features

mpiannucci: If desired, we could add it to the wasm ci to test it with no features enabled so we have coverage for all features and none

paraseba: love it, yes!

mpiannucci:
I feature gated the tests that use storage features

PR #1637, opened 2026-02-11
Update contributing.md with duplicate mention of minio docker setup

jbusecke:
I missed the minio docker compose instructions because I clicked right into the Rust workflow. I think they are slim enough to copy, but can also move them to a separate section and reference them in Python and Rust testing sections if that

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -269,6 +269,10 @@ just compile-tests
 
 #### Testing
 
+The full Python test suite depends on S3 and Azure compatible object stores.
+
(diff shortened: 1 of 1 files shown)


TomNicholas:
Let's make it a separate section and reference it. It would be nice to include a little more detail on what is actually involved to setup and use docker too.

jbusecke:
And how to debug issues (which I seem to have more of haha). Will work on this once I got the testing running.

jbusecke:
Closed via #1640

PR #1638, opened 2026-02-12
Optimize fetch of repo object

### LINKED ISSUE #1375: Optimize fetch of `repo` object, downloading it only if not modified
Many times we fetch the repo object but we have the value of the previous etag. In some other places we could store this value to have it available. With this, we can instruct the S3 library to fetch the object only if it wasn't modified.

Related, can `update_repo_info` be made more efficient by doing a first attempt without fetching the object?

li-em:
Fix #1375

- Add a repo info cache to `AssetManager`, so we can avoid re-fetching it if the etag didn't change.
- Add `get_object_conditional` method to `Storage` trait, and move most of the implementation of `get_object_range` to a more general (and private) `get_object_range_conditional` outside the trait, to avoid implementing most of the same logic in two places.
- `get_object_conditional` sets `if-none-match` header with etag, and returns a `GetModifiedResult` enum with variants for new data and version, or just a sentinel value saying it is on latest version.

### DIFF
--- a/icechunk/src/storage/s3.rs
+++ b/icechunk/src/storage/s3.rs
@@ -909,32 +909,34 @@ impl S3Storage {
                 };
 
                 match sdk_err.as_service_error() {
-                Some(e) if e.is_no_such_key() => {
-                    Err(StorageErrorKind::ObjectNotFound.into())
+                    Some(e) if e.is_no_such_key() => {
+                        Err(StorageErrorKind::ObjectNotFound.into())
+                    }
+                    Some(_)
+                        if sdk_err
+                            .raw_response()
+                            .is_some_and(|x| x.status().as_u16() == 404) =>
+                    {
+                        // needed for Cloudflare R2 public bucket URLs
+                        // if object doesn't exist we get a 404 that isn't parsed by the AWS SDK
+                        // into anything useful. So we need to parse the raw response, and match
+                        // the status code.
+                        Err(StorageErrorKind::ObjectNotFound.into())
+                    }
+                    Some(_)
+                        // aws_sdk_s3 doesn't return an error when
+                        // status 304 (Not Modified) happens, so
+                        // check the http status code here and
+                        // return None to make it easy to catch
+                        // downstream
+                        if sdk_err
+                            .raw_response()
+                            .is_some_and(|x| x.status().as_u16() == 304) =>
+                    {
+                        Ok(None)
+                    }
+                    _ => Err(s3_get_err(sdk_err)),
                 }
-                Some(_)
-                    // aws_sdk_s3 doesn't return an error when status 304 (Not Modified)
-                    // happens, so check the http status code here and return None
-                    // catch it easily downstream
-                    if sdk_err
-                        .raw_response()
-                        .is_some_and(|x| x.status().as_u16() == 304) =>
-                {
-                    Ok(None)
-                }
-                Some(_)
-                    if sdk_err
-                        .raw_response()
-                        .is_some_and(|x| x.status().as_u16() == 404) =>
-                {
-                    // needed for Cloudflare R2 public bucket URLs
-                    // if object doesn't exist we get a 404 that isn't parsed by the AWS SDK
-                    // into anything useful. So we need to parse the raw response, and match
-                    // the status code.
-                    Err(StorageErrorKind::ObjectNotFound.into())
-                }
-                _ => Err(s3_get_err(sdk_err)),
-            }
             }
         }
     }
--- a/icechunk/src/storage/object_store.rs
+++ b/icechunk/src/storage/object_store.rs
@@ -3,10 +3,9 @@
 //! Supports local filesystem, in-memory, Azure Blob, and Google Cloud Storage.
 
 use crate::private;
-use crate::storage::{
-    s3_config::{S3Credentials, S3Options},
-    strip_quotes,
-};
+#[cfg(feature = "object-store-s3")]
+use crate::storage::s3_config::{S3Credentials, S3Options};
+use crate::storage::strip_quotes;
 use async_trait::async_trait;
 use bytes::Bytes;
 use chrono::{DateTime, TimeDelta, Utc};


icechunk/src/storage/mod.rs
```diff
@@ -502,6 +502,12 @@ pub trait Storage: fmt::Debug + fmt::Display + private::Sealed + Sync + Send {
         settings: &Settings,
     ) -> StorageResult<DateTime<Utc>>;
 
+    async fn get_object_etag(
+        &self,
+        path: &str,
+        settings: &Settings,
+    ) -> StorageResult<Option<ETag>>;
```
paraseba: I wonder: when the person open a repo, we'll have to fetch the repo object anyway, would it be better to change the get_object style methods to also get the etag? Because I believe that would be "free"

icechunk/src/asset_manager.rs
```diff
@@ -478,7 +485,29 @@ impl AssetManager {
         &self,
     ) -> RepositoryResult<(Arc<RepoInfo>, VersionInfo)> {
         self.fail_unless_spec_at_least(SpecVersionBin::V2dot0)?;
-        fetch_repo_info(self.storage.as_ref(), &self.storage_settings).await
+
+        // check etag from version_info here
+        let remote_info =
```
paraseba: the problem is this has two roundtrips to object store in the worst case, while still being a full one in the common case.

I think a better approach here is to always call `get_object` on the object, but only fetch it if modified: See `If-None-Match` [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax)

That gives you single roundtrip in the worst case and somewhat fresher objects.

icechunk/src/asset_manager.rs
```diff
@@ -19,7 +19,7 @@ use std::{
 };
 use tokio::{
     io::{AsyncBufRead, AsyncReadExt},
-    sync::Semaphore,
+    sync::{RwLock, Semaphore},
```
paraseba: I think I really like the idea of putting this "cache" in the asset manager! It keeps everything else unaware of this optimization.

icechunk/src/storage/mod.rs
paraseba: See my comment later in asset_manager, it will make this more clear.

icechunk/src/asset_manager.rs
```diff
@@ -82,6 +82,11 @@ pub struct AssetManager {
 
     #[serde(skip)]
     request_semaphore: Semaphore,
+
+    #[serde(skip)]
+    repo_version: Arc<RwLock<Option<VersionInfo>>>,
+    #[serde(skip)]
+    current_repo: Arc<RwLock<Option<Arc<RepoInfo>>>>,
```
paraseba: Is it better to have a single lock holding both things?
```rust
repo_cache: Arc<RwLock<Option<(VersionInfo, Arc<RepoInfo>)>>>

Comment (paraseba) [reply icechunk/src/asset_manager.rs]: I wonder if this can be a std::sync::RwLock instead. Because I think you can read the current value, unlock, fetch the object, and then re-lock if needed.

File: icechunk/src/storage/mod.rs [hunk 5]
```diff
@@ -502,6 +507,13 @@ pub trait Storage: fmt::Debug + fmt::Display + private::Sealed + Sync + Send {
         settings: &Settings,
     ) -> StorageResult<DateTime<Utc>>;
 
+    async fn get_object_conditional(
+        &self,
+        path: &str,
+        settings: &Settings,
+        previous_version: Option<VersionInfo>,
```
paraseba: Question, how about adding an `Option<VersionInfo>` argument to `get_object`? That would save us yet one more method on this trait, but, does it make the code much more complicated? If it doesn't I think I would prefer it.

icechunk/src/storage/s3.rs
```diff
@@ -777,6 +779,34 @@ impl Storage for S3Storage {
...
+            .await
+        {
+            Ok(Some((stream, new_version))) => {
+                let reader = StreamReader::new(stream.map_err(std::io::Error::other));
+                Ok(GetModifiedResult::Modified { data: Box::pin(reader), new_version })
+            }
+            Ok(None) => Ok(GetModifiedResult::OnLatestVersion),
+            Err(sdk_err) => match sdk_err.kind {
+                StorageErrorKind::S3GetObjectError(e)
+                    if e.raw_response().is_some_and(|x| x.status().as_u16() == 304) =>
```
paraseba: Are some object stores showing this behavior? Can you add a comment to the code explaining why we need to do this?

icechunk/src/asset_manager.rs
```diff
@@ -478,7 +484,32 @@ impl AssetManager {
         &self,
     ) -> RepositoryResult<(Arc<RepoInfo>, VersionInfo)> {
         self.fail_unless_spec_at_least(SpecVersionBin::V2dot0)?;
-        fetch_repo_info(self.storage.as_ref(), &self.storage_settings).await
+
+        #[allow(clippy::expect_used)]
+        let repo_cache = self.repo_cache.read().expect("Poison lock").clone();
```
paraseba: We have a few of this type of errors in `RepositoryError`. Let's add an enum case for `PoisonError` and avoid the `expect` call.

paraseba: This `clone` here looks a bit funny, do we really need it?

paraseba: Ah I see, you are cloning so you can release the guard. Please add a code comment that says something like `lock is released immediately`

icechunk/src/asset_manager.rs
```diff
@@ -1244,29 +1303,37 @@ async fn fetch_repo_info_backup(
...
+        repo.expect("Logic bug, must have a repo_info here")
+    })
 }
 
 pub async fn fetch_repo_info_from_path(
     storage: &(dyn Storage + Send + Sync),
     storage_settings: &storage::Settings,
     path: &str,
-) -> RepositoryResult<(Arc<RepoInfo>, VersionInfo)> {
+    previous_version: Option<VersionInfo>,
```
paraseba: Can we take by ref here so we don't need to clone the VersionInfo object?

icechunk/src/asset_manager.rs
```diff
@@ -478,7 +484,32 @@ impl AssetManager {
...
+            Ok(Some((repo_info, version_info))) => {
+                #[allow(clippy::expect_used)]
+                let mut repo_cache = self.repo_cache.write().expect("Poison lock");
+                *repo_cache = Some((repo_info.clone(), version_info.clone()));
+
+                return Ok((repo_info, version_info));
+            }
+            Ok(None) => {
+                #[allow(clippy::expect_used)]
+                return Ok(repo_cache.expect(
```
paraseba, line 498: I'd love to see this enforced by the type system. Is there a restructuring of the code that would allow us to avoid this `expect`? Don't put more than 10 min of thought into it.

icechunk/src/asset_manager.rs
```diff
@@ -516,7 +547,15 @@ impl AssetManager {
             &self.storage_settings,
             None,
         )
-        .await
+        .await?;
+
+        #[allow(clippy::expect_used)]
```
paraseba: Shouldn't `write_repo_info` take care of updating the cache?

icechunk/src/storage/mod.rs
li-em: I didn't want to change `get_object` because it is a public method, but since IC2 is still alpha and not released that should be fine? It will brake any users of the Rust crate.

I do agree that limiting the number of methods in the trait is better

icechunk/src/storage/s3.rs
li-em: `object_store` returns a proper `Error` for not modified, but `aws_sdk_s3` only returns the status code 304. Certainly a comment would be useful :sweat_smile:

icechunk/src/storage/mod.rs
paraseba: Oh this is a great comment @li-em . I definitely never gave you this context, but the reality is that we still don't recommend usage of the Rust crate, and we don't promise semantic versioning. We consider `Storage` almost as a private thing. So feel free to change it as needed.

icechunk/src/asset_manager.rs
li-em: But then I would need to pass the cache along to `write_repo_info`, since it doesn't have access to it (it is a function, not a method in `AssetManager`, where the repo cache is

paraseba: Maybe we need another member helper function?

icechunk/src/asset_manager.rs
```diff
@@ -488,22 +521,38 @@ impl AssetManager {
             .await
     }
 
-    #[instrument(skip(self, info))]
-    pub async fn create_repo_info(
+    pub async fn write_repo_info(
```
li-em, line 526: @paraseba I didn't expose all the parameters in the `write_repo_info` function into the method, because we can pick them from `AssetManager`. Only added the three that changed in the current invocations (`info`, `version`, and `backup_path`)

icechunk/src/asset_manager.rs
li-em: Added a `write_repo_info`  method with just the arguments we can´t pick from the asset manager

icechunk/src/repository.rs
```diff
@@ -144,6 +144,8 @@ pub enum RepositoryErrorKind {
         "repository version error, this operation requires a repository that is at least version {minimum_spec_version}, please upgrade your on-disk format before executing this operation"
     )]
     BadRepoVersion { minimum_spec_version: SpecVersionBin },
+    #[error("Poisoned lock")]
```
paraseba: let's make the error string more understandable to users, maybe "concurrency error, lock cannot be acquired" or something

PR #1639, opened 2026-02-12
Support async refreshable credential callbacks in Python bindings

### LINKED ISSUE #1424: Allow async credential refresh functions
### Problem

Currently icechunk storages only allow non-async functions to be passed in as credential refresh functions.
It would be nice to allow async functions.

### Proposed Solution

If an async function is passed in, icechunk should run the credential refresh function on an async event loop.

In a web backend, i'd like you to run the refresh function on the "main" loop, as that's likely what my database access is tied to.

I've written some examples of cases that should be handled, there's probably more ways users be weird.

```python
import asyncio
import icechunk


async def refresh():
    return {"creds": "creds"}


async def get_repo_async():
    storage = icechunk.s3_storage(get_credentials=refresh)
    repo = await icechunk.Repository.open_async(storage=storage)
    return repo


def get_repo():
    storage = icechunk.s3_storage(get_credentials=refresh)
    repo = icechunk.Repository.open(storage=storage)
    return repo


# Nice example where "binding" the repo to the event loop it was created on
# would make sense, and running the refresh function on that same loop would work
async def normal():
    repo = await get_repo_async()
    await repo.list_branches_async()  # <- refresh should be called on the same loop as the repo was created on?
    repo.list_branches()  # <- trying to put refresh() on the same loop would deadlock things i think?


# Less nice where the repo is put on different loops
def new_loop():
    repo = asyncio.run(get_repo_async())  # create repo on loop 1
    asyncio.run(repo.list_branches_async())  # <- what to do here? explode?


# Creating a repo sync but then trying to use it on a running loop
def sync_repo_async_usage():
    # create repo sync, should it create its own loop for refresh to run on?
    # Should it reject async refresh when created?
    repo = get_repo()
    asyncio.run(repo.list_branches_async())  # <- what to do here? explode?

```

mpiannucci:
Adds support for async refreshable credential callbacks in the Python bindings, allowing users to provide async def functions for credential refresh that work correctly across sync and async contexts via pyo3-async-runtimes' TaskLocals bridge.

Closes #1424

### DIFF
--- a/icechunk-python/src/sync.rs
+++ b/icechunk-python/src/sync.rs
@@ -27,7 +27,7 @@ pub(crate) fn would_deadlock_current_loop(
 pub(crate) fn ensure_not_running_event_loop(py: Python<'_>) -> PyResult<()> {
     match pyo3_async_runtimes::get_running_loop(py) {
         Ok(_) => Err(PyValueError::new_err(
-            "deadlock: synchronous API called from a running event loop thread; use the async API or run the sync call in a worker thread",
+            "synchronous API called from a running event loop thread (deadlock risk); this usage is disallowed. Use the async API or run the synchronous call in a worker thread",
         )),
         Err(err) if err.is_instance_of::<PyRuntimeError>(py) => Ok(()),
         Err(err) => Err(err),
--- a/icechunk-python/tests/conftest.py
+++ b/icechunk-python/tests/conftest.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import Literal, cast
 
 import boto3
@@ -20,6 +21,12 @@ def parse_repo(
             storage=in_memory_storage(),
             spec_version=spec_version,
         )
+
+
+async def parse_repo_async(
+    store: Literal["local", "memory"], path: str, spec_version: int | None
+) -> Repository:
+    return await asyncio.to_thread(parse_repo, store, path, spec_version)
 
 
 @pytest.fixture(scope="function")
--- a/icechunk-python/tests/test_can_read_old.py
+++ b/icechunk-python/tests/test_can_read_old.py
@@ -11,6 +11,7 @@ intentionally changed, the repository files must be regenerated. For that, run t
 file as a python script: `python ./tests/test_can_read_old.py`.
 """
 
+import asyncio
 import shutil
 from datetime import UTC, datetime
 from typing import Any, cast
@@ -86,8 +87,10 @@ async def write_a_split_repo(path: str) -> None:
     )
 
     print(f"Writing repository to {store_path}")
-    repo = mk_repo(create=True, store_path=store_path, config=config)
-    session = repo.writable_session("main")
+    repo = await asyncio.to_thread(
+        mk_repo, create=True, store_path=store_path, config=config
+    )
+    session = await repo.writable_session_async("main")
     store = session.store
 
     root = zarr.group(store=store)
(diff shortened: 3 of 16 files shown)


icechunk-python/python/icechunk/repository.py
```diff
@@ -21,6 +22,24 @@
 from icechunk.store import IcechunkStore
 
 
+def _raise_if_running_loop(
```
mpiannucci: The idea here being that you should call sync icechunk methods from an async context because it could result in a deadlock

icechunk-python/python/icechunk/credentials.py
```diff
@@ -46,16 +48,50 @@
...
+            async def _as_coroutine() -> CredentialType:
+                return await awaitable
+
+            coroutine = _as_coroutine()
+
+        try:
+            return asyncio.run(coroutine)
+        except RuntimeError as err:
+            coroutine.close()
+            if "asyncio.run() cannot be called from a running event loop" in str(err):
```
paraseba, line 75: Interesting.... this is a bit of a breaking change, because everybody (including Earthmover) uses True for scatter

paraseba (commented):
This is a miracle of engineering hehe, I understand little of it, but that's OK. I hope we can make Samantha's tests pass and merge.

icechunk-python/python/icechunk/credentials.py
mpiannucci: I'll think on this..

mpiannucci:
This is working well now, but it required adding throwing an exception when calling synbc api from async context

I think we need to change this to warn that this is can lead to deadlocks and that in a future version of icechunk this behavior may be disallowed. 

A good example of the consequence of this is the xarray test failure

mpiannucci:
@dcherian i dont understand the xarray test suite, do you ahve any context on this?

```
________________ TestIcechunkStoreFilesystem.test_load_async[3] ________________

self = <tests.run_xarray_backends_tests.TestIcechunkStoreFilesystem object at 0x7ff67a78a050>

    @pytest.mark.asyncio
    @pytest.mark.skipif(
        not has_zarr_v3,
        reason="zarr-python <3 did not support async loading",
    )
    async def test_load_async(self) -> None:
>       await super().test_load_async()

.venv/lib/python3.11/site-packages/xarray/tests/test_backends.py:2659: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.11/site-packages/xarray/tests/test_backends.py:595: in test_load_async
    with assert_loads() as ds:
../../../../_temp/uv-python-dir/cpython-3.11.14-linux-x86_64-gnu/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/xarray/tests/test_backends.py:583: in assert_loads
    with self.roundtrip(expected) as actual:
../../../../_temp/uv-python-dir/cpython-3.11.14-linux-x86_64-gnu/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/xarray/tests/test_backends.py:2648: in roundtrip
    with self.create_zarr_target() as store_target:
../../../../_temp/uv-python-dir/cpython-3.11.14-linux-x86_64-gnu/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
tests/run_xarray_backends_tests.py:51: in create_zarr_target
    with self.create_repo() as repo:
../../../../_temp/uv-python-dir/cpython-3.11.14-linux-x86_64-gnu/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
tests/run_xarray_backends_tests.py:102: in create_repo
    yield Repository.create(local_filesystem_storage(tmpdir))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #1640, opened 2026-02-12
Update contributor docs

jbusecke:
I yesterday tried to get a development environment working and noticed a bunch of things that weren't quite working, so this PR is trying to bring the docs to the newest state of the development workflow

- More detailed explanation of the docker storage container setup + more prominent mention in Python **and** Rust Section
- Moved python docs to `icechunk-python` folder (sorry this is causing quite a large diff) + Moved contributing material about docs below Python/Rust testing.
- Set default `-n auto` for parallel python tests by default
- Stolen `--livereload` temporary bugfix from https://github.com/earth-mover/icechunk/pull/1621/changes
- Several smaller fixes

## TODO:
- [x] Figure out the issue with `strict`
- [x] Add link to @ianhi 's docs-output-filter
- [x] Document the ulimit adjustment that was needed on my machine
- [x] Add instructions to test the docker container via GUI? 

I noticed two consistently failing python tests which will be adressed in separate issues/PRs:
- `tests/test_zarr/test_store/test_icechunk_store.py::TestIcechunkStore::test_get_many - pyo3_async_runtimes.RustPanic: rust future panicked: unknown error` -> #1642 #1643 
- `tests/test_zarr/test_stateful.py::test_zarr_hierarchy - hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.` -> #1641

PR #1641, opened 2026-02-12
Prevent hypothesis from generating invalid zarr keys

jbusecke:
Claude fix for the issues encountered in #1640.

I frankly have no idea if this makes sense, but it does fix the issue on my machine and passes Ci. 

Maybe @TomNicholas or @ianhi can take a look and let me know if this is ok?


ianhi:
closing because I missed this and then accidentally pushed a fix directly to main. sorry about htat!

https://github.com/earth-mover/icechunk/commit/3a9e0c910e40498afc5d770f791af2f3483ace94

PR #1645, opened 2026-02-13
Docs: Update VirtualChunkContainer examples

tylanderson:
Updates doc examples for `VirtualChunkContainer` so `url_prefix` ends with `/` and argument to store are provided as `opts` not options.

Also adds a note about `url_prefix` trailing slash at the top of the Virtual Reference API docs.


ianhi:
thanks @tylanderson!

the docs failure is due to use swapping where our docs live to make local builds a bit easier. If you can rebase this onto main then the change is just move the changes to `icechunk-python/docs/docs.virtual.md`

If you prefer I can also do the rebase and push. whatever works best for you.

paraseba:
Thanks @tylanderson  !!

PR #1647, opened 2026-02-13
Bump pre-commit ruff version + define options in pyproject.toml

### LINKED ISSUE #1644: Contributor Guide instructions for linting do not work
### What happened?

Follow up on #1640 

When I follow the instructions and run `uv run ruff check python` I get the following error. 

```
(icechunk) juliusbusecke@Juliuss-MacBook-Air icechunk-python % uv run ruff check python
UP043 [*] Unnecessary default type arguments
    --> python/icechunk/_icechunk_python.pyi:2046:30
     |
2044 |     async def getsize_prefix(self, prefix: str) -> int: ...
2045 |
2046 | class PyAsyncStringGenerator(AsyncGenerator[str, None], metaclass=abc.ABCMeta):
     |                              ^^^^^^^^^^^^^^^^^^^^^^^^^
2047 |     def __aiter__(self) -> PyAsyncStringGenerator: ...
2048 |     async def __anext__(self) -> str: ...
     |
help: Remove default type arguments

UP043 [*] Unnecessary default type arguments
    --> python/icechunk/_icechunk_python.pyi:2085:32
     |
2083 |         ...
2084 |
2085 | class PyAsyncSnapshotGenerator(AsyncGenerator[SnapshotInfo, None], metaclass=abc.ABCMeta):
     |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2086 |     def __aiter__(self) -> PyAsyncSnapshotGenerator: ...
2087 |     async def __anext__(self) -> SnapshotInfo: ...
     |
help: Remove default type arguments

Found 2 errors.
[*] 2 fixable with the `--fix` option.
```

Should we adjust the linting flags here?

jbusecke:
Closes #1644

I encountered some inconsistent behavior between running the linting locally and with pre-commit. This simply bumps the pre-commit version and pins the ruff version in `pyproject.toml` to the same version. I also set the `--fix` and `--show-fixes` options on the `pyproject.toml` rather than in pre-commit to have one central place of control.

Some linting output has changed, please let me know if we should stick with the older version for consistency.


icechunk-python/pyproject.toml
```diff
@@ -139,6 +139,8 @@ warn_unreachable = true
 enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 
 [tool.ruff]
+fix = true
+show-fixes = true
```
paraseba, line 143: Is this standard in other projects? Isn't it weird having tools changing your code without you explicitly asking? I'm OK with if this is common.

paraseba (approved):

icechunk-python/pyproject.toml
jbusecke: Hmm that is a good question actually. My goal here was to align pre-commit and manually running ruff, but after thinking about it, maybe this is a bad idea.

jbusecke: But FWIW pre-commit was already doing this!

jbusecke:
@paraseba I think that this issue will be solved by #1674 and then Ill refactor this to a pure version bump for ruff!

jbusecke:
Closing this as stale

PR #1649, opened 2026-02-13
Fix Python benchmarks

### LINKED ISSUE #1399: Benchmark and test IC2 performance under different workloads

dcherian:
Closes #1399


dcherian:
Results in #1399 


This stuff is quite a mess; and should be rethought. But it's good for gaining confidence that we haven't regressed w.r.t IC1

PR #1653, opened 2026-02-17
Retry MinimumThroughput and other StreamingErrors

### LINKED ISSUE #1569: Tweak stalled stream strategy
Currently we have a pretty long timeout (60s), and no retries.

We should experiment with shorter timeouts and aggressive retries.

dcherian:
Closes #1569

I am opening this as an RFC.

My conclusions:
- The diagnosis in https://github.com/earth-mover/icechunk/issues/1569#issuecomment-3909797248 is correct. The error is thrown while streaming data, and _outside_ the AWS SDK's retry loop.
- I attempted to write a (complicated) unit test but failed.  Even upstream AWS tests start up a ['faulty download server'](https://docs.rs/crate/aws-sdk-s3/latest/source/tests/stalled-stream-protection.rs#248)
- The approach here, cooked up by Claude, is to use [toxiproxy](https://github.com/Shopify/toxiproxy) to simulate a crappy connection. 
- Then I used [`backon`](https://docs.rs/backon/latest/backon/index.html) to implement exponential backoff in `fetch_chunk` around `get_object_concurrently` as proposed in https://github.com/earth-mover/icechunk/issues/1569#issuecomment-3909826030. 
- ^ actually lets us retry other StreamingErrors too (remove the toxics in the test described below would result in a different `IncompleteBody` error), so that is good.
- Re the test: It took some work to actually trigger the error. Apparently we need a combination of "toxics"
  - The test writes a chunk and then fetches it.
  - First we verify that we throw the 0B/s error.
  - Then we fetch the chunk, and remove the toxic after 7s concurrently. 
  - Thanks to the retry, we do successfully fetch the chunk eventually.
 
I will add inline comments re: some decisions to make.

-----

PS: Did you realize that the 90s had at least two bigs hits using the word "toxic"?
- Britney Spears - Toxic: https://www.youtube.com/watch?v=LOZuxwVk7TU
- System of a Down - Toxicity: https://www.youtube.com/watch?v=iywaBOMvYLI

The latter is a good representation of my day trying to track this down.

### DIFF
--- a/icechunk/tests/test_stalled_stream.rs
+++ b/icechunk/tests/test_stalled_stream.rs
@@ -33,7 +33,7 @@ use bytes::Bytes;
 /// This test requires MinIO and Toxiproxy running via docker compose:
 /// ```bash
 /// docker compose up -d
-/// cargo test test_stalled_stream --test test_stalled_stream -- --ignored --nocapture
+/// cargo test test_stalled_stream --test test_stalled_stream -- --nocapture
 /// ```
 use std::time::Duration;
 


icechunk/tests/setup_stalled_stream_test.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
dcherian: If we like this idea of using toxiproxy; i can have claude merge with our existing docker compose stuff.

icechunk/src/storage/s3.rs
```diff
@@ -686,10 +689,33 @@ impl Storage for S3Storage {
...
+        .retry(
+            ExponentialBuilder::new()
+                .with_max_times(settings.retries().max_tries().get().into())
+                .with_min_delay(Duration::from_millis(
+                    settings.retries().initial_backoff_ms().into(),
+                ))
+                .with_max_delay(Duration::from_millis(
+                    settings.retries().max_backoff_ms().into(),
+                )),
+        )
```
dcherian, line 708: reusing existing retry settings here. Sound OK?

icechunk/src/storage/s3.rs
```diff
@@ -686,10 +689,33 @@ impl Storage for S3Storage {
...
-            .await
+
+        (|| async {
+            self.get_object_concurrently(settings, key.as_str(), range)
+                .await?
+                .to_bytes((range.end - range.start) as usize)
+                .await
+        })
+        .retry(
+            ExponentialBuilder::new()
```
dcherian, line 700: new dep but potentially handles wasm well? https://docs.rs/backon/latest/backon/#default-sleeper

icechunk/src/storage/s3.rs
```diff
@@ -686,10 +689,33 @@ impl Storage for S3Storage {
...
+                    settings.retries().max_backoff_ms().into(),
+                )),
+        )
+        .when(|e: &ICError<StorageErrorKind>|
+              // Sadly ThroughputBelowMinimum is buried inside a boxed error,
+              // so check the debug output (yuck!)
+              // We actually check for StreamingError to catch more generic errors too
+              // For example, the integration test raises UnexpectedEof when toxics are removed.
+              format!("{e:?}").contains("StreamingError"))
+        .notify(|_err, duration| {
```
dcherian, line 715: Gah! but the errors are opaque 🤷🏾‍♂️

paraseba (approved):

dcherian:
CI errors should be fixed on #1657 and then merged in here

PR #1654, opened 2026-02-17
Fix redirect test failure (only happening on macos)

jbusecke:
When running the local rust tests I was getting: 
```
ailures:

---- test_redirect_storage stdout ----
Error: ICError { kind: BadRedirect("Redirect Storage response must be a redirect, no location header detected"), context: SpanTrace [] }


failures:
    test_redirect_storage

test result: FAILED. 15 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.32s

error: test failed, to rerun pass `-p icechunk --test test_storage`
error: Recipe `test` failed on line 6 with exit code 101
```
Thanks to the suggestion of @li-em and some clauding I was able to fix this by replacing all occurences of `localhost` to `127.0.0.1` (apparently macos defaults to `::1`).

### DIFF
--- a/icechunk-python/tests/test_redirect_storage.py
+++ b/icechunk-python/tests/test_redirect_storage.py
@@ -49,7 +49,7 @@ def test_era5_with_redirect() -> None:
     try:
         port = find_free_port()
         (server, thread) = run_server(port)
-        storage = ic.redirect_storage(f"http://localhost:{port}")
+        storage = ic.redirect_storage(f"http://127.0.0.1:{port}")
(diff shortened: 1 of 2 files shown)


jbusecke:
I am going to dabble a bit with #1656 to see if I can actually expose the bug in CI before fixing it.

jbusecke:
Since #1656 did not expose this issue I assume this is a behavior that was either changed after macos 15 or is somehow unique to my machine. So ill go to merge this anyways now.

PR #1656, opened 2026-02-17
Messing around with macos tests

jbusecke:
This is an attempt to expand the full testing (not just unit tests) to macos runners.

Apparently it is very difficult (if not impossible?) to run docker within (already virtualized?) runners (see [here](https://github.com/douglascamata/setup-docker-macos-action#arm64-processors-m-series-used-on-macos-15-images-and-beyond-are-unsupported)). 

My initial hope was to catch #1654 here before implementing a fix, but it seems like this is a behavior that was either changed after macos 15 or is somehow unique to my machine.

Either way I thought it could be good to test fully against at least one macos version?

The remaining question is if its worth carrying these tests as they are significantly slower (still working on that). 

**Positive side effect of this endeavour**: With claudes help I was able to reorganize the order and start compiling the tests while the docker setup finishes in parallel, which saves ~2 min on each test:

## Before
image
image

## After
image
image

### DIFF
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -29,7 +29,7 @@ env:
 jobs:
   rust:
     name: Rust CI
-    timeout-minutes: 35
+    timeout-minutes: 30
     runs-on: ${{ matrix.os }}
     defaults:
       run:
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -126,10 +126,12 @@ jobs:
           just test
 
       - name: Run doc tests
+        if: matrix.os != 'macos-latest'
         run: |
           just doctest
 
       - name: Run examples
+        if: matrix.os != 'macos-latest'
         run: |
           just run-all-examples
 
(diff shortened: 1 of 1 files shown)


jbusecke:
The first successful macos full test took almost 30 min: 
image

jbusecke:
Brought the macos test down to 
image

jbusecke:
Happy to revert the full macos-15-intel test for the sake of time, but could leave in the reordering of steps to speed up the other CI runs.

li-em:
~Would it be worth to run full tests only on `macos-latest`, and just to unit tests in `macos-15-intel` (like before)?~

Sorry, should have reviewed it properly before, you mention it is not possible to run docker on arm :see_no_evil:

jbusecke:
Yeah its a real bummer TBH, but maybe that gets solved when all the runners are m4 or whatever...not holding my breath.

jbusecke:
huh wait, why is it breaking now?

jbusecke:
Ok seems like (atleast the ubuntu failure was a flaky one).

paraseba (approved):

jbusecke:
Hopelessly outdated at this point. If this is still of interest I can revive it, but will close for now.

PR #1659, opened 2026-02-17
Add configurable retries for `update_repo_info`

### LINKED ISSUE #1655: Configurable number and type of retry for `update_repo_info`
`AssetManager::update_repo_info` is the most important method in IC2, it consistently modifies the entry point to the repo. Because all actions modify this object, there may be high contention over it, for this reason, modifications must be retried.

The number of retries is currently hardcoded in multiple places. What we want:
* make the number of retries configurable
* have the ability to configure different numbers for different operations. For example the required number of retries for GC could be much higher than for branch creation. 
* I deally we would need some time of exponential back-off on the retries. Probably a couple of retries without wait, and then some randomized waits that get larger

mpiannucci:
Closes #1655 

- **Configurable retries**: Replaced the 11 hardcoded `100`-retry call sites with a new `RepoUpdateRetryConfig` that allows setting different retry limits per operation category (commit, refs, GC), with sensible defaults (100 for commit/refs, 500 for GC).
- **Exponential backoff with jitter**: The retry loop in `update_repo_info` now sleeps between attempts instead of tight-looping — the first 5 retries are immediate, then backoff grows exponentially with full jitter up to a configurable max, reducing contention under concurrent writes.
- Integrates with existing `RetriesSettings`
- Uses the same library used in PR #1653 for S3 retries

### DIFF
--- a/icechunk/src/config.rs
+++ b/icechunk/src/config.rs
@@ -391,31 +391,14 @@ impl ManifestConfig {
 }
 
 /// Retry configuration for repo info update operations.
-///
-/// Different operation types can have different retry settings.
-/// If a specific category is `None`, the `default` settings are used.
-/// If `default` is also `None`, hardcoded defaults are used.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
 pub struct RepoUpdateRetryConfig {
-    /// Default retry settings applied when a specific category is not set.
+    /// Default retry settings for all repo update operations.
     #[serde(default)]
     pub default: Option<RetriesSettings>,
-
-    /// Retry settings for commit operations (flush, commit).
-    #[serde(default)]
-    pub commit: Option<RetriesSettings>,
-
-    /// Retry settings for ref operations (branch/tag create/delete/reset, metadata).
-    #[serde(default)]
-    pub refs: Option<RetriesSettings>,
-
-    /// Retry settings for GC operations (delete_snapshots, expire).
-    #[serde(default)]
-    pub gc: Option<RetriesSettings>,
 }
 
 static DEFAULT_REPO_UPDATE_RETRIES: OnceLock<RetriesSettings> = OnceLock::new();
-static DEFAULT_REPO_UPDATE_GC_RETRIES: OnceLock<RetriesSettings> = OnceLock::new();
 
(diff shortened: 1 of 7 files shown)


icechunk/src/config.rs
```diff
@@ -385,6 +390,92 @@ impl ManifestConfig {
     }
 }
 
+/// Retry configuration for repo info update operations.
+///
+/// Different operation types can have different retry settings.
+/// If a specific category is `None`, the `default` settings are used.
+/// If `default` is also `None`, hardcoded defaults are used.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct RepoUpdateRetryConfig {
```
paraseba, line 395: This is a bit of a weird mix of options. How about this, for now, we keep `default` only. They can always use `reopen` if they need a specific setting for an operation. Later, as we need more detail, we add more options for each operation, we don't need to break format for that.

paraseba:
> — the first 5 retries are immediate,

Is this implemented @mpiannucci ?

paraseba:
I think implementing a couple retries without wait would be useful, because in the vast majority of the times, it will just work without delaying. But I couldn't find the code that does that, is it the library natively doing it? If so, let's add a comment explaining it

mpiannucci:
I will verify thanks!

icechunk/src/asset_manager.rs
```diff
@@ -521,11 +506,29 @@ impl AssetManager {
...
         &self,
+        retry_settings: &storage::RetriesSettings,
         mut update: impl FnMut(Arc<RepoInfo>, &str) -> RepositoryResult<Arc<RepoInfo>>,
     ) -> RepositoryResult<VersionInfo> {
+        let max_attempts = retry_settings.max_tries().get() as usize;
+
+        // The first few retries are immediate (no delay) since brief contention
+        // typically resolves within one or two attempts. After that, we switch to
+        // exponential backoff with jitter.
+        let immediate_retries: u64 = 5;
```
mpiannucci, line 520: @paraseba Let me know if this is the behavior you are looking for, and if need this to be configurable?

paraseba: This is great, I don't think we need to configure this

Issue #1666, opened 2026-02-17
Enable test in test_virtual_refs.rs


paraseba:
I think this fixme can be lifted already:

```
    // FIXME: enable this once object_store can access public buckets without credentials
    // otherwise we get an error in GHA
```

TomNicholas:
I think since @li-em 's bugfix to anon access this should work

PR #1669, opened 2026-02-18
Delete FIXMEs and re-enable test

paraseba:
Closes: #1666


;
mpiannucci, line 1: Dont think this should be here

paraseba: wtf, thank you!

Issue #1672, opened 2026-02-18
minio -> rustfs


paraseba:

li-em:
There is also garage as another option: https://garagehq.deuxfleurs.fr/

PR #1673, opened 2026-02-18
Migrate tests from MinIO to RustFS

paraseba:
Also fix some lint warnings on a test

Closes: #1672

### DIFF
--- a/icechunk/src/virtual_chunks.rs
+++ b/icechunk/src/virtual_chunks.rs
@@ -598,7 +598,6 @@ impl ChunkFetcher for S3Fetcher {
                 ))
             }
             Some(Checksum::ETag(etag)) => {
-                dbg!("!!!!!!!", strip_quotes(&etag.0));
                 b = b.if_match(strip_quotes(&etag.0));
             }
             None => {}


icechunk/tests/test_virtual_refs.rs
```diff
@@ -288,7 +288,7 @@ async fn write_chunks_to_minio(chunks: impl Iterator<Item = (String, Bytes)>) {
         client
             .put_object()
             .bucket(bucket_name.clone())
-            .key(key)
+            .key(key.strip_prefix('/').unwrap())
```
li-em, line 291: Is this a behavior difference between minio and rustfs? Why is the stripping needed now?

paraseba: yeah ... other object stores have the ability to swallow the extra / in the key :shrug:

icechunk/src/virtual_chunks.rs
```diff
@@ -598,7 +598,7 @@ impl ChunkFetcher for S3Fetcher {
                 ))
             }
             Some(Checksum::ETag(etag)) => {
-                b = b.if_match(&etag.0);
```
paraseba, line 601: RustFS is very picky about etags

icechunk-python/tests/test_virtual_ref.py
```diff
@@ -97,7 +97,7 @@ async def test_write_minio_virtual_refs(
             location=f"s3://testbucket/{prefix}/chunk-1",
             offset=0,
             length=4,
-            etag_checksum="bad etag",
+            etag_checksum="00000000000000000000000000000000",
```
paraseba, line 100: RustFS doesn't like arbitrary strings as etags

PR #1661, opened 2026-02-17
Pixi for docs and tests

abkfenris:
Adds Pixi configuration for doc building and serving in addition to running pytest.

I’ve tried tweaking the ReadTheDocs config to play nice as well.

Works on #1648

PR #1674, opened 2026-02-18
Set up pixi for development

li-em:
Extends #1661 to add dependencies for running the Rust parts too.

Trying out with the code quality github actions check

### DIFF
--- a/icechunk-python/pixi.lock
+++ b/icechunk-python/pixi.lock
@@ -5864,7 +5864,7 @@ packages:
 - pypi: ./
   name: icechunk
   version: 2.0.0a0
-  sha256: c8eed05da65e48bdae3ef2746e541d88147b7cf0866d274445dab60f8df8fb61
+  sha256: efa6a75b6350fa2fc66f0cd1c3d6928908e3101caab6c1c6cd987e47ecb2924b
   requires_dist:
   - zarr>=3.1.0
   requires_python: '>=3.11'


.github/workflows/code-quality.yaml
```diff
@@ -33,20 +34,20 @@ jobs:
...
-      - name: Install Rust toolchain
-        run: |
-          rustup update --no-self-update ${{ env.RUST_CHANNEL }}
-          rustup component add --toolchain ${{ env.RUST_CHANNEL }} rustfmt rust-src clippy
-          rustup default ${{ env.RUST_CHANNEL }}
-
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@v2
         with:
           key: ${{ env.RUST_CHANNEL }}
```
li-em, line 40: Must fix this, rust channel doesn't make much sense anymore

jbusecke:
I have rearranged the logic here a bit and have now defined all developer 'tasks' in the Justfile. This seems more consistent at this point. I also updated the Python (+docs) instructions, but have not yet gotten to the Rust (and the special python test cases like upstream).

icechunk-python/docs/docs/ingestion/index.md
```diff
@@ -0,0 +1 @@
+# Index of ingestion
```
jbusecke, line 1: WHY IS THIS DAMN PAGE SHOWING UP ALL THE TIME?

icechunk-python/docs/docs/contributing.md
```diff
@@ -22,19 +22,22 @@ This guide describes the local development workflow for:
...
+=== "pixi (Recommended)"
+    Activate the developer shell (all following commands should be run within)
 
-#### Setting up your development environment
+    ```bash
+    pixi shell -m icechunk-python/pyproject.toml -e all
+    ```
+    and then in the shell
+    ```bash
+    pixi run -e all init
```
jbusecke: Thinking of making this a just command too and then have a subset of the commands depend on it? Thoughts @li-em ?

icechunk-python/docs/docs/contributing.md
```diff
@@ -111,7 +126,26 @@ cd icechunk-python
...
+
+    ```bash
+    just ruff-format
+    just ruff
+    just mypy
+    just py-pre-commit
+    ```
+
+=== "uv"
+    TBW
```
jbusecke: Its pretty rough to keep all of these variants around. How do people feel about just instructing people to use pixi, or else they kind of have to know what they are doing? Maybe that is too exclusionary...

icechunk-python/pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = ["zarr>=3.1.0"]
 dev = [
     { include-group = "test" },
     "mypy",
-    "ruff",
+    "ruff==0.8.0",
```
jbusecke, line 30: I will unpin this in a separate PR, so we can cleanly separate changes in code due to the version change (see also #1647)

icechunk-python/docs/docs/contributing.md
abkfenris: Pixi shouldn't need the `-e all` for defined tasks as `pixi run init` should know which environment it should run in, and it supports dependencies for it's built in tasks via https://pixi.prefix.dev/latest/workspace/advanced_tasks/#depends-on

li-em: yup! And I think we can name the `all` environment as `default`, which also avoids the `-e all` argument in the CLI

Justfile
```diff
@@ -74,6 +74,27 @@ pre-commit-python:
...
+pytest *args:
+  cd icechunk-python && pytest {{args}}
+
+docs-serve *args:
+  mkdocs serve -f icechunk-python/docs/mkdocs.yml --livereload {{args}}
+
+docs-build *args:
+  mkdocs build -f icechunk-python/docs/mkdocs.yml {{args}}
+
 create-deepak-env name:
```
jbusecke: @dcherian should we refactor this to pixi too?

icechunk-python/docs/docs/contributing.md
jbusecke: Testing that now

abkfenris: I use a variant of that technique for OceanHackWeek, so that we can split up the dependencies by tutorial/project to track where they came from but still install them all: https://github.com/oceanhackweek/jupyter-image/blob/293f282b7a7b1b5d7bfe2f52f4c995a8fd04fdb2/py-base/pixi.toml#L155-L165

Justfile
dcherian: we can delete that

icechunk-python/docs/docs/contributing.md
jbusecke: the default env works, but `pixi run init` does not:
I am getting
```
(icechunk) juliusbusecke@arm64-apple-darwin20 icechunk % pixi run init
 WARN Using local manifest /Users/juliusbusecke/Code/icechunk/pixi.toml rather than /Users/juliusbusecke/Code/icechunk/icechunk-python/pyproject.toml from environment variable `PIXI_PROJECT_MANIFEST`
init: command not found
```

This works now.
```
pixi shell -m icechunk-python/pyproject.toml init
# and then in the shell
pixi run -m icechunk-python/pyproject.toml init
``` 

which still feels awkward. Is there a way I can run the task automatically when starting the shell? Or alternatively set env variables? That way I can replace the `-m ...` I think.

Justfile
jbusecke: Removed:
```
create-deepak-env name:
  mamba create -y -n icechunk-{{name}} python=3.12 ipykernel ipdb
  mamba activate icechunk-{{name}}
  just coiled-ice-create {{name}}

coiled-ice-create version:
  pip install coiled arraylake icechunk=='{{version}}' watermark xarray bokeh
```

pixi.lock
abkfenris, line 1: This might be confusing pixi without a matching `pixi.toml` or `pyproject.toml` committed.

icechunk-python/docs/docs/contributing.md
abkfenris: From that warning, I think you may have conflicting manifests confusing things. 

Do you need to use `pixi shell` and `maturin develop`? I think re-enabling
```toml
[tool.pixi.pypi-dependencies]
icechunk = { path = ".", editable = true }
```
With make `maturin` automatically run for any pixi tasks (or starting a shell).

If that's slowing some things down, icechunk could be set as a feature and then only included in the environments for specific tasks.

dcherian:
I added a new `stubtest` check in https://github.com/earth-mover/icechunk/pull/1675. Perhaps that should be pixi-fied too

icechunk-python/docs/docs/contributing.md
jbusecke: That was it! Thanks a lot for this!

jbusecke:
@dcherian we were thinking of leaving the GHA pixification to the code quality checks only for now (as a demo basically). Maybe its better to layer on another PR replacing the entire CI to go through pixi too?

jbusecke:
Hmmm what happened to the RTD build....

abkfenris:
> Hmmm what happened to the RTD build....

Icechunk isn't getting built before the docs since
```toml
[tool.pixi.pypi-dependencies]
icechunk = { path = ".", editable = true }
```
is commented out.

jbusecke:
Thank you so much for the help (and patience) @abkfenris, I clearly need another coffee (and stop listening to claude haha).

abkfenris:
Well they did build a really effective lying machine. 

I'm glad for the distraction from folks trying to rage bait me into writing a SHEF encoder/decoder for them (among so many other things).

jbusecke:
Ok seems like all the CI is passing, but I am running into some issues with the pypi netcdf install in the docs build. I suspect this is due to the fact that my local python is 3.14 whereas the rtd build is 3.12?

abkfenris:
> Ok seems like all the CI is passing, but I am running into some issues with the pypi netcdf install in the docs build. I suspect this is due to the fact that my local python is 3.14 whereas the rtd build is 3.12?

It should be using the Python pulled in by `Pixi` even with a version specified on RTD.

```sh
❯ pixi tree -i python

mkdocstrings_python 2.0.2 

python 3.14.3 
├── pkgconfig 1.5.5 
├── nodeenv 1.10.0 
│   └── pre-commit 4.5.1 
├── ukkonen 1.1.0 
│   └── identify 2.6.16 
│       └── pre-commit 4.5.1 (*)
├── platformdirs 4.9.2 
│   ├── mkdocs_get_deps 0.2.0 
│   │   └── mkdocs 1.6.1 
```

jbusecke:
adding netcdf4 to the pixi deps fixes this on my mac (might have been specific to the pypi wheels for mac, or the non-existence thereof?)

jbusecke:
... and broke the CI...great.

abkfenris:
Hmm, it looks like it swapped for free-threaded builds of at least some packages.

```diff
-      - pypi: https://files.pythonhosted.org/packages/41/3c/a36c2450754618e62008bf7435ccb0f88053e07592e6028a34776213d877/markupsafe-3.0.3-cp314-cp314-manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_28_x86_64.whl
+      - pypi: https://files.pythonhosted.org/packages/50/09/c419f6f5a92e5fadde27efd190eca90f05e1261b10dbd8cbcb39cd8ea1dc/markupsafe-3.0.3-cp314-cp314t-manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_28_x86_64.whl
```

abkfenris:
Including Python

```diff
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pycparser-2.22-pyh29332c3_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/python-3.14.3-hb06a95a_101_cp314.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/python_abi-3.14-8_cp314.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/pyyaml-6.0.3-py314h807365f_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/python-3.14.3-h6718eea_1_cp314t.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/python_abi-3.14-8_cp314t.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pyyaml-6.0.3-pyh7db6752_1.conda
```

That's not right

li-em:
> Hmm, it looks like it swapped for free-threaded builds of at least some packages.

WTH, we definitely don't want that :joy:

abkfenris:
Try explicitly `pixi add python` so it doesn't solve it from other packages.

abkfenris:
I tossed a post in Prefix's discord because free-threaded Python should not have been pulled in there: https://discord.com/channels/1082332781146800168/1474142415702065459/1474142415702065459

icechunk-python/docs/docs/ingestion/index.md
ianhi: this is my white whale

jbusecke:
Ill be OOO today and Monday, but pick this up Tues.

.github/workflows/code-quality.yaml
jbusecke: @claude adapt this to the proposed changes within pixi

icechunk-python/docs/docs/contributing.md
```diff
@@ -92,16 +85,42 @@ cd icechunk-python
     maturin develop
     ```
 
+All build, test, and code quality tasks are managed with [`just`](https://github.com/casey/just). Run `just --list` to see all available recipes with descriptions.
```
jbusecke, line 88: I figured this would have helped me in the beginning. I also added proper 'docs' so that just --list returns sensible output

icechunk-python/docs/docs/contributing.md
```diff
@@ -355,64 +392,28 @@ pre-commit run rust-pre-commit-ci --hook-stage manual
...
 
 This builds the site to `docs/.site` directory.
 
 **Tips**:
 
-- Use `mkdocs serve --dirty` to only rebuild changed files (faster for iterative development)
+- Use `just docs-serve --dirty` to only rebuild changed files (faster for iterative development)
 - You may need to restart if you make changes to `mkdocs.yml`
-- For debugging the doc build logs, check out [docs-output-filter](https://github.com/ianhi/docs-output-filter) (you can run `uv run docs-output-filter -- mkdocs serve --livereload` once installed). *This also works to debug remote builds like RTD with the `--url` flag* 😍
+- For debugging the doc build logs, check out [docs-output-filter](https://github.com/ianhi/docs-output-filter) (you can run `docs-output-filter -- just docs-serve` once installed). *`docs-output-filter` also works to debug remote builds like RTD with the `--url` flag*
```
jbusecke, line 416: @ianhi is this ok?

icechunk-python/pyproject.toml
```diff
@@ -185,3 +185,39 @@ ban-relative-imports = "all"
...
+uv = ">=0.10.4,<0.11"
+openssl = ">=3.6.1,<4"
+pkgconfig = ">=1.5.5,<2"
+cargo-deny = ">=0.19.0,<0.20"
+prek = ">=0.3.3,<0.4"
+netcdf4 = ">=1.7.4,<2"
+cargo-nextest = ">=0.9.106,<0.10"
+
+[tool.pixi.target.osx-arm64.activation.env]
+DYLD_LIBRARY_PATH = "$CONDA_PREFIX/lib"
```
jbusecke: @li-em is this kosher to do? It works for me locally but seems janky...

Justfile
```diff
@@ -4,62 +4,64 @@ profile := "dev"
...
 
+[doc("Run Rust doc tests only")]
 doctest *args='':
   cargo test --workspace --profile {{profile}} --doc {{args}}
 
-# run all tests with logs enabled
+[doc("Run all Rust tests with RUST_LOG enabled (e.g. `just test-logs debug`)")]
 test-logs level *args='':
-  RUST_LOG=icechunk={{level}} cargo nextest run --no-fail-fast --cargo-profile {{profile}} --workspace --all-targets {{args}} -- --nocapture
+  export DYLD_LIBRARY_PATH="${CONDA_PREFIX:-}/lib" && RUST_LOG=icechunk={{level}} cargo nextest run --no-fail-fast --cargo-profile {{profile}} --workspace --all-targets {{args}} -- --nocapture
```
jbusecke, line 17: Not 100% sure if this here or above fixed the issue. Ill test locally.

icechunk-python/pyproject.toml
jbusecke: I have removed the `DYLD_LIBRARY_PATH` from here and left it in the justfile only

Justfile
jbusecke: Ill leave this in for now unless someone has a neater solution that does not require users to modify their mac?

jbusecke:
I am happy with this for now. Ill wait for the last look of @li-em before merging. I am planning to push a couple smaller PRs soon that will:
- update the ruff versions
- move other parts of the CI to run through pixi/just (I can hold off on that until after IC2)
- Work through the contributor guide beyond python.

Issue #1607, opened 2026-02-03
explore using `stubtest`


dcherian:
we sometimes have issues with our stub files being out of date. `stubtest` might help. https://mypy.readthedocs.io/en/stable/stubtest.html

We should explore.

PR #1675, opened 2026-02-18
Add stubtest to CI to catch stub/runtime mismatches

dcherian:
Closes https://github.com/earth-mover/icechunk/issues/1607

Big diff but mostly minor since we're just fixing the stubs and we wrap that interface with the public interface in `repository.py`

Most of whats left on the allowlist are apparently pyO3 quirks


mpiannucci (approved):

dcherian:
> how is the big txt file maintained?

This was generated by Claude using the `--allowlist` flag to the CLI. I think as we add more to the interface we may need to manually add entries there. There are docs here: https://mypy.readthedocs.io/en/stable/stubtest.html#allowlist

Issue #1412, opened 2025-11-18
Move configuration into the repo info object


paraseba:
This would allow us to:
* track configuration changes in the op log
* apply configuration changes consistently with other repo changes

A trade off is that fetching config would require downloading the whole repo object now. Do you think this is a concern @TomNicholas ?

TomNicholas:
I do think that probably is a performance concern, yes.

mpiannucci:
Did we decide we need to do this for consistency? Regardless of the possible perf hit?

And to be clear, this means config.yaml goes away?

paraseba:
> Did we decide we need to do this for consistency? Regardless of the possible perf hit?

Yes, I think the small performance hit on loading config-only is a corner use case, that can be improved with optimistic loading without or with backed up config when needed. We can also use or improve `reopen`

> And to be clear, this means config.yaml goes away?

Yes. We need to add to the migration too, but feel free to split into different issues. LMK if it would be useful to iterate on a tiny design doc.

PR #1676, opened 2026-02-18
Log config changes in ops log

### LINKED ISSUE #1660: Track configurations changes in ops log

mpiannucci:
Fixes #1660


dcherian (approved):
LGTM

icechunk/src/repository.rs
```diff
@@ -543,12 +543,27 @@ impl Repository {
...
     #[instrument(skip_all)]
     pub async fn save_config(&self) -> RepositoryResult<storage::VersionInfo> {
-        Repository::store_config(
+        let version = Repository::store_config(
             self.storage().clone(),
             self.config(),
             &self.config_version,
         )
-        .await
+        .await?;
```
paraseba, line 551: This is good enough at this stage, but it's inconsistent (process can die between `store_config` and this). To achieve consistency we need to implement in conjunction with #1412 .

But I'm OK with merging at this stage and later move the config to the repo object, that's why I split it in two tickets.

mpiannucci: Ok will incorporate into pr for that issue next!

PR #1681, opened 2026-02-19
Two version icechunk stateful tests

ianhi:
early open to test out the CI actually running

This got a bit more complicated trying to deal with the all little areas where there are small API differences (i.e. additions like `allow_empty`) and making sure that we set up the storage objects properly. 

We have to use on-disk storage as you cannot share memory stores across library versions.

I didn't add checking between the two actors as they ultimately both get checked against the model so that seemed redundant.

closes: https://github.com/earth-mover/icechunk/issues/1321

also I had a decent amoutn of free time while I waited for hypothesis tests to run. so i also wrote a small blog about this because I thought it was fun: https://earthmover.io/?p=3155&preview=true

### DIFF
--- a/icechunk-python/tests/test_stateful_compat.py
+++ b/icechunk-python/tests/test_stateful_compat.py
@@ -35,7 +35,7 @@ repo_ops_mod.IcechunkError = (ic.IcechunkError, ic_v1.IcechunkError)  # type: ig
 
 
 import hypothesis.strategies as st
-from hypothesis import note, settings
+from hypothesis import note
 from hypothesis.stateful import initialize, precondition, rule, run_state_machine_as_test
 
 from tests.test_stateful_repo_ops import (
@@ -46,9 +46,7 @@ from tests.test_zarr.test_stateful import (
 )
 
 
-class CrossVersionVersionControlStateMachine(
-    VersionControlStateMachine
-):
+class CrossVersionVersionControlStateMachine(VersionControlStateMachine):
     """Two-actor version control test: one actor is v2, the other is v1.
 
     Uses filesystem storage since v1 and v2 packages can't share storage objects.
@@ -97,9 +95,7 @@ def test_two_actors_cross_version() -> None:
     )
 
 
-class CrossVersionTwoActorZarrHierarchyStateMachine(
-    ModifiedZarrHierarchyStateMachine
-):
+class CrossVersionTwoActorZarrHierarchyStateMachine(ModifiedZarrHierarchyStateMachine):
     """Two-actor Zarr hierarchy test: one actor is v2, the other is v1.
 
     Uses filesystem storage since v1 and v2 packages can't share storage objects.
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -419,7 +419,7 @@ class VersionControlStateMachine(RuleBasedStateMachine):
                 self.sync_store.set(path, value)
 
     @rule()
-    @precondition(lambda self: getattr(self.repo, "spec_version", 1) == 1)
+    @precondition(lambda self: self.model.spec_version == 1)
     def upgrade_spec_version(self) -> None:
         # don't test simple cases of catching error upgradging a v2 spec
         # that should be covered in unit tests
@@ -440,9 +440,7 @@ class VersionControlStateMachine(RuleBasedStateMachine):
         """
         assert self.storage is not None, "storage must be initialized"
         self.repo = self.actor.open(self.storage, config=config)
-        note(
-            f"Reopened repository (spec_version={getattr(self.repo, 'spec_version', 1)})"
-        )
+        note(f"Reopened repository (spec_version={self.model.spec_version})")
 
         # Reopening discards uncommitted changes - reset model to last committed state
         branch = (
@@ -471,7 +469,7 @@ class VersionControlStateMachine(RuleBasedStateMachine):
     # https://github.com/earth-mover/icechunk/issues/1532
     @precondition(
         lambda self: (self.model.changes_made)
-        and (getattr(self.repo, "spec_version", 1) >= 2)
+        and (self.model.spec_version >= 2)
         and len(self.model.commits) > 1
     )
     def amend(self, message: str) -> str:
@@ -635,10 +633,7 @@ class VersionControlStateMachine(RuleBasedStateMachine):
     # TODO: v1 has bugs in expire_snapshots, only test for v2
     # https://github.com/earth-mover/icechunk/issues/1520
     # https://github.com/earth-mover/icechunk/issues/1534
-    @precondition(
-        lambda self: bool(self.model.commits)
-        and getattr(self.repo, "spec_version", 1) == 2
-    )
+    @precondition(lambda self: bool(self.model.commits) and self.model.spec_version == 2)
     @rule(
         data=st.data(),
         delta=st.timedeltas(
--- a/icechunk-python/tests/test_zarr/test_stateful.py
+++ b/icechunk-python/tests/test_zarr/test_stateful.py
@@ -35,6 +35,8 @@ PROTOTYPE = default_buffer_prototype()
 class ModelStore(MemoryStore):
     """MemoryStore with move and copy methods for testing."""
 
+    spec_version: int
+
     async def move(self, source: str, dest: str) -> None:
         """Move all keys from source to dest.
 
@@ -92,6 +94,7 @@ class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
         super().__init__(temp_store)
         # Replace parent's MemoryStore with our ModelStore that has move()
         self.model = ModelStore()
+        self.model.spec_version = 1
         zarr.group(store=self.model)
 
     @initialize(spec_version=st.sampled_from([1, 2]))
@@ -107,6 +110,7 @@ class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
         else:
             self.repo = self.actor.create(self.storage)
         self.store = self.repo.writable_session("main").store
+        self.model.spec_version = spec_version
 
         super().init_store()
 
@@ -208,7 +212,7 @@ class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
             )
 
     @rule()
-    @precondition(lambda self: getattr(self.repo, "spec_version", 1) == 1)
+    @precondition(lambda self: self.model.spec_version == 1)
     @precondition(lambda self: not self.store.session.has_uncommitted_changes)
     def upgrade_spec_version(self) -> None:
         """Upgrade repository from spec version 1 to version 2."""
@@ -217,9 +221,10 @@ class ModifiedZarrHierarchyStateMachine(ZarrHierarchyStateMachine):
         # Reopen to pick up the upgraded spec version
         self.repo = self.actor.open(self.storage)
         self.store = self.repo.writable_session("main").store
+        self.model.spec_version = 2
 
     @rule(data=st.data(), num_moves=st.integers(min_value=0, max_value=5))
-    @precondition(lambda self: getattr(self.repo, "spec_version", 1) >= 2)
+    @precondition(lambda self: self.model.spec_version >= 2)
     @precondition(lambda self: not self.store.session.has_uncommitted_changes)
     @precondition(lambda self: bool(self.all_arrays) or bool(self.all_groups))
     def move_operations(self, data: st.DataObject, num_moves: int) -> None:


icechunk-python/python/icechunk/testing/strategies.py
```diff
@@ -64,14 +65,19 @@ def repository_configs(
     ),
     inline_chunk_threshold_bytes: st.SearchStrategy[int] | None = None,
     splitting: st.SearchStrategy[ic.ManifestSplittingConfig] | None = None,
+    ic_module: ModuleType | None = None,
 ) -> ic.RepositoryConfig:
+    ice = ic_module or ic
```
ianhi, line 71: @dcherian the thing that felt bad is this need to use a local copy of the module in places like this or in the superclasses. But I don't really see an easier way around this, with multiple versions you have to use different objects or everythign breaks. and unless you want to completely copy the underlying functions then you have to push some reslience to multi version up into the base classes.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -156,7 +166,13 @@ def commit_with_check(self, data: st.DataObject) -> None:
...
 
         allow_empty = not self.store.session.has_uncommitted_changes
-        self.store.session.commit("foo", allow_empty=allow_empty)
+        # Try with allow_empty, fall back without it for v1 compatibility
+        try:
+            self.store.session.commit("foo", allow_empty=allow_empty)
+        except TypeError as e:
+            if "allow_empty" not in str(e):
+                raise
+            self.store.session.commit("foo")
```
ianhi: Another spot where the two version tests are affecting the base class

icechunk-python/python/icechunk/testing/strategies.py
```diff
@@ -64,14 +65,19 @@ def repository_configs(
...
+    kwargs: dict[str, Any] = {
+        "inline_chunk_threshold_bytes": draw(inline_chunk_threshold_bytes)
         if inline_chunk_threshold_bytes is not None
         else None,
-        manifest=manifest,
-    )
+        "manifest": manifest,
+    }
+    # num_updates_per_repo_info_file is v2-only
+    if hasattr(ice.RepositoryConfig(), "num_updates_per_repo_info_file"):
```
dcherian: can we do an explicit version check here?

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -361,6 +377,51 @@ def pickle_objects(self) -> None:
         pickle.loads(pickle.dumps(self.repo))
 
 
+class TwoActorZarrHierarchyStateMachine(ModifiedZarrHierarchyStateMachine):
```
dcherian: I think we can just fold this in to the other one.

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -361,6 +377,51 @@ def pickle_objects(self) -> None:
...
+
+
+def test_two_actors() -> None:
+    def mk_test_instance_sync() -> ModifiedZarrHierarchyStateMachine:
+        return TwoActorZarrHierarchyStateMachine(in_memory_storage())
+
+    run_state_machine_as_test(  # type: ignore[no-untyped-call]
+        mk_test_instance_sync, settings=settings(report_multiple_bugs=False)
+    )
+
```
dcherian: Certainly we don't need to run this test in its current form

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -370,6 +431,15 @@ def mk_test_instance_sync() -> ModifiedZarrHierarchyStateMachine:
     )
 
 
+def test_zarr_hierarchy_two_actors() -> None:
```
dcherian: dupe?

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,162 @@
...
+
+Skipped entirely if icechunk_v1 is not installed.
+"""
+
+import os
+import tempfile
+
+# Set log filter before importing icechunk packages so both v1 and v2
+# pick it up during tracing initialization.
+os.environ.setdefault("ICECHUNK_LOG", "warn,icechunk::storage::object_store=error")
```
dcherian: what's this for?

ianhi: All the local filesystem warnings you get. With them it's super annoying to look through the error logs on a failed test.

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: ? why not. this is two actors both v2, the compat is two actors of two different versions. So not fully redundant.

icechunk-python/tests/test_zarr/test_stateful.py
dcherian: Again can we use a version check here please?

icechunk-python/tests/test_zarr/test_stateful.py
dcherian: because we switch between actors, it's no different from the current test which also "switches" actors with `reopen_with_config`.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -55,6 +60,17 @@
...
+# Branches/tags are stored as files at ``refs/<name>``.  On case-insensitive
+# file systems (macOS APFS default), 'T' and 't' collide.
+if sys.platform == "darwin":
+    ref_name_text = st.text(
+        st.sampled_from(string.ascii_lowercase + string.digits),
+        min_size=1,
+        max_size=5,
+    )
+else:
+    ref_name_text = simple_text
```
dcherian:
```suggestion
ref_name_text = simple_text.map(lambda x: x.lower() if sys.platform == "darwin" else x)
```

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,162 @@
...
+        self.storage = self.actor_storage_objects[actor_name]
+
+        note(f"Creating repository with spec_version={spec_version}, actor={self.actor}")
+
+        try:
+            self.repo = self.actor.create(self.storage, spec_version=spec_version)
+        except TypeError as e:
+            if "spec_version" not in str(e):
+                raise
+            self.repo = self.actor.create(self.storage)
```
dcherian: Let's use an explicit version check here too

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,162 @@
...
+        spec_version: Literal[1, 2],
+        use_in_memory_storage: bool,
+    ) -> str:
+        return super().initialize(data, spec_version, use_in_memory_storage)  # type: ignore[return-value]
+
+
+def test_two_actors_cross_version() -> None:
+    run_state_machine_as_test(  # type: ignore[no-untyped-call]
+        CrossVersionTwoActorVersionControlStateMachine,
+        settings=settings(report_multiple_bugs=False),
```
dcherian:
```suggestion
```
This is a debugging thing

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -373,11 +382,20 @@ def initialize(self, data: st.DataObject, spec_version: Literal[1, 2]) -> str:
         self.model.num_updates = 1
 
         # initialize with some data always
-        # TODO: always setting array metadata, since we cannot overwrite an existing group's zarr.json
```
dcherian, line 376: ?

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -791,7 +814,91 @@ def check_ancestry(self) -> None:
...
+        super().reopen_repository(data)
+
+
+def test_two_actors() -> None:
+    def mk_test_instance_sync() -> TwoActorVersionControlStateMachine:
+        return TwoActorVersionControlStateMachine()
+
+    run_state_machine_as_test(  # type: ignore[no-untyped-call]
+        mk_test_instance_sync, settings=settings(report_multiple_bugs=False)
+    )
```
dcherian:
```suggestion
```

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -791,7 +814,91 @@ def check_ancestry(self) -> None:
...
+    ) -> str:
+        choice = data.draw(st.sampled_from(list(self.actors.keys())))
+        note(f"initializing with actor {choice!r}")
+        self.actor = self.actors[choice]
+        self.ic = self.actor_modules[choice]
+
+        # Create storage - have to differentiate so that in multi-actor
+        # tests with different icechunk versions nothing breaks. icechunk_v1 cannot
+        # share in-memory storage with icehunk 2
+        if use_in_memory_storage:
```
dcherian: I would do something like this instead in the superclass

Image

And since the smethod names don't change you can just do `self.ic.local_filesystem_storage`

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -791,7 +814,91 @@ def check_ancestry(self) -> None:
...
+
+    def __init__(self) -> None:
+        super().__init__(actor=None)
+        self.actors = {"one": Repository, "two": Repository}
+        self.actor_modules: dict[str, Any] = {"one": icechunk, "two": icechunk}
+        self.actor_storage_objects = {}
+        self.on_disk_storage_factory: dict[str, Callable[[str], Storage]] = defaultdict(
+            lambda: local_filesystem_storage
+        )
+
```
dcherian:
```suggestion
```

I don't think either of these is needed. See below

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,162 @@
...
+        try:
+            self.repo = self.actor.create(self.storage, spec_version=spec_version)
+        except TypeError as e:
+            if "spec_version" not in str(e):
+                raise
+            self.repo = self.actor.create(self.storage)
+        self.store = self.repo.writable_session("main").store
+
+        # Call grandparent's init_store for zarr-level init
+        ZarrHierarchyStateMachine.init_store(self)
```
dcherian: This looks very wrong to me. You should just be calling methods on super()

ianhi: I rememebr thinking exactly the same but still ending up with this. Though it was right before I took a nap so maybe I wasn't thinking straight 🤔

ianhi: ok i remember. it's basically a question of where the logic above these lines lives. it can either live here. or it can live in the base class. If it lives here then we need to skip these lines:

https://github.com/earth-mover/icechunk/blob/fdb9ac18de11bc2d528c5f7ecd3074f301ea6904/icechunk-python/tests/test_zarr/test_stateful.py#L94-L98

and just call the grandparent

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: hmm, but you still need the if in order to correctly populate the other actors storage. In the filesystem case with multiple versions the each need to call their own method.

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: yes. I think i wrote a couple of these before I started carting around the icechunk object. so i wasn't able to do version checks. didn't cross my mind after. will add

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: the other meaning the cross version? or its super calss?

icechunk-python/tests/test_zarr/test_stateful.py
ianhi: fair enouhg. i keep getting in this mental trap of conceptualizing them as running in parallel. rather than interspersed.

icechunk-python/tests/test_stateful_compat.py
ianhi: sure. but we also have that everywhere else. I can also remove there, or keep here to be consistent. dont feel strongly either way.

dcherian: remove it everywhere, it's a mistake

icechunk-python/tests/test_stateful_repo_ops.py
dcherian: when you open with the other actor, can't you just remake storage with the right module?

icechunk-python/tests/test_stateful_compat.py
dcherian: i'd refactor so we always call methods on super()

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: ok after thinking more you can remove the factory object. but I really don't think you can escape keeping track of the storage for the multi version. It's complicated by the fact that you can't recreate the in memory storage object as needed like yoiu can with the on disk version.

ianhi: at least I dont think you can.

ianhi: we could simplify by only ever doing multiactor tests with on disk. which I'm realizing i fell into the same parallel timing vs hand off mental trap here :/

icechunk-python/tests/test_stateful_compat.py
ianhi: done. had to simplify a bit by just removing the two actor without multi version. but i think that that is fine and ultimately achieved the right thing.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -613,7 +635,10 @@ def delete_branch(self, branch: str) -> None:
     # TODO: v1 has bugs in expire_snapshots, only test for v2
     # https://github.com/earth-mover/icechunk/issues/1520
     # https://github.com/earth-mover/icechunk/issues/1534
-    @precondition(lambda self: bool(self.model.commits) and self.repo.spec_version == 2)
+    @precondition(
+        lambda self: bool(self.model.commits)
+        and getattr(self.repo, "spec_version", 1) == 2
```
dcherian:
```suggestion
        and self.model.spec_version == 2
```

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -449,7 +471,7 @@ def commit(self, message: str) -> str:
     # https://github.com/earth-mover/icechunk/issues/1532
     @precondition(
         lambda self: (self.model.changes_made)
-        and (self.repo.spec_version >= 2)
+        and (getattr(self.repo, "spec_version", 1) >= 2)
```
dcherian:
```suggestion
        and self.model.spec_version >= 2
```

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -55,6 +56,10 @@
     max_size=5,
 )
 
+# Branches/tags are stored as files at ``refs/<name>``.  On case-insensitive
+# file systems (macOS APFS default), 'T' and 't' collide.
+ref_name_text = simple_text.map(lambda x: x.lower() if sys.platform == "darwin" else x)
```
dcherian, line 61: hmmm cant figure out why i havent run in to this before!

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -399,7 +419,7 @@ def set_doc(self, path: str, value: Buffer) -> None:
                 self.sync_store.set(path, value)
 
     @rule()
-    @precondition(lambda self: self.repo.spec_version == 1)
+    @precondition(lambda self: getattr(self.repo, "spec_version", 1) == 1)
```
dcherian:
```suggestion
    @precondition(lambda self: self.model.spec_version == 1)
```

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -410,7 +430,7 @@ def upgrade_spec_version(self) -> None:
 
     @rule(data=st.data())
     def reopen_repository(self, data: st.DataObject) -> None:
-        config = data.draw(repository_configs())
+        config = data.draw(repository_configs(ic_module=self.ic))
```
dcherian, line 433: This is the only ugly thing I see, but I can also see its unavoidable. Nice work!

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,169 @@
...
+from tests.test_stateful_repo_ops import (
+    VersionControlStateMachine,
+)
+from tests.test_zarr.test_stateful import (
+    ModifiedZarrHierarchyStateMachine,
+)
+
+
+class CrossVersionVersionControlStateMachine(
+    VersionControlStateMachine,
```
dcherian:
```suggestion
    VersionControlStateMachine
```
total nit :)

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,169 @@
...
+        note(f"switching to actor {choice!r}")
+        self.actor = self.actors[choice]
+        self.ic = self.actor_modules[choice]
+        self.storage = self._make_storage()
+        super().reopen_repository(data)
+
+
+def test_two_actors_cross_version() -> None:
+    run_state_machine_as_test(  # type: ignore[no-untyped-call]
+        CrossVersionVersionControlStateMachine,
```
dcherian:
```suggestion
        CrossVersionVersionControlStateMachine
```

more nits :P vertical space is at a premium on laptop monitors!

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+
+
+class CrossVersionTwoActorZarrHierarchyStateMachine(
+    ModifiedZarrHierarchyStateMachine,
```
dcherian:
```suggestion
    ModifiedZarrHierarchyStateMachine
```

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -0,0 +1,169 @@
...
+    @precondition(lambda self: False)
+    def upgrade_spec_version(self) -> None:
+        pass
+
+    def _make_storage(self) -> ic.Storage:
+        return self.ic.local_filesystem_storage(self._storage_path)  # type: ignore[no-any-return]
+
+    @initialize(
+        data=st.data(),
+        target=VersionControlStateMachine.branches,
```
dcherian, line 72: one more (minor) reason to not use a Bundle

ianhi: ?

icechunk-python/tests/test_stateful_repo_ops.py
ianhi: do this everywhre i guess?

ianhi: i did this everywhere

icechunk-python/tests/test_stateful_compat.py
ianhi: ackk. i applied this. and then unthinkingly agreed with whatever ruff wanted and seemingly end up back with this :/ we can always try harder to bring this back in a future PR

PR #1694, opened 2026-02-20
Move repo config into repo object

### LINKED ISSUE #1691: Add configuration to repo object
Serialize using flexbuffers, update all sync to use this config

mpiannucci:
Closes #1691

Makes all repo config updates records atomic

Backward compatible by trying to read `config.yaml` when the repo object config is null

### DIFF
--- a/icechunk/src/repository.rs
+++ b/icechunk/src/repository.rs
@@ -32,6 +32,7 @@ use tracing::{Instrument, debug, error, instrument, trace};
 use crate::{
     Storage, StorageError,
     asset_manager::AssetManager,
+    change_set::ChangeSet,
     config::{
         Credentials, DEFAULT_MAX_CONCURRENT_REQUESTS, ManifestPreloadCondition,
         RepositoryConfig,
@@ -46,7 +47,7 @@ use crate::{
             ManifestFileInfo, NodeData, NodeType, Snapshot, SnapshotInfo,
             SnapshotProperties,
         },
-        transaction_log::{Diff, DiffBuilder},
+        transaction_log::{Diff, DiffBuilder, TransactionLog},
     },
     refs::{self, Ref, RefError, RefErrorKind},
     session::{Session, SessionErrorKind, SessionResult},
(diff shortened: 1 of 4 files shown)


icechunk/src/migrations/mod.rs
```diff
@@ -225,6 +225,12 @@ pub async fn migrate_1_to_2(
     info!("Found {} non-dangling snapshots", all_snapshots.len());
 
     info!("Creating repository info file");
+    let config = repo.config().clone();
+    let config_bytes: Vec<u8> = flexbuffers::to_vec(&config).map_err(|e| {
```
mpiannucci, line 235: I think this correctly handles migration, but i may be naive

icechunk/src/format/repo_info.rs
```diff
@@ -441,10 +451,29 @@ impl RepoInfo {
...
         .expect("Cannot generate initial snapshot")
     }
 
+    /// Read the raw config bytes from the FlatBuffer (for pass-through in mutations).
+    pub(crate) fn config_bytes_raw(&self) -> IcechunkResult<Option<Vec<u8>>> {
+        Ok(self.root()?.config().map(|v| v.bytes().to_vec()))
+    }
+
+    /// Read the repository configuration from the repo info.
+    /// Returns `None` for repos created before config was embedded.
```
paraseba: This should return None for repositories that have the default configuration no?

icechunk/src/migrations/mod.rs
paraseba: Maybe after the migration is "verified correct", whatever that means, we should delete the config yaml file.

icechunk/src/repository.rs
```diff
@@ -248,15 +248,20 @@ impl Repository {
...
 
             if spec_version >= SpecVersionBin::V2dot0 {
                 let snap_info = new_snapshot.as_ref().try_into()?;
-                let repo_info =
-                    Arc::new(RepoInfo::initial(spec_version, snap_info, num_updates));
+                let repo_info = Arc::new(RepoInfo::initial(
+                    spec_version,
+                    snap_info,
+                    num_updates,
+                    Some(config_ref),
```
paraseba: This is a bit lossy. In the previous version, if config is equal to the default, it's not saved to repo storage. That is a pretty good thing because it gives a different semantics, it basically says "I want the defaults, and I'm ok with the defaults even if they change". In this new scenario two things happen:
* every repo uses storage to store config even if it's the default
* we don't know if what is stored means "I really want this" or "give me the defaults"

Can we maintain the previous semantics of not storing the default config?

icechunk/src/repository.rs
```diff
@@ -311,18 +324,8 @@ impl Repository {
...
-        );
-
-        let storage_c = Arc::clone(&storage);
-        let find_spec_version = tokio::spawn(Self::fetch_spec_version(storage_c));
-
-        let (config_res, spec_version_res) = try_join!(fetch_config, find_spec_version)?;
 
-        // fail if repo doesn't exist
-        let spec_version = match spec_version_res {
+        let spec_version = match Self::fetch_spec_version(Arc::clone(&storage)).await {
```
paraseba: Can we launch the fetch of the spec version concurrently with the "optimistic" fetch of config.yaml? We need to ignore errors on config.yaml if we end up discovering that we are in a V2 repo. I just don't want to make IC2 (the library) slower to open IC1 repos.

mpiannucci:
@paraseba Addressed comments. Now we dont store config by default like in v1 and concurrently download the config + spec version

PR #1695, opened 2026-02-20
Protect repo object updates from time skew

### LINKED ISSUE #1635: Time skew when we update the repo object
We cannot allow an update where the timestamp is <= than the top of the ops log.

mpiannucci:
Closes #1635

### DIFF
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -70,6 +70,9 @@ pub struct AssetManager {
 
     max_concurrent_requests: u16,
 
+    #[serde(default = "_default_true")]
+    use_repo_info_cache: bool,
+
     #[serde(skip)]
     manifest_cache_size_warned: AtomicBool,
     #[serde(skip)]
@@ -91,6 +94,10 @@ pub struct AssetManager {
     repo_cache: RwLock<Option<(Arc<RepoInfo>, VersionInfo)>>,
 }
 
+const fn _default_true() -> bool {
+    true
+}
+
 impl private::Sealed for AssetManager {}
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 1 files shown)


dcherian (approved):

icechunk/src/format/repo_info.rs
```diff
@@ -1309,6 +1320,7 @@ mod tests {
         let snap2 = SnapshotInfo {
             id: id2.clone(),
             parent_id: Some(id1.clone()),
+            flushed_at: DateTime::from_timestamp_micros(2_000_000).unwrap(),
```
dcherian, line 1323: ~How did this work earlier?~ Ah missed the `..`

icechunk/src/format/mod.rs
```diff
@@ -330,6 +331,10 @@ pub enum IcechunkFormatErrorKind {
     Path(#[from] PathError),
     #[error("invalid timestamp in file")]
     InvalidTimestamp,
+    #[error(
+        "update timestamp `{new_time}` is not newer than the latest update `{latest_time}`"
+    )]
+    InvalidUpdateTimestampOrdering { latest_time: DateTime<Utc>, new_time: DateTime<Utc> },
```
paraseba: This is the error people will see if they just have their laptops on the wrong time zone or something, so we want it to be very clear. There is a similar protection with an error in Session:
```rust
    #[error(
        "snapshot timestamp is invalid, please verify if the machine clock has drifted: local clock: `{snapshot_time}`, object store clock: `{object_store_time}`"
    )]
    InvalidSnapshotTimestamp {
        object_store_time: DateTime<Utc>,
        snapshot_time: DateTime<Utc>,
    },

PR #1697, opened 2026-02-20
Make the repo info cache in AssetManager optional

### LINKED ISSUE #1684: Make the repo info cache in AssetManager optional

li-em:
Fix #1684

- Avoid the repo info cache when creating an asset manager with `AssetManager::new_no_cache`
- Still default to using the cache for all other use cases

### DIFF
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -519,7 +519,14 @@ impl AssetManager {
                         "Logic bug in fetch_repo_info, repo_cache should exist here",
                     ));
                 } else {
-                    todo!("return error")
+                    // It is very surprising to get to this branch:
+                    // - repo info cache is not being used
+                    // - but when conditionally retrieving the repo info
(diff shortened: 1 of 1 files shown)


icechunk/src/asset_manager.rs
```diff
@@ -485,19 +502,32 @@ impl AssetManager {
...
+                    ));
+                } else {
+                    // It is very surprising to get to this branch:
+                    // - repo info cache is not being used
+                    // - but when conditionally retrieving the repo info
+                    //   without a version, which should fetch and return
+                    //   the repo info, we got back a result saying it
+                    //   didn't change!
+                    // This is very much panic or unreachable territory...
+                    unreachable!()
```
li-em, line 529: @paraseba I can make a proper error here, instead of forcing a panic

paraseba: I think this is fine, the argument is convincing

Issue #1689, opened 2026-02-19
object store error Operation `put_opts` with `opts.attributes` specified not yet implemented by LocalFileSystem


jackiryan:
### What happened?

My issue is similar to this one but I could not implement the workaround since I am using local_filesystem_storage: https://github.com/earth-mover/icechunk/issues/1431

I am attempting to process a set of repos containing virtual chunks, essentially the same NEX-GDDP-CMIP6 data on S3 referenced in [this example from the virtualizarr docs](https://virtualizarr.readthedocs.io/en/stable/usage.html).

To mitigate the intermittent network issues, I attempted to add retry settings to the RepositoryConfig:
```python
        storage = icechunk.local_filesystem_storage(vds_name)
        config = icechunk.RepositoryConfig(
            storage=icechunk.StorageSettings(
                retries=icechunk.StorageRetriesSettings(
                    max_tries=10,
                    initial_backoff_ms=500,
                    max_backoff_ms=5000,
                ),
            )
        )
        config.set_virtual_chunk_container(
            icechunk.VirtualChunkContainer(f"{bucket_url}/", icechunk.s3_store(region="us-west-2"))
        )
        credentials = icechunk.containers_credentials(
            {bucket_url: icechunk.s3_credentials(anonymous=True)}
        )
```

This works great when I open a readonly session to an existing icechunk store, however since I am using local filesystem storage for my initial buildout, I get an error when creating a new one:
```python
# This works
repo = icechunk.Repository.open(storage, config, credentials)
# This throws an error
repo = icechunk.Repository.open_or_create(storage, config, credentials)
```

```
    PyRepository.open_or_create(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        storage,
        ^^^^^^^^
        config=config,
        ^^^^^^^^^^^^^^
        authorize_virtual_chunk_access=authorize_virtual_chunk_access,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
icechunk.IcechunkError:   × object store error Operation `put_opts` with `opts.attributes` specified not yet implemented by LocalFileSystem(file:///foo/bar).
  │ 
  │ context:
  │    0: icechunk::storage::object_store::update_config
  │            with previous_version=VersionInfo { etag: None, generation: None }
  │              at icechunk/src/storage/object_store.rs:320
  │    1: icechunk::repository::store_config
[...]

PR #1699, opened 2026-02-20
FIX: don't override all storage defaults when user passes a config

ianhi:
Fixes: https://github.com/earth-mover/icechunk/issues/1689 (will need a backport to IC1)

The core bug was that if the user passed any storage config we would throw away all storage defaults for that backend, even if the user hadn't overridden them. This PR creates has us create default config with the correct storage backend defaults and then uses that for the merge.

Not super confident on if this was the correct way to structure the rust syntax as the fix. I went through a few iterations trying to make clippy happy and ended up with this. But happy to change.

Will try to write a test as well but wanted some more knowledgable rust person's feedback on the clones and expect calls if they are done properly

### DIFF
--- a/icechunk/benches/manifest.rs
+++ b/icechunk/benches/manifest.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::unwrap_used)]
 // NOTE: All expensive setup must go INSIDE the bench_with_input / iter_custom closure,
 // not before it. Criterion's filter (the CLI argument) is only checked inside
 // bench_with_input. Any work done before that call runs unconditionally for every
@@ -355,6 +356,7 @@ fn benchmark_commit_split_manifests(c: &mut Criterion) {
 
 /// Benchmarks committing half a million chunks
 /// by sequential appending such that each commit is a new manifest.
+/// The sequential repeated commit really stresses `get_existing_node` and flatbuffer decoding.
 fn benchmark_append_split_manifests(c: &mut Criterion) {
     let mut group = c.benchmark_group("append_split_manifests");
     group.sample_size(10).sampling_mode(criterion::SamplingMode::Flat);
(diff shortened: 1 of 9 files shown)


icechunk/src/config.rs
```diff
@@ -452,6 +452,16 @@ pub const DEFAULT_MAX_CONCURRENT_REQUESTS: u16 = 256;
 pub const DEFAULT_NUM_UPDATES_PER_REPO_INFO_FILE: u16 = 100;
 
 impl RepositoryConfig {
+    /// Merge backend storage defaults into this config. The backend defaults
+    /// become the base, with any existing storage settings overriding them.
+    pub fn with_storage_defaults(mut self, defaults: storage::Settings) -> Self {
+        self.storage = Some(match self.storage {
+            Some(s) => defaults.merge(s),
```
paraseba: shouldn't this line be `s.merge(defaults)`? We want `s` to take precedence, don't we?

icechunk/src/config.rs
```diff
@@ -452,6 +452,16 @@ pub const DEFAULT_MAX_CONCURRENT_REQUESTS: u16 = 256;
 pub const DEFAULT_NUM_UPDATES_PER_REPO_INFO_FILE: u16 = 100;
 
 impl RepositoryConfig {
+    /// Merge backend storage defaults into this config. The backend defaults
+    /// become the base, with any existing storage settings overriding them.
+    pub fn with_storage_defaults(mut self, defaults: storage::Settings) -> Self {
```
paraseba: This is a weird signature. It's sometimes used for builders, but I don't think we use this pattern anywhere for config. Why not just take `&mut self`? Or even just merge with the defaults you want?

icechunk/src/repository.rs
```diff
@@ -221,13 +221,13 @@ impl Repository {
...
-            config.map(|c| RepositoryConfig::default().merge(c)).unwrap_or_default();
-        let storage_settings = match config.storage() {
-            Some(s) => s.clone(),
-            None => storage.default_settings().await?,
-        };
+        // Merge the given config with the defaults, including backend storage defaults
+        let default_config = RepositoryConfig::default()
+            .with_storage_defaults(storage.default_settings().await?);
+        let config = config.map(|c| default_config.merge(c)).unwrap_or(default_config);
+        #[allow(clippy::expect_used)]
```
paraseba: this is very hard to follow, and it requires an `expect`.

I feel the ultimate problem here is that in the original code `RepositoryConfig::default` is non the right default. We need the default that is "best" for the given storage instead. Then `merge` takes care of everything else I think?

icechunk/src/config.rs
ianhi: I'm pretty sure that as currently written in the config `default.merge(user)`is the correct way to prefer the user. see:

https://github.com/earth-mover/icechunk/blob/fdb9ac18de11bc2d528c5f7ecd3074f301ea6904/icechunk/src/storage/mod.rs#L311-L316

and the last line of the match calls this which will prefer theirs (user) over mine 

https://github.com/earth-mover/icechunk/blob/fdb9ac18de11bc2d528c5f7ecd3074f301ea6904/icechunk/src/storage/mod.rs#L223-L225

ianhi: otherwise this wouldn't have fixed the bug

ianhi: Maybe that should be changed? but it would have to be changed everywhere to be consistent.

icechunk/src/repository.rs
ianhi: I wrote a new RepositoryConfig::default_with_storage method which I think solves this properly?

ianhi: also switched to `ok_or` from expect. I don't think we can escape rust telling us that the storage might not be some so at elast one is necessary

ianhi:
I also had to update the tests to account for this new behavior. the most confusing change is adding in an explicit max concurrent =1 to the manifest split test, with these fixes it was getting the backend default of 5, instead of the fallback of 1 which was breaaking the asserts.

icechunk/src/repository.rs
```diff
@@ -222,13 +224,16 @@ impl Repository {
...
+        // Merge two layers of config (In order of preference):
+        //   - User-provided config (passed to open())
+        //   - Backend storage defaults (e.g. S3 retry/concurrency settings)
+        let default_config =
+            RepositoryConfig::default_with_storage(storage.default_settings().await?);
+        let config = config.map(|c| default_config.merge(c)).unwrap_or(default_config);
+        let storage_settings = config
+            .storage()
+            .cloned()
+            .ok_or(RepositoryErrorKind::StorageSettingsMissing)?;
```
paraseba: Instead of introducing a new public error that should never be thrown I'd recommend an `expect`. But I think the core of the problem is your helper funtion `default_with_storage`, it doesn't "prove" to the compiler that it's returning with an storage. I think if you get rid of the function you'll get equally clear code without needed an expect.

ianhi: > it doesn't "prove" to the compiler that it's returning with an storage. 

How should I approach proving this to the compiler? It seems like there is always an option somewhere

paraseba: It's tricky, I'd have to think about it. Easiest is of course to get rid of the function because whit what it returns it's "too late". There could be some very advanced tricks, where you make your object parametric on what is or is not null, but definitely not worth going into those complications.

ianhi: Ok in collaboration with claude (and me carefully goingg through each line for explanations) came up with an explicit merge of storage settings and wrapping in Somes. code mostly by claude, names and comments by me

paraseba (approved):

ianhi:
should we backport this?

PR #1706, opened 2026-02-23
Amend checks when moves are involved

### LINKED ISSUE #1665: We cannot amend a rearrangement session with a normal session
Currently we don't support calling amend to "merge" a move session with a normal session. This can be seen in a `FIXME` in transaction_log.rs: `// FIXME: verify no moves in origins`.

We need to add checks for this, make sure it never happens, and fail nicely. Optionally amending a move with another move could be fine, but support for this would be optional for now I think

li-em:
Fix #1665

Move the previous transaction log check to the beginning of `do_flush`, avoid doing all the work before failing if amending with regular and rearrange sessions.

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -4447,13 +4447,78 @@ mod tests {
         let err = result.unwrap_err();
         assert!(matches!(err.kind, SessionErrorKind::RearrangeSessionOnly));
 
-        // Rearrange session, amend on top of a move commit: should fail?
+        // Rearrange session, amend on top of a move commit: should be fine
         let mut session = repo.rearrange_session("main").await?;
         session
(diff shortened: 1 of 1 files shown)


icechunk/src/session.rs
```diff
@@ -4389,6 +4419,45 @@ mod tests {
...
+        let mut session = repo.writable_session("main").await?;
+        session
+            .add_group("/fail".try_into().unwrap(), Bytes::copy_from_slice(b""))
+            .await?;
+        let result = session.amend("amend after move", None, false).await;
+        assert!(result.is_err());
+        let err = result.unwrap_err();
+        assert!(matches!(err.kind, SessionErrorKind::RearrangeSessionOnly));
+
+        // Rearrange session, amend on top of a move commit: should fail?
```
```suggestion
        // Rearrange session, amend on top of a move commit
```

PR #1708, opened 2026-02-24
Set parent of GC-ed snapshots to `INITIAL_SNAPSHOT_ID` when applicable

dcherian:
Replicates a bug detected by stateful testing.

a. Imagine a sequence of commits: 1, 2, 3, 4, 5
b. Reset branch from (5) to (2)
c. GC commits older than (4)
d. At this point (3) is deleted but (4) and (5) are preserved. 
e. The parent of (4) has not been rewritten to None, but is still (3). Any attempt to trace ancestry from (5) or (4) back will fail.
f. Next the repo_info creation will end up tracing that ancestry and fail with SnapshotNotFound

### DIFF
--- a/icechunk-python/python/icechunk/__init__.py
+++ b/icechunk-python/python/icechunk/__init__.py
@@ -255,7 +255,10 @@ def to_dict(config: ManifestSplittingConfig) -> SplitSizesDict:
 
 
 def upgrade_icechunk_repository(
-    repo: Repository, *, dry_run: bool = True, delete_unused_v1_files: bool = True
+    repo: Repository,
+    *,
+    dry_run: bool,
+    delete_unused_v1_files: bool = True,
 ) -> None:
     """
     Migrate a repository to the latest version of Icechunk.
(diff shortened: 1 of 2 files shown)


icechunk/src/ops/gc.rs
```diff
@@ -433,7 +434,20 @@ async fn delete_snapshots_from_repo_info(
...
             .all_snapshots()?
-            .filter_ok(|si| keep_snapshots.contains(&si.id))
+            .filter_map_ok(|si| {
+                if keep_snapshots.contains(&si.id) {
+                    match &si.parent_id {
+                        Some(parent) if keep_snapshots.contains(parent) => Some(si),
+                        // break the parent if it is going to be GC-ed
+                        // this is necessary for the case where history was edited with reset_branch
+                        // See test_gc.rs::test_gc_reset_branch for an example
+                        Some(_) => Some(SnapshotInfo { parent_id: None, ..si }),
```
dcherian: This isn't right is it? Because `expire` will detect a new parent using `root_to_snaps` and `new_parent`

icechunk/src/ops/gc.rs
```diff
@@ -430,10 +431,31 @@ async fn delete_snapshots_from_repo_info(
...
+                    match &si.parent_id {
+                        Some(parent) if keep_snapshots.contains(parent) => Some(si),
+                        // rewrite the parent if it is going to be GC-ed
+                        // this is necessary for the case where history was edited with reset_branch
+                        // See test_gc.rs::test_gc_reset_branch for an example
+                        // Note if a commit was left dnagling by expire it's parent will already have been rewritten
+                        // So here, we can either set None or INITIAL_SNAPSHOT_ID
+                        // We choose to set INITIAL_SNAPSHOT_ID until we consistently support
+                        // anonymous snapshots throughout the codebase.
+                        Some(_) => Some(SnapshotInfo {
```
paraseba, line 450: Oh setting to INITIAL is a great idea!

PR #1711, opened 2026-02-24
ListBucket permissions testing

### LINKED ISSUE #1680: We should be able to open an IC2 repo without ListBucket permissions
today it fails because we are doing a check against IC1 repos, but we can ignore errors if it turns out to be IC2

li-em:
Fix #1680 

Add a new `rustfs_init` container to set up RustFS options not available through configuration:
- create two users, `readonly` and `modify`
- put a policy in `testbucket` to only allow `GetObject` (no `ListBucket`) for `readonly` user, and `{Get,Put,Delete}Object` plus `ListBucket` for `modify` user

New users are necessary because the default one (`minio123`) is a super user and skips all policy checks in RustFS...

Add a new argument `check_clean_repo` (default: `true`) to `Repository::create`, so we can give advanced control when permissions are more restricted (need `{Get,Put,Delete}Object`, but no `ListBucket`, for example).

On the testing side, define a `Permission` enum both in Rust and Python, and avoid using `minio123` directly in tests, instead opting for retrieving a set of access and secret key for each test, with `ReadOnly` and `Modify` permissions matching the new users and the expected level of permissions for the test.

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -41,7 +41,7 @@ jobs:
           manylinux: auto
 
       - name: Upload wheels
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: test-wheels
           path: icechunk-python/dist
@@ -69,7 +69,7 @@ jobs:
           docker compose up -d rustfs_init
 
       - name: Download wheels
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: test-wheels
           path: icechunk-python/dist
@@ -146,7 +146,7 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Download wheels
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: test-wheels
           path: icechunk-python/dist
@@ -205,7 +205,7 @@ jobs:
           docker compose up -d rustfs_init
 
       - name: Download wheels
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: test-wheels
           path: icechunk/icechunk-python/dist
@@ -290,7 +290,7 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Download wheels
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: test-wheels
           path: icechunk-python/dist
--- a/.github/workflows/python-ci.yaml
+++ b/.github/workflows/python-ci.yaml
@@ -174,7 +174,7 @@ jobs:
           args: --release --out dist --find-interpreter
           manylinux: ${{ matrix.platform.manylinux }} # https://github.com/PyO3/maturin-action/issues/245
       - name: Upload wheels
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: wheels-linux-${{ matrix.branch == 'main' && 'main' || 'support-v1.x' }}-${{ matrix.platform.target }}
           path: icechunk-python/dist
@@ -216,7 +216,7 @@ jobs:
           args: --release --out dist --find-interpreter
           manylinux: musllinux_1_2
       - name: Upload wheels
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: wheels-musllinux-${{ matrix.branch == 'main' && 'main' || 'support-v1.x' }}-${{ matrix.platform.target }}
           path: icechunk-python/dist
@@ -253,7 +253,7 @@ jobs:
           target: ${{ matrix.platform.target }}
           args: --release --out dist --find-interpreter
       - name: Upload wheels
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: wheels-windows-${{ matrix.branch == 'main' && 'main' || 'support-v1.x' }}-${{ matrix.platform.target }}
           path: icechunk-python/dist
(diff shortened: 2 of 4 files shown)


compose.yaml
```diff
@@ -22,6 +28,54 @@ services:
...
+        restart: true
+    entrypoint:
+      - /bin/sh
+      - '-c'
+      - |
+        set -ex
+
+        rc alias set local http://rustfs:9000 minio123 minio123
+        rc admin user add local basic basicuser
+
```
paraseba, line 47: Executing this container make take a while, because it installs tools and stuff, do we need to modify the workflows that are currently waiting on RustFS to be up, to also wait until these keys are created?

icechunk-python/src/repository.rs
```diff
@@ -832,14 +832,15 @@ impl PyRepository {
 /// python threads can make progress in the case of an actual block
 impl PyRepository {
     #[classmethod]
-    #[pyo3(signature = (storage, *, config = None, authorize_virtual_chunk_access = None, spec_version = None))]
+    #[pyo3(signature = (storage, *, config = None, authorize_virtual_chunk_access = None, spec_version = None, check_clean_root = true))]
```
paraseba, line 835: I think we also need to modify repository.py? Or maybe you are leaving that for a future PR and relying on the default here. I kind of like, for now, to not expose that flag to users. Eventually we'll need it for python tests I guess.

icechunk-python/tests/test_session.py
```diff
@@ -192,3 +195,45 @@ def test_session_mode() -> None:
...
+    repo = Repository.create(storage=write_storage, config=config)
+    session = repo.writable_session("main")
+    store = session.store
+    group = zarr.group(store=store, overwrite=True)
+    air_temp = group.create_array("air_temp", shape=(1, 4), chunks=(1, 1), dtype="i4")
+    air_temp[0, 2] = 42
+    session.commit("init")
+
+    # Opening the repo with a storage without ListBucket permissions
+    repo = Repository.open(storage=readonly_storage, config=config)
```
paraseba, line 249: Just to convince me that this works, can you add to the test:
* create another repo with spec_version = 1
* open it with the credentials without list
* verify that it fails

icechunk/src/repository.rs
```diff
@@ -448,8 +449,14 @@ impl Repository {
...
             Self::open(config, storage, authorize_virtual_chunk_access).await
         } else {
-            Self::create(config, storage, authorize_virtual_chunk_access, create_version)
-                .await
+            Self::create(
+                config,
+                storage,
+                authorize_virtual_chunk_access,
+                create_version,
+                false,
```
paraseba: I think this should be `true`? Or even better, add the flag to this method too and forward it to `create`

icechunk/src/session.rs
```diff
@@ -4858,7 +4863,8 @@ mod tests {
         let in_mem_storage = new_in_memory_storage().await?;
         let storage: Arc<dyn Storage + Send + Sync> = in_mem_storage.clone();
         let repo =
-            Repository::create(None, Arc::clone(&storage), HashMap::new(), None).await?;
+            Repository::create(None, Arc::clone(&storage), HashMap::new(), None, true)
+                .await?;
```
paraseba, line 4867: Let's add a test that does this:
* create a repo with check_clean_root = true
* try to create the repo again in the same prefix with `check_clean_true = true`, show that it fails
* try to create the repo again in the same prefix with `check_clean_true = false`, show that it succeeds (even if it breaks the previous repo)

icechunk/src/strategies.rs
```diff
@@ -79,7 +79,7 @@ prop_compose! {
 
     runtime.block_on(async {
         let storage = new_in_memory_storage().await.expect("Cannot create in memory storage");
-        let repository = Repository::create(None, storage, HashMap::new(), Some(version))
+        let repository = Repository::create(None, storage, HashMap::new(), Some(version), false)
```
paraseba: I think we want true here

icechunk/tests/common/mod.rs
```diff
@@ -9,9 +9,26 @@ use icechunk::{
     storage::{new_r2_storage, new_tigris_storage},
 };
 
+pub(crate) enum Permission {
```
paraseba, line 12: awesome!

icechunk-python/tests/test_session.py
li-em: Added another test for v1 repo. Repo can be opened without list, but `.list_branches()` fails for v1 (and works for v2, since it is read from repo info instead of listing)

icechunk/src/session.rs
li-em: Done in 1432a801, but the last one fails because it is trying to write a new repo info without having an updated local one. But second fails as expected

compose.yaml
li-em: good point, it is still working in CI but better to do a new check

li-em: replaced `curl --silent --fail http://rustfs:9000/health` with `docker compose ps --status exited --filter status==0 | grep rustfs`

icechunk/src/repository.rs
li-em: yup, updated!

icechunk-python/src/repository.rs
li-em: oops, I missed the change in repository.py. Added there and also for `open_or_create`

.github/workflows/publish-rust-library.yml
```diff
@@ -23,11 +23,11 @@ jobs:
       - uses: actions/checkout@v6
       - name: Stand up RustFS
         run: |
-          docker compose up -d rustfs
+          docker compose up -d rustfs_init
       - name: Wait for RustFS to be ready
         run: |
           for _ in {1..10}; do
-              if curl --silent --fail http://rustfs:9000/health; then
+              if docker compose ps --status exited --filter status==0 | grep rustfs ; then
```
paraseba, line 30: awesome trick

PR #1713, opened 2026-02-25
Track rebase attempts in snapshot metadata

### LINKED ISSUE #1505: Record number of rebase attempts
It's useful to diagnose the number of rebase attempts when using `commit_rebasing`.

We could do this by either having a protected `icechunk` namespace inside `properties`. Alternatively, we could accept another closure which accepts `num: u16` (the current rebase attempt number) and returns a JsonValue that gets merged with `properties`.

mpiannucci:
Records the number of rebase attempts into the snapshot metadata in the `icechunk` object: 

```json
{
  "icechunk": {
    "rebase_attempts": 1
  },
  "user_key": "user_value"
}
```

Currently this is only tracked when the user uses commit_while_rebasing. Should we also track other rebases in the stateful session?

Closes #1505

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -161,8 +161,12 @@ class Model:
     def __getitem__(self, key: str) -> Buffer:
         return cast(Buffer, self.store[key])
 
-    def upgrade(self) -> None:
-        self.num_updates += 1
+    def upgrade(self, dry_run: bool) -> None:
+        if not dry_run:
+            self.spec_version = 2
+            # The ops log starts fresh after migration from v1,
+            # so reset to 1 (just the RepoMigratedUpdate entry).
+            self.num_updates = 1
 
     @property
     def has_commits(self) -> bool:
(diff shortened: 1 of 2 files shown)


paraseba:
Let's ask @dcherian about tracking in other forms of commits

dcherian:
> Should we also track other rebases in the stateful session?

Oh you mean manual rebases? IDK. The user can handle that. My request was definitely for `commit_while_rebasing`.

icechunk-python/tests/test_conflicts.py
```diff
@@ -93,6 +93,9 @@ def test_rebase_no_conflicts(repo: icechunk.Repository) -> None:
     np.testing.assert_array_equal(array_c[:], 1)
     assert array_c.attrs["repo"] == 2
 
+    info = next(iter(repo.ancestry(branch="main")))
+    assert info.metadata["icechunk"] == {"rebase_attempts": 1}
```
dcherian: We'll need to protect the `icechunk` namespace then? And prevent users from setting it? IIRC there's a `__root` attribute we set on the first commit; perhaps that should go under the `icechunk` namespace too?

cc @paraseba

mpiannucci: For this PR, users can set it if they want, we will merge the properties

mpiannucci:
Should we move __root and splitting_config to the icechunk object? Should we name the interior something other than `icechunk`?

paraseba:
> Should we move __root and splitting_config to the icechunk object? 
Yeah, I think we should have a single namespace. Maybe we call  it `__icechunk` ? I don't know.

so something like

```
"__icechunk": { is_root: false, splitting_config: ....}
```

What do you guys think?

I think I'm ok with renaming these metadata keys, don't imagine anybody is using them but our own code.

dcherian:
I like that; and i agree `splitting_config` should go in there

mpiannucci:
yes i like these suggestions. Will implement thank you!

PR #1714, opened 2026-02-25
fix bugs with upgrading the repo in stateful tests

ianhi:
Should fix the remaining issues @li-em was seeing

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -164,7 +164,9 @@ class Model:
     def upgrade(self, dry_run: bool) -> None:
         if not dry_run:
             self.spec_version = 2
-            self.num_updates += 1
+            # The ops log starts fresh after migration from v1,
+            # so reset to 1 (just the RepoMigratedUpdate entry).
+            self.num_updates = 1
 
     @property
     def has_commits(self) -> bool:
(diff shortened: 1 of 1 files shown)


icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -212,19 +212,21 @@ def commit_with_check(self, data: st.DataObject) -> None:
                 f"Expected: {get_expect.to_bytes()!r}"
             )
 
-    @rule()
+    @rule(delete_unused_v1_files=st.booleans())
```
dcherian:
```suggestion
    @rule(dry_run=st.booleans(), delete_unused_v1_files=st.booleans())
```

ianhi: what is the value of doing that in the tests? not sure what it tests?

ianhi: but you are right. i added. also found a further bug in the stateful repo ops

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -212,19 +212,22 @@ def commit_with_check(self, data: st.DataObject) -> None:
...
+        note(f"upgrading spec from 1 to 2 with {delete_unused_v1_files=}")
+        self.ic.upgrade_icechunk_repository(
+            self.repo, dry_run=dry_run, delete_unused_v1_files=delete_unused_v1_files
+        )
         # Reopen to pick up the upgraded spec version
         self.repo = self.actor.open(self.storage)
         self.store = self.repo.writable_session("main").store
-        self.model.spec_version = 2
+        if not dry_run:
+            self.model.spec_version = 2
```
dcherian, line 228:
```suggestion
            assert self.repo.spec_version == 2
            self.model.spec_version = 2
```

ianhi:
alarming that this slipped thorhugh till now. but thats randomness i guess. actually not so because i think we used to check based on the repo spec not the model

ianhi:
@dcherian can you check the last commit to make sure its ok

ianhi:
Nvm due to the power of auto merge

PR #1717, opened 2026-02-25
GC: Make sure "non pointed but newer" snapshots are accessible.

dcherian:
Previously, we could land in situations where a snapshot was preserved but a referenced manifest was deleted (it was older).

### DIFF
--- a/icechunk/src/ops/gc.rs
+++ b/icechunk/src/ops/gc.rs
@@ -260,15 +260,13 @@ async fn chunks_retained(
 pub async fn find_retained(
     asset_manager: Arc<AssetManager>,
     config: &GCConfig,
+    snaps: impl Stream<Item = RepositoryResult<Arc<Snapshot>>>,
 ) -> GCResult<(HashSet<ChunkId>, HashSet<ManifestId>, HashSet<SnapshotId>)> {
-    let all_snaps =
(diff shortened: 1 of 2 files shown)


icechunk/src/ops/gc.rs
```diff
@@ -366,7 +361,21 @@ pub async fn garbage_collect(
...
-    keep_snapshots.extend(non_pointed_but_new);
+    let all_snaps =
+        pointed_snapshots(Arc::clone(&asset_manager), &config.extra_roots).await?;
+    let am = Arc::clone(&asset_manager);
+    let non_pointed_snaps = stream::iter(non_pointed_but_new.into_iter().map(Ok))
+        .and_then(move |id| {
+            let am = Arc::clone(&am);
+            async move { am.fetch_snapshot(&id).await }
+        });
+
```
dcherian, line 372: This was Claude's take on solving a lifetime issue. The simple `stream.iter(non_pointed_but_new).iter()).then(|id| asset_manager.fetch_snapshot(id))` didn't work

paraseba: yeah, that looks about right

paraseba: yeah, that looks about right

PR #1722, opened 2026-02-26
Add C API and Julia bindings for Zarr Store

asinghvi17:
Hey!  A while ago I had opened https://github.com/earth-mover/icechunk/issues/381 to ask for a C compatible API.  I threw Claude at it now and this is what it generated.  I'm happy to split the Julia port out into a separate PR if you feel that makes things easier.

The idea here is that there is an `icechunk-c` crate which exposes a C API for icechunk, which is then wrapped by the `icechunk-jl` package.  In practice, if this is merged we will have an `Icechunk_jll` in [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil), which is Julia's binary building infrastructure. An `Icechunk.jl` would depend on `Icechunk_jll` instead, though we could override that locally for CI purposes in this repo.

I had this implemented in a blocking way for the C API since that seemed simplest, but happy to pursue another path if that's what folks here would want.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/icechunk-c/build.rs
+++ b/icechunk-c/build.rs
@@ -8,11 +8,7 @@ fn main() {
 
     let output_path = std::path::Path::new(&crate_dir).join("include").join("icechunk.h");
 
-    match cbindgen::Builder::new()
-        .with_crate(&crate_dir)
-        .with_config(config)
-        .generate()
-    {
+    match cbindgen::Builder::new().with_crate(&crate_dir).with_config(config).generate() {
         Ok(bindings) => {
             bindings.write_to_file(&output_path);
         }


rabernat:
Hi @asinghvi17 - apologies for the long silence on this PR. 

This PR generated a lot of excitement and discussion from our team. We deeply want to support the Julia community, and having C bindings unlocks some much more than just that. Like you, we are also very excited about the potential of AI tools to accelerate development.

However, as you know, the rapid growth of agents poses complex and nuanced questions for the open source community. There are many benefits AND also many concerns.

Rather than just looking at this PR in isolation, we decided it was best for us to actually come up with a policy. Here it is: https://icechunk.io/en/latest/ai-policy/

You can read the full policy, but the core principle is pretty straighforward:

> If you submit a pull request, you are responsible for understanding and having fully reviewed the changes. You must be able to explain why each change is correct and how it fits into the project. Strive to minimize changes to ease the burden on reviewers...

Your description of how this PR was created:

> I threw Claude at it now and this is what it generated.

...does not create the impression that you fully understand what is happening here. Therefore, the burden is shifted to our team in terms of reviewing and understanding these 2000 lines of code.

We have an extremely high quality bar for this project. We're committed to building something that organizations like NASA can depend on for critical, production work. We stand behind every line of code and every paragraph of the docs. Accepting this PR in its current form wouldn't allow us to maintain that same standard. (Not because the code is bad! Just because we can't _actively affirm_ that it's good.)

However, as mentioned above, we really like what this PR unlocks, and we want to support it. There are basically two paths going forward.

1. You commit to fully understanding the Claude-generated code, work with other devs to facilitate a review process that is not overly-burdensome, and you join the Icechunk dev team as the owner and maintainer of this part of the code base. Or...
2. For now, we simply move this project into another repo as a standalone project, in line with its more experimental status.

Doing 2 now does not preclude doing 1 later.

If we choose to go with 2, we would be happy to host the project via a standalone repo as part of the Earthmover org and bless this as an "official" Icechunk project.

We're very grateful for your work and we want to collaborate. 🙏  Hopefully our position makes sense to you.

PR #1724, opened 2026-02-26
remove (re)opens in actor switching

ianhi:
We already have reopening logic on the super classes. I wanted to remove the (re)open in the actor switch as that potentially does some re-init which might mask a subtle bug. This way its a bit more like two concurrent actors who have both opened the repo and are doing operations to it.

### DIFF
--- a/icechunk-python/tests/test_stateful_compat.py
+++ b/icechunk-python/tests/test_stateful_compat.py
@@ -12,9 +12,11 @@ Skipped entirely if icechunk_v1 is not installed.
 """
 
 import tempfile
+import types
+from dataclasses import dataclass
+from typing import Any
 
 import pytest
-from packaging.version import Version
 
 import icechunk as ic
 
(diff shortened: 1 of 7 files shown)


icechunk-python/tests/test_stateful_compat.py
```diff
@@ -131,19 +145,28 @@ def init_store(self, spec_version: int, data: st.DataObject) -> None:
...
         super().init_store(spec_version=spec_version)
+        self.actor_name = actor_name
+        self.repos[actor_name] = self.repo
 
     @rule(data=st.data())
     def switch_actor(self, data: st.DataObject) -> None:
+        # Draw an actor and use their storage object
+        actor_name = data.draw(st.sampled_from(list(self.actors.keys())))
+        if actor_name == self.actor_name:
+            return
```
dcherian:
```suggestion
        actor_name = data.draw(st.sampled_from(list(k for k in self.actors.keys() if k!=self.actor_name)))
```

let's just filter it out

ianhi: 👍

ianhi:
I actually havve a branch on top of this that goes way farther, and starts allowing for simple rebases with an aribitrary number of actors working in parallel on a repo. so might just swap to that

Issue #1648, opened 2026-02-13
Planning Issue: Replacing pre-commit with pixi


jbusecke:
@li-em and I are planning to get together next week and refactor the CI + docs with regard to code quality control. 

Our goal is to streamline the complex install/configuration steps for local development. Currently planning to refactor things with pixi. Ill update this issue as we go.

jbusecke:
Reminder to remove this sort of instruction:
````
    Now you can run tests and other commands:

    ```bash
    # Run tests (Rust changes will automatically trigger incremental rebuild)
    uv run pytest

    # Run type checking
    uv run mypy python tests

    # Run linting
    uv run ruff check python
    ````
````

from the install to the testing section (it was confusing to me that this was only given in the uv install instructions).

**Comment by jbusecke** [comment]
Also [this](https://github.com/earth-mover/icechunk/blob/7c017cdb842045b9885962099e5eba77a0fdc1a8/icechunk-python/docs/docs/contributing.md?plain=1#L244) is a typo

Should be 

```
!!! important

    The full ~~Python~~ Rust test suite depends on S3 and Azure compatible object stores. See [here](#docker-setup-for-local-storage-testing) for detailed instructions.
```

abkfenris:
I've got a branch hanging around for building the docs with Pixi from a little while ago if that would be useful.

jbusecke:
Yeah that would probably save us some time @abkfenris, but just if we can find it haha.

abkfenris:
@jbusecke an initial pass with docs and some test support is #1661 to get you started.

jbusecke:
Nice! Ill take a look together with @li-em tomorrow.

li-em:
Initial support merged in #1674 , follow-ups:
- [ ] https://github.com/earth-mover/icechunk/pull/1727
- [ ] move other parts of the CI to run through pixi/just (I can hold off on that until after IC2)
- [ ] Work through the contributor guide beyond python.
- [ ] Tweak rust cache usage in code quality CI check 
- [ ] #1698 
- [ ] move most `pre-commit` checks to call `just` instead: https://github.com/earth-mover/icechunk/pull/1727#issuecomment-3969066747

PR #1727, opened 2026-02-26
Bump ruff to newest version and resulting changes

jbusecke:
The ruff version was quite old `0.8.0` vs `0.15.3`. I bumped the version and needed a few small changes resulting from the new version. 

Let me know if these changes are not desired and I can add `RUF059` to the ignore list.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
       - id: mixed-line-ending
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.8.0'
+    rev: 'v0.15.3'
     hooks:
       - id: ruff-format
       - id: ruff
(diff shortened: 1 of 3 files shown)


jbusecke:
I think @paraseba has pointed out this exact error before, so for now I assume its upstream of this change.

dcherian (approved):

jbusecke:
Ok @li-em actually just pointed out that we still have [ruff in pre-commit](https://github.com/earth-mover/icechunk/blob/74b20b196f4cfff30a22e353b29717388cf6430c/.pre-commit-config.yaml#L12-L17). I missed that in #1674, apologies. 

The situation is quite confusing at the moment. We have two sets of checks here:
**pre-commit exclusive**
- codespell, whitespace etc

**just wrapped in pre-commit**
- all the rust stuff is already just a local call to just, we don't really need PC at all here its just a convienience for people who like to have checks run at commit time

Since we do not wan't or can't implement everything in PC, I propose we go the other way around with as many of the checks as possible: Implement them as a just formula, and call that formula from pre-commit. 
We could go even further and ditch PC entirely if we can replace the 'builtin' hooks with equivalent checks. 

Curious to hear what folks think.

li-em:
At the moment it is more pressing to fix the `ruff` version mismatch, because it is annoying to get CI failures because it uses a different version.

I added @jbusecke to https://github.com/earth-mover/icechunk/issues/1648#issuecomment-3968760700 to continue the conversation

PR #1729, opened 2026-02-27
Resolve the issue of concurrent changes during GC

paraseba:
This change accounts for the case where the repo is updated during a GC run.

- [x] Do the same for expiration

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -272,6 +272,11 @@ class Model:
     def get(self, key: str) -> Buffer | None:
         return self.store.get(key)
 
+    def delete_doc(self, key: str) -> None:
+        if key in self.store:
+            del self.store[key]
+            self.changes_made = True
+
     def upgrade(self, dry_run: bool) -> None:
         if not dry_run:
             # only reachable snapshots are migrated over
@@ -546,12 +551,10 @@ class VersionControlStateMachine(RuleBasedStateMachine):
         # instead we explicitly add a rule to make a lot of commits
         commit_ids = []
         for should_amend in amend_or_commit:
-            # annoyingly we need to ensure we overwrite with different data, otherwise icechunk doesn't like it.
-            path, meta = data.draw(
-                st.tuples(metadata_paths, v3_array_metadata).filter(
-                    lambda pm: self.model.get(pm[0]) != pm[1]
-                )
-            )
+            # We need to ensure we overwrite with different data, otherwise icechunk doesn't register a change.
+            # Delete first then set, to avoid slow hypothesis filter rejection.
+            path, meta = data.draw(st.tuples(metadata_paths, v3_array_metadata))
+            self.delete_doc(path)
             self.set_doc(path, meta)
             msg = data.draw(st.text(max_size=MAX_TEXT_SIZE))
             if (
@@ -574,6 +577,12 @@ class VersionControlStateMachine(RuleBasedStateMachine):
             # not at branch head, modifications not possible.
             with pytest.raises(IcechunkError, match="read-only store"):
                 self.sync_store.set(path, value)
+
+    def delete_doc(self, path: str) -> None:
+        note(f"deleting path {path!r}")
+        if self.model.branch is not None:
+            self.sync_store.delete(path)
+            self.model.delete_doc(path)
 
     @precondition(lambda self: self.model.spec_version >= 2)
     @rule(meta=simple_attrs)


paraseba:
I don't know how to test this. Our options are, 
* we don't test this concurrent case, and reviewers look really hard at it and convince themselves it works.
* we spend a good amount of time to introduce tools that would allow to test this
* we write a best effort test that hopefully does writes concurrently while we run a GC, but for that we need to find a way to make GC run for a while

icechunk/src/ops/gc.rs
```diff
@@ -323,6 +324,65 @@ pub async fn garbage_collect(
...
+                        RepositoryErrorKind::RepoUpdateAttemptsLimit(max_attempts as u64)
+                            .into(),
+                    ));
+                }
+            },
+            Err(err) => {
+                return Err(err);
+            }
+        }
+    }
```
dcherian: This could be `garbage_collect_one_attempt.retry(backoff).map_err(|e| GCError::... .into())?`? https://docs.rs/backon/latest/backon/struct.Retry.html

icechunk/src/ops/gc.rs
```diff
@@ -430,24 +493,45 @@ pub async fn garbage_collect(
     Ok(summary)
 }
 
+/// Updates the repo object eliminating snapshots
+/// Returns true if the operation was successful, if it returns false, GC should be retried
```
dcherian:
```suggestion
/// Returns Ok() if the operation was successful, if it returns Err(), GC should be retried
```

?

icechunk/src/ops/gc.rs
```diff
@@ -456,25 +540,57 @@ async fn delete_snapshots_from_repo_info(
...
+                    //
+                    // we don't need to add to final_snaps, we are dropping it
+                }
+                (false, false) => {
+                    // this is a new snapshot
+                    if let Some(parent) = &si.parent_id
+                        && drop_snapshots.contains(parent)
+                    {
+                        // this is a new snapshot created since we started GC
+                        // but we are traying to drop its parent. Case 2b
```
dcherian:
```suggestion
                        // but we are trying to drop its parent. Case 2b
```

icechunk/src/ops/gc.rs
```diff
@@ -456,25 +540,57 @@ async fn delete_snapshots_from_repo_info(
...
+                        return Err(RepositoryErrorKind::RepoInfoUpdated.into());
+                    } else {
+                        // a new snapshot with the root as parent,
+                        // root is always retained
```
dcherian: not just root no? If the parent is in keep_snapshots, we hit this branch, correct?

icechunk/src/ops/gc.rs
```diff
@@ -456,25 +540,57 @@ async fn delete_snapshots_from_repo_info(
...
+            }
+        }
+
+        // TODO: quite inefficient
+        let final_snap_ids: HashSet<_> = final_snaps.iter().map(|si| &si.id).collect();
+        for (_, pointed_snap) in repo_info.tags()?.chain(repo_info.branches()?) {
+            if !final_snap_ids.contains(&pointed_snap) {
+                return Err(RepositoryErrorKind::RepoInfoUpdated.into());
+            }
+        }
```
dcherian, line 574: IIUC we are ignoring an update where a tag is deleted and a snapshot can be GC-ed. But that's quite minor.

icechunk/src/ops/gc.rs
```diff
@@ -490,7 +606,11 @@ async fn delete_snapshots_from_repo_info(
         Ok(Arc::new(new_repo_info))
     };
 
-    let _ = asset_manager.update_repo_info(retry_settings, do_update).await?;
+    let retry_settings = storage::RetriesSettings {
+        max_tries: Some(NonZeroU16::MIN),
```
dcherian, line 599: so `1`? 😆

icechunk/src/ops/gc.rs
```diff
@@ -690,6 +810,10 @@ pub async fn expire(
     }
 }
 
+/// Since expire_v2 is a relatively fast operation (repo object only) we retry it if the repo info
+/// object was modified since it started
+/// Since expire_v2 is a relatively fast operation (repo object only) we retry it if the repo info
+/// object was modified since it started
```
dcherian, line 803:
```suggestion
```

icechunk/src/ops/gc.rs
```diff
@@ -698,14 +822,78 @@ pub async fn expire_v2(
...
+        .with_max_delay(std::time::Duration::from_millis(
+            retry_config.max_backoff_ms() as u64
+        ))
+        .with_max_times(max_attempts)
+        .with_jitter()
+        .build();
+
+    let mut attempts: u64 = 1;
+    loop {
+        match expire_v2_one_attempt(
```
dcherian: same here, could probably do `expire_...().retry(backoff)...`

icechunk/src/ops/gc.rs
dcherian: But seriously, why not use retry_settings?

dcherian:
> we spend a good amount of time to introduce tools that would allow to test this

shuttle now supports tokio apparently but I haven't tried it: https://github.com/awslabs/shuttle/pull/238

icechunk/src/ops/gc.rs
paraseba: just changed it, so much better! I copied the code from the wrong place :facepalm:

icechunk/src/ops/gc.rs
paraseba: intentional, to prove it's not Claude

icechunk/src/ops/gc.rs
paraseba: ups, right, adjusting the comment

icechunk/src/ops/gc.rs
paraseba: exactly, there are a few cases like that that will have to be GC'ed in the next pass. I only want to make sure I don't delete something I shouldn't, extra garbage is fine.

icechunk/src/ops/gc.rs
paraseba: I'm retrying outside of this, this has to do a single attempt and fail immediately

PR #1730, opened 2026-02-27
Design doc for more IC2 format changes

paraseba:


design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+  location: string;
+
+  // NEW OPTIONAL FIELD
+  compressed_location: [uint8]
+}
+```
+
+Notice in the end, this will be zstd compression of zstd data, because the full
+manifest is compressed with zstd before writing it. We should evaluate if the
+parameters of this second compression should change.
```
TomNicholas, line 100: What is the point of the second compression once we already have the first compression? Shouldn't we try to just make sure the first compression does a decent job?

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+As an example, if the repo has a VCC like
+
+```yaml
+  s3:
+    name: my-virtual-icechunk
+    url_prefix: s3://testbucket/my-repo/chunks
+    store: !s3_compatible
+      region: us-east-1
+      anonymous: false
+```
```
TomNicholas, line 41: So looking at this VCC (and therefore looking at the whole repo config), there is no way for me to know if this repo contains any relative virtual refs? That seems like it might cause problems?

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+Authorization of VCCs will still be based on prefixes and not names, to avoid
+security issues. VCC names will be used as a way to refer to a VCC in manifests.
+So VCC names must be unique per repo (except when name is not present).
+
+To maintain compatibility, we still allow null names, but those VCCs without
+name cannot be used for relative virtual chunk resolution.
+
+### Reinterpretation of `ChunkRef::location`
+
+Now, the `location` field in `ChunkRef` can be a url with protocol `vcc://`.
```
TomNicholas, line 22: Is it a good idea to overload the meaning of protocol like this? What if I want a relative `file:///` ref? Or what if someone invents an AWS alternative called VCC and wants to use that protocol?

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Design
+
+### Virtual Chunk Container (VCC) names
+
+We re introduce virtual chunk container names. We had them at the beginning,
+but we drop them for security reasons. Most of the code still exists.
+Authorization of VCCs will still be based on prefixes and not names, to avoid
+security issues. VCC names will be used as a way to refer to a VCC in manifests.
+So VCC names must be unique per repo (except when name is not present).
```
TomNicholas, line 15: Is this already a requirement? If not it invalidates previously-written configs.

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+Hostnames in urls with protocol `vcc` must much the name of a VCC as defined in
+the repo configuration, and if it doesn't those relative chunk refs
+will generate an error when dereferenced.
+
+### Compression of `location` field
+
+The `location` field in `ChunkRef` will be compressed using
+[Zstandart dictionary compression](https://facebook.github.io/zstd/#small-data),
+globally across each whole manifest. Other compression options are available
+but Icechunk already uses zstd.
```
TomNicholas, line 56: IIUC using Zstd for this gives good byte-level compression but sacrifices the ability to do efficient (O(1)) random access to a ref.

Alternatives are treating the data as columnar and doing dictionary encoding on the values (which arrow does), or doing [FSST encoding](https://spiraldb.com/post/compressing-strings-with-fsst) on the values.

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+Beyond the format changes, we have to:
+
+- Compression, uncompression and configuration
+- Make virtual chunk resolution work for both absolute and relative locations
+- Make `Store::set_virtual_ref` use relative locations for IC2 repositories, if
+there is a matching VCC. Potentially we may want to add explicit flags to force
+use of relative or absolute urls only.
+- Make `Session.all_virtual_chunk_locations` work for absolute and relative
+locations
+- Review tests and error messages to use the right combination of relative and
```
TomNicholas, line 119: It might be hard to do this if we can't know whether a particular VCC is being used for any virtual refs until get-time

design-docs/014-virtual-chunk-ref-efficiency.md
TomNicholas: But OTOH maybe there's no need for per-element random access here since we're already deserializing the entire manifest

design-docs/014-virtual-chunk-ref-efficiency.md
paraseba: because with the new compression, we keep stuff compressed *in memory*, so each location stays compressed and we decompress it on demand when/if somebody asks for it.

design-docs/014-virtual-chunk-ref-efficiency.md
paraseba: Right, but that is no worse than what we have today. I expect we'll solve this with the new "extra" space also defined in this PR. Not implementing it now though.

design-docs/014-virtual-chunk-ref-efficiency.md
paraseba: I think a `file:///` would still work, you just need to define the VCC with `file:///`. Regarding new protocols ... i don't know, that may be an issue. It's just that I didn't want to change too much or use more extra space in the manifest. Other ideas?

design-docs/014-virtual-chunk-ref-efficiency.md
paraseba: Yes, this was a requirement at some point

design-docs/014-virtual-chunk-ref-efficiency.md
paraseba: we are compressing the locations one by one, with a shared dictionary (this is the same approach FSST takes). Chunks are never searched by location, only by index. But there is in fact a higher cost to get the location, we need to decompress it. My thinking is decompression time will be negligible when compared to actually fetching the chunk.

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+```flatbuffers
+table Manifest {
+  // existing field
+  id: ObjectId12 (required);
+
+  // existing field
+  arrays: [ArrayManifest] (required);
+
+  // NEW OPTIONAL FIELD
+  location_dictionary: [uint8]
```
paraseba: I should also store the compression algorithm here, in case we want to change it later

design-docs/014-virtual-chunk-ref-efficiency.md
```diff
@@ -0,0 +1,124 @@
...
+in manifest generation time is minimal in real world scenarios.
+
+Ideally compression parameters should be enabled on an array per array
+basis, as we do with preloading or splitting.
+
+To store the compression dictionary and the compressed location we add in
+the manifest flatbuffer:
+
+```flatbuffers
+table Manifest {
```
paraseba, line 75: should we create the compression dict for the whole manifest or each Array independently?

design-docs/013-feature-flags.md
```diff
@@ -0,0 +1,94 @@
...
+}
+```
+
+The full list of feature flag strings is hardcoded in the code. We introduce new
+API:
+
+```python
+
+class FeatureFlagStatus(Enum):
+  Enabled
```
paraseba: This should just be `Set` and `Unset`

design-docs/014-virtual-chunk-ref-efficiency.md
dcherian: I think a new protocol is fine.  Could call it `_vcc` to avoid an conflicts.

design-docs/014-virtual-chunk-ref-efficiency.md
dcherian: I expect whole manifest to be useful for real-world applications. storing an array per folder is pretty common.

design-docs/013-feature-flags.md
```diff
@@ -0,0 +1,94 @@
...
+  def feature_flags(self) -> list[FeatureFlag]:
+    ...
+
+  def update_feature_flag(self, feature_name: str, status: FeatureFlagStatus):
+    ...
+```
+
+Icechunk code has (something isomorphic to) a hardcoded list of `FeatureFlag`.
+
+The presence of a string in the `features` array means that the opposite of
```
dcherian: yikes, very confusing! Can we instead make each entry be either `enable-` or `disable-` . Also, boolean blindness... do we expect that `feature` will never be an Enum in future?

design-docs/015-extra-data-in-manifests-and-snapshots.md
```diff
@@ -0,0 +1,151 @@
...
+## Design
+
+We add extra information to the different objects using flexbuffers maps.
+Icechunk 2 already uses flexbuffers for serialization of repo and snapshot
+metadata.
+
+This document doesn't dictate what data can or should be added in the
+flexbuffer map.
+
+### Manifest references
```
dcherian, line 19: What about `ChunkRef` or do we plan to have chunk-specific "extra information" handled at the Zarr level?

design-docs/013-feature-flags.md
mpiannucci: Agree, this is very confusing

design-docs/015-extra-data-in-manifests-and-snapshots.md
```diff
@@ -0,0 +1,151 @@
+# Extra data in manifests and snapshots
+
+We have several candidates to add "extra information" to snapshots,
+manifests and manifest refs. We currently don't have a natural place to put this
+information. Examples are statistics, pre computed quantities, and others.
```
mpiannucci, line 5: I woudl love to understand more about this. Currently i dont quite understand the use case

design-docs/013-feature-flags.md
```diff
@@ -0,0 +1,94 @@
+# Feature flags
+
+We want to add to the format the ability to soft-block the repo from executing
```
mpiannucci, line 3: Is this basically a proxy for permissions? What good does it do to turn off things like amend or commit unless this is basically a permissions whitelist ala LLM agents

design-docs/015-extra-data-in-manifests-and-snapshots.md
TomNicholas: I would quite like it to be possible to add chunk-level information later, in case we ever need per-chunk codecs...

design-docs/015-extra-data-in-manifests-and-snapshots.md
TomNicholas: e.g. I want to know the `max` of an entire zarr array, if the `max` of every chunk were previously saved, this is then super cheap to calculate. These "summary statistics" is an approach used e.g. in Parquet.

PR #1731, opened 2026-02-27
Generate fake ops log during migration

### LINKED ISSUE #1693: Should migration create a "fake" ops_log?

mpiannucci:
Closes #1693 

The V1→V2 migration now generates a synthetic ops log by reconstructing commit, branch, tag, and deletion events from the snapshot graph, so downstream applications can see the repository's history trajectory

Assumptions we make to simplify things: 
* All branches and tags are created using the first snapshot in the ancestry of the snapshot they are pointing toward
* Deleted tags are included, but deleted branches cannot be

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -1069,18 +1069,11 @@ class VersionControlStateMachine(RuleBasedStateMachine):
             return
         actual_ops = list(self.repo.ops_log())
 
-        # ops should be ordered: strictly decreasing for normal repos,
-        # non-increasing after migration (synthetic entries can share timestamps)
-        if self.model.migrated:
-            assert all(
-                first.updated_at >= second.updated_at
-                for (first, second) in itertools.pairwise(actual_ops)
-            )
-        else:
-            assert all(
-                first.updated_at > second.updated_at
-                for (first, second) in itertools.pairwise(actual_ops)
-            )
+        # ops should be strictly decreasing in both normal and migrated repos
+        assert all(
+            first.updated_at > second.updated_at
+            for (first, second) in itertools.pairwise(actual_ops)
+        )
 
         if self.model.migrated:
             # The model tracks RepoMigratedUpdate + all post-migration ops.
--- a/icechunk/src/migrations/mod.rs
+++ b/icechunk/src/migrations/mod.rs
@@ -244,8 +244,20 @@ fn generate_migration_ops_log(
         ));
     }
 
-    // Sort chronologically, then reverse for newest-first order
+    // Sort chronologically (oldest first)
     entries.sort_by_key(|(_, time)| *time);
+
+    // Ensure strictly increasing timestamps: if an entry collides with or is
+    // before the previous one, bump it by 1μs.  This mirrors the real ops log
+    // where every operation gets a unique timestamp.
+    for i in 1..entries.len() {
+        let prev_time = entries[i - 1].1;
+        if entries[i].1 <= prev_time {
+            entries[i].1 = prev_time + TimeDelta::microseconds(1);
+        }
+    }
+
+    // Reverse for newest-first order
     entries.reverse();
     entries
 }
@@ -748,13 +760,13 @@ mod tests {
             "TagCreatedUpdate (idx {deleted_tag_created_idx}) should be newer than NewCommitUpdate (idx {commit_idx})"
         );
 
-        // Verify timestamps are non-increasing (newest first)
+        // Verify timestamps are strictly decreasing (newest first)
         for window in ops_log.windows(2) {
             let (time_a, _, _) = &window[0];
             let (time_b, _, _) = &window[1];
             assert!(
-                time_a >= time_b,
-                "ops log timestamps must be non-increasing: {time_a} should be >= {time_b}"
+                time_a > time_b,
+                "ops log timestamps must be strictly decreasing: {time_a} should be > {time_b}"
             );
         }
 


icechunk/src/migrations/mod.rs
```diff
@@ -223,16 +355,79 @@ pub async fn migrate_1_to_2(
...
+    );
+    info!("Generated {} ops log entries", ops_log.len());
+
+    let previous_updates: Vec<crate::format::IcechunkResult<_>> =
+        ops_log.into_iter().map(|(ut, dt)| Ok((ut, dt, None))).collect();
+
+    // Use a page size large enough to hold all synthetic entries (+1 for RepoMigratedUpdate)
+    // so nothing is truncated. There are no backup files to chain to, so overflow would
+    // silently drop entries. Subsequent operations will use the configured page size.
+    let migration_page_size = (previous_updates.len() as u16 + 1)
```
mpiannucci, line 446: Is this the behavior that we want? Or are we ok with truncating the log

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+/// Read a deleted tag's snapshot ID directly from V1 storage,
+/// bypassing the delete marker check in `fetch_tag`.
+async fn fetch_deleted_tag_snapshot_id(
+    repo: &Repository,
+    tag_name: &str,
+) -> Option<SnapshotId> {
+    let ref_path = format!("{V1_REFS_FILE_PATH}/tag.{tag_name}/ref.json");
+    match repo.storage().get_object(repo.storage_settings(), &ref_path, None).await {
+        Ok((mut reader, ..)) => {
+            let mut data = Vec::with_capacity(1_024);
```
paraseba: Not that it matters, but I think it will always be 35 bytes, so you can make it 40 instead of 1024

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+        }
+        Err(_) => None,
+    }
+}
+
+/// Generate a synthetic ops log from V1 repository data.
+///
+/// Reconstructs a plausible history of operations from the snapshot graph,
+/// branch/tag state. Uses snapshot `flushed_at` timestamps for commit events.
+/// Tag/branch creation events use the `flushed_at` of the snapshot they point to.
```
paraseba, line 150: Just in case, why don't we make the tag/branch events 1 millisecond later?

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -1063,14 +1066,25 @@ def check_ops_log(self) -> None:
...
+            # non-None backup paths must be unique
+            all_backups = [
+                op.backup_path for op in actual_ops if op.backup_path is not None
+            ]
+            assert len(all_backups) == len(set(all_backups))
+
+            assert self.model.ops_log[::-1] == actual_ops
+            assert isinstance(
+                actual_ops[-1], ic.RepoInitializedUpdate | ic.RepoMigratedUpdate
+            )
```
dcherian: I think these need to be updated
```suggestion
    assert self.model.ops_log[::-1] == actual_ops
    assert isinstance(actual_ops[-1], ic.RepoInitializedUpdate)
```

icechunk/src/format/repo_info.rs
```diff
@@ -361,11 +361,12 @@ impl RepoInfo {
                 (ut, dt, backup_path)
             })
         });
-        // assert backup path is only present when there are previous updates
+        // assert backup path is only present when there are previous updates,
+        // unless backup_path is None (e.g. during migration when creating a fresh
+        // repo info with a synthetic ops log and no previous files to reference).
         assert!(
-            last_update.is_none() && backup_path.is_none()
```
dcherian, line 366: not sure why this had to change; and the comment is confusing to me. `backup_path` is also None for the RepoInitializedUpdate?

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+            },
+            snap.flushed_at,
+        ));
+    }
+
+    // Followed by a BranchCreatedUpdate + NewCommitUpdate for each unique snapshot
+    // in the ancestry of each branch. BranchCreatedUpdate is timestamped to the
+    // root snapshot in the branch's ancestry (the first snapshot).
+    // Snapshots shared across branches are attributed to the first branch processed.
+    for (name, ancestry) in branches {
```
paraseba, line 201: I think real repos don't have a BranchCreated update for "main". Are we excluding it here?

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+        let branch_root_time = ancestry.last().map(|s| s.flushed_at).unwrap_or(root_time);
```
paraseba: the problem is all branches are going to have the time of the root snapshot. But ... I guess we don't care? Let's at least separate them so they are not exactly = to the snapshot.

icechunk/src/migrations/mod.rs
dcherian: Can't we stream out batches of size `num_updates_per_repo_info_file` instead?

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+                    snap.flushed_at,
+                ));
+            }
+        }
+    }
+
+    // Each tag gets a TagCreatedUpdate at the flushed_at of the snapshot it points to
+    for (name, snap_id) in tags {
+        let time = snap_times.get(snap_id).copied().unwrap_or(root_time);
+        entries.push((UpdateType::TagCreatedUpdate { name: name.to_string() }, time));
```
paraseba: let's make the time a ms later than the snapshot

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+    }
+
+    // Deleted tags: TagCreatedUpdate + TagDeletedUpdate, both at the pointed snapshot's time
+    for (name, snap_id) in deleted_tags {
+        let time = snap_times.get(snap_id).copied().unwrap_or(root_time);
```
paraseba, line 233: make it a bit later, basically, I don't want times to be = anywhere, because I'm afraid of users or ourselves doing == comparisons

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+            },
+            time,
+        ));
+    }
+
+    // Sort each update chronologically by timestamp, preserving insertion order for ties
+    entries.sort_by_key(|(_, time)| *time);
+
+    // Deduplicate timestamps: ensure strictly increasing, then reverse order as required by UpdateInfo.previous_updates
+    let micros = TimeDelta::microseconds(1);
```
paraseba: lol ... aaaaand I got here. I'm leaving all the previous messages asking for this as a self-shame.

icechunk/src/migrations/mod.rs
```diff
@@ -113,6 +120,131 @@ async fn validate_start(repo: &Repository) -> MigrationResult<()> {
...
+    // main_ancestry is tip-to-root, so last element is the root
+    let root = &main_ancestry[main_ancestry.len() - 1];
+    let root_time = root.flushed_at;
+
+    // Track which snapshots already have a NewCommitUpdate to avoid duplicates
+    // when branches share non-main ancestry (e.g. branch-b forked from branch-a).
+    let mut emitted_snap_ids: HashSet<_> =
+        main_ancestry.iter().map(|s| s.id.clone()).collect();
+
+    let mut entries: Vec<(UpdateType, DateTime<Utc>)> = Vec::new();
```
dcherian: The eager construction worries me. Could it break for repos with very many commits?

icechunk-python/tests/test_stateful_repo_ops.py
mpiannucci: You are saying that after migration the ops log should match right?

icechunk/src/migrations/mod.rs
```diff
@@ -223,16 +355,79 @@ pub async fn migrate_1_to_2(
...
+        repo.ancestry(&VersionInfo::SnapshotId(snap_id.clone()))
+            .await?
+            .try_collect::<Vec<_>>()
+            .await?
+    } else {
+        Vec::new()
+    };
+
+    // Collect non-main branch ancestries
+    let mut branch_ancestries: Vec<(&str, Vec<SnapshotInfo>)> = Vec::new();
```
paraseba: So, since ancestry is so slow in IC1, there is a huge performance optimization you can do, and I also do for GC I think: only traverse ancestries until you find some already seen snapshot. At that point, you already have the rest of the ancestry. Because branches don't go too far from their root, this gives you a speed linear in the number of branches in repos with long histories

icechunk/src/migrations/mod.rs
dcherian:
```suggestion
    let mut entries: Vec<(UpdateType, DateTime<Utc>)> = Vec::with_capacity(main_ancestry.len() + branches.len() + tags.len()
```

maybe?

mpiannucci: Maybe? Im not sure how else to do it tho

icechunk/src/migrations/mod.rs
paraseba: I think this is good enough, we need to keep the whole log and there is not much advantage in splitting the updates, because next change is going to split already.

icechunk/src/migrations/mod.rs
mpiannucci: Its more than that actually but thats close

icechunk-python/tests/test_stateful_repo_ops.py
dcherian: Yup. now that we make a fake one.

dcherian: Though I guess ordering may be wonky. Hmm... maybe not worth it? or we can compare set equality for migrated repos (assuming these types implement `__hash__`).  We could just compare `len` for migrated repos if it becomes too complicated

paraseba:
> We need to regenerate the test_can_read_old repos:
> 
> ```shell
> rm -rf tests/data/{test-repo-v2,split-repo-v2,test-repo-v2-migrated,split-repo-v2-migrated}
> python tests/test_can_read_old.py
> pytest
> ```

actually, we should only rebuild the `-migrated` ones.

icechunk/src/format/repo_info.rs
paraseba: I didn't understand this either

icechunk/src/migrations/mod.rs
paraseba: I was thinking something else ... the migration already traverses through all the snapshots, can we reuse that traversal? Otherwise we are doubling the time it takes

icechunk/src/migrations/mod.rs
paraseba: I think it's fine, there won't be more than a few thousands

icechunk/src/migrations/mod.rs
mpiannucci: Yes we exclude the default branch

icechunk/src/format/repo_info.rs
mpiannucci: I tried to write the comment how i understand it, let me know know if it still doesnt make sense

mpiannucci:
Okay this is ready for another review, i think i addressed everything

icechunk/src/format/repo_info.rs
dcherian: makes sense to me now. Thank you!

dcherian (approved):

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -1066,14 +1069,29 @@ def check_ops_log(self) -> None:
...
-        # ops should be ordered
-        assert all(
-            first.updated_at > second.updated_at
-            for (first, second) in itertools.pairwise(actual_ops)
-        )
+        # ops should be ordered: strictly decreasing for normal repos,
+        # non-increasing after migration (synthetic entries can share timestamps)
+        if self.model.migrated:
+            assert all(
+                first.updated_at >= second.updated_at
```
paraseba: Can we avoid `=` here? I think IC never adds more than one element to the ops log at once, so they never have the same time. (we should have a check for that hopefully)

Issue #1741, opened 2026-03-02
Delete and recreate node causes rebase errors


ianhi:
If you both delete and re-create a node in a session that is behind main then the rebase will fail, even though this is should a valid op.

If you modified the group in the meantime it should still of course fail - but thats a different error than this. 

(Found with multi user stateful tests !)

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "icechunk>=2.0.0a0",
#     "zarr",
# ]
#
# [tool.uv]
# prerelease = "allow"
# ///

import icechunk as ic
import zarr

print(f"icechunk {ic.__version__}")


def show_tree(store: ic.IcechunkStore, label: str = "") -> None:
    if label:
        print(label)
    print(zarr.open_group(store=store, mode="r").tree())
    print()


ic.set_logs_filter("warn,icechunk::storage::object_store=error")


storage = ic.in_memory_storage()

# Init: root group + group 'a'
repo = ic.Repository.create(storage, spec_version=2)
session = repo.writable_session("main")
zarr.open_group(store=session.store, path="a", mode="w")
session.commit("init: root + a")
show_tree(repo.readonly_session("main").store, "after init")

# Two separate repo objects (like two separate processes)
repo_a = ic.Repository.open(storage)
repo_b = ic.Repository.open(storage)

# A opens session (will become stale)
session_a = repo_a.writable_session("main")

# B: add 'b', commit
session_b = repo_b.writable_session("main")
zarr.open_group(store=session_b.store, path="b", mode="w")
session_b.commit("B: add b")
show_tree(repo.readonly_session("main").store, "MAIN:")


# A: delete 'a' and re-add it (disjoint from B)
print("delete and re-create group A")
del zarr.open_group(store=session_a.store, path="")["a"]
show_tree(session_a.store)

# commetningn thiss allows everything to wrok
zarr.open_group(store=session_a.store, path="a", mode="w")
show_tree(session_a.store)

print("DIFF\n==========")
print(session_a.status())
print("==========")

# A tries to commit
print("trying to rebase A...")
session_a.rebase(ic.ConflictDetector())
session_a.commit("A: delete+readd a")
show_tree(repo.readonly_session("main").store, "MAIN:")
```

```
after init
/
└── a

MAIN:
/
├── a
└── b

delete and re-create group A
/

/
└── a

DIFF
==========
Groups created:
    /a
Groups deleted:
    /a
==========
trying to rebase A...
trying to rebase A...
[...]

PR #1742, opened 2026-03-02
FIX: allow delete + recreate in rebase

ianhi:
Fixes: #1741

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -390,7 +390,7 @@ impl Session {
         &self,
         chunk_location: &VirtualChunkLocation,
     ) -> Option<&VirtualChunkContainer> {
-        self.virtual_resolver.matching_container(chunk_location.url())
+        self.virtual_resolver.matching_container(chunk_location)
     }
 
     /// Compute an overview of the current session changes
@@ -1089,13 +1089,19 @@ impl Session {
     pub async fn all_virtual_chunk_locations(
         &self,
     ) -> SessionResult<impl Stream<Item = SessionResult<String>> + '_> {
-        let stream =
-            self.all_chunks().await?.try_filter_map(|(_, info)| match info.payload {
-                ChunkPayload::Virtual(reference) => {
-                    ready(Ok(Some(reference.location.url().to_string())))
-                }
-                _ => ready(Ok(None)),
-            });
+        let resolver = &self.virtual_resolver;
+        let stream = self.all_chunks().await?.try_filter_map(move |(_, info)| match info
+            .payload
+        {
+            ChunkPayload::Virtual(reference) => {
+                let expanded = match resolver.expand_location(reference.location.url()) {
+                    Ok(abs) => abs,
+                    Err(e) => return ready(Err(e.into())),
+                };
+                ready(Ok(Some(expanded)))
+            }
+            _ => ready(Ok(None)),
+        });
         Ok(stream)
     }
 
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -238,18 +238,24 @@ class VirtualChunkContainer:
 
     Attributes
     ----------
+    name: str | None
+        Optional name for this container. When set, chunks can use relative ``vcc://name/path``
+        locations instead of full absolute URLs.
     url_prefix: str
         The prefix of urls that will use this containers configuration for reading virtual references.
     store: ObjectStoreConfig
         The storage backend to use for the virtual chunk container.
     """
 
-    name: str
+    name: str | None
     url_prefix: str
     store: ObjectStoreConfig
 
     def __new__(
-        cls, url_prefix: str, store: _AnyObjectStoreConfig
+        cls,
+        url_prefix: str,
+        store: _AnyObjectStoreConfig,
+        name: str | None = None,
     ) -> VirtualChunkContainer:
         """
         Create a new `VirtualChunkContainer` object
@@ -260,6 +266,9 @@ class VirtualChunkContainer:
             The prefix of urls that will use this containers configuration for reading virtual references.
         store: ObjectStoreConfig
             The storage backend to use for the virtual chunk container.
+        name: str | None
+            Optional name for this container. When set, chunks can use relative ``vcc://name/path``
+            locations instead of full absolute URLs.
         """
 
 class VirtualChunkSpec:
@@ -1429,7 +1438,11 @@ class RepositoryConfig:
         ...
     def set_virtual_chunk_container(self, cont: VirtualChunkContainer) -> None:
         """
-        Set the virtual chunk container for the repository.
+        Add or update a virtual chunk container in the repository configuration.
+
+        For named containers, the name is the identity: if a container with the
+        same name already exists (even with a different url_prefix), it will be
+        replaced. For unnamed containers, the url_prefix is the key.
 
         Parameters
         ----------
--- a/icechunk-python/src/config.rs
+++ b/icechunk-python/src/config.rs
@@ -560,8 +560,13 @@ pub struct PyVirtualChunkContainer {
 #[pymethods]
 impl PyVirtualChunkContainer {
     #[new]
-    pub fn new(url_prefix: String, store: PyObjectStoreConfig) -> Self {
-        Self { name: None, url_prefix, store }
+    #[pyo3(signature = (url_prefix, store, name = None))]
+    pub fn new(
+        url_prefix: String,
+        store: PyObjectStoreConfig,
+        name: Option<String>,
+    ) -> Self {
+        Self { name, url_prefix, store }
     }
 }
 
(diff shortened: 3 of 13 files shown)


icechunk/src/conflicts/detector.rs
```diff
@@ -41,8 +41,18 @@ impl ConflictSolver for ConflictDetector {
         )
         .try_filter_map(|(path, _)| async {
             match previous_repo.get_node(path).await {
-                Ok(_) => {
-                    Ok(Some(Conflict::NewNodeConflictsWithExistingNode(path.clone())))
+                Ok(existing_node) => {
+                    /*
+                    If the current changeset deletes then recreates a node
```
dcherian: I think this is easier to read as a `filter` added on line 40. Use `filter_map` to combine with the following `map`

icechunk/src/session.rs
```diff
@@ -5031,6 +5032,60 @@ mod tests {
...
+    /// Test that delete-then-recreate exactly the same node
+    /// does NOT conflict
+    ///
+    /// This session: delete group + recreate group at same path with samem metadata
+    /// Previous commit: add a different sibling group
+    async fn test_no_conflict_on_delete_then_recreate() -> Result<(), Box<dyn Error>> {
+        let (mut ds1, mut ds2) = get_sessions_for_conflict().await?;
+
+        let path: Path = "/foo/bar".try_into().unwrap();
+        ds1.add_group("/foo/baz".try_into().unwrap(), user_data()).await?;
```
```suggestion
        ds1.add_group("/foo/quux".try_into().unwrap(), user_data()).await?;
```

confusing otherwise!

icechunk/src/session.rs
```diff
@@ -5031,6 +5032,60 @@ mod tests {
...
+        ds1.commit("add sibling group", None).await?;
+
+        ds2.delete_group(path.clone()).await?;
+        ds2.add_group(path.clone(), Bytes::from("group-metadata")).await?;
+        ds2.commit("delete+re-add group", None).await.unwrap_err();
```
```suggestion
        assert!(ds2.commit("delete+re-add group", None).await.is_err());
```

icechunk/src/conflicts/detector.rs
```diff
@@ -41,8 +41,18 @@ impl ConflictSolver for ConflictDetector {
...
+                    at this path then that is ok, not a conflict
```
dcherian: i dont' see how modifying the content is a conflict unless they both modify the metadata.

icechunk/src/session.rs
dcherian: But shouldn't this be fine. `ds1` did not edit this group...

icechunk/src/session.rs
```diff
@@ -5031,6 +5032,60 @@ mod tests {
...
+        let (mut ds1, mut ds2) = get_sessions_for_conflict().await?;
+
+        let path: Path = "/foo/bar".try_into().unwrap();
+        ds1.update_group(&path, Bytes::from("updated")).await?;
+        ds1.commit("update group metadata", None).await?;
+
+        let node = ds2.get_node(&path).await.unwrap();
+        ds2.delete_group(path.clone()).await?;
+        ds2.add_group(path.clone(), user_data()).await?;
+        ds2.commit("delete+re-add group", None).await.unwrap_err();
```
```suggestion
        assert!(ds2.commit("delete+re-add group", None).await.is_err());
```

This one makes sense to me.

icechunk/src/conflicts/detector.rs
ianhi: isn't them both changing the values in a chunk a conflict? thats what i was thinking about when i wrote this

dcherian: IIUC this is checking node conflicts (so deletes/metadata changes) not chunk value changes

icechunk/src/conflicts/detector.rs
```diff
@@ -41,8 +41,18 @@ impl ConflictSolver for ConflictDetector {
         )
         .try_filter_map(|(path, _)| async {
             match previous_repo.get_node(path).await {
-                Ok(_) => {
-                    Ok(Some(Conflict::NewNodeConflictsWithExistingNode(path.clone())))
+                Ok(existing_node) => {
```
paraseba: I need to understand why we think this is not a conflict. I created a node, and somebody concurrently did the same. How do we know they both did it with the same intention and semantics? Even if the metadata is the same, it feels like a conflict to me, I think i want the system to tell me there is something wrong. Particularly in a default detector, in custom detectors maybe this is desired behavior.

But I may be wrong, please explain more.

ianhi: I was understanding this not as someone else created the node but:

snapshot 1: create A
Create two new sessions based of snapshot 1

Session 1: create B -> commit (snapshot 2 on main)

Session 2: (based on snapshot 1)
    - delete A
    - recreate A
    - commit

This leads to a conflict with Session 1's actions - even though session 1 didn't touch node A at all. 

If we had only deleted A then there would not have been a conflict - this new code is just also allowing the recreate after the delete.

icechunk/src/conflicts/detector.rs
ianhi: Testing with a script -> claude format as a table. theh comment justs means that this change is **not** worrying about chunk values, that that is handled elsewhere. i.e. alleviating concerns that this change woudl accidentally allow a conflict to pass through.

Key difference is in the top left box. the other conflicts also change slightly, but still throw a valid conflict.

### Before 

| | S2: delete+recreate group | S2: delete+recreate array + write chunk values |
|---|---|---|
| **S1: unrelated change** | Conflict (`NewNodeConflictsWithExistingNode`) | Conflict (`NewNodeConflictsWithExistingNode`) |
| **S1: modifies chunk values** | — | Conflict (`NewNodeConflictsWithExistingNode` + `DeleteOfUpdatedArray`) |
| **S1: updates group metadata** | Conflict (`NewNodeConflictsWithExistingNode` + `DeleteOfUpdatedGroup`) | — |

### After (with PR fix)

| | S2: delete+recreate group | S2: delete+recreate array + write chunk values |
|---|---|---|
| **S1: unrelated change** | No conflict | No conflict |
| **S1: modifies chunk values** | — | Conflict (`DeleteOfUpdatedArray`) |
| **S1: updates group metadata** | Conflict (`DeleteOfUpdatedGroup`) | — |

icechunk/src/session.rs
ianhi: this part is a not a conflict error, but a "hey you need to rebase" i can either just immediately rebase or make it more specific. proably will do the latter

ianhi: I added more specific error checking

icechunk/src/conflicts/detector.rs
ianhi: I added a prefilter intstead. nice idea

icechunk/src/conflicts/detector.rs
```diff
@@ -37,7 +37,15 @@ impl ConflictSolver for ConflictDetector {
...
 
         let new_nodes_explicit_conflicts = stream::iter(
-            current_changes.new_nodes().map(Ok),
+            current_changes
+                .new_nodes()
+                // Filter out new nodes that are delete+recreate (not a conflict).
+                // Content conflicts for these are detected lower down in this function.
+                .filter(|(path, _)| {
+                    !current_changes.deleted_groups().any(|(p, _)| p == *path)
+                        && !current_changes.deleted_arrays().any(|(p, _)| p == *path)
```
paraseba: this is quadratic on number of deleted nodes. Can you use `current_changes.is_deleted` instead?

icechunk/src/conflicts/detector.rs
ianhi: @paraseba would like to get your input on this before moving forward here. But Im pretty sure that what I have here is the correct behavior.

paraseba: I think you are right @ianhi . But this takes time to think about, and I don't remember much of the details. For example, my confusion came from misunderstanding  what it means for a node to be both deleted and new.

Not detecting conflicts is always a worst case scenario, so we need to be very conservative. False positive conflicts are much better than false negatives. So, maybe you can create some manual unit test that verifies we are not missing anything with this change? But it's also OK if you think we are already covered by existing stateful tests.

ianhi: > So, maybe you can create some manual unit test that verifies we are not missing anything with this change?

my big concern was that session 1 also modified that node, that we would still have a conflict. I added a unit test for that - and it does still raise a conflict. See my table under deepaks comment

for everything else I feel moderately ok with stateful testing

paraseba: Good enough for me @ianhi , let's merge once the minor issue is fixed

icechunk/src/session.rs
```diff
@@ -5186,6 +5187,74 @@ mod tests {
         Ok(())
     }
 
+    #[tokio_test]
+    /// Test that delete-then-recreate exactly the same node
+    /// does NOT conflict
+    ///
+    /// This session: delete group + recreate group at same path with samem metadata
```
dcherian: Why does the metadata matter? The parallel session made an independent change, so this is a conflict that gets trivially resolved (as in this test)

ianhi: does not. but at first i thought it did and then forgot to remove...

icechunk/src/conflicts/detector.rs
```diff
@@ -36,8 +36,19 @@ impl ConflictSolver for ConflictDetector {
...
+            .collect();
+
         let new_nodes_explicit_conflicts = stream::iter(
-            current_changes.new_nodes().map(Ok),
+            current_changes
+                .new_nodes()
+                // Filter out new nodes that are delete+recreate (not a conflict).
+                // Content conflicts for these are detected lower down in this function.
+                .filter(|(path, _)| !deleted_paths.contains(path))
+                .map(Ok),
```
ianhi, line 51: @paraseba I could not use `is_deleted` because that checks against both path and nodeid. but the nodeid will be different. So not just precomputing the set of deleted paths and using that. Is there a better way to do this? Other than computing the deleted paths in the changeset itself?.

Issue #1738, opened 2026-03-02
Release IC2 library defaulted to IC1 format


paraseba:
To derisk IC2 and make migration easier for users, we want to release an IC1.x version that contains `main`, but it only writes IC2 format by default.

For this we need, at least:

- [ ] Confidence that we are not going to break any IC1 repos, for this we need more tests targeting IC1
- [ ] Confidence Icechunk or users are not going to upgrade by mistake a repo from IC1 -> IC2. For this we need at least to add confirmation arguments to the migration functions, and probably display warning and ask for keyboard confirmation.
- [ ] Create IC1 repos by default on `create`, `open_or_create` and any other functions. Include warning saying IC2 is not production ready if users explicitly say they want IC2. 
- [ ] Confidence we are not going to degrade performance on IC1 repos. For this we need at least micro benchmarks
- [ ] Update any tests and workflows, to make sure we are testing both in IC1 and IC2, we don't want to lose IC2 testing just because the default is to create IC1 repos.

ianhi:
I am a bit worried that the cost of doing this will outweigh the benefits.

It leads the user to the question of why did we even have to release Icechunk python version 2 at all. It creates a confusing situation where the major versino no longer maps 1 to 1 with the on disk format. Which contradicts what we wrote down here: https://icechunk.io/en/latest/version-policy/#icechunk-python-library

> make migration easier for users

Maybe I'm missing something but it's not obvious to me why this makes migration easier for users. It seems like it would silently allow them new options in a new environment without the helpful signal that now you are on IC2 and should expect new options.

dcherian:
I will update this comment with progress.

> Confidence that we are not going to break any IC1 repos, for this we need more tests targeting IC1

https://github.com/earth-mover/icechunk/pull/1743
https://github.com/earth-mover/icechunk/pull/1739

> Create IC1 repos by default on create, open_or_create and any other functions. Include warning saying IC2 is not production ready if users explicitly say they want IC2.

> Confidence Icechunk or users are not going to upgrade by mistake a repo from IC1 -> IC2. For this we need at least to add confirmation arguments to the migration functions, and probably display warning and ask for keyboard confirmation.

We already have `dry_run=True` by default. I can update to add a warning; not sure if we really need keyboard confirmation on top of that since a user will have to type out `dry_run=True`

> Update any tests and workflows, to make sure we are testing both in IC1 and IC2, we don't want to lose IC2 testing just because the default is to create IC1 repos.

[...]

PR #1743, opened 2026-03-02
Parameterize rust tests over V1, V2

dcherian:
Greatly helped by Claude.

I manually audited every change. FWIW I didn't parameterize over the various conflict detection tests in `session.rs` but basically parameterized everything else.

I only run V2 with `test_large_manifests`. It uses ludicrous amounts of memory on my machine. I shall profile and attempt to fix in a future PR.

xref #1738


icechunk/src/repository.rs
```diff
@@ -1894,13 +1895,23 @@ mod tests {
         );
     }
 
-    #[tokio::test]
-    async fn test_repository_persistent_config() -> Result<(), Box<dyn Error>> {
+    #[tokio_test]
+    #[rstest]
+    #[case::v1(SpecVersionBin::V1dot0)]
+    #[case::v2(SpecVersionBin::V2dot0)]
```
li-em: with the [`rstest_reuse` crate](https://github.com/la10736/rstest/tree/master/rstest_reuse#composition-and-values) this can be 

```suggestion
    #[template]
    #[rstest]
    #[case::v1(SpecVersionBin::V1dot0)]
    #[case::v2(SpecVersionBin::V2dot0)]
    fn spec_version(#[case] spec_version: SpecVersionBin) {}

    #[tokio_test]
    #[apply(spec_version)]
```

and `#[apply(spec_version)]` in all the other tests, instead of listing the cases again.

icechunk/Cargo.toml
```diff
@@ -88,6 +88,7 @@ port_check = "0.3.0"
 pretty_assertions = "1.4.1"
 proptest = "1.10.0"
 proptest-state-machine = "0.7.0"
+rstest = "0.24"
```
li-em: any reason not to use `0.26.1`?

icechunk/src/repository.rs
dcherian: great!

icechunk/tests/test_virtual_refs.rs
```diff
@@ -45,6 +48,12 @@ use object_store::{
 };
 use pretty_assertions::assert_eq;
 
+#[template]
+#[rstest]
+#[case::v1(SpecVersionBin::V1dot0)]
+#[case::v2(SpecVersionBin::V2dot0)]
```
paraseba, line 54: no `apply` trick here?

dcherian: Reusing would require adding rstest as dependency for use in these integration tests. So I just duplicated the definition. The template is used later in this file.

PR #1749, opened 2026-03-03
throw when opening v2 repo from v1 icechunk

mpiannucci:


icechunk/src/storage/mod.rs
```diff
@@ -573,6 +574,16 @@ pub trait Storage: fmt::Debug + fmt::Display + private::Sealed + Sync + Send {
         }
     }
 
+    async fn has_v2_repo_info(&self) -> StorageResult<bool> {
+        let mut stream = self.list_objects(&self.default_settings(), "").await?;
```
mpiannucci: local filesystem cant list exact keys? IDK if there is a better simple way to do this vs this way

paraseba: Shouldn't we use a `get` instead of list? We do know the exact path. We want this to work even without list permissions.

mpiannucci: yeah trying that, but the storage trait is really opinionated

paraseba: Yeah, Storage in IC1 sucks. Feel free to add a new method like `has_object` that does a metadata GET request.

mpiannucci: Done

icechunk/src/repository.rs
```diff
@@ -254,6 +258,10 @@ impl Repository {
         authorize_virtual_chunk_access: HashMap<String, Option<Credentials>>,
     ) -> RepositoryResult<Self> {
         debug!("Opening Repository");
+        if storage.has_v2_repo_info().await? {
+            return Err(RepositoryErrorKind::RepositoryIsV2.into());
```
paraseba, line 262: can we add the same for commits? When the commit starts you spawn this check, right before it's about to update the branch you fail if the check told you it's v2.

mpiannucci: done

PR #1750, opened 2026-03-03
Refactor UpdateType and improve repr

### LINKED ISSUE #1683: Python's UpdateType shourd render `updated_at` in its repr

li-em:
Fix #1683

repr looking like this:
```
Update(kind=BranchCreated(name=120), updated_at=2026-03-03 20:10:28.979033 UTC, backup_path=None)
```

Did a big cleanup on `UpdateType` and `Update` along the way. Now `UpdateType` is an enum with all the different kinds of updates, and `Update` is a struct with `kind, updated_at, backup_path` fields.

## Pending

- [x] re-expose some getters for update types with data
- [x] fix stateful tests to account for new types

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -97,10 +97,10 @@ class UpdateModel:
     ictype = ic.UpdateType
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, self.ictype):
+        if not isinstance(other, ic.Update) or not isinstance(other.kind, self.ictype):
             return NotImplemented
         return all(
-            getattr(self, f.name) == getattr(other, f.name)
+            getattr(self, f.name) == getattr(other.kind, f.name)
             for f in fields(self)
             # we check ictype with isinstance above
             if f.name != "ictype"
@@ -1094,10 +1094,7 @@ class VersionControlStateMachine(RuleBasedStateMachine):
         all_backups = [op.backup_path for op in actual_ops if op.backup_path is not None]
         assert len(all_backups) == len(set(all_backups))
 
-        assert all(
-            isinstance(a.kind, m.ictype)
-            for m, a in zip(self.model.ops_log[::-1], actual_ops, strict=True)
-        )
+        assert self.model.ops_log[::-1] == actual_ops
         assert isinstance(
             actual_ops[-1].kind,
             ic.UpdateType.RepoInitialized | ic.UpdateType.RepoMigrated,


icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -99,84 +99,89 @@ class UpdateModel:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, self.ictype):
             return NotImplemented
-        return all(getattr(self, f.name) == getattr(other, f.name) for f in fields(self))
+        return all(
+            getattr(self, f.name) == getattr(other, f.name)
+            for f in fields(self)
+            # we check ictype with isinstance above
+            if f.name != "ictype"
+        )
```
li-em, line 107: @dcherian this and the other change at the end of this file feel wrong...

icechunk-python/tests/test_timetravel.py
```diff
@@ -1092,18 +1096,15 @@ async def test_long_ops_log(spec_version: int | None) -> None:
     updates = [update async for update in repo.ops_log_async()]
     assert len(updates) == NUM_BRANCHES + 1
 
-    t = type(updates[0])
-    assert t == ic.BranchCreatedUpdate
+    assert isinstance(updates[0].kind, ic.UpdateType.BranchCreated)
```
li-em, line 1099: The section on [complex enums](https://pyo3.rs/v0.27.2/class.html#complex-enums) in pyo3 docs says:
> WARNING: `Py::new` and `.into_pyobject` are currently inconsistent. Note how the constructed value is not an instance of the specific variant. For this reason, constructing values is only recommended using `.into_pyobject`.

But getting `.into_pyobject` was turning into a mess (unless I missed something obvious), so turned equality checks into `isinstance` calls in tests

icechunk-python/stubtest_allowlist.txt
```diff
@@ -181,9 +169,29 @@ icechunk._icechunk_python.Storage
...
+icechunk._icechunk_python.UpdateType.NewCommit.__new__
+icechunk._icechunk_python.UpdateType.NewDetachedSnapshot.__match_args__
+icechunk._icechunk_python.UpdateType.NewDetachedSnapshot.__new__
+icechunk._icechunk_python.UpdateType.RepoInitialized.__match_args__
+icechunk._icechunk_python.UpdateType.RepoMigrated.__match_args__
+icechunk._icechunk_python.UpdateType.RepoMigrated.__new__
+icechunk._icechunk_python.UpdateType.TagCreated.__match_args__
+icechunk._icechunk_python.UpdateType.TagCreated.__new__
+icechunk._icechunk_python.UpdateType.TagDeleted.__match_args__
+icechunk._icechunk_python.UpdateType.TagDeleted.__new__
```
li-em, line 197: Added these following other examples on the allowlist, but can go back to type stubs and try to find a better solution

icechunk-python/tests/test_feature_flags.py
```diff
@@ -56,20 +56,20 @@ def test_set_and_unset_flags_on_repo() -> None:
...
-    assert ops[0].id == flag_id
-    assert ops[0].new_value is None
-    assert isinstance(ops[1], ic.FeatureFlagChangedUpdate)
-    assert ops[1].id == other_id
-    assert ops[1].new_value is True
-    assert isinstance(ops[2], ic.FeatureFlagChangedUpdate)
-    assert ops[2].id == flag_id
-    assert ops[2].new_value is False
-    assert isinstance(ops[3], ic.RepoInitializedUpdate)
+    assert isinstance(ops[0].kind, ic.UpdateType.FeatureFlagChanged)
```
paraseba, line 63: We need to check with @samantha-earthmover to see if she needs to update notifications code with this change.

icechunk-python/stubtest_allowlist.txt
dcherian: IMO it's not worth the time. Feels like pyO3 should solve it

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -1089,9 +1094,13 @@ def check_ops_log(self) -> None:
         all_backups = [op.backup_path for op in actual_ops if op.backup_path is not None]
         assert len(all_backups) == len(set(all_backups))
 
-        assert self.model.ops_log[::-1] == actual_ops
```
dcherian, line 1092: the idea behind using equality was to check other data like "branch name" in BranchCreatedUpdate. Is there a way to do that?

icechunk-python/tests/test_stateful_repo_ops.py
dcherian: i commented down there. The code here can be whatever it needs for that equality check to be nicer

icechunk-python/tests/test_stateful_repo_ops.py
li-em: OK, found the right incantation for `__eq__` to revert this assertion to what it was before

PR #1753, opened 2026-03-03
Optimize manifest writes

dcherian:
xref #1399 

TODO with criterion
- [x] compare against main
- [ ] compare against IC1

-----

We were duplicating the ChangeSet's `set_chunks` `splits.len()` times earlier!

### `criterion` benchmarks against `main`

20% improvement for large numbers of manifests (as expected)
```
commit_split_manifests/inline/100
                 change:
                        time:   [−8.0487% −7.0303% −5.9660%] (p = 0.00 < 0.05)
                        thrpt:  [+6.3445% +7.5619% +8.7533%]
                        Performance has improved.
commit_split_manifests/virtual/1000
                 change:
                        time:   [−18.580% −17.039% −15.620%] (p = 0.00 < 0.05)
                        thrpt:  [+18.512% +20.538% +22.820%]
commit_split_manifests/inline/1000
                 change:
                        time:   [−17.519% −16.096% −14.842%] (p = 0.00 < 0.05)
                        thrpt:  [+17.429% +19.184% +21.240%]
```

### `cargo-instruments`
Total mem allocations down from 15GB -> 10 GB for 
```sh
cargo instruments -t Allocations  --profile bench --example large_manifests
```

which writes 10 million virtual chunk refs at one go and then flushes them to 10 manifests.

**Note** Only the first commit is reflected in this benchmark. I shall post tomorrow with the criterion results.

### Before
image

### After
image


icechunk/src/session.rs
```diff
@@ -2010,10 +2010,22 @@ impl<'a> FlushProcess<'a> {
             if self.change_set.array_manifest(node_id).is_some() {
                 let chunks = stream::iter(
                     self.change_set
-                        .new_array_chunk_iterator(node_id, node_path)
+                        .array_chunks_iterator(node_id, node_path)
```
dcherian, line 2027: This is the big change. Previously we cloned then filtered; now we filter then clone. The clones are needed for `ChunkInfo` I tried to return a `ChunkInfoRef` structure from this iterator that only held references but it was quite invasive. The core issue being we need to chain this iterator with the one from manifest files (which returns owned objects).

icechunk/src/format/manifest.rs
```diff
@@ -55,6 +55,7 @@ impl ManifestExtents {
         Self(ranges.into_iter().collect())
     }
 
+    #[inline(always)]
     pub fn contains(&self, coord: &[u32]) -> bool {
```
dcherian, line 59: TODO: inline the caller too. This made a ~10% difference

PR #1756, opened 2026-03-04
Add NodeSnapshot cache on Snapshot

dcherian:
Opening early for quick comments on approach.

Solid 50% improvement on our microbenchmarks


icechunk/src/format/snapshot.rs
```diff
@@ -302,10 +306,17 @@ impl ManifestFileInfo {
...
+// #[derive(PartialEq)]
 pub struct Snapshot {
     buffer: Vec<u8>,
     spec_version: SpecVersionBin,
+    node_cache: Cache<Path, Arc<NodeSnapshot>, FileWeighter>,
+}
+
+impl Weighter<Path, Arc<NodeSnapshot>> for FileWeighter {
+    fn weight(&self, _: &Path, _: &Arc<NodeSnapshot>) -> u64 {
+        1
```
paraseba: i don't think you need a weighter if they are all approximately the same weight, which i think is the case.

icechunk/src/format/snapshot.rs
```diff
@@ -383,7 +394,12 @@ impl Snapshot {
             &ROOT_OPTIONS,
             buffer.as_slice(),
         )?;
-        Ok(Snapshot { buffer, spec_version })
+        Ok(Snapshot {
+            buffer,
+            spec_version,
+            // FIXME: use config parameter
+            node_cache: Cache::with_weighter(1, 10, FileWeighter),
```
paraseba: Is the idea to keep only 1 node? Maybe we don't need the overhead of a `Cache` for that? But also, maybe using `Cache` is simplest/safest

paraseba: I know the cache needs to be small, because we keep many snapshots in the asset manager cache, so we'll keep all those caches around.

dcherian: oh good point, didn't realize that. I was thinking something small like 1-5. Shall I add a config option?

mpiannucci: I dont like having more than one cache instance, is there a reason this doesnt use the existing cache?

dcherian: The other caches are on AssetManager and basically store bytes for snapshots, manifests, chunks etc. 

This one caches `NodeSnapshot` instances that are used on _every_ set & get. It caches them on the `Snapshot` instance.

icechunk/src/format/snapshot.rs
```diff
@@ -302,10 +306,17 @@ impl ManifestFileInfo {
     }
 }
 
-#[derive(PartialEq)]
+// #[derive(PartialEq)]
```
dcherian: @paraseba not sure about this change. That said I also didn't see where we needed `PartialEq` on `Snapshot`

paraseba: probably for tests, but I'm ok with removing it

icechunk/src/format/snapshot.rs
paraseba: Yeah @mpiannucci this is a very local cache, basically it helps with things like 1k requests to the same node, which is very common, and currently require 1k binary searches over all nodes. We have discussed this in the past with @TomNicholas too.

@dcherian I think 1-5 size is reasonable, and I don't think you need a config  for now. Let's give it a try.

Here is an idea, does this cache library (or other) offer a ttl option? We could be very aggressive with that and we wouldn't have to worry about size. If caches die after 2 seconds of not being accessed, all Snapshots in the asset manager cache don't use this space.

dcherian: yeah adding a config is pretty invasive...

icechunk/src/format/snapshot.rs
```diff
@@ -302,10 +303,10 @@ impl ManifestFileInfo {
     }
 }
 
-#[derive(PartialEq)]
 pub struct Snapshot {
     buffer: Vec<u8>,
     spec_version: SpecVersionBin,
+    node_cache: Cache<Path, Arc<NodeSnapshot>>,
```
paraseba, line 314: can you add a comment explaining the purpose of this cache? It's  very usual to make many consecutive requests to the same node, etc

paraseba (approved):

PR #1763, opened 2026-03-04
tweak Pixi config to avoid long CI runtimes

li-em:
the Code Quality CI runtime is pretty long, because we end up rebuilding the icechunk package multiple times. Try to limit it and better reuse cache between rust and maturin.

### DIFF
--- a/icechunk-python/docs/docs/contributing.md
+++ b/icechunk-python/docs/docs/contributing.md
@@ -37,7 +37,14 @@ We use [pixi](https://pixi.prefix.dev/latest/) to manage both the python and rus
 
     ```bash
     pixi shell -m icechunk-python/pyproject.toml
+    just develop
     ```
+
+    `just develop` will set up the `maturin-import-hook` for fast incremental Rust compilation
+    and do the first build with `maturin develop --uv`. `just develop` can be skipped in future
+    developer shell activations, but should also run fast after the first invocation (and if
+    there no changes in the Rust code).
+
     !!! Usage without pixi shell
         You could run all commands below using this pattern:
         ```shell


ianhi:
Does this require any update to the contributing docs?

icechunk-python/docs/docs/contributing.md
```diff
@@ -37,7 +37,14 @@ We use [pixi](https://pixi.prefix.dev/latest/) to manage both the python and rus
...
     ```bash
     pixi shell -m icechunk-python/pyproject.toml
+    just develop
     ```
+
+    `just develop` will set up the `maturin-import-hook` for fast incremental Rust compilation
+    and do the first build with `maturin develop --uv`. `just develop` can be skipped in future
+    developer shell activations, but should also run fast after the first invocation (and if
+    there no changes in the Rust code).
+
```
li-em, line 47: @ianhi do you agree with the wording here?

icechunk-python/pyproject.toml
```diff
@@ -102,6 +102,8 @@ features = ["pyo3/extension-module"]
 module-name = "icechunk._icechunk_python"
 python-source = "python"
 exclude = ["README.md"]
+editable-profile = "dev"
+cargo-extra-args = ["--workspace"]
```
li-em, line 106: These two configs are to line up `maturin` with regular `cargo` invocations thru `just`, so it avoids rebuilding as much as possilbe.

icechunk-python/pyproject.toml
```diff
@@ -191,9 +193,6 @@ icechunk = "icechunk._icechunk_python:cli_entrypoint"
 channels = ["conda-forge"]
 platforms = ["osx-arm64", "linux-64", "linux-aarch64", "win-64"]
 
-[tool.pixi.pypi-dependencies]
-icechunk = { path = ".", editable = true }
-
```
li-em, line 196: This ended up forcing the IC compilation when activating the pixi shell, but wasn't well integrated with the maturin import hook. Opted to move to a `pixi shell` and then `just develop`, because `just develop` also works for other options (like `uv`, `conda/mamba`, and `nix`).

.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
         args: ["--fix", "--show-fixes", "icechunk-python/"]
 
   - repo: https://github.com/rhysd/actionlint
-    rev: v1.7.8
+    rev: v1.7.11
```
li-em, line 20: Bumping this because `actionlint` had outdated definitions for `rust-cache` and was failing, even if the `rust-cache` configs were valid.

.github/workflows/code-quality.yaml
```diff
@@ -37,7 +36,7 @@ jobs:
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@v2
         with:
-          key: ${{ env.RUST_CHANNEL }}
+          add-job-id-key: "false"
```
li-em, line 39: Planting the seed to share the rust cache between jobs 🙏

PR #1764, opened 2026-03-04
More optimizations

dcherian:
A collection of changes that improves sets, & commits.

1. `tracing` Spans 😠 introduce a consistent 20% tax on everything by Debug formatting things. We use a cheaper formatter for `Path`. We _should_ do something about `ObjectId` but we'd need to rewrite base32 encoding as far as I (& Claude) can tell.
2. To reduce `ObjectId` formatting overhead, I moved the span to functions that execute _actual_ fetches not the AssetManager method that returns a cached value if present.
3. Optimize construction of Array NodeData from Snapshots (much fewer Vec allocations & deallocations)
4. Switch ManifestSplits to be a Vec of Vec of _sorted_ "edges" along each axis. This lets us do binary search to locate manifests for each coord, instead of linear search. I should've done this right from the start 🤦🏾‍♂️ . I was distracted by the cool `Vec<ManifestExtents>` type :)
5. Fastpath for constructing Path from Snapshot, where we know the String is valid.
6. Concurrent manifest fetching + avoid cloning payload again when flushing existing nodes.

Comparison to main:
```
append_split_manifests/set/inline
                        time:   [426.85 ms 429.40 ms 432.56 ms]
                        change: [−73.772% −73.205% −72.719%] (p = 0.00 < 0.05)
                        Performance has improved.

append_split_manifests/commit/inline
                        time:   [118.16 ms 122.81 ms 128.02 ms]
                        change: [−51.844% −50.015% −48.011%] (p = 0.00 < 0.05)
                        Performance has improved.

append_split_manifests/set/virtual
                        time:   [1.0116 s 1.0149 s 1.0193 s]
                        change: [−41.568% −41.119% −40.706%] (p = 0.00 < 0.05)
                        Performance has improved.


commit_split_manifests/inline/100
                 change:
                        time:   [−23.604% −21.441% −19.593%] (p = 0.00 < 0.05)
                        thrpt:  [+24.367% +27.293% +30.897%]
                        Performance has improved.

commit_split_manifests/inline/1000
                 change:
                        time:   [−71.571% −71.141% −70.754%] (p = 0.00 < 0.05)
                        thrpt:  [+241.92% +246.51% +251.75%]
                        Performance has improved.

commit_split_manifests/virtual/1000
                 change:
                        time:   [−67.699% −67.436% −67.191%] (p = 0.00 < 0.05)
                        thrpt:  [+204.79% +207.08% +209.58%]
                        Performance has improved.

set_chunks/inline/100000
                 change:
                        time:   [−37.877% −31.675% −24.727%] (p = 0.00 < 0.05)
                        thrpt:  [+32.849% +46.359% +60.970%]
                        Performance has improved.

get_chunks/100 
                 change:
                        time:   [−50.286% −49.958% −49.589%] (p = 0.00 < 0.05)
                        thrpt:  [+98.370% +99.833% +101.15%]
                        Performance has improved.

get_chunks/1000 
                 change:
                        time:   [−56.124% −55.269% −54.588%] (p = 0.00 < 0.05)
                        thrpt:  [+120.21% +123.56% +127.92%]
                        Performance has improved.
```

### DIFF
--- a/icechunk/src/format/mod.rs
+++ b/icechunk/src/format/mod.rs
@@ -503,6 +503,8 @@ pub mod format_constants {
     pub const ICECHUNK_COMPRESSION_ZSTD: &str = "zstd";
 }
 
+// The impl of Debug for Utf8UnixPathBuf is expensive and triggered often by tracing Spans
+// This implemnetation is much cheaper, and removes formatting from our samply profiles.
 impl fmt::Debug for Path {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         fmt::Display::fmt(self, f)


icechunk/src/format/manifest.rs
```diff
@@ -134,39 +134,44 @@ pub struct ManifestRef {
...
-pub struct ManifestSplits(pub Vec<ManifestExtents>);
+pub struct ManifestSplits(pub Vec<Vec<u32>>);
 
 impl ManifestSplits {
     /// Used at read-time
-    pub fn from_extents(extents: Vec<ManifestExtents>) -> Self {
-        assert!(!extents.is_empty());
-        Self(extents)
-    }
+    // pub fn from_extents(extents: Vec<ManifestExtents>) -> Self {
```
paraseba: delete

icechunk/src/session.rs
```diff
@@ -238,15 +238,17 @@ where
...
-    pub fn find<'a>(&'a self, coord: &'a ChunkIndices) -> Option<&'a ManifestExtents> {
-        debug_assert_eq!(coord.0.len(), self.0[0].len());
-        find_coord(self.iter(), coord).map(|x| x.1)
-    }
-
-    #[inline(always)]
-    pub fn position(&self, coord: &ChunkIndices) -> Option<usize> {
-        debug_assert_eq!(coord.0.len(), self.0[0].len());
-        find_coord(self.iter(), coord).map(|x| x.0)
+    pub fn find<'a>(&'a self, coord: &'a ChunkIndices) -> Option<ManifestExtents> {
```
paraseba, line 236: let's add some explanation to this algorithm

paraseba (approved):

dcherian:
@paraseba please review this last commit too: https://github.com/earth-mover/icechunk/pull/1764/changes/9cbeffcfd05af5b4d4b1cf29f9987250ac87ed1c

icechunk/src/session.rs
```diff
@@ -2062,47 +2062,42 @@ impl<'a> FlushProcess<'a> {
...
         for mref in previous_manifests {
-            let manifest =
-                fetch_manifest(&mref.object_id, old_snapshot, &self.asset_manager)
-                    .await?;
-
-            for item in manifest.iter(node_id.clone()) {
-                match item {
-                    Ok((idx, payload)) => {
-                        if !modified_chunks.contains_key(&idx) && extent.contains(&idx.0)
+            futs.push(fetch_manifest(&mref.object_id, old_snapshot, &self.asset_manager));
```
dcherian: I think having these in memory is fine? They're likely to be cached in any case

icechunk/src/session.rs
```diff
@@ -2062,47 +2062,42 @@ impl<'a> FlushProcess<'a> {
...
+            .await?;
 
-        // add modified chunks
-        for (idx, maybe_payload) in modified_chunks.iter() {
-            if let Some(payload) = maybe_payload.as_ref() {
-                all_chunks_vec.push(Ok(ChunkInfo {
-                    node: node_id.clone(),
-                    coord: idx.clone(),
-                    payload: payload.clone(),
-                }));
```
dcherian, line 2113: removed these clones since we've already cloned them earlier. Now we pass those in by value.

icechunk/src/format/mod.rs
```diff
@@ -503,6 +503,12 @@ pub mod format_constants {
     pub const ICECHUNK_COMPRESSION_ZSTD: &str = "zstd";
 }
 
+impl fmt::Debug for Path {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
paraseba, line 509: Can you add a comment explaining why this is preferred, or you risk people deleting it.

icechunk/src/session.rs
```diff
@@ -2059,47 +2062,42 @@ impl<'a> FlushProcess<'a> {
...
-                match item {
-                    Ok((idx, payload)) => {
-                        if !modified_chunks.contains_key(&idx) && extent.contains(&idx.0)
+            futs.push(fetch_manifest(&mref.object_id, old_snapshot, &self.asset_manager));
+        }
+        let mut all_chunks_vec = futs
+            .try_fold(Vec::new(), |mut acc, manifest| async {
+                acc.extend(manifest.iter(node_id.clone()).filter_map_ok(
+                    |(idx, payload)| {
+                        if extent.contains(&idx.0) && !modified_chunks.contains_key(&idx)
```
paraseba: Is this the right order for the `&&` operation? I think extent.contains will much most of the time, because most users don't have splitting?

icechunk/src/session.rs
paraseba: Hm ... I'm not sure, this potentially loads all manifests for the array. How about something like this: call `fetch_manifest` but don't return the manifest itself, just it's id. Then you call fetch manifest again, and those that fitted in the cache will be fast. We would have to think about ordering, should it be ordered features instead for this to work?

icechunk/src/session.rs
dcherian: you're right; i added a comment.

icechunk/src/session.rs
dcherian: OK i use `.buffer_unordered(1)` to have one in flight while we iterate through the other.

PR #1765, opened 2026-03-04
Allow relative virtual chunks

paraseba:
To use relative urls users need to:

* Create named virtual chunk containers
* When they add virtual references use locations with protocol `vcc://`
* Make the hostname in the url equal to some VCC name
* If validate_container=True when they set a virtual ref, and the url is `vcc://foo/bar.nc`, `foo` must match some VCC name. The final url will be `{foo url prefix}/bar.nc`


icechunk/src/format/manifest.rs
```diff
@@ -272,7 +278,47 @@ impl VirtualChunkLocation {
...
+            .filter(|s| !s.is_empty())
+            .collect::<Vec<_>>()
+            .join("/");
+        Ok(VirtualChunkLocation(format!(
+            "{VCC_RELATIVE_URL_SCHEME}{container_name}/{path}"
+        )))
+    }
+
+    /// Parse a location that may be either `vcc://` relative or an absolute URL.
+    pub fn from_url(path: &str) -> Result<VirtualChunkLocation, VirtualReferenceError> {
```
TomNicholas, line 301: This method is just for API backwards compatibility?

icechunk/src/format/manifest.rs
```diff
@@ -292,6 +338,8 @@ impl VirtualChunkLocation {
             host.to_string()
         } else if scheme == "file" {
             "".to_string()
+        } else if scheme == "vcc" {
+            return Err(VirtualReferenceErrorKind::NoContainerForName(path.into()).into());
```
TomNicholas, line 328: I'm confused - how does ::from_url not also trigger this Err?

icechunk/src/config.rs
```diff
@@ -575,10 +575,30 @@ impl RepositoryConfig {
...
 impl RepositoryConfig {
-    pub fn set_virtual_chunk_container(&mut self, cont: VirtualChunkContainer) {
-        let mut before = self.virtual_chunk_containers.clone().unwrap_or_default();
+    pub fn set_virtual_chunk_container(
+        &mut self,
+        cont: VirtualChunkContainer,
+    ) -> Result<(), String> {
+        let before = self.virtual_chunk_containers.clone().unwrap_or_default();
+        // Check name uniqueness: if the new container has a name, no existing
+        // container with the same name should have a different url_prefix
```
TomNicholas: I don't understand how its okay to have two containers with the same name at all. How does IC know which VCC to resolve to?

TomNicholas: Okay actually this is just for updating. But it's still confusing - when I overwrite an existing container by setting a new container with the same name as the old one, is that meant to be an overwrite (in which case there should be no restrictions), or are you trying to force me to delete the old container entirely before I can set the new one (in which case two containers with the same name should not be allowed ever).

icechunk/src/virtual_chunks.rs
```diff
@@ -1162,4 +1232,119 @@ mod tests {
...
+            s3_store_config(),
+        )
+        .unwrap();
+        let resolver = VirtualChunkResolver::new(
+            [cont].into_iter(),
+            Default::default(),
+            Default::default(),
+        );
+
+        let result = resolver.expand_location("vcc://unknown/chunk.nc");
```
TomNicholas, line 1321: Might also be worth testing case of

`"vcc://chunk.nc"`

where there is no name, but it's using a relative protocol.

icechunk/src/format/manifest.rs
paraseba: basically, yes, but it should work for normal and vcc

icechunk/src/format/manifest.rs
paraseba: If i understand your question correctly: Because you only get here if the url doesn't have  a host, which in the case of vcc:// is expected to be the container name

icechunk/src/config.rs
paraseba: So the issue here is that we have two identities for VCCs and they both need to work for different cases. The url prefix and the name. I changed the approach: if the container you pass here has a name it will be used as the key and replace any other of the same name. If it doesn't have a name, it will use the url prefix as key.

paraseba:
Thanks  for the thorough review @TomNicholas , I implemented what you asked for, and I think set_virtual_container is much more intuitive now. I'll merge once tests pass.

PR #1769, opened 2026-03-05
update version policy: major version tracks on-disk format

ianhi:
As discussed in the icechunk sync today

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -101,6 +101,7 @@ jobs:
           PYTHON_TAG="cp$(echo ${{ env.PYTHON_VERSION }} | tr -d '.')"
           WHEEL=$(ls dist/*-${PYTHON_TAG}-*.whl)
           uv pip install "$WHEEL" --group test
+          uv pip install "hypothesis @ git+https://github.com/ianhi/hypothesis.git@flaky-feedback#subdirectory=hypothesis-python"
           # Install specific versions based on matrix
           if [ "${{ matrix.deps-version.name }}" = "minimum" ]; then
             echo "Installing minimum versions:"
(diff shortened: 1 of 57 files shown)


icechunk-python/docs/docs/version-policy.md
```diff
@@ -1,82 +1,49 @@
...
-
-**Therefore any update to the Icechunk format specification version will be accompanied by a _major_ release of the icechunk python library.**
-
-For example, when the new [Icechunk version 2](https://earthmover.io/blog/evolving-our-tensor-storage-engine-a-preview-of-icechunk-2) format specification is first supported, it will be via a new major icechunk-python release, with a version updated from `1.x.y` to `2.0.0`.
+## Library - Data format compatibility
 
 ### Reading
 
-The latest release of `icechunk` Python will make best efforts to always be able to read data written against any previous version of the Icechunk Spec.
+Any major version of an Icechunk library can read data written with its own or any prior major version.
```
paraseba: I wouldn't like to make this promise. It means Icechunk code grows forever, we need to keep all the code to be able to read. Since we also want to test, we also need to keep around all the code needed to write.

Thoughts?

ianhi: to me this feels like the most important guarantee. its what i would want as a user. Also seems inline with equivalents. e.g. sqlite, hdf, iceberg

ianhi: >  Since we also want to test, we also need to keep around all the code needed to write.

i think this is not true. as long as we keep third-wheel around

paraseba: those examples are only partially true, but also, those projects have at least 10x the number of contributors that IC has.

From sqlite docs:
>  In other words, since 2004 all SQLite releases have been backwards compatible, though not necessarily forwards compatible. 

That leaves sqlite 1 and sqlite 2 out of the backwards compatibility window.

When formats are mature it's easier to offer this guarantees.

PR #1770, opened 2026-03-05
New inspect_manifest functionality

paraseba:


icechunk/src/inspect.rs
```diff
@@ -320,6 +321,83 @@ pub async fn repo_info_json(
...
+                (_, ChunkPayload::Virtual(_)) => num_virtual += 1,
+            }
+        }
+        arrays.push(ArrayManifestInspect {
+            node_id: node_id_str,
+            num_chunk_refs: num_inline + num_native + num_virtual,
+            num_inline,
+            num_native,
+            num_virtual,
+        });
```
rabernat, line 368: Sorry to just leave a drive-by comment but I randomly ended up on this PR and had thoughts...

Parquet files store all kinds of statistics in the footer so you don't have to actually read the whole file to know what's in it.

Couldn't we generate all of these stats _when we write the file_ and then store them in a header / footer? That way it's very cheap to get this information.

PR #1771, opened 2026-03-05
Honor can_write in Storage

### LINKED ISSUE #1371: Honor the `Storage::can_write` method
Find everywhere where we write and make sure have proper errors when the storage can not write

li-em:
Fix #1371

First commit tried to put checks in `put_object`, `delete_batches`, and `copy_object`, the lowest level methods in the Storage trait. But that was way too slow, especially with repo info updates (which can retry many times).

Moved them up the call stack to the `AssetManager` methods (`write_new_{manifest,snapshot,tx_log}`, `write_repo_info`, `write_chunk`), and to `garbage_collect` and `expire` in GC ops.

### DIFF
--- a/icechunk/src/ops/gc.rs
+++ b/icechunk/src/ops/gc.rs
@@ -375,7 +375,7 @@ pub async fn garbage_collect(
         .await
 }
 
-pub async fn garbage_collect_one_attempt(
+async fn garbage_collect_one_attempt(
     asset_manager: Arc<AssetManager>,
     config: &GCConfig,
     num_updates_per_repo_info_file: u16,
@@ -864,7 +864,7 @@ pub async fn expire_v2(
 }
 
 #[instrument(skip(asset_manager))]
-pub async fn expire_v2_one_attempt(
+async fn expire_v2_one_attempt(
     asset_manager: Arc<AssetManager>,
     older_than: DateTime<Utc>,
     expired_branches: ExpiredRefAction,
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -710,6 +710,12 @@ impl AssetManager {
         bytes: Bytes,
     ) -> RepositoryResult<()> {
         trace!(%chunk_id, size_bytes=bytes.len(), "Writing chunk");
+        if !self.storage.can_write().await? {
+            return Err(RepositoryErrorKind::ReadonlyStorage(
+                "Cannot write chunk".to_string(),
+            )
+            .into());
+        }
 
         let path = format!("{CHUNKS_FILE_PATH}/{chunk_id}");
         let _permit = self.request_semaphore.acquire().await?;


icechunk/src/ops/expiration_v1.rs
```diff
@@ -172,12 +171,6 @@ pub async fn expire(
     let storage = asset_manager.storage().as_ref();
     #[allow(deprecated)]
     let storage_settings = asset_manager.storage_settings();
-    if !storage.can_write().await? {
-        return Err(GCError::Repository(
-            RepositoryErrorKind::ReadonlyStorage("Cannot expire snapshots".to_string())
-                .into(),
-        ));
-    }
```
li-em, line 180: This check happens at the `expire` method in `icechunk/src/ops/gc.rs` now

icechunk/src/ops/gc.rs
```diff
@@ -373,13 +380,6 @@ pub async fn garbage_collect_one_attempt(
     config: &GCConfig,
     num_updates_per_repo_info_file: u16,
 ) -> GCResult<GCSummary> {
-    if !asset_manager.can_write_to_storage().await? {
-        return Err(GCError::Repository(
-            RepositoryErrorKind::ReadonlyStorage("Cannot garbage collect".to_string())
-                .into(),
-        ));
-    }
-
```
li-em, line 382: Check only once on the `garbage_collect` method, instead of at every attempt

paraseba: does this function need to be public? we should probably make it private, same for the equivalent for expiration

paraseba:
how about Session::set_chunk_ref ?

li-em:
> I think we are missing `Repository::writable_session` and `Repository::rearrange_session` maybe? Or maybe those already include the check?

They already include the check:
- https://github.com/earth-mover/icechunk/blob/54aedbdec0d1e5782e2104b7d5c22bcf1b962fbe/icechunk/src/repository.rs#L1795-L1801
- https://github.com/earth-mover/icechunk/blob/54aedbdec0d1e5782e2104b7d5c22bcf1b962fbe/icechunk/src/repository.rs#L1821-L1826

icechunk/src/ops/gc.rs
li-em: you mean `pub(crate)`, not `pub` for `asset_manager.can_write_to_storage`? or make `expire` and `garbage_collect` private in `icechunk/src/ops/gc.rs`?

li-em:
> how about Session::set_chunk_ref ?

Does `set_chunk_ref` write to storage directly, or only indirectly? The places I found have `.write_chunk` (which has the check as of latest commit) followed by `.set_chunk_ref`, or in the case of inline/virtual they would be written to the manifest (and tx_log), which would be covered in the `write_*` methods in the asset manager. Or did I miss something?

icechunk/src/ops/gc.rs
li-em: ah, sorry, it's the `_one_attempt` versions. Done!

paraseba (approved):
lgtm, let's merge

PR #1775, opened 2026-03-06
Optimize `get_existing_node`: avoid cloning `Vec<ManifestRef>`

dcherian:
Two big changes here for 2X improvements on appending split manifests.

- We avoid cloning `Vec<ManifestRef` in `get_existing_node`. This is a big win for our microbenchmarks.
- We avoid a surprisingly expensive equality check for `Path` by having `moved_from` and `moved_to` return Enums that let us know whether the node's path needs to be updated.

```
append_split_manifests/set/inline
                        change: [−43.791% −42.009% −40.391%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
append_split_manifests/commit/inline
                        change: [−42.075% −37.057% −31.993%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high severe
append_split_manifests/set/virtual
                        change: [−55.134% −54.300% −53.442%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
       - id: mixed-line-ending
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.8.0'
+    rev: 'v0.15.3'
     hooks:
       - id: ruff-format
       - id: ruff
--- a/icechunk-python/benchmarks/lib.py
+++ b/icechunk-python/benchmarks/lib.py
@@ -60,7 +60,7 @@ def get_task_chunk_shape(
         if c == s or left is None:
             task_chunk_shape.append(c)
         else:
-            q, r = divmod(s, c)
+            q, _ = divmod(s, c)
             if q > left:
                 task_chunk_shape.append(left * c)
             else:
--- a/icechunk-python/pixi.lock
+++ b/icechunk-python/pixi.lock
@@ -926,7 +926,7 @@ environments:
       - pypi: https://files.pythonhosted.org/packages/14/25/b208c5683343959b670dc001595f2f3737e051da617f66c31f7c4fa93abc/rich-14.3.3-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/04/54/6f679c435d28e0a568d8e8a7c0a93a09010818634c3c3907fc98d8983770/roman_numerals-4.1.0-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/b7/de/f7192e12b21b9e9a68a6d0f249b4af3fdcdff8418be0767a627564afa1f1/rpds_py-0.30.0-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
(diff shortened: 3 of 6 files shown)


icechunk/src/format/mod.rs
```diff
@@ -517,6 +517,26 @@ impl Display for Path {
     }
 }
 
+// Derived PartialEq on Path is surprisingly expensive.
+// PartialEq on the inner Utf8UnixPathBuf splits the paths in to components
+// and then checks equality!
+// Because our paths are canonicalized and absolute, we can impl a fastpath.
```
paraseba: I'll need to be convinced we never rely on the higher quality comparison

dcherian: My thinking is that we check for canonical & absolute path but yes, hard to prove. 

What do you think of having `moved_from` returning an Enum instead of Option

```
Moved(Path)
Unmoved
Deleted
```

Currently both Moved and Unmoved are modeled by `Some` and Deleted is `None`

icechunk/src/session.rs
```diff
@@ -1867,67 +1867,74 @@ async fn get_existing_node(
...
+        WasMoved::Deleted => unreachable!(),
+    };
+
+    match snapshot.get_node(renamed_path.as_ref()) {
+        Ok(node) => {
+            let node = Arc::unwrap_or_clone(node);
+
+            let node = match node.node_data {
+                // this overly verbose match arm allows us to avoid cloning `manifests`
+                // at all, by not holding a reference to `node`
```
dcherian: This is actually quite important for arrays with many manifests.

icechunk/src/change_set.rs
```diff
@@ -823,83 +844,88 @@ mod tests {
...
-            &Path::new("/other").unwrap(),
-        );
+        assert!(matches!(
+            mt.moved_from(&Path::new("/foo/bar/old/inner").unwrap()),
+            Moved(p) if p.as_ref() == &Path::new("/foo/bar/new/inner").unwrap()
+        ));
+        assert!(matches!(
+            mt.moved_from(&Path::new("/foo/bar/old").unwrap()),
+            Moved(p) if p.as_ref() == &Path::new("/foo/bar/new").unwrap()
+        ));
```
dcherian: the path changes are worrying; i can't figure out what the old version did

Cargo.toml
```diff
@@ -26,12 +26,24 @@ lto = "thin"
...
 opt-level = 2
 lto = "off"
 codegen-units = 8
 debug = 1
 
+# Optimized profile for benchmarking with maturin/PyO3 compatibility
+# maturin doesn't work well with --profile bench for ... reasons.
+# keep this in sync with `bench` above.
+[profile.perf]
+inherits = "release"
```
li-em, line 41: Can likely use `inherits = "bench"` and avoid problems keeping them in sync

dcherian: that doesn't work. `bench` uses the "test harness" which doesn't work with maturin according to claude.

PR #1776, opened 2026-03-06
Compress virtual chunk locations.

paraseba:
It only compresses on IC2, if there are enough virtual chunks and the
locations are long enough.

### DIFF
--- a/icechunk-python/stubtest_allowlist.txt
+++ b/icechunk-python/stubtest_allowlist.txt
@@ -134,8 +134,6 @@ icechunk._icechunk_python.RebaseFailedError
 icechunk._icechunk_python.RepoUpdateRetryConfig
 icechunk._icechunk_python.RepositoryConfig
 icechunk._icechunk_python.RepositoryConfig.previous_file
-icechunk._icechunk_python.RepositoryConfig.repo_update_retries
-icechunk._icechunk_python.S3Credentials
 icechunk._icechunk_python.S3Credentials.Anonymous
 icechunk._icechunk_python.S3Credentials.Anonymous.__match_args__
 icechunk._icechunk_python.S3Credentials.FromEnv


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -840,13 +840,58 @@ class ManifestSplittingConfig:
         """
         ...
 
+class ManifestVirtualChunkLocationCompressionConfig:
+    """Configuration for zstd dictionary-based compression of virtual chunk location URLs in manifests."""
+
+    def __new__(
+        cls,
+        min_virtual_chunks_to_compress: int | None = None,
+        dictionary_max_training_samples: int | None = None,
```
dcherian, line 848:
```suggestion
        *,
        dictionary_max_training_samples: int | None = None,
```

icechunk-python/src/config.rs
```diff
@@ -1453,32 +1453,116 @@ impl From<&PyManifestSplittingConfig> for ManifestSplittingConfig {
...
+    #[pyo3(get, set)]
+    pub dictionary_max_size_bytes: Option<u32>,
+    #[pyo3(get, set)]
+    pub compression_level: Option<i32>,
+}
+
+#[pymethods]
+impl PyManifestVirtualChunkLocationCompressionConfig {
+    #[new]
+    #[pyo3(signature = (min_virtual_chunks_to_compress=None, dictionary_max_training_samples=None, dictionary_max_size_bytes=None, compression_level=None))]
```
dcherian:
```suggestion
    #[pyo3(signature = (min_virtual_chunks_to_compress=None, *, dictionary_max_training_samples=None, dictionary_max_size_bytes=None, compression_level=None))]
```

dcherian: Perhaps this should be `min_num_chunks` instead of `min_virtual_chunks_to_compress`. The "virtual" and "compress" is already in `ManifestVirtualChunkLocationCompressionConfig`

icechunk-python/src/config.rs
```diff
@@ -1453,32 +1453,116 @@ impl From<&PyManifestSplittingConfig> for ManifestSplittingConfig {
...
+    #[pyo3(get, set)]
+    pub virtual_chunk_location_compression:
+        Option<Py<PyManifestVirtualChunkLocationCompressionConfig>>,
 }
 
 #[pymethods]
 impl PyManifestConfig {
     #[new]
-    #[pyo3(signature = (preload=None, splitting=None))]
+    #[pyo3(signature = (preload=None, splitting=None, virtual_chunk_location_compression=None))]
```
dcherian, line 1547:
```suggestion
    #[pyo3(signature = (*, preload=None, splitting=None, virtual_chunk_location_compression=None))]
```

icechunk/flatbuffers/manifest.fbs
```diff
@@ -51,6 +57,15 @@ table Manifest {
...
   arrays: [ArrayManifest] (required);
+
+  // zstd dictionary used to decompress compressed_location fields in ChunkRefs
+  // Introduced in spec version 2.0
+  location_dictionary: [uint8];
+
+  // compression algorithm used for compressed_location fields
+  // 0 = none, 1 = zstd dictionary compression
+  // Introduced in spec version 2.0
+  compression_algorithm: uint8 = 1;
```
dcherian, line 76: A spec file should be updated. Are you planning to do that in one go at the end?

icechunk/src/format/manifest.rs
```diff
@@ -377,6 +378,10 @@ pub struct ChunkInfo {
     pub payload: ChunkPayload,
 }
 
+const COMPRESSION_ALG_NONE: u8 = 0;
+const COMPRESSION_ALG_ZSTD_DICT: u8 = 1;
+const MAX_DECOMPRESSED_LOCATION_SIZE: usize = 1_024;
+
```
dcherian, line 385: Descriptive comment here for `MAX_DECOMPRESSED_LOCATION_SIZE` would be helpful. I started here and it's not clear what these mean. Should `COMPRESSION_ALG` be an enum instead?

icechunk/src/format/manifest.rs
```diff
@@ -408,22 +413,56 @@ impl Manifest {
...
+    }
+
     pub fn from_sorted_vec(
         manifest_id: &ManifestId,
         sorted_chunks: Vec<ChunkInfo>,
-    ) -> Option<Self> {
+        virtual_chunks_compression_config: Option<
+            &ManifestVirtualChunkLocationCompressionConfig,
+        >,
+    ) -> IcechunkResult<Option<Self>> {
```
dcherian, line 442: ~What can cause a failure now?~

I see it's `train_location_dictionary` (so hard to spot a `?` in the diff :) )

icechunk/src/format/manifest.rs
```diff
@@ -492,11 +556,29 @@ impl Manifest {
         self.root().arrays().iter().map(|am| NodeId::from(am.node_id().0))
     }
 
+    pub fn uses_location_compression(&self) -> bool {
+        self.root().compression_algorithm() == COMPRESSION_ALG_ZSTD_DICT
```
dcherian:
```suggestion
        self.root().compression_algorithm() != COMPRESSION_ALG_NONE
```

Is this more future-proof?

icechunk/src/format/manifest.rs
```diff
@@ -408,22 +413,56 @@ impl Manifest {
...
+            &ManifestVirtualChunkLocationCompressionConfig,
+        >,
+    ) -> IcechunkResult<Option<Self>> {
+        let location_compression_dict =
+            train_location_dictionary(&sorted_chunks, virtual_chunks_compression_config)?;
+        // we generate all compressed locations if needed
+        let compressed_locations =
+            match (&location_compression_dict, virtual_chunks_compression_config) {
+                (Some(d), Some(config)) => compress_locations(&sorted_chunks, d, config),
+                _ => vec![None; sorted_chunks.len()],
```
dcherian, line 449: Seems weird to me that the config can be None here. Doesn't that mean the default, which is to compress? Did I misunderstand the change to config.rs?

icechunk/src/format/manifest.rs
dcherian: Hehe I guess `MAX_DECOMPRESSED_LOCATION_SIZE` is max length of path in bytes; definitely needs a comment.

icechunk/src/format/manifest.rs
```diff
@@ -578,6 +671,7 @@ impl Iterator for PayloadIterator {
 
 fn ref_to_payload(
     chunk_ref: generated::ChunkRef<'_>,
+    decompressor: Option<&mut zstd::bulk::Decompressor<'static>>,
```
dcherian, line 680: so weird it neds to be mutable.

icechunk/src/format/manifest.rs
dcherian: Aha, the default is not to compress! Would be good to explicitly write that out somewhere, maybe in `config.rs`

icechunk/src/format/manifest.rs
```diff
@@ -618,9 +731,131 @@ fn checksum(payload: &generated::ChunkRef<'_>) -> Option<Checksum> {
...
+                let j = rng.random_range(0..=virtual_count);
+                if j < max_samples {
+                    reservoir[j] = loc;
+                }
+            }
+            virtual_count += 1;
+        }
+    }
+
+    if virtual_count < min_chunks {
```
dcherian, line 787: we can early-exit if `chunks.len() < min_chunks` , though this may have been checked earlier

icechunk/src/format/manifest.rs
```diff
@@ -618,9 +731,131 @@ fn checksum(payload: &generated::ChunkRef<'_>) -> Option<Checksum> {
...
+        return Ok(None);
+    }
+
+    let sample_bytes: Vec<&[u8]> = reservoir.iter().map(|s| s.as_bytes()).collect();
+
+    // zstd doesn't like it if many samples are too small, in that case we don't compress
+    let small_count = sample_bytes.iter().filter(|s| s.len() < 8).count();
+    if small_count >= sample_bytes.len() / 2 {
+        tracing::warn!(
+            "Skipping zstd dictionary training: at least half of the {} samples are smaller than 8 bytes",
```
dcherian:
```suggestion
            "Skipping virtual chunk location compression: at least half of the {} samples are smaller than 8 bytes",
```

icechunk/src/format/manifest.rs
```diff
@@ -618,9 +731,131 @@ fn checksum(payload: &generated::ChunkRef<'_>) -> Option<Checksum> {
...
+
+/// Compress virtual chunk locations in parallel using a pre-trained zstd dictionary.
+///
+/// Returns one entry per chunk: `Some(compressed_bytes)` for virtual chunks,
+/// `None` for inline/ref chunks.
+fn compress_locations(
+    chunks: &[ChunkInfo],
+    dict: &[u8],
+    config: &ManifestVirtualChunkLocationCompressionConfig,
+) -> Vec<Option<Vec<u8>>> {
```
dcherian, line 830: this could fail while compressing no?

icechunk/src/format/manifest.rs
```diff
@@ -618,9 +731,131 @@ fn checksum(payload: &generated::ChunkRef<'_>) -> Option<Checksum> {
...
+                        })
+                        .collect::<Vec<_>>()
+                })
+            })
+            .collect();
+
+        #[allow(clippy::expect_used)]
+        handles
+            .into_iter()
+            .flat_map(|h| h.join().expect("Cannot join threads compressing locations"))
```
dcherian:
```suggestion
            .flat_map(|h| h.join().expect("Cannot join threads compressing virtual chunk locations"))
```

icechunk/src/format/manifest.rs
dcherian: I see we fall back to uncompressed which is `None`. A commment describing return type would be helpful

icechunk/src/repository.rs
```diff
@@ -3811,4 +3816,121 @@ mod tests {
...
+        }
+
+        // Migrate to IC2
+        migrate_1_to_2(repo, false, true).await.unwrap();
+        let repo =
+            Repository::open(Some(config), Arc::clone(&storage), HashMap::new()).await?;
+        assert_eq!(repo.spec_version(), SpecVersionBin::V2dot0);
+
+        // Rewrite manifests (now with IC2 compression enabled)
+        rewrite_manifests(
```
dcherian, line 3968: nice!

icechunk/flatbuffers/manifest.fbs
paraseba: yeah ... the spec is very outdated

icechunk-python/src/config.rs
paraseba: I think this would break api

icechunk/src/format/manifest.rs
paraseba: The intention was the default to be compression for IC2, and of course we cannot compress in IC1. Did I get it wrong?

icechunk/src/format/manifest.rs
paraseba: yeah, decompress takes mut self ... weird I bet related to C API

icechunk/src/format/manifest.rs
paraseba: great!

icechunk-python/src/config.rs
dcherian: My personal preference is that this is an OK break. At the very least we can stick `*` `after splitting`

PR #1778, opened 2026-03-06
Warm snapshot cache to speed up migration

mpiannucci:
The migration now spawns a background task that lists all snapshot objects from storage, sorts them by creation time (latest first), and fetches them N at a time to warm the asset manager cache. This makes the sequential ancestry walk mostly hits cached snapshots instead of making individual network round-trips

### DIFF
--- a/icechunk-python/docs/docs/virtual.md
+++ b/icechunk-python/docs/docs/virtual.md
@@ -164,6 +164,10 @@ ds.sst.isel(time=26, zlev=0).plot(x='lon', y='lat', vmin=0)
 
 While `VirtualiZarr` is the easiest way to create virtual datasets with Icechunk, the Store API that it uses to create the datasets in Icechunk is public. `IcechunkStore` contains a [`set_virtual_ref`](./reference.md#icechunk.IcechunkStore.set_virtual_ref) method that specifies a virtual ref for a specified chunk.
 
+!!! important "URL prefix must end with a trailing slash"
+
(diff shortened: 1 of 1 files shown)


icechunk/src/migrations/mod.rs
```diff
@@ -381,10 +381,68 @@ pub async fn migrate_1_to_2(
...
+    // Prefetch snapshots concurrently to warm the asset manager cache.
+    // We list all snapshot objects, sort by creation time (latest first),
+    // and fetch them N-at-a-time in the background while the ancestry
+    // walk proceeds. This turns what would be sequential network
+    // round-trips into mostly cache hits.
+    let asset_manager = Arc::clone(repo.asset_manager());
+    let prefetch_handle = {
+        let am = Arc::clone(&asset_manager);
+        tokio::spawn(async move {
+            const PREFETCH_CONCURRENCY: usize = 10;
```
mpiannucci: Making this 64 made my migration time 1/5 on my local machine but im not sure what the right default is or a place for a config for this would be

mpiannucci:
One downside of this appraoch: we now fetch every snapshot in the store up front, but migration still walks reachable ancestry again afterward. Since the snapshot cache is size-limited by weighted node count, not snapshot count, large repos will churn the cache and often pay for these reads twice, including for dangling snapshots the migration never needs.

icechunk/src/migrations/mod.rs
paraseba: maybe the migration function needs to take more options. 10 sounds low to me, I think 64 is ok

paraseba: I think having this as an option is important in case there are repos that cannot fit 64 snapshots in memory

PR #1780, opened 2026-03-06
Backports for IC1

ianhi:
backport several PRs (1 PR per commit)

mostly. clean backports. for a few there were a few changes necessary:

  #1699 (storage defaults bug fix)

  - No async default_settings(): On v1.x, storage.default_settings() is sync. On main it's
  storage.default_settings().await?. So the fix pattern is the same but without await.
  - No V2 code paths: Main has separate SpecVersionBin::V2 branches in open() for reading config from
  repo info objects. V1.x only has the config.yaml path, so the fix is simpler — just three call sites
  (create(), new(), store_config()) vs main's additional V2 open path and reopen().
  - Test approach: Main's test directly checks repo.storage_settings (exposed property). V1.x doesn't
  expose that, so we test via round-trip write/read with partial storage config to verify it works
  end-to-end.

  #1742 (delete + recreate in rebase)

  - Removed the moves() guard: Main adds an early return checking
  previous_change.moves().next().is_some() and returning MoveOperationCannotBeRebased. Moves don't exist
  in IC1, so this was dropped entirely. Only the deleted_paths filter on new_nodes() was backported.


dcherian:
We can grab #1795 too and see if that makes the AL integration tests happier.

PR #1782, opened 2026-03-06
Add space in the format to store extra information

paraseba:
## Description

We add a new `extra` field pretty much where it makes sense. There is no significant space impact because these are all flatbuffers tables and the field is optional.

Code-wise, we don't do anything. All these fields will remain empty until we find a use for them after Icechunk 2.0.


ianhi:
image
?

paraseba:
@ianhi jujutsu ftw

PR #1784, opened 2026-03-06
Fix snapshot node cache.

dcherian:
The `quick_cache` library we use has a footgun where caches of size 1 are caches of size 0. Instead we size the node cache on a Snapshot to 2.

https://github.com/arthurprs/quick-cache/issues/105

### DIFF
--- a/icechunk/src/format/snapshot.rs
+++ b/icechunk/src/format/snapshot.rs
@@ -308,14 +308,19 @@ impl ManifestFileInfo {
     }
 }
 
+// It is _extremely_ common to set/get chunks in large batches sequentially.
+// Without a cache we incur the cost of repeatedly decoding the flatbuffer.
+// This cache is very small (size 2 currently), because we could cache a large number of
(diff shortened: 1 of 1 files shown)


paraseba (approved):

icechunk/src/format/snapshot.rs
```diff
@@ -308,14 +308,19 @@ impl ManifestFileInfo {
...
 }
 
+// It is _extremely_ common to set/get chunks in large batches sequentially.
+// Without a cache we incur the cost of repeatedly decoding the flatbuffer.
+// This cache is very small (size 2 currently), because we could cache a large number of
+// snapshots. The size of 2 is fine for typically workloads that iterate sequentially
+// through nodes.
+// We cannot choose 1 because of a bug in the `quick_cache` dependency
+// where a size-1 cache is effectively no cache
+// https://github.com/arthurprs/quick-cache/issues/105
```
mpiannucci, line 318: What a find

PR #1787, opened 2026-03-09
[fix] Update repo info cache when changes made

paraseba:
## Description
Also in this PR tracing logs to know if repo info cache is in use


icechunk/src/asset_manager.rs
```diff
@@ -633,9 +649,16 @@ impl AssetManager {
             )
```
li-em, line 649: An alternative is to call `self.write_repo_info` here, which does the cache update

PR #1793, opened 2026-03-09
Minimum supported Rust version policy

li-em:
Drafting an MSRV policy after https://github.com/earth-mover/icechunk/pull/1786, where `rust-version` was set for each crate.


Cargo.toml
```diff
@@ -3,6 +3,10 @@ members = ["icechunk", "icechunk-python", "icechunk-macros"]
 default-members = ["icechunk"]
 resolver = "2"
 
+[workspace.package]
+rust-version = "1.91.1"
+edition = "2024"
+
```
li-em, line 9: Opted for moving these config keys to the workspace, and use `rust-version.workspace = true` in each member `Cargo.toml`, so we have to update in just one place.

.github/workflows/rust-ci.yaml
```diff
@@ -50,6 +50,13 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 
+      - name: Check if MSRV is up to date
+        run: |
+          set -e
+          grep 'rust-version = "${{ env.RUST_CHANNEL }}"' Cargo.toml
+          # shellcheck disable=SC2016
+          grep 'The current MSRV is `${{ env.RUST_CHANNEL }}`' icechunk-python/docs/docs/contributing.md
```
li-em, line 58: A quick check just so we don't forget to update config and docs when bumping `RUST_CHANNEL` here

flake.nix
```diff
@@ -116,6 +116,7 @@
                 pkgs.cargo-nextest # test runner
                 pkgs.cargo-deny
                 pkgs.cargo-edit
+                pkgs.cargo-msrv
```
li-em, line 119: This is useful to check how deps are treating MSRV, and if we are holding too many updates by staying on an older version.

Cargo.toml
li-em: we can move a bunch more to avoid duplication in members: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
(but in a future PR)

paraseba (approved):

PR #1794, opened 2026-03-09
Add TimeoutSettings to storage and wire through to S3 client

dcherian:
Adds `connect_timeout_ms`, `read_timeout_ms`, `operation_timeout_ms`, and `operation_attempt_timeout_ms` to storage Settings. These are applied to the AWS SDK `TimeoutConfig` when constructing the S3 client, giving control over SDK-level timeouts separate from retry settings.

## Description

I added this when debugging the 0B/s errors but forgot to push earlier. It helped surface #1700 too because the AWS logging prints out timeout info as soon as the client is initialized :)

No strong feelings on merging, but it seems useful to have this be configurable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2082,7 +2082,7 @@ dependencies = [
 
 [[package]]
 name = "icechunk"
-version = "0.3.14"
+version = "2.0.0-alpha.3"
 dependencies = [
  "anyhow",
  "assert_fs",
@@ -2157,7 +2157,7 @@ dependencies = [
 
 [[package]]
 name = "icechunk-python"
-version = "2.0.0-alpha.0"
+version = "2.0.0-alpha.3"
 dependencies = [
  "async-stream",
  "async-trait",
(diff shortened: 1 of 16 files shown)


icechunk-python/src/config.rs
```diff
@@ -991,12 +1062,13 @@ impl Eq for PyStorageSettings {}
 
 #[pymethods]
 impl PyStorageSettings {
-    #[pyo3(signature = ( concurrency=None, retries=None, unsafe_use_conditional_create=None, unsafe_use_conditional_update=None, unsafe_use_metadata=None, storage_class=None, metadata_storage_class=None, chunks_storage_class=None, minimum_size_for_multipart_upload=None))]
+    #[pyo3(signature = ( concurrency=None, retries=None, timeouts=None, unsafe_use_conditional_create=None, unsafe_use_conditional_update=None, unsafe_use_metadata=None, storage_class=None, metadata_storage_class=None, chunks_storage_class=None, minimum_size_for_multipart_upload=None))]
```
paraseba: since we don't have a `*` should this go at the end to maintain compatibility?

PR #1796, opened 2026-03-09
bump all versions to 2.0.0-alpha3

ianhi:
## Description

bump all version to 2.0.0-alpha.3 which is the next tag in the series.

I used a version of deepaks xarray release claude skill to compile the changelog

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -854,7 +854,7 @@ class ManifestVirtualChunkLocationCompressionConfig:
 
         Parameters
         ----------
-        num_chunks: int | None
+        min_num_chunks: int | None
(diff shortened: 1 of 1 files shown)


Changelog.python.md
```diff
@@ -1,5 +1,42 @@
 # Changelog
 
+## Icechunk 2.0.0-alpha.3
```
paraseba: Is the intention to list all new features in IC2? Maybe it's better to keep alphas outside of the main changelog for now?

paraseba:
@li-em what do you think of this move to alpha on the rust side?

li-em:
> @li-em what do you think of this move to alpha on the rust side?

I did a test publishing a new crate, and it only picks alpha versions if you are explicit about it, and `0.x` is considered stable too. So anyone trying to `cargo add icechunk` will get `0.3.22` instead of the new alpha

PR #1797, opened 2026-03-09
Check repository status before doing any operations

### LINKED ISSUE #1335: Add repo status
Offline, readonly, etc

li-em:
Fix #1335

Expose RepoStatus from the flatbuffers definition all the way to Python.

Add checks for status at session creation (writable and rearrange), and GC operations (garbage collection and expire)

### DIFF
--- a/icechunk/src/storage/latency.rs
+++ b/icechunk/src/storage/latency.rs
@@ -175,32 +175,14 @@ mod tests {
     use crate::storage::new_in_memory_storage;
 
     #[tokio::test]
-    async fn test_latency() {
+    async fn storage_with_latency() {
         let backend = new_in_memory_storage().await.unwrap();
         let storage = LatencyStorage::new(backend, 0, 0);
         let settings = storage.default_settings().await.unwrap();
 
-        // Write a test object so we can read it back
-        storage
-            .put_object(&settings, "test/key", "hello".into(), None, vec![], None)
-            .await
-            .unwrap();
-
-        // Baseline: measure write and read with no latency
-        let start = std::time::Instant::now();
-        storage
-            .put_object(&settings, "test/key", "hello".into(), None, vec![], None)
-            .await
-            .unwrap();
-        let write_baseline = start.elapsed();
-
-        let start = std::time::Instant::now();
-        let _ = storage.get_object_range(&settings, "test/key", None).await.unwrap();
-        let read_baseline = start.elapsed();
-
-        // Add 50ms write latency, 30ms read latency
-        storage.set_write_delay_ms(50);
-        storage.set_read_delay_ms(30);
+        // Add 500ms write latency, 300ms read latency
+        storage.set_write_delay_ms(500);
+        storage.set_read_delay_ms(300);
 
         let start = std::time::Instant::now();
         storage
@@ -213,17 +195,14 @@ mod tests {
         let _ = storage.get_object_range(&settings, "test/key", None).await.unwrap();
         let read_with_latency = start.elapsed();
 
-        let write_added = write_with_latency.saturating_sub(write_baseline);
-        let read_added = read_with_latency.saturating_sub(read_baseline);
-
         // Allow some wiggle room for scheduling jitter
         assert!(
-            write_added.as_millis() >= 45,
-            "expected ~50ms write latency, got {write_added:?}"
+            write_with_latency.as_millis() >= 500,
+            "expected ~500ms write latency, got {write_with_latency:?}"
         );
         assert!(
-            read_added.as_millis() >= 25,
-            "expected ~30ms read latency, got {read_added:?}"
+            read_with_latency.as_millis() >= 300,
+            "expected ~300ms read latency, got {read_with_latency:?}"
         );
     }
 }


icechunk-python/src/repository.rs
```diff
@@ -447,6 +447,84 @@ impl PyGCSummary {
 
 impl_pickle!(PyGCSummary);
 
+#[pyclass(name = "RepoAvailability", eq, eq_int)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) enum PyRepoAvailability {
+    Online,
+    ReadOnly,
+    Offline,
```
paraseba: I propose we leave `Offline` out for now

paraseba: Or maybe we keep it and we throw an informative "not implemented yet" error if they try to set the repo offline?

icechunk/src/format/repo_info.rs
```diff
@@ -19,6 +19,40 @@ use super::{
...
+            RepoAvailability::ReadOnly => generated::RepoAvailability::ReadOnly,
+            RepoAvailability::Offline => generated::RepoAvailability::Offline,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RepoStatus {
+    pub availability: RepoAvailability,
+    pub set_at: u64,
```
paraseba: I think set_at should be a `DateTime<Utc>` here?

icechunk/src/format/repo_info.rs
```diff
@@ -478,7 +519,7 @@ impl RepoInfo {
             &Default::default(),
             UpdateInfo {
                 update_type: UpdateType::RepoInitializedUpdate,
-                update_time: update_time.unwrap_or(Utc::now()),
+                update_time: update_time.unwrap_or_else(Utc::now),
```
li-em, line 554: I updated `.unwrap_or` calls in this file to use `.unwrap_or_else` instead, which is a small optimization:

https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or
> Arguments passed to unwrap_or are eagerly evaluated; if you are passing the result of a function call, it is recommended to use [unwrap_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else), which is lazily evaluated.

icechunk/src/format/repo_info.rs
```diff
@@ -19,6 +19,40 @@ use super::{
 use chrono::{DateTime, Utc};
 use flatbuffers::{VerifierOptions, WIPOffset};
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum RepoAvailability {
+    Online,
```
paraseba, line 25: awesome

icechunk/src/format/repo_info.rs
```diff
@@ -19,6 +19,40 @@ use super::{
 use chrono::{DateTime, Utc};
 use flatbuffers::{VerifierOptions, WIPOffset};
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
```
paraseba, line 23: i don't remember ... why do we need Serialize/Deserialize here?

icechunk/src/format/repo_info.rs
```diff
@@ -487,6 +528,11 @@ impl RepoInfo {
             config_bytes.as_deref(),
             None::<std::iter::Empty<u16>>,
             None::<std::iter::Empty<u16>>,
+            &RepoStatus {
+                availability: RepoAvailability::Online,
+                set_at: update_time.unwrap_or_else(Utc::now),
```
paraseba, line 565: I think this should be the update_time you get as argument and not `now`

icechunk/src/format/repo_info.rs
```diff
@@ -1016,6 +1087,37 @@ impl RepoInfo {
...
+        let snaps: Vec<_> = self.all_snapshots()?.try_collect()?;
+        Self::from_parts(
+            spec_version,
+            self.all_tags()?,
+            self.all_branches()?,
+            self.deleted_tags()?,
+            snaps,
+            &self.metadata()?,
+            UpdateInfo {
+                update_type: UpdateType::ConfigChangedUpdate,
```
paraseba: ahhh hm ... I  think we should add a new udpate type for this. But feel free to create a ticket and do it in a follow up PR

icechunk/src/migrations/mod.rs
```diff
@@ -532,6 +532,11 @@ pub async fn migrate_1_to_2(
         Some(config_bytes.as_slice()),
         None::<std::iter::Empty<u16>>,
         None::<std::iter::Empty<u16>>,
+        &RepoStatus {
+            availability: RepoAvailability::Online,
+            set_at: Utc::now(),
```
paraseba: How feasible would it be to put in here, instead of now(), the timestamp of the first snapshot in the repo? I don't want to traverse the whole ancestry again, we would need to get it as we migrate above.

icechunk/src/repository.rs
```diff
@@ -764,6 +766,40 @@ impl Repository {
         Ok(())
     }
 
+    #[instrument(skip_all)]
+    pub async fn get_status(&self) -> RepositoryResult<RepoStatus> {
+        self.asset_manager().fail_unless_spec_at_least(SpecVersionBin::V2dot0)?;
```
paraseba, line 753: great!

icechunk/src/repository.rs
```diff
@@ -764,6 +766,40 @@ impl Repository {
...
+        let (repo, _) = self.asset_manager().fetch_repo_info().await?;
+        Ok(repo.status()?)
+    }
+
+    #[instrument(skip(self, status))]
+    pub async fn set_status(&self, status: &RepoStatus) -> RepositoryResult<()> {
```
paraseba, line 759: don't skip status in the instrumentation, it's useful.

icechunk/src/repository.rs
```diff
@@ -1800,6 +1836,15 @@ impl Repository {
             )
             .into());
         }
+        if self.spec_version() >= SpecVersionBin::V2dot0 {
```
paraseba: it feels like we need a function like `fail_if_status_readonly()?`

icechunk/src/format/repo_info.rs
li-em: The reason was missing in this draft: printing the status in Python (with `inspect` and `UpdateType`)

icechunk/src/format/repo_info.rs
li-em: Done in this PR

icechunk/src/format/repo_info.rs
li-em: Sorry, but this is already the `update_time` we get as an argument, no?

icechunk/src/repository.rs
li-em: ended up calling it `fail_unless_online_status`, to try to account for offline in the future

icechunk/src/asset_manager.rs
```diff
@@ -619,6 +620,7 @@ impl AssetManager {
             &str,
             VersionInfo,
         ) -> RepositoryResult<Arc<RepoInfo>>,
+        skip_online_check: bool,
```
li-em, line 653: @paraseba ended up adding this as an argument for `update_repo_info`, and only place where this is `true` is when calling `repository.set_status`. Updated all other calls to `false`

icechunk-python/src/repository.rs
li-em: Left offline out for now

icechunk/src/migrations/mod.rs
li-em: Does https://github.com/earth-mover/icechunk/pull/1797/commits/8dbb0867e8668be7e3e27dda474a76b72e1fa283 match your thinking?

icechunk-python/tests/test_session.py
```diff
@@ -274,3 +276,111 @@ def test_repository_open_no_list_bucket(any_spec_version: int | None) -> None:
         assert repo.list_branches() == set(["main", "new_branch"])
         assert list(repo.list_tags()) == ["new_tag"]
         assert len(list(repo.ops_log())) == 5
+
+
+def test_readonly_repo_status_blocks_writable_session() -> None:
+    repo = Repository.create(
```
paraseba, line 285: can we test that it blocks branch/tag creation, gc, and commit on a session that was open before it became readonly?

icechunk/src/format/repo_info.rs
paraseba: ups, sorry

icechunk/src/ops/gc.rs
```diff
@@ -332,6 +332,19 @@ pub async fn garbage_collect(
         ));
     }
 
+    // Check repo status (only available on IC2)
+    if matches!(asset_manager.spec_version(), SpecVersionBin::V2dot0) {
```
paraseba: We should use the Ord instance here, anything > 1.0

icechunk/src/asset_manager.rs
paraseba: looks great

icechunk/src/asset_manager.rs
```diff
@@ -640,6 +642,14 @@ impl AssetManager {
         let mut attempts: u64 = 1;
         loop {
             let (repo_info, repo_version) = self.fetch_repo_info().await?;
+            if !skip_online_check
+                && repo_info.status()?.availability != RepoAvailability::Online
+            {
+                return Err(RepositoryErrorKind::ReadonlyRepository(
+                    "Cannot update repo info".to_string(),
```
paraseba: let's give them a better explanation that includes in the text: the current status, when was set, and the reason if available

icechunk/src/repository.rs
```diff
@@ -1774,6 +1850,19 @@ impl Repository {
         Ok(session)
     }
 
+    async fn fail_unless_online_status(&self, error_msg: &str) -> RepositoryResult<()> {
+        if self.spec_version() >= SpecVersionBin::V2dot0 {
+            let status = self.get_status().await?;
+            if status.availability != RepoAvailability::Online {
+                return Err(RepositoryErrorKind::ReadonlyRepository(
```
paraseba: see my other comment on giving them a better explanation

icechunk/src/repository.rs
```diff
@@ -741,7 +751,49 @@ impl Repository {
...
+                backup_path,
+                num_updates,
+            )?))
+        };
+
+        let _ = self
+            .asset_manager
+            .update_repo_info(
+                self.config.repo_update_retries().retries(),
+                do_update,
```
paraseba: let's add a comment to this `true` explaining why we need it

icechunk/src/repository.rs
li-em: Added an `error_msg` in `RepoStatus` to print it consistently when something goes wrong

icechunk/src/repository.rs
li-em: I reverted the change in `update_repo_info` having a new bool argument, and instead added an unsafe `update_repo_info_unchecked` method instead. It is too easy to flip the boolean and do something wrong in the future... This forces an `unsafe {}` block and a comment justifying why the unchecked method is being called.

icechunk-python/tests/test_session.py
li-em: Expanded the 4 current tests to check this by opening a session, changing status, and then trying operations. Then change status back to online, and ops should work again.

icechunk/src/storage/latency.rs
```diff
@@ -213,17 +195,14 @@ mod tests {
...
         let read_with_latency = start.elapsed();
 
-        let write_added = write_with_latency.saturating_sub(write_baseline);
-        let read_added = read_with_latency.saturating_sub(read_baseline);
-
         // Allow some wiggle room for scheduling jitter
         assert!(
-            write_added.as_millis() >= 45,
-            "expected ~50ms write latency, got {write_added:?}"
+            write_with_latency.as_millis() >= 495,
```
ianhi:
```suggestion
            write_with_latency.as_millis() >= 500,
```

icechunk/src/storage/latency.rs
```diff
@@ -213,17 +195,14 @@ mod tests {
...
+            "expected ~50ms write latency, got {write_with_latency:?}"
         );
         assert!(
-            read_added.as_millis() >= 25,
-            "expected ~30ms read latency, got {read_added:?}"
+            read_with_latency.as_millis() >= 295,
+            "expected ~30ms read latency, got {read_with_latency:?}"
```
ianhi:
```suggestion
            write_with_latency.as_millis() >= 500,
            "expected ~50ms write latency, got {write_with_latency:?}"
);
assert!(
            read_with_latency.as_millis() >= 300,
            "expected ~30ms read latency, got {read_with_latency:?}"
```

PR #1809, opened 2026-03-11
Fix compat config

mpiannucci:
we dont want this

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,9 +198,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-config"
-version = "1.8.12"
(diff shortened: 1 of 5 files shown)


icechunk/src/repository.rs
```diff
@@ -389,24 +389,14 @@ impl Repository {
 
         let (persisted_config, config_version) = if spec_version >= SpecVersionBin::V2dot0
         {
-            // For V2+ repos, read config from the repo info object.
-            // Only fall back to config.yaml for old V2 repos that predate embedded config.
+            // V2+ repos: config is always embedded in the repo info object.
             let (repo_info, _) = temp_am.fetch_repo_info().await?;
             match repo_info.config()? {
```
paraseba: I don't think this `match` make sense anymore.  For V2 you can just do `(repo_info.config(), for_creation())` and you are done

PR #1815, opened 2026-03-12
Check for some important conflicts on session merge

paraseba:
The checks are not complete, but are much better than they used to be in IC1.

Fixes: #1814


icechunk-python/tests/test_dask.py
```diff
@@ -96,7 +96,7 @@ def test_xarray_to_icechunk_nested_pickling(
 
             newds = ds + 1
             session = repo.writable_session("main")
-            to_icechunk(newds, session=session, mode="w")
+            to_icechunk(newds, session=session, mode="a")
```
paraseba: @dcherian could you check this? My new conflict checks don't let this pass, but I'm not sure if this is correct, and if this would be too much of an API change.

icechunk-python/tests/test_distributed_append.py
```diff
@@ -16,9 +16,12 @@
 PLOT = False
 SPLIT_EVERY = 128
 DIMS = ("x", "y")
-IC_STORAGE = ic.local_filesystem_storage(
-    f"/tmp/test/icechunk_data_corrupted/{str(datetime.datetime.now()).split(' ')[-1]}",
-)
+
+
+def mk_ic_storage():
+    return ic.local_filesystem_storage(
```
paraseba, line 22: Same here @dcherian , but I'm more confident this was a test bug

icechunk-python/tests/test_dask.py
dcherian: the intent with mode="w" is clear the store first, then write. I think `w-` would raise if the store isn't empty. So this is a BIG breaking change.

icechunk-python/tests/test_distributed_append.py
dcherian: i guess the clash here is because the tests use fixtures and so are parameterized?

icechunk/src/change_set.rs
```diff
@@ -68,9 +68,133 @@ impl EditChanges {
...
-        // path, different objects, or if the same path is added and deleted, etc.
+    fn merge(&mut self, other: EditChanges) -> SessionResult<()> {
+        // TODO: the conflict  detection is not comprehensive
+
+        // check if both created same group with different metadata
+        let shared: Vec<_> = self
+            .new_groups
+            .iter()
+            .filter_map(|(k, v)| {
+                let theirs = other.new_groups.get(k);
```
dcherian: `.filter(|k| other.new_groups.contains_key(k))`? Could do the same below too, or maybe I'm missing something?

icechunk-python/tests/test_dask.py
dcherian: Hmm.... I don't see how your chnages modify the behaviour

icechunk/src/change_set.rs
paraseba: I have two allow two things in here, one is if the other doesn't have the key, then allow also if both edited to the same metadata. That's the || in the condition

icechunk/src/repository.rs
```diff
@@ -4034,4 +4034,109 @@ mod tests {
...
+                Some(vec!["x".into()]),
+                Bytes::from_static(b"{}"),
+            )
+            .await?;
+        session.commit("init", None).await?;
+
+        let mut session = repo.writable_session("main").await?;
+
+        // zarr-python's mode="w" will delete, then re-create
+        session.delete_array(array_path.clone()).await?;
```
dcherian, line 4212: TODO: also just delete all chunks

icechunk/src/session.rs
```diff
@@ -362,7 +364,8 @@ impl Session {
     }
 
     pub fn read_only(&self) -> bool {
-        self.branch_name.is_none()
+        // self.branch_name.is_none()
```
dcherian: i guess we should track `read_only` explicitly on Session

icechunk-python/python/icechunk/xarray.py
```diff
@@ -313,15 +313,15 @@ def to_icechunk(
 
     as_dataset = _make_dataset(obj)
 
+    # really just a copy
+    base = session.fork()
+    anonsnap = session.flush("foo")
+    session = repository.writable_session(branch=None, snapshot_id=anonsnap)
```
dcherian: oh this sucks. We don't have access to the repository here.

icechunk/src/repository.rs
```diff
@@ -4101,4 +4034,109 @@ mod tests {
...
+        session
+            .add_array(
+                array_path.clone(),
+                shape,
+                Some(vec!["x".into()]),
+                Bytes::from_static(b"{}"),
+            )
+            .await?;
+
+        // we "backup" the current changeset
```
dcherian: this wouldn't be needed if we had `Session::from_transaction_log(snapshot: &SnapshotId)`

icechunk/src/repository.rs
```diff
@@ -4101,4 +4034,92 @@ mod tests {
...
+            .add_array(
+                array_path.clone(),
+                shape,
+                Some(vec!["x".into()]),
+                Bytes::from_static(b"{}"),
+            )
+            .await?;
+
+        // create a _clean_ writable session from that snapshot
+        let clean = session.fork().await?;
```
dcherian: this creates a new clean session on top of the anonymous snapshot constructed from a _clone_ of session. So session is still live and can be merged in to later at the end of the test

dcherian: I think the only other option is to take ownership of Session, return the anonymous SnapshotId, and have the user re-create the base session from that SnapshotId. This feels like we are exposing implementation details to the user.

icechunk-python/tests/test_dask.py
```diff
@@ -38,17 +38,23 @@ def test_store_dask(any_spec_version: int | None) -> None:
...
         # Can't fork a dirty session
-        with pytest.raises(ValueError):
-            fork = session.fork()
+        # with pytest.raises(ValueError):
+        #     fork = session.fork()
+
+        # session.commit("i am forced to commit to make progress")
+        # session = repo.writable_session("main")
+
+        # Can we enable merging for non ForkSession?
```
dcherian: Can we enable merging for non ForkSession?

icechunk-python/tests/test_session.py
```diff
@@ -41,8 +36,7 @@ async def test_session_fork(use_async: bool, any_spec_version: int | None) -> No
         zarr.group(session.store)
         assert session.has_uncommitted_changes
 
-        with pytest.raises(ValueError):
-            session.fork()
+        fork = session.fork()
```
dcherian, line 45:
```suggestion
        // test that useless fork doesn't break anything
        fork = session.fork()
```

icechunk/src/session.rs
```diff
@@ -397,6 +402,22 @@ impl Session {
         self.virtual_resolver.matching_container(chunk_location)
     }
 
+    pub async fn fork(&self) -> SessionResult<Self> {
+        // TODO: why do we allow Clone?
+        let snap = self.clone().flush("foo", None).await?;
```
dcherian: disallow commits on fork sessions

icechunk/src/session.rs
```diff
@@ -397,6 +402,22 @@ impl Session {
...
+
+        Ok(Session::create_writable_session(
```
dcherian, line 598: add a test for anonymous snapshots being preserved by GC if they are new enough

icechunk/src/repository.rs
```diff
@@ -4101,4 +4034,92 @@ mod tests {
...
+        )
+        .await?;
+        // merge the distributed writers' sessions
+        s1.merge(s2).await?;
+
+        // merge that in to the base
+        // TODO: we might reconstruct this from the transaction log
+        // Session::from_transaction_log?
+        session.merge(s1).await?;
+
```
dcherian: - check `s1` & `s2` cannot commit
- check refs after commit
- update comment

icechunk-python/tests/test_dask.py
```diff
@@ -76,8 +76,9 @@ def test_xarray_to_icechunk_nested_pickling(
             session = repo.writable_session("main")
 
             to_icechunk(ds, session=session, mode="w")
-            # TODO: repeated writes will fail the merge check
-            to_icechunk(ds, session=session, mode="w")
+            with pytest.raises(IcechunkError):
```
dcherian: this one could be weird in interactive sessions; but 🤷🏾‍♂️

dcherian:
@li-em can you take a look at the unit tests I just added please?

icechunk-python/python/icechunk/xarray.py
```diff
@@ -326,22 +328,26 @@ def to_icechunk(
...
+        # where we previously forced a user to commit.
+        # See test_distributed_append.py for an example.
+        fork: ForkSession = session.fork()
+        writer.writer.targets = [
+            # Any new arrays have already been created in the `write_metadata` step above
+            # so now we just use `mode="a"`.
+            zarr.open_array(fork.store, path=target.path, mode="a")
+            for target in writer.writer.targets
+        ]
+
```
dcherian, line 344: @paraseba I'm much happier with this. We basically place no restrictions on the user now. For example, no changes were needed in `test_dask.py` and `test_distributed_append.py`

PR #1818, opened 2026-03-12
Add user agent to storage backends

### LINKED ISSUE #1721: Set user-agent when interacting with object stores
Icechunk should set a useragent header when talking to icechunk stores that enables providers to identify it. This will help us establish usage prevalence ect

mpiannucci:
Closes #1721 

standardizes all user agents to include "icechunk/{version}"

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -229,7 +229,7 @@ jobs:
           PYTHON_TAG="cp$(echo ${{ env.PYTHON_VERSION }} | tr -d '.')"
           WHEEL=$(ls dist/*-${PYTHON_TAG}-*.whl)
           uv pip install "$WHEEL" --group test
-          uv pip install pytest-mypy-plugins
+          uv pip install "pytest-mypy-plugins<4"
           # Install specific xarray version based on matrix
           if [ "${{ matrix.xarray-version.version }}" != "latest-release" ]; then
             echo "Installing xarray ${{ matrix.xarray-version.version }}"


icechunk/src/storage/object_store.rs
```diff
@@ -57,6 +57,14 @@ use std::{
...
 };
 use tokio::sync::{OnceCell, RwLock};
+
+#[cfg(any(
+    feature = "object-store-s3",
+    feature = "object-store-gcs",
+    feature = "object-store-azure",
+    feature = "object-store-http"
+))]
+static ICECHUNK_USER_AGENT: &str = concat!("icechunk/", env!("CARGO_PKG_VERSION"));
```
paraseba: We should use the python version here instead, not sure how to get it, but that's what people know about? If that's too hard maybe `icechunk-rust/...`

mpiannucci: Yeah its confusing, we can do some plumbing to make this happen if we want it. IDK worth thinking about.

dcherian: I'm OK with just injecting `rust` in this string, and then making it easy to grab rust library version from the python library

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -2988,6 +2988,15 @@ def spec_version() -> int:
     """
     ...
 
+def user_agent() -> str:
+    """
+    The user-agent string sent with icechunk storage requests.
+
+    Returns:
+        str: The user-agent string (e.g., "icechunk/2.0.0-alpha.3")
```
dcherian:
```suggestion
        str: The user-agent string (e.g., "icechunk-rust/2.0.0-alpha.3")
```

icechunk/src/lib.rs
```diff
@@ -65,6 +65,13 @@ pub use storage::new_s3_storage;
 pub use storage::{ObjectStorage, Storage, StorageError, new_in_memory_storage};
 pub use store::Store;
 
+/// Returns the user-agent string for icechunk HTTP requests.
+///
+/// Format: `icechunk-rust-<version>` (e.g., `icechunk-rust-2.0.0-alpha.3`).
+pub fn user_agent() -> &'static str {
+    concat!("icechunk-rust-", env!("CARGO_PKG_VERSION"))
```
mpiannucci, line 72: We use dashes as a separator because of restrictions from aws native library

Issue #1824, opened 2026-03-12
Track down `list_prefix` bug in V1.X


dcherian:
Internal slack link: https://earthmoverworkspace.slack.com/archives/C07C8TCQQRG/p1773332775722249

Just a reminder that we need to track down this `list_prefix` bug. Which seems to be broken after v1.1.6.

ianhi:
1.1.6 last working version. Seems lilke culprit is: https://github.com/earth-mover/icechunk/pull/1257

PR #1825, opened 2026-03-12
Fix bug in list_prefix implementation

paraseba:
Certain paths with non alphanumeric characters were hiding others.

For example:

`foo-bar` would hide `foo/baz`

Data is safe, but Icechunk wasn't able to return those nodes.

Fixes: #1824

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -3,6 +3,7 @@ name: Code Quality Checks
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+  merge_group:
   push:
     branches:
       - main
@@ -55,7 +56,7 @@ jobs:
 
       - name: Run clippy linting
         run: |
-          just profile=ci lint --workspace
+          just profile=ci lint --workspace --all-targets
 
       - name: Run doc tests
         run: |
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -7,6 +7,7 @@ on:
     branches: [main, 'support/**']
   pull_request:
     branches: [main, 'support/**']
+  merge_group:
 
 permissions:
   contents: read
--- a/.github/workflows/dependency-check.yaml
+++ b/.github/workflows/dependency-check.yaml
@@ -3,6 +3,7 @@ name: Dependency Check
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+  merge_group:
   push:
     branches:
       - main
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -7,6 +7,7 @@ on:
       - 'support/**'
   pull_request:
     types: [opened, reopened, synchronize, labeled]
+  merge_group:
   workflow_dispatch:
 
 concurrency:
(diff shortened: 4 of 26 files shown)


icechunk/src/format/snapshot.rs
```diff
@@ -685,26 +685,37 @@ impl Iterator for NodeIterator {
 
     fn next(&mut self) -> Option<Self::Item> {
         let nodes = self.snapshot.root().nodes();
-        if self.next_index < nodes.len() {
+        loop {
+            // we need to looop over all nodes starting from the index
```
ianhi: loooooooooop :)

paraseba: man, I do the the triple o loop at least every two weeks :facepalm:

icechunk/src/format/snapshot.rs
```diff
@@ -685,26 +685,37 @@ impl Iterator for NodeIterator {
...
                     if let Some(after_prefix) =
-                        res.path.to_string().strip_prefix(self.prefix.as_str())
+                        node_path.strip_prefix(self.prefix.as_str())
                         && (after_prefix.is_empty() || after_prefix.starts_with('/'))
                     {
                         self.next_index += 1;
-                        Some(Ok(res))
+                        return Some(Ok(res));
+                    } else if node_path.as_str() > self.prefix.as_str()
+                        && !node_path.starts_with(self.prefix.as_str())
```
dcherian, line 708: i see, you're relying on alphabetical sorting to quick-exit but previously you'd skip over `foo-bar` if you were looking at `foo`

ianhi (dismissed):
I'm seeing list_prefix fail (the writable here implies that it is pre-commit) - but it does seem to fix `list_dir`

see also #1826 

```
E               AssertionError: list_prefix('0') [writable]:
E                   missing: ['0_c/zarr.json']
E                   extra:   []
E               assert {'0/zarr.json..._c/zarr.json'} == {'0/zarr.json'}
E
E                 Extra items in the left set:
E                 '0_c/zarr.json'
E                 Use -v to get more diff
E               Falsifying example: test_list_prefix(
E                   tree=GroupNode(children={'0': GroupNode(children={}),
E                     '0_c': ArrayNode(shape=(1,), dtype=dtype('int32'))}),
E               )

```

ianhi:
this is the simplest tree that seems to break things:

```
    root = zarr.open_group(mem, mode="w")
    root.create_group("0")
    root.create_array("0_c", shape=(1,), dtype="i4")
```

`list_prefix("0")` gives a different result than on a zarr memory store.

Also feel free to clear my review if/when this gets fixed its before I am awake 😬 (but also i am taking melatonin tongiht to drag myself back

ianhi:
ok concrete minimal example:

```python
import asyncio

import zarr
from icechunk import Repository, in_memory_storage


async def run(store):
    root = zarr.open_group(store, mode="w")
    root.create_group("0")
    root.create_array("0_c", shape=(1,), dtype="i4")
    print(sorted([k async for k in store.list_prefix("0")]))


async def main():
    print("MemoryStore:")
    await run(zarr.storage.MemoryStore())

    print("IcechunkStore:")
    repo = Repository.create(storage=in_memory_storage())
    await run(session.store)

asyncio.run(main())
```

```
MemoryStore:
['0/zarr.json', '0_c/zarr.json']
IcechunkStore:
['0/zarr.json']
```

paraseba:
> MemoryStore:
> ['0/zarr.json', '0_c/zarr.json']
> IcechunkStore:
> ['0/zarr.json']
> ```

this is known behavior.

For example, Icechunk has this error:
```
        if path != Path::root() {
            let _ = session.get_node(&path).await.map_err(|_| {
                StoreErrorKind::BadKeyPrefix {
                    prefix: prefix.to_owned(),
                    message: "Only prefixes pointing to a group or array are allowed"
                        .to_string(),
                }
            })?;
        }
```

In your example it doesn't trigger because there is in fact a group there

PR #1840, opened 2026-03-13
Concurrency permutation testing with shuttle \m/

dcherian:
The `shuttle_tokio` crate swaps out tokio for a custom scheduler that then tries random orderings of every `await` point. 

https://github.com/awslabs/shuttle

https://dl.acm.org/doi/10.1145/3477132.3483540

- It isn't too invasive, all our current CI passes
- I added a new CI job since it needs `--features shuttle`

I reverted #1798 and it managed to catch a race

```
called `Result::unwrap()` on an `Err` value: 
ICError { 
    kind: RepositoryError(FormatError(InvalidUpdateTimestamp { latest_time: 2026-03-13T16:14:17.141941Z, new_time: 2026-03-13T16:14:17.141722Z }
)), ]
```

It was *extremely* hard to come up with test cases. I didn't manage to write one to catch #1791

### DIFF
--- a/icechunk/tests/test_shuttle.rs
+++ b/icechunk/tests/test_shuttle.rs
@@ -280,10 +280,7 @@ async fn assert_action_postcondition(
                 .map_ok(|info| info.id)
                 .try_collect()
                 .await?;
-            assert!(
-                anc.contains(&snap),
-                "commit {snap:?} not found in ancestry of branch {branch:?}"
-            );
+            assert!(anc.contains(&snap));
         }
         AddBranch(branch, snap) => {
             assert!(repo.list_branches().await?.contains(&branch));
@@ -299,9 +296,17 @@ async fn assert_action_postcondition(
         DeleteTag { tag, .. } => {
             assert!(!repo.list_tags().await?.contains(&tag));
         }
-        Amend { branch, new_snap, .. } => {
+        Amend { branch, new_snap, previous_snap } => {
             let tip = repo.lookup_branch(&branch).await?;
             assert_eq!(tip, new_snap, "amend snapshot should be branch tip for {branch}");
+            let anc: HashSet<SnapshotId> = repo
+                .ancestry(&VersionInfo::BranchTipRef(branch.clone()))
+                .await?
+                .map_ok(|info| info.id)
+                .try_collect()
+                .await?;
+            assert!(anc.contains(&new_snap));
+            assert!(!anc.contains(&previous_snap));
         }
         ResetBranch { branch, to_snap, .. } => {
             let tip = repo.lookup_branch(&branch).await?;


icechunk/Cargo.toml
```diff
@@ -75,10 +76,10 @@ async-compression = { version = "0.4.41", features = ["tokio", "zstd"] }
 serde_bytes = "0.11.19"
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-tokio = { version = "1.50.0", features = ["rt-multi-thread", "macros"] }
+tokio = { version = "1.50.0", features = ["rt-multi-thread", "macros", "sync", "time", "io-util", "rt"] }
 
 [target.'cfg(target_family = "wasm")'.dependencies]
-tokio = { version = "1.50.0", features = ["rt", "macros"] }
+tokio = { version = "1.50.0", features = ["rt", "macros", "sync", "time", "io-util"] }
```
dcherian, line 82: Claude claims we are using these already.

icechunk/tests/test_shuttle.rs
```diff
@@ -0,0 +1,416 @@
...
+                let action = action.clone();
+                async move { execute_action(repo, action, branch).await }
+            })
+        })
+        .collect::<Vec<_>>();
+
+    let mut results = vec![];
+    for handle in handles {
+        results.push(handle.await??);
+    }
```
li-em: Should this be a `join_all` instead of a loop?

```suggestion
    let handles = futures::future::join_all(handles).await;

    let mut results = vec![];
    for handle in handles {
        results.push(handle??);
    }
```

Or does shuttle take care of this?

dcherian: Turns out shuttle doesn't provide that one.

icechunk/tests/test_shuttle.rs
```diff
@@ -0,0 +1,416 @@
...
+        assert!(
+            time_a > time_b,
+            "ops log timestamps must be strictly decreasing: {time_a} should be > {time_b}"
+        );
+    });
+
+    // all non-None backup paths must be unique
+    let backup_paths: Vec<_> =
+        log.iter().filter_map(|(_, _, path)| path.as_ref()).collect();
+    let unique: HashSet<_> = backup_paths.iter().collect();
```
paraseba: I think only the last op in the ops log can have null backup?

icechunk/tests/test_shuttle.rs
```diff
@@ -0,0 +1,416 @@
...
+    repo: Arc<Repository>,
+    actions: &[Action],
+    branches: &[String],
+) -> Result<(), Box<dyn Error + Send + Sync>> {
+    let path: Path = "/array".try_into()?;
+    for (action, branch) in actions.iter().zip(branches.iter()) {
+        match action {
+            Action::AddBranch => {}
+            Action::Amend => {
+                repo.create_branch(branch, &repo.lookup_branch("main").await?).await?;
```
paraseba: i'm confused by amend vs commit

dcherian: this function "sets" up the repo. So for `amend` we create one commit here that then gets amended in the future later.

icechunk/tests/test_shuttle.rs
dcherian: Actually you're right and I was confused. Thanks!

PR #1846, opened 2026-03-13
Feat: add `LatencyStorage` + Fix: Flaky stateful test failures

### LINKED ISSUE #1746: Add LatencyStorage wrapper
will help with benchmarking

xref #1399

ianhi:
Closes #1746 

Hopefully fix the flaky failures we have been seeing by not drawing times from a strategy.

I also swapped the order to put what I view as the simpler cases first in the sampled_from which from my reading of https://hypothesis.readthedocs.io/en/latest/reference/strategies.html#hypothesis.strategies.sampled_from

might improve shrinking performance

### DIFF
--- a/icechunk-python/src/config.rs
+++ b/icechunk-python/src/config.rs
@@ -2246,6 +2246,10 @@ impl PyStorage {
 
     /// List objects in the storage backend, optionally filtered by a key prefix.
     ///
+    /// Returns a list of ``(key, size_in_bytes)`` tuples for each object found.
+    /// When ``prefix`` is ``None`` or empty, all objects under the repository root are listed.
+    /// Custom ``settings`` can be provided to override the storage's default settings.
+    ///
     /// Deprecated: use ``list_objects_metadata`` instead, which also returns
     /// the ``created_at`` timestamp.
     #[pyo3(signature = (settings=None, prefix=None))]
@@ -2281,6 +2285,9 @@ impl PyStorage {
     }
 
     /// List objects with full metadata, optionally filtered by a key prefix.
+    ///
+    /// When ``prefix`` is ``None`` or empty, all objects under the repository root are listed.
+    /// Custom ``settings`` can be provided to override the storage's default settings.
     ///
     /// Returns
     /// -------


dcherian:
Do you understand how this is meaningfully different --- you're still making a state-dependent draw? Can you with `--hypothesis-show-statistics` and report how many times GC is actually doing something? I much prefer the previous order where we actually deleted things.

dcherian:
Like the real thing that's flaky here is that if we draw `commit_time` and add 200ms we may or may not expire/GC it. 

What we could do is call the Icechunk code; get back the list of expired snaps; then expire those in the model. and assert none of these expired snaps are newer than drawn time + some offset?

ianhi:
> Do you understand how this is meaningfully different

you mean in the ordering? I concede that maybe the shrinking properties aren't the thing to prioritize. But the flakiness 

> Like the real thing that's flaky here is that if we draw commit_time and add 200ms we may or may not expire/GC it.

ah wait I was thinking part of the flakiness was that we were selecting a **different** time value cutoff in different runs. and try to protect against that. i.e. the. actual commit times are shifted in time. is that wrong?

dcherian:
Hmm... i see what you're saying but it'd be good to understand if that is indeed the case (does hypothesis care about actual values drawn?)! If so, then can we keep the previous ordering and just switch to drawing integers. 

Also IIRC i tried that and it didn't help but perhaps I'm misremembering.

ianhi:
Ok yea confirmed with a script. I really had it in my head that hypothesis would complain if we generated a different actual value on a second run - and I was after fixing that. i.e. 2026-10-01:20:10:34 vs the next data gen is at  2026-10-01:20:10:35 but I was wrong about that

ianhi:
the flakiness here is just in the commit vs flusshed time difference?

dcherian:
> the flakiness here is just in the commit vs flusshed time difference?

IIUC yes. and it manifests itself as being indeterminate in whether a draw actually expires a snapshot or not. So we should think of workarounds. Honestly it'd be easier if we just exposed `flushed_at`. This is available in V2 using the ops_log I think

ianhi:
Ok looked into this more deeply. It's not commit vs flushed as we actually get flushed in python:

https://github.com/earth-mover/icechunk/blob/a146a84355d16099cd190e58a16e7b535e9dbab4/icechunk-python/tests/test_stateful_repo_ops.py#L303-L305

https://github.com/earth-mover/icechunk/blob/a146a84355d16099cd190e58a16e7b535e9dbab4/icechunk-python/src/repository.rs#L215-L223

Rather it is `flushed_at` vs `created_at`. The latter of which is what is actually used to expire

https://github.com/earth-mover/icechunk/blob/a146a84355d16099cd190e58a16e7b535e9dbab4/icechunk/src/ops/gc.rs#L154-L159


So I think there may be a real possibility of a bug for a user how tries to GC based on the snapshots they have access to.


1. I look up a specific snapshot
2. I use that snapshots written_at + .0001 to do GC
3. the snapshot i used here is not expired

That's not catastrophic but does seem like a bug to me


I think we only see this on CI because the github actions file system is slower than laptops.

dcherian:
No I've been down this road when I first wrote this haha. Much of the complexity in this test stems from me trying to understand what was going on and being quite strict about it. 

I think the way to proceed here is to see if `created_at` is available somehow at the python layer (`ops_log`? maybe the transaction log for a commit?), and store that explicitly on the model; then draw from that for the expiration & GC steps.

ianhi:
> I think the way to proceed here is to see if created_at is available somehow at the python layer (ops_log? maybe the transaction log for a commit?), and store that explicitly on the model; then draw from that for the expiration & GC steps.

I agree. i kept at this bit a longer, and its actually pretty easy to expose in `list_objects` and it does seem to fix the bug. Which i can consistently repro with a janky latency store addition to our memory store. will push it  up on monday :)

icechunk-python/python/icechunk/testing/__init__.py
```diff
@@ -0,0 +1,14 @@
+"""Testing utilities for icechunk.
+
+This module provides tools for testing icechunk behavior under conditions
+that are hard to reproduce with real storage backends, such as high latency.
```
dcherian, line 4: Could point people toward the Zarr wrapper stores too

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -833,13 +867,19 @@ def delete_tag(self, tag: str) -> None:
             self.model.delete_tag(tag)
 
     def _draw_older_than(self, data: st.DataObject) -> datetime.datetime:
+        # Draw cutoffs from storage-level created_at (not written_at/flushed_at)
+        # because that is what Rust GC compares against. The +10ms offset is
+        # past the concurrent snapshot/transaction write gap (sub-ms) but tight
+        # enough that not all snapshots fall below the cutoff.
+        created_at_times = sorted(
```
dcherian: let's maintain this on the model and populate it on every commit step instead. That way it's cheap to search through.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -520,9 +532,31 @@ def _initialize_model_and_data(self, data: st.DataObject) -> None:
...
 
-    @initialize(data=st.data(), target=branches, spec_version=st.sampled_from([1, 2]))
-    def initialize(self, data: st.DataObject, spec_version: Literal[1, 2]) -> str:
-        self.storage = self._make_storage()
+    @initialize(
+        data=st.data(),
+        target=branches,
+        spec_version=st.sampled_from([1, 2]),
+        # Both latencies are zero (~75%) or both non-zero (~25%)
+        # to exercise the flushed_at vs created_at timing gap in GC.
```
dcherian, line 540: I wouldn't bother here. It's not meaningfully important. Instead I would leave a comment desribing that LatencyStorage can be used for debugging

ianhi:
Ok here is an update on this PR:

It is now three commits that do:

1. Add a LatencyStorage wrapper
2. Expose `created_at` via the new `list_objects_metadata` (new to avoid breaking API change)
3. Fix the actual flaky test using `created_at`


I added the Latency Store to the test as it was extremely helpful in debugging the flaky failure, but would happily remove except from a long overnight CI test run - it does make the test slower, even after I remove the read latency.

### FlakyFailure

My understanding of the FlakyFailure is that:

First we chose a cutoff for GC of commit time + 200ms. The 200 ms accounts for the fact that rust uses `created_at` not flushed at for GC. 

But in a higher latency env (e.g. Github actions) one of two sources of flakiness could happen:

1. `created_at - flushed_at` > 200 ms
2. A second commit could happen and be flushed (but not created) at within those 200 ms

In either case our model and and actual GC would diverge.


The fix for this was to expose the`created_at`. After that I significantly shrunk the time gap to 1 microsecond. Which led to second failure even with 0 artificial latency. With too small a gap on top of created time we started missing the invariannt that checks that snapshots and transactions match on disk. This is because  the snapshot and transaction log are written async on commit:

https://github.com/earth-mover/icechunk/blob/a146a84355d16099cd190e58a16e7b535e9dbab4/icechunk/src/session.rs#L2598-L2634

There are two fixes for this

**Simple:**

Continue using a small time buffer (1 ms) small enough to not trigger the other flaky bugs, but large enough to hopefully always capture both files for GC

**More complex + Robust**

In the draw_older_than take the max of transaction log and snapshot created at times.

icechunk/src/storage/latency.rs
```diff
@@ -0,0 +1,215 @@
...
+    ) -> StorageResult<VersionedUpdateResult> {
+        self.backend.copy_object(settings, from, to, content_type, version).await
+    }
+
+    async fn list_objects<'a>(
+        &'a self,
+        settings: &Settings,
+        prefix: &str,
+    ) -> StorageResult<BoxStream<'a, StorageResult<ListInfo<String>>>> {
+        self.backend.list_objects(settings, prefix).await
```
dcherian, line 125: we should `sleep` on all of these; it's very confusing otherwise!

dcherian:
> The fix for this was to expose the created_at. 

👏🏾 👏🏾 

> Continue using a small time buffer (1 ms) small enough to not trigger the other flaky bugs

This is fine; just add a comment describing the tx log vs snapshot difference.

ianhi:
oh noo. v1 does not have `list_object_metadata` :/ I guess I just always need to do that with v2?

actually maybe they dont run due to preconditions

icechunk/src/storage/latency.rs
ianhi: done

icechunk/src/storage/latency.rs
```diff
@@ -0,0 +1,219 @@
...
+    async fn put_object(
+        &self,
+        settings: &Settings,
+        path: &str,
+        bytes: Bytes,
+        content_type: Option<&str>,
+        metadata: Vec<(String, String)>,
+        previous_version: Option<&VersionInfo>,
+    ) -> StorageResult<VersionedUpdateResult> {
+        tokio::time::sleep(Duration::from_millis(self.write_delay_ms())).await;
```
paraseba: NIT: maybe private methods `sleep_for_read()` and `sleep_for_write()` ?

icechunk/src/storage/latency.rs
paraseba: This may be OK, but it's not perfect. The underlying list is a stream, every 1k keys it needs to go back to the store, and those gets are not getting a sleep, only the first one.

That said, what you have is probably enough.

Issue #960, opened 2025-05-27
Make storage cache extensible


mpiannucci:
### Problem

The current cache is always in memory, in a distributed or serverless environment it would be very useful to allow for out of memory caches like redis, dynamo db or another key value store like cloudflare kv. This would allow multiple workers to share a chunk cache and improve efficiency

mpiannucci:
A file cache would be a nice option as well

mpiannucci:
* We will use a cache typetag trait for representing caches
* By default icechunk will use in memory cache when not specified
* Caches are NOT serialized to config, only loadable at runtime

j-kapp:
Hi @mpiannucci, I'm not sure if this is the correct place to ask this, but I see that there hasn't been much recent movement on this ticket so I'm just wondering if it's still going to be implemented? 😅 

For context, I am considering using icechunk to read zarr data from slow cheap object storage that is far from compute. When chunks are read I would like to cache them locally on the compute server (or as close as possible) so that when the same chunk is read again it is much faster. I assume that would be a very common use case for icechunk users. Zarr v3 seems to provide that exact functionality with the experimental [CacheStore](https://zarr.readthedocs.io/en/stable/user-guide/experimental/#cachestore) feature, but I have not found similiar functionality in icechunk (apart from in-memory caching). I guess I could implement some kind of custom caching layer (similiar to what was described [here](https://github.com/earth-mover/icechunk/discussions/802)), but it would be ideal (less work for me haha 😆 ) if this could be handled automatically by icechunk.

Any insights would be much appreciated. Thank you!

ianhi:
> but I have not found similiar functionality in icechunk

Hi @j-kapp great question. To be clear Icechunk does ship with built in caching, configurable via this https://icechunk.io/en/stable/guides/configuration/#caching but as Matt notes it is in memory

> Zarr v3 seems to provide that exact functionality with the experimental [CacheStore](https://zarr.readthedocs.io/en/stable/user-guide/experimental/#cachestore) feature, 

Because Icechunk ultimately provides a zarr store interface this should be useable with Icechunk today! You can wrap the `session.store` in the `CacheStore`. 

Something like the following

```python
session = repo.readonly_session("main")
cache_store = LocalStore(path_to_on_disk_cache)
cached_store = CacheStore( 
   store=session.store,    
   cache_store=cache_store,
)
```

[...]

PR #1849, opened 2026-03-16
Design Doc : Add pluggable cache backend

mpiannucci:
See comments for points of specific emphasis

#960

### DIFF
--- a/design-docs/016-pluggable-cache-backend.md
+++ b/design-docs/016-pluggable-cache-backend.md
@@ -6,14 +6,55 @@ Icechunk currently uses in-memory LRU caches (`quick_cache`) in `AssetManager` f
 
 ## Design
 
-### 1. `CacheBackend` Trait
+### 1. Crate Layout
+
+Following the same pattern as storage (`icechunk-storage` for the trait, `icechunk-s3` / `icechunk-arrow-object-store` for backends), caching is split into separate crates:
+
+| Crate | Contents |
+|---|---|
+| `icechunk-cache` | `CacheBackend` trait, error types, `InMemoryCache` |
+| `icechunk-cache-redis` | `RedisCache` implementation (optional, behind feature flag) |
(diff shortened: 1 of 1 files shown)


design-docs/016-pluggable-cache-backend.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Motivation
+
+Icechunk currently uses in-memory LRU caches (`quick_cache`) in `AssetManager` for snapshots, manifests, transactions, and chunks. This works well for long-lived processes but falls short in serverless environments where process memory is ephemeral. We need to support external cache backends (Redis, disk, dynamo, object storage, etc.) that persist across invocations, while keeping the in-memory backend as the default.
+
+## Design
+
+### 1. `CacheBackend` Trait
+
+A new trait abstracts over cache storage. Since external backends require serialized data, **all backends operate on bytes**. Deserialization happens in `AssetManager` on every cache hit.
```
mpiannucci, line 48: This will introduce a performance hit. The alternative would be a specific cache interface with types specified so in memory could cache everything as full types

design-docs/016-pluggable-cache-backend.md
```diff
@@ -0,0 +1,164 @@
...
+
+### 2. Backend Implementations
+
+**InMemoryCache** (default) — wraps `quick_cache` as today. Config: max capacity per cache category (reuses existing `CachingConfig` knobs). Single `Cache<String, Bytes, ...>` under the hood, with weight-based eviction.
+
+**RedisCache** — uses `redis` crate with async connection pool. Config: connection URL, optional TTL, optional max memory hint.
+
+**FileCache** (future) — local disk-backed cache using a path on a local filesystem. Config: directory path, max size.
+
+### 3. `CachingConfig` — Current State and Evolution
```
mpiannucci: This is the biggest area of contention for me: We use quick cache with its specific weighting system and limits but not all cache backends will work with that kind of an interface

PR #1856, opened 2026-03-17
Process nodes in parallel during the flush process

paraseba:
There is a new `max_concurrent_manifests` argument to several methods, a good number to pass would be the number of manifests modified in the commit, or the maximum number of manifests that can comfortably fit in memory, whatever is lower.

This is  not exposed to python, which is still using `max_concurrent_manifests=1` (sequential processing).

### DIFF
--- a/icechunk/tests/test_gc.rs
+++ b/icechunk/tests/test_gc.rs
@@ -564,7 +564,7 @@ async fn test_gc_reset_branch() -> Result<(), Box<dyn std::error::Error>> {
             )
             .await?;
         let snap = session
-            .commit(&format!("commit {i}"))
+            .commit(format!("commit {i}"))
             .max_concurrent_nodes(8)
             .execute()
             .await?;


mpiannucci:
I wonder if this makes more sense as a repo setting or an env var to avoid polluting the commit method signature? Thihnking through the intended uses, to me it seems that it will be a property of the end system i am running on more than the repo itself or the script calling commit. But i may be missing some perspective

icechunk/src/session.rs
```diff
@@ -2448,69 +2484,87 @@ async fn do_flush(
...
+                )
+                .await
             }
-        } else {
-            trace!(path=%node.path, "Node has no changes, keeping the previous manifest");
-            // Array wasn't deleted but has no changes in this session
-            flush_data.copy_previous_manifest(&node, old_snapshot.as_ref());
-        }
+        }))
+        .buffer_unordered(max_concurrent_manifests)
```
dcherian: Calling it `max_concurrent_manifests` is a bit misleading since it's really `max_concurrent_nodes`. and would be useless when writing a single node with many manifests. Should we switch to a queue-based approach instead, where we queue up a future per manifest and then crunch through that concurrently?

icechunk/src/session.rs
```diff
@@ -2315,19 +2306,63 @@ impl<'a> FlushProcess<'a> {
...
                         .clone()
                 }));
-                for mr in array_refs.iter() {
-                    let new_ref = mr.clone();
-                    self.manifest_refs
-                        .entry(node.id.clone())
-                        .and_modify(|v| v.push(new_ref.clone()))
-                        .or_insert_with(|| vec![new_ref]);
-                }
+                result.manifest_refs.extend(array_refs.iter().cloned());
```
dcherian: Does `into_iter()` not work here? 

Seems to work if you remove the `&` from the match

```suggestion
                result.manifest_refs.extend(array_refs.into_iter());
```

icechunk/src/session.rs
```diff
@@ -2037,276 +2071,233 @@ impl<'a> FlushProcess<'a> {
...
 
-    /// Record the previous manifests for an array that was not modified in the session
-    fn copy_previous_manifest(&mut self, node: &NodeSnapshot, old_snapshot: &Snapshot) {
+            Ok(Some(result))
+        } else {
+            Ok(None)
+        }
+    } else {
+        trace!(path=%node.path, "Node has no changes, keeping the previous manifest");
         match &node.node_data {
```
```suggestion
        match node.node_data {
```

icechunk/src/session.rs
```diff
@@ -2315,19 +2306,63 @@ impl<'a> FlushProcess<'a> {
                         )
                         .clone()
```
```suggestion
```

dcherian:
> avoid polluting the commit method signature

I agree. Shall we limit this to commit_with_options on the Rust side? That would reduce the blast radius by quite a bit.

I think it's ok on rewrite_manifests and amend since those are used rarely.

paraseba:
So, I thought quite a bit and went back and forth between some of the options mentioned here. It's all subjective, but these were my conclusions:
* this is not a config for the repo, because it depends on how/where you are running. For example it depends on the type of commit you are doing.
* commit_with_options is the obvious choice, but there are a bunch of things that need this flag too, like rewrite manifest, amend, etc etc, so it would be weird not having it in commit. Unless with introduce amend_with_options and all that stuff.
* It's very hard/impossible to give this value a default, it depends on size of manifests, memory, etc


None of these are conclusive, just where I ended up after trying some of the options @dcherian and @mpiannucci are proposing. I'm very open to changes

mpiannucci:
Im coming around to your way of thinking about this: on some level the amount of parallelism also depends on the number of manifests that will be committed. We should talk through all the implications

mpiannucci (approved):

icechunk/src/session.rs
paraseba: done

icechunk/src/session.rs
paraseba: done

icechunk/src/session.rs
paraseba: done

PR #1858, opened 2026-03-18
Benchmark virtual chunks set/get with prefixes

### LINKED ISSUE #1696: Benchmark and optimize virtual chunk reads

dcherian:
Closes #1696
xref #1399

Simply testing against data in minio shows that fetching chunks is what takes time. I didn't add any latency here (it makes it too slow) but the option is now available.

There is some 10-20% overhead on _every_ chunk get.
- for 1000 manifest node, this is equally split between copying NodeSnapshot, dropping NodeSnapshot, creating Manifest decompressor
- for 1 manifest node, basically all the overhead is constructing the Manifest decompressor.

I don't think this overhead matters much.

image

Despite that, the second commit does some micro-optimization around creating virtual chunk refs by allocating a String once and pushing in to it. :) This will help a small amount on bulk sets.

Also optimize rewriting manifests by avoid parsing the VCL again.

### DIFF
--- a/icechunk/benches/manifest.rs
+++ b/icechunk/benches/manifest.rs
@@ -100,6 +100,7 @@ fn default_storage_kind() -> StorageKind {
 
 /// Parse `ICECHUNK_BENCH_LATENCY_MS` env var; returns `Some(ms)` if set.
 fn toxiproxy_latency_ms() -> Option<u64> {
+    #[allow(clippy::expect_used)]
     std::env::var("ICECHUNK_BENCH_LATENCY_MS")
         .ok()
         .map(|v| v.parse::<u64>().expect("ICECHUNK_BENCH_LATENCY_MS must be an integer"))


icechunk/src/format/manifest.rs
```diff
@@ -698,15 +715,15 @@ fn ref_to_payload(
                 e,
             ))
         })?;
-        let location = VirtualChunkLocation::from_url(location_str)?;
+        let location = VirtualChunkLocation::from_trusted(location_str.to_string());
```
dcherian: this.... was insane. `Url::parse` is quite expensive! Mostly relevant for `rewrite_manifests`

icechunk/src/format/manifest.rs
```diff
@@ -265,6 +265,11 @@ impl VirtualChunkLocation {
         self.0.as_str()
     }
 
+    /// Wrap a pre-validated location string without re-parsing.
+    fn from_trusted(s: String) -> Self {
+        VirtualChunkLocation(s)
```
paraseba, line 270: Maybe put this in VirtualChunkLocation? Is it worth making it `unsafe {}`?

paraseba: ups, done

PR #1861, opened 2026-03-18
Fix bug in ops log after migration

paraseba:
If we migrate a repo with more than 100 "operations", next time we write to it we don't have a backup path to "overflow" from the ops log. This is because migration (correctly) creates updates without a backup path.

The solution implemented here is to temporarily grow the size of the per file ops log, until we can point to the first version of the repo object, the one migrated. After that per file ops log goes back to the expected size.


paraseba:
the new test fails in main and passes in this PR

mpiannucci:
great catch seba 🙏

PR #1863, opened 2026-03-18
Copy object race condition

paraseba:
We use copy_object to backup repo_info. It needs to be consistent, and reject the copy if source has changed. object_store doesn't support it, so we do get+put instead of copy 🤷.

Created a ticket in object_store https://github.com/apache/arrow-rs-object-store/issues/667

### DIFF
--- a/icechunk/src/storage/object_store.rs
+++ b/icechunk/src/storage/object_store.rs
@@ -402,10 +402,10 @@ impl Storage for ObjectStorage {
             let result = self.get_client(settings).await.get_opts(&from, opts).await;
             match result {
                 Ok(result) => {
-                    let bytes =
-                        result.bytes().await.map_err(|e| -> StorageError {
-                            Box::new(e).into()
-                        })?;
+                    let bytes = result
+                        .bytes()
+                        .await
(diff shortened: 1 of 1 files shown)


icechunk/src/storage/object_store.rs
```diff
@@ -389,15 +389,48 @@ impl Storage for ObjectStorage {
...
+                Err(object_store::Error::Precondition { .. }) => {
+                    Ok(VersionedUpdateResult::NotOnLatestVersion)
+                }
+                Err(object_store::Error::NotFound { .. }) => {
+                    Err(StorageErrorKind::ObjectNotFound.into())
+                }
+                Err(err) => Err(Box::new(err).into()),
+            }
+        } else {
+            match self.get_client(settings).await.copy(&from, &to).await {
```
dcherian, line 425: should we raise and say we need the unsafe conditional udpate option to proceed?

paraseba: no, because this is what people in things like Jasmine use. They know it's not safe (even for simple commits) but it's worth the risk.

dcherian (approved):

PR #1865, opened 2026-03-18
Consolidate benchmarks into a single binary

dcherian:
Combine the separate `manifest` and `asset_manager` bench targets into one `main` binary to reduce compile times and simplify the bench CLI.


dcherian:
latest commits add integration with tracing-chrome & tracing-samply. I'm not sure which is better yet, so i am just leaving both in for now.

Issue #1862, opened 2026-03-18
`shift_array` interaction with empty chunks


ianhi:
If you shift an empty chunk onto a filled chunk then the filled chunk will (I believe incorrectly) retain it's original value. 

Consdier: 

Fill value of -1 an array with only the middle element written (chunksize=1)

`[-1, 2, -1]`

now shift by `(1,)`

you currently get
`[-1, 2, 2]`

but I think you should get `[-1, -1, 2]`

```python
import icechunk as ic
import zarr

repo = ic.Repository.create(storage=ic.in_memory_storage())
session = repo.writable_session("main")
arr = zarr.open_array(
    session.store,
    path="data",
    mode="w",
    shape=(3,),
    chunks=(1,),
    dtype="f4",
    fill_value=-1,
)
arr[1] = 2.0  # only c/1 written; c/0 and c/2 are fill
session.commit("initial")

print("array with only item 1 filled - fill vlaue=-1")
print("before shift:", arr[:])  # [-1, 2, -1]

"""Sparse array shift: only c/1 written, c/0 and c/2 never written. Shift +1."""
print("shift (1,)")
session = repo.writable_session("main")
arr = zarr.open_array(session.store, path="data")
session.shift_array("/data", (1,))

print("after shift: ", arr[:])  # expect [-1, -1, 2] — c/1 vacated, c/2 gets c/1's data
# icechunk gives [-1, 2, 2] — c/1 keeps stale data
```

Issue #1737, opened 2026-03-02
More sophisticated stateful/property testing settings


dcherian:
Hypothesis allows us to run with different settings in different contexts using "profiles". Here is an example I wrote up for Zarr : https://github.com/zarr-developers/zarr-python/blob/b8094d8765e4f6de5121806ed8f209ffac315168/tests/conftest.py#L234-L256 . Mostly it sets up hypothesis to explore many more examples in nightly runs (`max_examples`, `stateful_step_count`). 

We might also set the default profile to test less to speed up local development.

The main deliverable here is to run extra examples nightly.

PR #1870, opened 2026-03-19
add hypothesis profiles

ianhi:
Including super long runs on the nightly tests

and always triggering `HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY` in CI runs

I made the nightly numbers big, but hopefully not too big. I don't have a great sense for what a good upper limit is.


icechunk-python/tests/conftest.py
```diff
@@ -1,12 +1,44 @@
...
+    parent=settings.get_profile("ci"),
+    max_examples=500,
+    stateful_step_count=200,
+    suppress_health_check=[HealthCheck.filter_too_much, HealthCheck.too_slow],
+)
+settings.register_profile(
+    "nightly",
+    parent=settings.get_profile("ci"),
+    max_examples=1000,
+    stateful_step_count=500,
```
dcherian, line 35:
```suggestion
    stateful_step_count=500,
    derandomize=False,
```

dcherian:
Nice, i thought i had created an issue for this but seem to have not done that.

Just one suggestion, from my time writing this for Zarr: https://github.com/zarr-developers/zarr-python/blob/4acebeba7be73e7e3ee1442e89741a3295cea8fc/tests/conftest.py#L234

dcherian:
> and always triggering HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY in CI runs

What does this do in CI? Do we backup the output to a cache?

icechunk-python/tests/conftest.py
```diff
@@ -1,12 +1,44 @@
...
+    suppress_health_check=[HealthCheck.filter_too_much, HealthCheck.too_slow],
+)
+if "HYPOTHESIS_PROFILE" in os.environ:
+    settings.load_profile(os.environ["HYPOTHESIS_PROFILE"])
```
dcherian: you could set this in the workflow YAML using `--hypothesis-profile` instead?

icechunk-python/tests/conftest.py
ianhi: that is the default in the inherited CI profile

ianhi:
> Nice, i thought i had created an issue for this but seem to have not done that.

whoops you did. i just forgot to link it: https://github.com/earth-mover/icechunk/issues/1737

> What does this do in CI? Do we backup the output to a cache?

In combination with this branch: https://github.com/HypothesisWorks/hypothesis/pull/4676

it will mean that a flaky strategy in a stateful test prints out it draw history. Which makes it way easier to debug a flaky issue.

icechunk-python/tests/conftest.py
dcherian: No the default is True: https://hypothesis.readthedocs.io/en/latest/reference/api.html#built-in-profiles . For the nightly we do want randomization on every run

ianhi: oh wait. why derandomize false? i was confused. i thought we would wnat this?

ianhi: ahh for nightly. fair enough

icechunk-python/tests/conftest.py
ianhi: that is better. swapped

ianhi:
immediately helpful output:

https://github.com/earth-mover/icechunk/actions/runs/23305810393/job/67780409517?pr=1870

ianhi:
the failure here is the same as #1862 which is patched in #1870

PR #1874, opened 2026-03-19
Create a SpecVersion enum for Python, and listing supported spec versions

### LINKED ISSUE #1413: Visibility on supported spec versions
Something like `ic.supported_spec_versions()` that listed out the `SpecVersionBin` enum with an indicator of which is default would be nice

### LINKED ISSUE #1541: `spec_version` in python is an `int`, should it be an enum?
The `int` type is a smell to me. At the rust level we have `V2dot0` or presumable `V2dot1` is a possibility and `2.1` is not an int. We could type it as a Number but why not use an enum like at the rust level instead?

li-em:
Fix #1541
Fix #1413 

Create a new `SpecVersion` enum for Python, mirroring the `SpecVersionBin` enum in Rust. It can be constructed from valid integers (1 and 2) when passed as argument in functions.

Change the `__repr__` for `SpecVersion` to point which variant is the current spec version.

Add a method `ic.supported_spec_versions()`  to list supported spec versions.

Avoid changing Python tests, allowing them to continue using integers for spec version instead of new enum.

### DIFF
--- a/icechunk/src/format/mod.rs
+++ b/icechunk/src/format/mod.rs
@@ -391,6 +391,8 @@ pub mod format_constants {
 
     use serde::{Deserialize, Serialize};
 
+    use super::IcechunkFormatErrorKind;
+
     /// Binary file type identifier in the file header.
     #[repr(u8)]
     #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -439,16 +441,21 @@ pub mod format_constants {
         V1 = 1u8,
         #[default]
         V2 = 2u8,
+        // When adding new versions here, don't forget to update the
+        // PySpecVersion enum in icechunk-python/src/repository.rs too!
     }
 
     impl TryFrom<u8> for SpecVersionBin {
-        type Error = String;
+        type Error = IcechunkFormatErrorKind;
 
         fn try_from(value: u8) -> Result<Self, Self::Error> {
             match value {
                 n if n == SpecVersionBin::V1 as u8 => Ok(SpecVersionBin::V1),
                 n if n == SpecVersionBin::V2 as u8 => Ok(SpecVersionBin::V2),
-                n => Err(format!("Bad spec version code: {n}")),
+                n => Err(IcechunkFormatErrorKind::InvalidSpecVersion {
+                    found: n,
+                    max_supported: Self::current() as u8,
+                }),
             }
         }
     }
@@ -671,11 +678,8 @@ mod tests {
         let future_version: u8 = 3;
         let result = SpecVersionBin::try_from(future_version);
         assert!(result.is_err());
+        let err = result.unwrap_err();
 
-        let err = IcechunkFormatErrorKind::InvalidSpecVersion {
-            found: future_version,
-            max_supported: SpecVersionBin::current() as u8,
-        };
         let msg = err.to_string();
         assert!(
             msg.contains("format version 3"),


icechunk-python/python/icechunk/__init__.py
```diff
@@ -303,6 +305,10 @@ def upgrade_icechunk_repository(
     return Repository(new_repo)
 
 
+def supported_spec_versions() -> list[SpecVersion]:
+    return [SpecVersion.v2, SpecVersion.v1]
```
li-em, line 309: Could try to list from enum here, but it might come a time in the future that we have some supported versions (but not all?)

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1896,6 +1897,16 @@ class ManifestFileInfo:
         """The number of chunk references contained in this manifest"""
         ...
 
+@final
+@total_ordering
```
li-em, line 1901: Total ordering allows us to do comparisons in Python (`SpecVersion.v2 > SpecVersion.v1 == True`)

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -1904,7 +1915,7 @@ class PyRepository:
         *,
         config: RepositoryConfig | None = None,
         authorize_virtual_chunk_access: dict[str, _AnyCredential | None] | None = None,
-        spec_version: int | None = None,
+        spec_version: SpecVersion | int | None = None,
```
li-em, line 1919: Take `int` as valid `spec_version`, but only in methods inputs; outputs should return `SpecVersion | None` instead of an int.

icechunk-python/src/repository.rs
```diff
@@ -887,6 +887,80 @@ fn mk_update(
...
 
+#[repr(u8)]
+#[pyclass(
+    eq,
+    eq_int,
+    ord,
+    rename_all = "snake_case",
+    skip_from_py_object,
+    name = "SpecVersion",
+    frozen
```
li-em, line 898: Small optimization, but since the enum is never mutated we can declare it as `frozen` here

icechunk-python/src/repository.rs
```diff
@@ -887,6 +887,80 @@ fn mk_update(
...
+    V1 = 1u8,
+    #[default]
+    V2 = 2u8,
+}
+
+impl<'py> FromPyObject<'_, 'py> for PySpecVersion {
+    type Error = PyErr;
+
+    /// Custom implementation that allows passing an integer or PySpecVersion
+    fn extract(ob: Borrowed<'_, 'py, PyAny>) -> Result<Self, Self::Error> {
```
li-em, line 911: `extract` does the validation from an `int` or `SpecVersion` passed along as function argument. Technically we could also supported strings in the future, but... better not =]

icechunk-python/src/repository.rs
```diff
@@ -887,6 +887,80 @@ fn mk_update(
...
+            SpecVersionBin::V2 => Self::V2,
+        }
+    }
+}
+
+#[pymethods]
+impl PySpecVersion {
+    pub(crate) fn __repr__(&self) -> String {
+        match self {
+            Self::V2 => "SpecVersion.v2 (current)".into(),
```
li-em, line 955: This is a bit cheesy, but allows us to print in the Python REPL or notebooks which one is the current spec version

icechunk-python/src/repository.rs
```diff
@@ -887,6 +887,80 @@ fn mk_update(
...
+            Self::V1 => "SpecVersion.v1".into(),
+        }
+    }
+
+    pub(crate) fn current(&self) -> Self {
+        Default::default()
+    }
```
li-em: Using `Default::default()` here so we don't ever have to update the code =]

icechunk-python/src/repository.rs
```diff
@@ -949,12 +1023,7 @@ impl PyRepository {
                     let config = config
                         .map(|c| c.try_into().map_err(PyValueError::new_err))
                         .transpose()?;
-                    let version = match spec_version {
-                        Some(n) => Some(
-                            SpecVersionBin::try_from(n).map_err(PyValueError::new_err)?,
-                        ),
-                        None => None,
-                    };
+                    let version = spec_version.map(|v| v.into());
```
li-em, line 1028: This validation happens in `extract` now

icechunk-python/tests/conftest.py
```diff
@@ -79,7 +81,9 @@ def write_chunks_to_minio(
 
 
 @pytest.fixture(
-    scope="function", params=[1, 2, None], ids=["spec-v1", "spec-v2", "no-spec-version"]
+    scope="function",
+    params=[1, SpecVersion.v2, None],
```
li-em, line 85: Leaving mixed `int` and `SpecVersion` here to try to test both when the fixture is used

icechunk-python/tests/test_spec_versions.py
```diff
@@ -101,3 +101,30 @@ def test_exists_and_fetch_spec_version_with_storage_settings() -> None:
...
+def test_version_int_or_specversion() -> None:
+    storage = ic.in_memory_storage()
+    repo = ic.Repository.create(storage, spec_version=ic.SpecVersion.v2)
+    assert repo.spec_version == ic.SpecVersion.v2
+    assert repo.spec_version == 2
+
+
+def test_version_supported_listing() -> None:
+    assert (
+        str(ic.supported_spec_versions()) == "[SpecVersion.v2 (current), SpecVersion.v1]"
```
li-em, line 130: This is how it ends up being printed in the REPL

icechunk/src/format/mod.rs
```diff
@@ -436,18 +436,18 @@ pub mod format_constants {
         Ord,
     )]
     pub enum SpecVersionBin {
-        V1dot0 = 1u8,
+        V1 = 1u8,
         #[default]
-        V2dot0 = 2u8,
+        V2 = 2u8,
```
li-em, line 443: I can revert these changes, but the `dot0` is a bit misleading because we don't intend to support non-integer versions

icechunk-python/src/repository.rs
```diff
@@ -887,6 +887,82 @@ fn mk_update(
...
+    ord,
+    rename_all = "snake_case",
+    skip_from_py_object,
+    name = "SpecVersion",
+    frozen
+)]
+#[derive(PartialEq, Default, Clone, PartialOrd, Debug)]
+pub enum PySpecVersion {
+    V1 = 1u8,
+    #[default]
```
dcherian, line 903: Can we add a comment to the Rust enum asking the editor to keep the Python enum in sync.

icechunk/src/format/mod.rs
paraseba: much better

icechunk-python/src/repository.rs
paraseba: love it

icechunk-python/src/repository.rs
li-em: Done! Also fixed the `TryFrom` impl to use the `InvalidSpecVersion` error, and updated the test to check it directly instead of creating a new error

PR #1876, opened 2026-03-20
Implement `fork` in Rust

dcherian:
We move `fork` to the rust `Session`. This allows to get rid of a bunch of state we were tracking on the Python `Session` 🥳.

`fork` is implemented by creating an anonymous snapshot `anon`, and returning a new Session whose parent is `anon` that allows writes but disallows commits. The user is expected to distribute these Sessions to workers, and merge them back in to the "base" Session which was forked initially.

- `Session.merge(ForkSesion)`, `ForkSession.merge(ForkSession)` are allowed
- `ForkSession.merge(Session)` will fail.

Thanks to this, we don't force a user to commit before executing distributed writes!

TODO:
- [x] add to stateful test suite (see #1018)
- [x] check GC behaviour for anonymous snapshots

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -389,9 +389,7 @@ impl<'a> CommitBuilder<'a> {
         }
 
         if self.amend {
-            self.session
-                .asset_manager
-                .fail_unless_spec_at_least(SpecVersionBin::V2dot0)?;
+            self.session.asset_manager.fail_unless_spec_at_least(SpecVersionBin::V2)?;
         }
 
         let commit_method =
@@ -818,7 +816,7 @@ impl Session {
     #[instrument(skip(self))]
     pub async fn move_node(&mut self, from: Path, to: Path) -> SessionResult<()> {
         // Icechunk 1 has no way to represent move in its on-disk format
-        self.asset_manager.fail_unless_spec_at_least(SpecVersionBin::V2dot0)?;
+        self.asset_manager.fail_unless_spec_at_least(SpecVersionBin::V2)?;
         // does the source node exist?
         let _ = self.get_node(&from).await?;
         // are we overwriting the destination node?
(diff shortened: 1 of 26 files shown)


icechunk-python/python/icechunk/store.py
```diff
@@ -67,19 +64,17 @@ def __eq__(self, value: object) -> bool:
...
     def __getstate__(self) -> object:
         # for read_only sessions we allow pickling, this allows distributed reads without forking
-        writable = not self.session.read_only
-        if writable and not self._for_fork:
+        session = self._store.session
+        if not session.read_only and not session.is_fork:
             raise ValueError(
                 "You must opt-in to pickle writable sessions in a distributed context "
                 "using Session.fork(). "
-                # link to docs
```
dcherian, line 75: TODO: fix this

icechunk/src/session.rs
dcherian, line 1296: this kept bugging me; we see 2 spans for a `commit`. now we have spans on "top-level" methods: `commit`, `commit_with_options`, `rewrite_manifests`, `flush`, `amend`.

icechunk-python/python/icechunk/session.py
```diff
@@ -445,13 +421,6 @@ async def commit_async(
         icechunk.NoChangesToCommitError
             If there are no changes to commit and allow_empty is False.
         """
-        if self._allow_changes:
```
dcherian, line 448: for better or worse I removed this "protection". i.e. we allow committing the base commit even after forking.

I think this is fine:
- it's trivial to override;
- maybe you want to commit the base in case the distributed workers fail.

icechunk-python/python/icechunk/session.py
```diff
@@ -673,21 +614,8 @@ def fork(self) -> "ForkSession":
...
-                "You should not need to fork a read-only session. Read-only sessions can be pickled and transmitted directly."
-            )
-        self._allow_changes = True
-        # force a deep-copy of the underlying Session,
-        # so that multiple forks can be created and
-        # used independently in a local session.
-        # See test_dask.py::test_fork_session_deep_copies for an example
-        return ForkSession(PySession.from_bytes(self._session.as_bytes()))
+        # TODO: Do we still need ForkSession?
+        return ForkSession(self._session.fork())
```
dcherian, line 618: I opted to keep this since the type is public and I've definitely seen it used in client code. Core logic is all in Rust though.

icechunk/src/session.rs
```diff
@@ -485,7 +499,7 @@ impl Session {
         storage: Arc<dyn Storage + Send + Sync>,
         asset_manager: Arc<AssetManager>,
         virtual_resolver: Arc<VirtualChunkResolver>,
-        branch_name: String,
+        branch_name: Option<String>,
```
paraseba, line 511: we should have a docstring for this argument. What does `none` mean? I assume a session on top of an anonymous snap

Issue #1873, opened 2026-03-19
resize+reindex inconsistent behavior


ianhi:
When you resize the deletion is in the change set which has an interaction with shift array. This means that where you put the commit impacts the end result

```python
import icechunk as ic
import zarr


def setup():
    repo = ic.Repository.create(storage=ic.in_memory_storage())
    session = repo.writable_session("main")
    arr = zarr.open_array(session.store, path="d", mode="w", shape=(3,), chunks=(1,), dtype="f4", fill_value=-1)
    arr[:] = [1.0, 2.0, 3.0]
    session.commit("init")
    return repo


def resize_then_shift(repo):
    session = repo.writable_session("main")
    arr = zarr.open_array(session.store, path="d")
    arr.resize((2,))
    session.shift_array("/d", (-1,))
    arr = zarr.open_array(session.store, path="d")
    print(f"resize(2,), shift(-1,):                 shape={arr.shape} values={arr[:]}")
    session.commit("resize+shift")
    arr = zarr.open_array(repo.writable_session("main").store, path="d")
    print(f"  after commit:                         shape={arr.shape} values={arr[:]}")


def resize_commit_then_shift(repo):
    session = repo.writable_session("main")
    arr = zarr.open_array(session.store, path="d")
    arr.resize((2,))
    session.commit("resize")
    session = repo.writable_session("main")
    session.shift_array("/d", (-1,))
    arr = zarr.open_array(session.store, path="d")
    print(f"resize(2,), commit, shift(-1,):         shape={arr.shape} values={arr[:]}")
    session.commit("shift")
    arr = zarr.open_array(repo.writable_session("main").store, path="d")
    print(f"  after commit:                         shape={arr.shape} values={arr[:]}")


repo = setup()
arr = zarr.open_array(repo.writable_session("main").store, path="d")
print(f"initial:                                shape={arr.shape} values={arr[:]}")
resize_then_shift(setup())
resize_commit_then_shift(setup())
```
```
initial:                                shape=(3,) values=[1. 2. 3.]
resize(2,), shift(-1,):                 shape=(2,) values=[ 2. -1.]
  after commit:                         shape=(2,) values=[ 2. -1.]
resize(2,), commit, shift(-1,):         shape=(2,) values=[2. 2.]
  after commit:                         shape=(2,) values=[2. 2.]
```

PR #1877, opened 2026-03-20
Handle empty chunks in reindex

ianhi:
Fix: https://github.com/earth-mover/icechunk/issues/1862
and fix: https://github.com/earth-mover/icechunk/issues/1873

Handle the case of reindex not shifting empty chunks by also allowing a `backward` argument to `reindex` 

I also added a small optimization of not calling the backwards function when we don't need to. This can give a nice speedup for dense arrays. and is especially important when using a python callback

(Big thanks to @paraseba for helping me through understandign the core of the bug and planning the solution)

## Behavior change!

This has changed (to my eyes fixed) the behavior with shift_array we discussed before where edge positions that are vacated retain stale data. We had that as performance optimization to not iterate all chunks, but with backwards it's essentially no cost to check now. So I've added that if an out of bounds position would shift to a chunk it is now deleted. This feels much more consistent to me.

## TODO

- ~There's still some testing stuff to work through here. Adding more tests for reindex + updating the stateful tests.~

- ~Performance optimization. If we keep track of what indexes we've already visted with forward we can avoid computing backwards when it is no necessary~
- Could also allow a single ForwardBackward function as an optimization.


icechunk/src/session.rs
```diff
@@ -876,6 +896,20 @@ impl Session {
                     .into());
                 }
             }
+            if let Some(ref backwards) = backwards {
+                if destinations.contains(&old_chunk_index) {
+                    continue;
+                }
```
ianhi, line 902: This gives a speedup for forward/backward calls - especially when they originate from python. At least for a dense array where we would otherwise have a lot of unnecessary backwards calls.

To benchmark I made a dense array with 100000 chunks and then reimplemented the shift array logic in python. results:

|                | Before | After |
  |---------------------|--------|-------|
  | forward only (py)       | 0.27s  | 0.27s |
  | forward + backward (py)  | 0.72s  | 0.35s |
  | shift_array (rust)  | 0.47s  | 0.31s |

PR #1888, opened 2026-03-23
Add upstream CI workflow to test zarrs against latest icechunk main

jhamman:
Adds a daily CI workflow that clones zarrs_icechunk, patches its Cargo.toml to use the local icechunk checkout via [patch.crates-io], and runs its build and test suite. On scheduled failures, auto-creates or updates a GitHub issue with build/test logs.


jhamman:
https://github.com/earth-mover/icechunk/actions/runs/23417937748/job/68116992722 shows the successful run for this new action.

PR #1896, opened 2026-03-23
Fix amend and moved nodes in transaction log

### LINKED ISSUE #1822: Verify amend on top of amend in the transaction log merge
#1706 checks for amends on top of rearrange session, but there is a missing case: amend on top of amend is allowed, but this case in not being dealt with in the tx log merge:

- https://github.com/earth-mover/icechunk/blob/c66cb108cbd795e0a6adb8e9d153f822bf6a8c91/icechunk/src/format/transaction_log.rs#L468-L469

Write a test that does this `rearrange + rearrange amend` and checks the diffs. If it fails we should implement the merge logic. If that gets too tricky, we just disallow amends on top of rearrange all together.

li-em:
Fix #1822

When amending multiple Rearrange sessions (those containing only moves) the transaction log was not merging these amended moves.

The transaction log includes the `NodeId` for the moved node, so use that to disambiguate moves (instead of using paths), and collapse moves for same node ID (replacing `to` field from older move).

Also remove any identity move (`/a -> /a`), especially after they are reproduced after collapsing multiple moves.

### DIFF
--- a/icechunk/src/repository.rs
+++ b/icechunk/src/repository.rs
@@ -4443,7 +4443,6 @@ mod tests {
         // Check if transaction log has no moves, since everything was moved
         // back to their initial location.
         let tx_log = repo.asset_manager.fetch_transaction_log(&amend_snap_id).await?;
-        dbg!(tx_log.moves().collect::<Vec<_>>());
         assert!(tx_log.moves().count() == 0);
 
         let diff = repo


ianhi:
I would love to add property test here that does the moves all in one commit on one store, and then in two commits amended in another store and then compares the transaction logs after the fact.

icechunk-format/src/transaction_log.rs
```diff
@@ -376,9 +376,75 @@ impl TransactionLog {
...
+        // ```
+        // Move {
+        //   from: /source,
+        //   to: /src,
+        // },
+        // ```
+
+        // Save moves into an inverted map (to -> from) to make it easy
+        // to check if we have overlapping moves.
+        let mut moved_map: BTreeMap<Path, Path> = Default::default();
```
ianhi: FWIW I am also constructing a BTreeMap of moves in the move fixes PR. which can maybe be reused here

ianhi: would the BTreeMap (reverse moves) or the HashMap (forward moves) be useful if it was exposed out of the movetracker? https://github.com/earth-mover/icechunk/pull/1910

li-em: Not quite, because those are at `icechunk`, and `icechunk-format` is a dependency of `icechunk` but not the other way around...

Could maybe move to `icechunk-types` and then be used in both, but in any case would need to rebuild here from transactions, so I don't think we can reuse the MoveTracker here

ianhi:
does this also handle intersecting move chains well? I suspect yes but good to think about 


for example if i move a child in. then later move it

something like:

given `a, b, c`

`a-> b/a`
`b->d`
`d/a-> c/e`

ianhi (changes requested):
I extracted my move hypothesis tests onto this PR and seem to be able to induce some panics here

```

thread 'tokio-rt-worker' (54590743) panicked at icechunk-format/src/transaction_log.rs:424:13:
assertion failed: {
    use itertools::MultiUnzip as _;
    let (from, to): (BTreeSet<&Path>, BTreeSet<&Path>) =
        moved_map.iter().filter_map(|(to, from)|
                    if to != from {
                        Some((from, to))
                    } else { None }).multiunzip();
    from.is_disjoint(&to)
}
FAILED
```


This is a variant of what hypothesis found:

```python
import icechunk as ic
import zarr

repo = ic.Repository.create(storage=ic.in_memory_storage())
root.create_group("foo")
root.create_array("foo/bar", shape=(1,), dtype="i4")
session.commit("init")

# First move: rename group (committed)
session = repo.rearrange_session("main")
session.move("/foo", "/f")
session.commit("commit msg")

# Second move: move child to a path that was the original source (amended)
session = repo.rearrange_session("main")
session.move("/f/bar", "/foo")
session.amend("amend message")  # panics here
```

```
icechunk.IcechunkError:   × session error: a concurrent task failed
  │
  │ context:
  │    0: icechunk::session::commit_inner
  │            with amend message max_concurrent_nodes=1 rewrite_manifests=false commit_method=Amend allow_empty=false
  │              at icechunk/src/session.rs:1554
  │
```

li-em:
> I extracted my move hypothesis tests onto this PR and seem to be able to induce some panics here

The interesting bit is that the assumption on the check is wrong, because it doesn't consider something moving to a place that was previously used. Removing the check the diff for the Python code you posted shows
```
    moved_nodes: [
        Move {
            from: /foo,
            to: /f,
        },
        Move {
            from: /f/bar,
            to: /foo,
        },
    ],
```
which is the expected value in the transaction log

---

I wonder if I should wait for #1910 to be merged and then improve this PR, in case there are more corner cases I didn't consider for the merging solution in this PR

ianhi:
> I wonder if I should wait for https://github.com/earth-mover/icechunk/pull/1910 to be merged and then improve this PR, in case there are more corner cases I didn't consider for the merging solution in this PR

I think that makes sense. I feel so much more confident about this stuff once hypothesis passes

ianhi:
ok without the debug assert there is an interesting failure. which is that a single commit actually does not do the same merging you are doing here. So amend fully merges the moves, but a single commit does not

```
E         diff (one commit): [('/0', '/1hU'), ('/xf4vq', '/0')]
E         diff (amended):    [('/xf4vq', '/1hU')]
```

the diff feels a little in reverse here. which we whould probably alslo fix it should be `('/xf4vq', '/0'), ('/0', '/1hU')` which is correctly collapsed in the

ianhi:
Just checked and the edge case with two moves  above:

given a, b, c
```
a-> b/a
b->d
d/a-> c/e
```

is consistent between amemnd and not amend, but neither are fully collapsed.

icechunk-format/src/transaction_log.rs
```diff
@@ -407,9 +407,84 @@ impl TransactionLog {
...
+                // if this is the first move we see, just insert.
+                // Otherwise we check to see if the current "from" is already in the moved_map
+                // (meaning this "from" is a "to" from another move)
+                // and if so remove it and replace with a new (to, old_from) pair,
+                if let Some(old_from) = moved_map.remove(&node_id) {
+                    moved_map.insert(node_id, Move { to, ..old_from });
+                } else {
+                    moved_map
+                        .insert(node_id.clone(), Move { to, from, node_id, node_type });
+                }
```
dcherian: Would the `.entry` API help clean this up?

li-em: definitely,
```rust
moved_map
    .entry(node_id.clone())
    .and_modify(|m| m.to = to.clone())
    .or_insert_with(|| Move { to, from, node_id, node_type });
```
is much easier to read =]

dcherian: Very clean!

icechunk/src/repository.rs
```diff
@@ -4233,4 +4233,368 @@ mod tests {
...
+            .readonly_session(&VersionInfo::SnapshotId(first_rearr_snap_id.clone()))
+            .await?;
+        assert!(session.get_group(&"/f".try_into().unwrap()).await.is_ok());
+        assert!(session.get_group(&"/f/bar".try_into().unwrap()).await.is_ok());
+        assert!(session.get_group(&"/foo/bar".try_into().unwrap()).await.is_err());
+        // Check if transaction log has two moves,
+        // one for parent /foo -> /f
+        // another for child /foo/bar -> /f/bar
+        let tx_log =
+            repo.asset_manager.fetch_transaction_log(&first_rearr_snap_id).await?;
```
dcherian, line 4510: what happens if you amend a commit with `f -> foo` do we get no moves? ;) We should...

icechunk/src/repository.rs
```diff
@@ -4233,4 +4233,368 @@ mod tests {
...
+        assert!(session.get_group(&"/a".try_into().unwrap()).await.is_err());
+        assert!(session.get_group(&"/b".try_into().unwrap()).await.is_err());
+        assert!(session.get_group(&"/b/a".try_into().unwrap()).await.is_err());
+        // Check if transaction log has just two moves
+        // one for /b -> /d
+        // another for /a -> /c/e
+        //   /a -> /d/a -> /c/e is a collapsed move
+        let tx_log =
+            repo.asset_manager.fetch_transaction_log(&third_amend_snap_id).await?;
+        dbg!(tx_log.moves().collect::<Vec<_>>());
```
dcherian:
```suggestion
```

icechunk/src/repository.rs
```diff
@@ -4233,4 +4233,368 @@ mod tests {
...
+        let diff = repo
+            .diff(
+                &VersionInfo::SnapshotId(Snapshot::INITIAL_SNAPSHOT_ID),
+                &VersionInfo::SnapshotId(third_amend_snap_id.clone()),
+            )
+            .await?;
+
+        assert_eq!(diff.moved_nodes.len(), 2);
+
+        dbg!(diff);
```
dcherian:
```suggestion
```

icechunk/src/repository.rs
```diff
@@ -4233,4 +4233,368 @@ mod tests {
...
+        assert!(session.get_group(&"/d/a".try_into().unwrap()).await.is_ok());
+        assert!(session.get_group(&"/a".try_into().unwrap()).await.is_err());
+        assert!(session.get_group(&"/b".try_into().unwrap()).await.is_err());
+        assert!(session.get_group(&"/b/a".try_into().unwrap()).await.is_err());
+        // Check if transaction log has two moves
+        // one for parent /b -> /d
+        // another for child /a -> /d/a
+        let tx_log =
+            repo.asset_manager.fetch_transaction_log(&second_amend_snap_id).await?;
+        dbg!(tx_log.moves().collect::<Vec<_>>());
```
dcherian:
```suggestion
```

dcherian: ok there are lots of these here; let's remove them or comment them out

icechunk-format/src/transaction_log.rs
```diff
@@ -407,9 +407,79 @@ impl TransactionLog {
...
+        // ```
+        // Move {
+        //   from: /source,
+        //   to: /src,
+        // },
+        // ```
+
+        // Save moves into a map (NodeId -> Move) to make it easy
+        // to check if we have overlapping moves.
+        let mut moved_map: BTreeMap<NodeId, Move> = Default::default();
```
ianhi: Could this just be a `HashMap`? I don't think we use any of the Btreemap properties?

ianhi:
> I suggest one more test foo -> f -> foo to see what happens

I agree. As I read it this will end up as a move of `foo->foo` which we should not include I think - and for that matter we maybe write those to disk for a single commit, whcih should also not be allowed

ianhi:
also what should happen if you make a move, commit, then revert the move in an amend? Should you have to pass `allow_empty=True`?

icechunk/src/repository.rs
li-em: Added a filter to remove identity moves (`/foo -> /f -> /foo`), including children

icechunk/src/repository.rs
li-em: yeah, these are useful for debugging, but can add as needed

icechunk-python/tests/test_move.py
```diff
@@ -176,12 +176,10 @@ def test_moves(
     # Moves in the tx log must be sorted by final path
     snap_after = repo.lookup_branch("main")
     diff = repo.diff(from_snapshot_id=snap_before, to_snapshot_id=snap_after)
-    assert_moves_sorted_by_final_path(diff.moved_nodes)
+    if diff.moved_nodes:
+        assert_moves_sorted_by_final_path(diff.moved_nodes)
```
li-em, line 180: @ianhi this caught an identity move (`/1 -> /0, /0 -> /1`), and then failed because `assert_moves_sorted_by_final_path` checks if `diff.moved_nodes` is empty. In this case it is indeed empty, so skipping the assert.

li-em: also added same check in `icechunk-python/tests/test_zarr/test_stateful.py`

icechunk/src/repository.rs
```diff
@@ -4369,6 +4369,102 @@ mod tests {
...
+            .await?;
+        assert!(session.get_group(&"/foo".try_into().unwrap()).await.is_ok());
+        assert!(session.get_group(&"/foo/bar".try_into().unwrap()).await.is_ok());
+        assert!(session.get_group(&"/f".try_into().unwrap()).await.is_err());
+        assert!(session.get_group(&"/f/bar".try_into().unwrap()).await.is_err());
+
+        // Check if transaction log has no moves, since everything was moved
+        // back to their initial location.
+        let tx_log = repo.asset_manager.fetch_transaction_log(&amend_snap_id).await?;
+        dbg!(tx_log.moves().collect::<Vec<_>>());
```
dcherian: these snuck in :)

dcherian (approved):
Nice!

icechunk/src/repository.rs
li-em: sorry, I love my `dbg!` checks...

PR #1901, opened 2026-03-23
Use assume in tree generation to prune invalid cases

li-em:
Pretty sure this is the wrong fix (because we want `children` to be as many as they were before), but need someone with more hypothesis strategy experience to fix it properly :upside_down_face:

### DIFF
--- a/icechunk-python/python/icechunk/testing/trees.py
+++ b/icechunk-python/python/icechunk/testing/trees.py
@@ -16,6 +16,7 @@ from dataclasses import dataclass, field
 
 import hypothesis.strategies as st
 import numpy as np
+from hypothesis import assume
 
 import zarr
 import zarr.abc.store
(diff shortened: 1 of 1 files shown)


icechunk-python/python/icechunk/testing/trees.py
```diff
@@ -235,7 +235,7 @@ def rebuild_with_names(
         )
         existing_names = existing_names | set(new_names)
         children: dict[str, Node] = {}
-        for name, child in zip(new_names, group.children.values(), strict=True):
+        for name, child in zip(new_names, group.children.values(), strict=False):
```
ianhi: this should be fine. they should alwayas be the same length. why are we requioring strict=False in lint? seems like an ok thing to allow?

dcherian:
This is hiding a real bug

```
[77](https://github.com/earth-mover/icechunk/actions/runs/23460025444/job/68259871010#step:7:278)
FAILED tests/test_store_properties.py::test_list_dir - ValueError: zip() argument 2 is longer than argument 1
while generating 'tree' from trees(
```

https://github.com/earth-mover/icechunk/actions/runs/23460025444/job/68259871010

icechunk-python/python/icechunk/testing/trees.py
```diff
@@ -235,6 +236,8 @@ def rebuild_with_names(
         )
         existing_names = existing_names | set(new_names)
         children: dict[str, Node] = {}
+        assume(len(new_names) == len(group.children))
+
```
li-em, line 240: Revert the strict zip change, use `assume` to make sure they are the same size.

(but probably there are more efficient ways than doing this...)

li-em:
Fixed properly in #1907

Issue #1632, opened 2026-02-09
Integrate with pyo3-object_store


vladidobro:
Hello Icechunk developers,
I have not seen a ticket for this

### Problem

Currently Icechunk reeimplements (if I read correctly) many of the functionality of obstore - python builders that create and wrap rust ObjectStores.
This leads to incompatibility between different libraries that use object_store from rust under the hood (such as polars), and creates possibly unnecessary burden for Icechunk in maintaining these wrappers and keeping up with upstream.

For example
- `icechunk.azure_storage(container: str)` vs `obstore.store.AzureStore(container_name: str)`
- there is no `icechunk.storage.from_url`
- refreshable credentials were added to icechunk only recently, even if obstore has them for some while

### Proposed Solution

Icechunk could integrate with https://crates.io/crates/pyo3-object_store.
> This provides Python builder classes so that Python users can easily create [Arc<dyn ObjectStore>][object_store::ObjectStore] instances, which can then be used in pure-Rust code.

This could result in improved user experience and less maintenance, but there are surely some disadvantages too.

### Additional context

I have briefly discussed it with @kylebarron in obstore issues: https://github.com/developmentseed/obstore/issues/531.

A similar ticket in polars: https://github.com/pola-rs/polars/issues/24063

mpiannucci:
Hi @vladidobro, this is something we want to support but it's tricky because of how icechunk serializes storage configurations. We will need some effort in figuring out the best way to integrate this, it's definitely possible. 

Also our refreshable credentials came along before obstores existed so that was another motivation. 

We are currently working to modularize the storage bit to make this easier but there are still hard bits left, especially around virtual chunks

Happy to talk about this more in depth if desired

vladidobro:
Hi @mpiannucci,

> Also our refreshable credentials came along before obstores existed so that was another motivation.

hm, sorry, I don't know how I got that mixed up.

> Happy to talk about this more in depth if desired

Thank you, I would love to hear about the tricky parts, but I am not sure it will be very useful, as I am not deeply neither Rust nor Icechunk internals.

kylebarron:
I created https://github.com/earth-mover/icechunk/pull/2160, which prototypes passing obstore stores into icechunk. This implementation is a bit... hacky? in the sense that it tries to extract the config from an obstore store and convert the parameters into something that icechunk natively implements.

[...]

PR #1902, opened 2026-03-23
feat: Azure refreshable credentials

maxrjones:
This PR implements refreshable Azure credentials following the same patterns used for GCS and AWS. Similar to those backends, I didn't add tests for the refreshable credentials. However, I did test that this works via the following uv script:

```python
# /// script
# dependencies = [
# "h5py>=3.15.1",
# "virtualizarr[icechunk,netcdf3] @ git+https://github.com/zarr-developers/VirtualiZarr.git",
# "icechunk @ git+https://github.com/maxrjones/icechunk.git@azure-refreshable-creds#subdirectory=icechunk-python",
# "requests",
# ]
# ///
"""
Demo: Azure refreshable credentials with Icechunk + VirtualiZarr.

Virtualizes a Planetary Computer NetCDF file into an Icechunk repo,
then reads it back using refreshable SAS token credentials.

Usage:
    uv run demo_azure_refreshable.py
"""

import warnings
from datetime import datetime

import requests
import xarray as xr

import icechunk
from icechunk import (
    AzureCredentials,
    AzureRefreshableCredential,
    AzureStaticCredentials,
    Credentials,
    ObjectStoreConfig,
    RepositoryConfig,
    VirtualChunkContainer,
    azure_refreshable_credentials,
    containers_credentials,
)
from obspec_utils.registry import ObjectStoreRegistry
from obstore.auth.planetary_computer import PlanetaryComputerCredentialProvider
from obstore.store import AzureStore
from virtualizarr.parsers import HDFParser

warnings.filterwarnings(
    "ignore",
    message="Numcodecs codecs are not in the Zarr version 3 specification*",
    category=UserWarning,
)

ACCOUNT = "ukmoeuwest"
CONTAINER = "deterministic"
PREFIX = "global/near-surface/20260302T0000Z"
SAS_TOKEN_URL = (
    f"https://planetarycomputer.microsoft.com/api/sas/v1/token/{ACCOUNT}/{CONTAINER}"
)


def get_credential() -> AzureRefreshableCredential:
    """Fetch a fresh SAS token from Planetary Computer.

    This function is pickled by Icechunk and called whenever the token expires.
    """
    resp = requests.get(SAS_TOKEN_URL)
    resp.raise_for_status()
    data = resp.json()
    expires_at = datetime.fromisoformat(data["msft:expiry"].replace("Z", "+00:00"))
    return AzureRefreshableCredential.SasToken(
        token=data["token"], expires_after=expires_at
    )


def main():
    https_url = f"https://{ACCOUNT}.blob.core.windows.net/{CONTAINER}/{PREFIX}"
    az_prefix = f"az://{CONTAINER}"
    pc = PlanetaryComputerCredentialProvider(https_url)
    store = AzureStore(credential_provider=pc)
    registry = ObjectStoreRegistry({az_prefix: store})

    config = RepositoryConfig()
    vcc = VirtualChunkContainer(
        url_prefix=f"{az_prefix}/",
        store=ObjectStoreConfig.Azure({"account": ACCOUNT}),
    )
    config.set_virtual_chunk_container(vcc)

    # Step 1: Virtualize Planetary Computer data into Icechunk
    print("Step 1: Virtualize Planetary Computer data into Icechunk")
    sas_token = str(pc()["sas_token"])
    storage = icechunk.in_memory_storage()
    repo = icechunk.Repository.create(
        storage,
        config=config,
        authorize_virtual_chunk_access={
            vcc.url_prefix: Credentials.Azure(
                AzureCredentials.Static(AzureStaticCredentials.SasToken(sas_token))
            ),
        },
    )

    parser = HDFParser()
    for list_stream in store.list():
        for entry in list_stream:
            url = f"{az_prefix}/{PREFIX}/{entry['path']}"
            print(f"  Virtualizing {entry['path']}...")
            ds = open_virtual_dataset(
                url=url, registry=registry, parser=parser, decode_times=False
            )
            ds.vz.to_icechunk(session.store)
            break
        break
    session.commit("Add virtual dataset")

    # Step 2: Reopen with refreshable credentials and read data
    print("\nStep 2: Load from Icechunk with refreshable Azure credentials")
    repo = repo.reopen(
        authorize_virtual_chunk_access=containers_credentials(
            {f"{az_prefix}/": azure_refreshable_credentials(get_credential)}
        ),
    )
    ds = xr.open_zarr(session.store, consolidated=False, decode_times=False)
    print(ds)
    print()
    for name in ds.data_vars:
        print(f"  {name}: max={ds[name].max().values}")


if __name__ == "__main__":
    main()
```


maxrjones:
This works, but I'm not super happy with the design because it duplicates what's already available in obstore and the manual adaptation is repetitive. It would be better if credential providers were interchangeable between icechunk and obstore, with an API like:

```python
repo = Repository.open(
    storage,
    authorize_virtual_chunk_access={
        "az://container/": PlanetaryComputerCredentialProvider(url),
    },
)
```

Could obspec provide credential provider protocols to enable icechunk and obstore to accept any callable matching the expectations @kylebarron? Would Icechunk be interested in a protocol based approach @TomNicholas?

mpiannucci:
I think that is best left to another PR since this works, and we just reorganized crates which may make this easier to reason about. 

See https://github.com/earth-mover/icechunk/issues/1632

I think we are at a point where we could enable this from Python but it's not trivial because of https://docs.rs/pyo3-object_store/latest/pyo3_object_store/#abi-stability

mpiannucci (approved):
Thanks for this @maxrjones . it looks great

paraseba:
Thank you @maxrjones !!

kylebarron:
@maxrjones asked

> Could obspec provide credential provider protocols to enable icechunk and obstore to accept any callable matching the expectations @kylebarron? Would Icechunk be interested in a protocol based approach @TomNicholas?

I don't understand part of the first question. Obspec has nothing to do with credentials. 

> I think we are at a point where we could enable this from Python but it's not trivial because of [docs.rs/pyo3-object_store/latest/pyo3_object_store#abi-stability](https://docs.rs/pyo3-object_store/latest/pyo3_object_store/#abi-stability)

1. This is unrelated to supporting a credential provider callback. You could easily allow the user to pass in a callback to customize how to fetch credentials. This means you could support any Azure authentication solely via the _user_ depending on the azure Python library.
2. If you supported [obspec protocols](https://developmentseed.org/obspec/latest/), then you have no need for sharing rust classes between two Python packages. The user could pass in an obstore class and you would call its Python API without trying to access it through Rust.
3. As noted in https://docs.rs/pyo3-object_store/latest/pyo3_object_store/#abi-stability, `pyo3-object_store` has workarounds (that have been used in production in `async-tiff` and others) to recreate a new `ObjectStore` instance _on the Rust side_ from a user's Python Obstore class, essentially by pickling. Here the primary downside is that you have two separate HTTP contexts, but it does work effectively.

maxrjones:
thanks for the quick reviews, Matt and Seba!

@kylebarron, probably this should move to a dedicated issue if the conversation continues, but I imagined protocols for credential providers could fall under "a Python protocol for interfacing with object storage". 🤷‍♂️ sounds unlikely though, and my needs are met via duplication anyways.

kylebarron:
I see the confusion; obspec defines the API for how to interface with a "store" instance _after_ you've created it. 

These are equivalent to the _methods_ on an obstore class. But obspec doesn't concern itself at all with how to _create_ such a store. This keeps it fully generic, because some other object store could come along with an entirely different credential process. 

The credentialization is _only_ an obstore concern.

Issue #1562, opened 2026-01-15
Issues with `move`


ianhi:
# Can't move an empty group

```python
import icechunk as ic
repo = ic.Repository.create(     
    storage=ic.storage.in_memory_storage(),
)                                                                
with repo.transaction('main', message='create') as store:
    root = zarr.group(store=store, overwrite=True)           
    root.create_group("foo/bar")
    root.create_group("baz/")
    
display(root.tree())

session = repo.rearrange_session("main")
session.move("/foo/bar/", "/baz/bar")
# session.move("/", "../")
root = zarr.open_group(store=session.store)
root.tree()

```

```python
/
├── baz
└── foo
    └── bar



---------------------------------------------------------------------------
IcechunkError                             Traceback (most recent call last)
Cell In[25], line 13
     10 display(root.tree())
     12 session = repo.rearrange_session("main")
---> 13 session.move("/foo/bar/", "/baz/bar")
     14 # session.move("/", "../")
     15 root = zarr.open_group(store=session.store)

File ~/Documents/dev/icechunk/icechunk-python/python/icechunk/session.py:157, in Session.move(self, from_path, to_path)
    156 def move(self, from_path: str, to_path: str) -> None:
--> 157     return self._session.move_node(from_path, to_path)

IcechunkError:   x session error: node not found at `/foo/bar/`: existing node not found
  | 
  | context:
  |    0: icechunk::session::get_node
  |            with path=Path(Utf8PathBuf { _encoding: "unix", inner: "/foo/bar/" })
  |              at icechunk/src/session.rs:742
  |    1: icechunk::session::move_node
  |            with from=Path(Utf8PathBuf { _encoding: "unix", inner: "/foo/bar/" }) to=Path(Utf8PathBuf { _encoding: "unix", inner: "/baz/bar" })
  |              at icechunk/src/session.rs:542
  | 
```

trying without the trailing slash just deletes teh group

```
/
├── baz
└── foo
    └── bar

moving: /foo/bar -> /baz/bar

/
├── baz
└── foo
```

# Trying to move the root group fails with the wrong error

I think this *should* fail, but not with an error about write operations, should be caught early that moving the root group is not allowed.

Same setup as above. But change the move to:

```python
session.move("/", "/new")
```

```python
/
├── baz
└── foo
    └── bar

# long error - truncated
# ultimately fails with

[...]

PR #1910, opened 2026-03-24
Do eager tracking of moves

ianhi:
Fixes: https://github.com/earth-mover/icechunk/issues/1562

We weren't correctly calculating `get_node` on any non root nodes when a node had been moved into that node from elsewhere.

So there are three plcaes to fix

1. movetracker

 The simplest fix for this seems to be to eagerly compute the consequences of all moves as we accumulate them, instead of at the end. This is very useful for the case of multi move sequences

`/a -> /b/a`,  `/b->/c`. when we list what is under `/c` we need to be able to capture `a`

I think the total amount of iteration is the same or less then before because when commit is called ultimately all nodes get iterated over to track if they moved or not.

I added two data structures to MoveTracker. To track moves based on final or original key. This is because we need efficient lookups by both at different points.

2. move_node

pass the subtree info to movetracker so it can properly compute everything

3. get_node

use the new moved_into to figure out what got movedinto a group from elsewhere.

### Testing

added lots of rust tests with weird edge cases I ran into while figuring this out. I used some helpers to remove some of the visual noise of unwrap etc as I find it much easier to follow that way. but happy to revert that stuff as it is not actually necessary.

Also added hypothesis based tree move tests which is how i found all those edge cases. passes consistently now. so pretty happy with that.

### DIFF
--- a/icechunk/src/change_set.rs
+++ b/icechunk/src/change_set.rs
@@ -114,9 +114,7 @@ pub enum MovedFrom<'a> {
 #[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub struct MoveTracker {
     moves: Vec<Move>,
-    #[serde(default)]
     nodes_by_original: HashMap<Path, Path>,
-    #[serde(default)]
     nodes_by_final: BTreeMap<Path, Path>,
 }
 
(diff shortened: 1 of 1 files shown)


icechunk-python/python/icechunk/testing/trees.py
```diff
@@ -76,15 +76,64 @@ def _write(group: zarr.Group, node: GroupNode) -> None:
 
 Node = ArrayNode | GroupNode
 
+# Fixed dtype/shape — we're testing path handling, not dtype serialization.
+DTYPE = np.dtype("i4")
+SHAPE = (1,)
+
+
+def tree_from_str(spec: str) -> GroupNode:
```
ianhi, line 85: exists solely to make the `@example`s for hypothesis more human readable.

ianhi:
there may be a bit of extra verbosity in the comments - a consequence of me slowly going through the rust and leaving (or having claude leave me) notes. believe ti or this is hte pared back version, but I can try to trim them down more

PR #1912, opened 2026-03-24
Move all dependencies to workspace, use workspace=true in members

li-em:
With the new internal crates it is becoming more complicated to track dependency versions. This PR moves all the dependencies to the workspace Cargo.toml, and use `dep.workspace = true` in the member's Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,85 @@ resolver = "2"
 rust-version = "1.94.0"
 edition = "2024"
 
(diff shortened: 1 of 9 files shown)


paraseba:
feel free to merge if you have been careful about the details of features and dependencies. Maybe claude can do a careful review.

li-em:
My control here was not changing the `Cargo.toml` at all =]

PR #1914, opened 2026-03-25
First pass at a V2 spec

paraseba:
This is only the beginning of the rewrite. I think it's good to merge after feedback, but at some point we should do a full rewrite of the spec. In this PR the focus was to keep as similar to V1 as possible.


ianhi:
just for review can you put the. contents back into spec.md so we can see the line by line diff more easily?

paraseba:
Good idea @ianhi , done. I'll bring the structure back before merging

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
 
 - `$ROOT` base URI (s3, gcs, local directory, etc.)
-- `$ROOT/config.yaml` optional persistent default configuration for the repository
-- `$ROOT/refs/` reference files
+- `$ROOT/repo` repo info file, entry point for all operations
 - `$ROOT/snapshots/` snapshot files
 - `$ROOT/manifests/` chunk manifests
 - `$ROOT/transactions/` transaction log files
 - `$ROOT/chunks/` chunks
+- `$ROOT/overwritten/` old versions of the repo info object, used as backup and to maintain the ops log
```
ianhi, line 123: ops log isn't defined yet i think?

icechunk-python/docs/docs/spec.md
ianhi, line 28: This is a little confusing because this is notn a goal in the sense that the other bullets are.

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
+This object is the main entry point for an Icechunk repository, and the only mutable object in an Icechunk V2 repo. Every read operation starts by fetching the repo info object. Every Icechunk update, from making a commit to deleting a tag, makes a conditional write on this object.
 
-References are very important in the Icechunk design.
-Creating or updating references is the point at which consistency and atomicity of Icechunk transactions is enforced.
-Different client sessions may simultaneously create two inconsistent snapshots; however, only one session may successfully update a reference to point it to its snapshot.
+This object stores:
 
-References (both branches and tags) are stored as JSON files, the content is a JSON object with:
+- The repository configuration.
+- All branches, tags and deleted tag names.
```
ianhi, line 137: branches are defined below this now instead of before mentioning. Same for tags i think

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
+  After creation, branches may be updated to point to a different snapshot.
+- **Tags** are _immutable_ references to a snapshot.
+  A repository may contain zero or more tags.
+  After creation, tags may never be updated, unlike in Git. Tag delete is allowed, but a new tag with the name of a deleted one cannot be added.
 
 #### Snapshot Files
 
-The snapshot file fully describes the schema of the repository, including all arrays and groups.
+The snapshot file fully describes the schema of the repository, including all arrays and groups. Each commit to an Icechunk
+repository creates a new snapshot file. This snapshot iniforms what are the available groups and arrays
```
ianhi, line 200:
```suggestion
repository creates a new snapshot file. This snapshot informs what are the available groups and arrays
```

not sure informs is quite the right word though

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
+in this commit, and provides a way to access the chunk manifests for each array.
```
ianhi, line 201: What does provides a way mean?

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
 
 The snapshot file is encoded using [flatbuffers](https://github.com/google/flatbuffers). The IDL for the
-on-disk format can be found in [the repository file](https://github.com/earth-mover/icechunk/tree/main/icechunk/flatbuffers/snapshot.fbs)
+on-disk format can be found in [the fbs file](https://github.com/earth-mover/icechunk/blob/404100b584fb7ac70de860bd430aa8291df98c4d/icechunk-format/flatbuffers/snapshot.fbs).
 
 The most important parts of a snapshot file are:
```
ianhi, line 206: Should we define all of it?

icechunk-python/docs/docs/spec.md
```diff
@@ -292,16 +300,19 @@ The transaction log file maintains information about the id of modified objects:
 - `updated_groups`: list of node ids.
 - `updated_arrays`: list of node ids.
 - `updated_chunks`: list of node ids and chunk indices.
+- `moved_nodes`: list of `from`, `to` paths, sorted by first move first.
```
ianhi, line 303: we should be more specific here. is it only top level moves, all moves including children?

ianhi: should also add that moves cannot co-exist with any of the updates above.

icechunk-python/docs/docs/spec.md
```diff
@@ -335,15 +347,30 @@ Usually, a client will want to read from the latest branch (e.g. `main`).
...
-   b. If unsuccessful, the tag already exists.
+1. Do conditional update to write the new repo info file that contains the new tag
+    1. If successful, the update succeeded and the tag is created.
+    1. If unsuccessful because there was some previous update, read the file again and retry
+    1. If unsuccessful because the snapshot doesn't exist, fail the update
+
+## Changes from spec version 1
+
+- The repo info object was introduced as a single point of consistency and atomicity. This file is in the `repo` path, inside the repository tree. Many new fields are introduced in this object, such as feature flags, status, spec version and others.
+- There is now a single mutable object: `repo`. Updates to this file first backup the original contents in `overwritten` prefix.
```
ianhi, line 368: is this the same thing said twice?

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
 #### Snapshot Files
 
-The snapshot file fully describes the schema of the repository, including all arrays and groups.
+The snapshot file fully describes the schema of the repository, including all arrays and groups. Each commit to an Icechunk
+repository creates a new snapshot file. This snapshot iniforms what are the available groups and arrays
+in this commit, and provides a way to access the chunk manifests for each array.
 
 The snapshot file is encoded using [flatbuffers](https://github.com/google/flatbuffers). The IDL for the
-on-disk format can be found in [the repository file](https://github.com/earth-mover/icechunk/tree/main/icechunk/flatbuffers/snapshot.fbs)
+on-disk format can be found in [the fbs file](https://github.com/earth-mover/icechunk/blob/404100b584fb7ac70de860bd430aa8291df98c4d/icechunk-format/flatbuffers/snapshot.fbs).
```
ianhi, line 204: on reflection maybe these flatbuffer defs should be inlined? or be part of the spec subsectionn on the docs. they are the most concrete and specific spec and deserve to be treated as such.

icechunk-python/docs/docs/spec.md
```diff
@@ -259,6 +260,13 @@ of the array and a list of `ChunkRef` sorted by the chunk coordinate.
 
 These three types of chunks references are encoded in the same flatbuffers table, using optional fields.
 
+For the case of virtual chunks, under certain conditions the `locations` can be compressed using zstd compression.
+When locations are compressed, a compression dictionary is computed globally across the whole manifest and:
```
ianhi, line 264: I think this should be either more, or less, specific. "certain conditions" is unclear.

ianhi: where should i check to find if they are compressed? Also is this inline chunks?

icechunk-python/docs/docs/spec.md
```diff
@@ -335,15 +347,30 @@ Usually, a client will want to read from the latest branch (e.g. `main`).
...
+1. Do conditional update to write the new repo info file that contains the new tag
+    1. If successful, the update succeeded and the tag is created.
+    1. If unsuccessful because there was some previous update, read the file again and retry
+    1. If unsuccessful because the snapshot doesn't exist, fail the update
+
+## Changes from spec version 1
+
+- The repo info object was introduced as a single point of consistency and atomicity. This file is in the `repo` path, inside the repository tree. Many new fields are introduced in this object, such as feature flags, status, spec version and others.
+- There is now a single mutable object: `repo`. Updates to this file first backup the original contents in `overwritten` prefix.
+- There is no longer a `refs` prefix, branches and tags are stored in the repo info object.
```
ianhi, line 369: Can we be specific in what part of the flatbuffer these are now.

icechunk-python/docs/docs/spec.md
```diff
@@ -335,15 +347,30 @@ Usually, a client will want to read from the latest branch (e.g. `main`).
...
+- A transaction id is written for the first commit that initializes the repo.
+- There is no longer a `config.yaml` file, repo configuration has been moved to the repo info object.
+- Snapshots no longer store its parent in the snapshot file. This information has been moved to the repo info object.
+- A new `extra` field has been added in the flatbuffers of most datastructures. This allows for future expansion without a format change.
```
ianhi, line 373: How owuld this work for an external implementor?

icechunk-python/docs/docs/spec.md
```diff
@@ -335,15 +347,30 @@ Usually, a client will want to read from the latest branch (e.g. `main`).
...
+- For each dimension of each array, instead of storing the `chunk_legth` in the snapshot, the `num_chunks` is now stored.
```
ianhi, line 374:
```suggestion
- For each dimension of each array, instead of storing the `chunk_length` in the snapshot, the `num_chunks` is now stored.
```

I'm not sure i understand the significance of this?

icechunk-python/docs/docs/spec.md
```diff
@@ -335,15 +347,30 @@ Usually, a client will want to read from the latest branch (e.g. `main`).
...
+This will help with rectilinear chunk grids.
+- Virtual chunk location can optionally be compressed inside the manifest.
```
ianhi, line 376: This should be more explicitly called out in the spec. or more directlyl referenced to the flatbuffer.

icechunk-python/docs/docs/spec.md
```diff
@@ -47,15 +47,12 @@ Icechunk requires that the storage system support the following operations:
 - **Conditional update** - For the commit process to be safe and consistent, the storage system must be able to atomically update a file only if the current version is known to the writer.
 - **Seekable reads** - Chunk file formats may require seek support (e.g. shards).
```
TomNicholas, line 48: This doesn't seem to be true? We certainly are able to use sharding without SEEK

icechunk-python/docs/docs/spec.md
```diff
@@ -47,15 +47,12 @@ Icechunk requires that the storage system support the following operations:
...
 - **Deletes** - Delete files that are no longer used (via a garbage-collection operation).
-- **Sorted List** - The storage system must allow the listing of directories / prefixes in lexicographical order.
 
 These requirements are compatible with object stores, like S3, as well as with filesystems.
 
-The storage system is not required to support random-access writes. Once written, most files are immutable until they are deleted. The exceptions to this rule are:
+The storage system is not required to support random-access writes. Once written, most files are immutable until they are deleted. The exceptions to this rule is:
```
TomNicholas, line 53: nit
```suggestion
The storage system is not required to support random-access writes. Once written, most files are immutable until they are deleted. The exception to this rule is:
```

icechunk-python/docs/docs/spec.md
```diff
@@ -47,15 +47,12 @@ Icechunk requires that the storage system support the following operations:
...
 
-- the repository configuration file doesn't track history, updates are done atomically but in place,
-- branch reference files are also atomically updated in place,
-- snapshot files can be updated in place by the expiration process (and administrative operation).
+- the main entry point to a repository, an object with path `repo` under the repository prefix.
```
TomNicholas, line 55:
```suggestion
the repository info file, which is the main entry point to a repository, stored as an object with path `repo` under the repository prefix.
```

icechunk-python/docs/docs/spec.md
TomNicholas, line 123: I don't love the choice of name for this directory... Why not `history/` or `log/`?

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
-Similar to Git, Icechunk supports the concept of _branches_ and _tags_.
-These references point to a specific snapshot of the repository.
+The repo info file is encoded using [flatbuffers](https://github.com/google/flatbuffers). The IDL for the
+on-disk format can be found in [the fbs file](https://github.com/earth-mover/icechunk/blob/404100b584fb7ac70de860bd430aa8291df98c4d/icechunk-format/flatbuffers/repo.fbs).
 
-- **Branches** are _mutable_ references to a snapshot.
-  Repositories may have one or more branches.
-  The default branch name is `main`.
-  Repositories must always have a `main` branch, which is used to detect the existence of a valid repository in a given path.
-  After creation, branches may be updated to point to a different snapshot.
```
TomNicholas, line 142: This information seems to have been lost?

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
+##### Updates to the repo info file
 
-This mechanism is used by Icechunk on commits.
-When a client checks out a branch, it keeps track of the "version" of the reference file for the branch.
-When it tries to commit, it attempts to conditionally update this file in an atomic "all or nothing" operation.
-If this succeeds, the commit is successful.
-If this fails (because another client updated that file since the session started), the commit fails.
-At this point, the client may choose to retry its commit (possibly re-reading the updated data) and then try the conditional update again.
+The repo info object is the only mutable object in an Icechunk repo. Before an attempt is made to overwrite it, the file
+is copied to the `overwritten` prefix in the repo. The copy will be stored with a path that looks something like
```
TomNicholas, line 166: So as the attempt to overwrite the repo info object can fail, there are potentially things in `overwritten/` that were never overwritten?

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
-##### Tags
+- The literal `repo.`
+- A number that is the current Unix timestamp in milliseconds subtracted from the timestamp
+in milliseconds corresponding to the timestamp `3000-01-01T00:00:00`. This gives a "last one first" ordering when
+listing the prefix, but Icechunk itself doesn't depend on this property.
+- 12 random bytes encoded as Crockford base 32.
 
-Tags are immutable. Their files follow the pattern `refs/tag.$TAG_NAME/ref.json`.
+These backups of the repo info file are only used by Icechunk for its `ops_log` functionality. These files are linked in
+the repo info object itself, and they form a single linked list for ops logs that go beyond the 1,000 operations limit.
```
TomNicholas, line 181: Really need to define the ops log before this.

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
-Tags can also be deleted once created, but we cannot allow a delete followed by a creation, since that would
-result in an observable mutation of the tag. To solve this issue, we don't allow recreating tags that were deleted.
-When a tag is deleted, its reference file is not deleted, but a new tombstone file is created in the path:
-`refs/tags.$TAG_NAME/ref.json.deleted`.
+- **Branches** are _mutable_ references to a snapshot.
+  Repositories may have one or more branches.
+  The default branch name is `main`.
+  Repositories must always have a `main` branch, which is used to detect the existence of a valid repository in a given path.
+  After creation, branches may be updated to point to a different snapshot.
+- **Tags** are _immutable_ references to a snapshot.
```
TomNicholas, line 193: I don't think we have defined snapshots yet either.

icechunk-python/docs/docs/spec.md
```diff
@@ -121,88 +115,93 @@ flowchart TD
...
-When a client checks out a branch, it obtains a specific snapshot ID and uses this snapshot as the basis for any changes it creates during its session.
-The client creates a new snapshot and then updates the branch reference to point to the new snapshot (a "commit").
-However, when updating the branch reference, the client must detect whether a _different session_ has updated the branch reference in the interim, possibly retrying or failing the commit if so.
-This is an "optimistic concurrency" strategy; the resolution mechanism can be expensive, but conflicts are expected to be infrequent.
+- A commit is attempted to a branch that was deleted by other session
+- A branch is reset but it was deleted by other session
 
-All major object stores support a "conditional update" operation.
-In other words, object stores can guard against the race condition which occurs when two sessions attempt to update the same file at the same time. Only one of those will succeed.
+##### Updates to the repo info file
```
TomNicholas, line 163: I feel like it would be easier to define what happens during an update, and the whole optional `overwritten/` directory, after describing the layout of a static repo directory.

icechunk-python/docs/docs/spec.md
TomNicholas: "details" / "records"

icechunk-python/docs/docs/spec.md
TomNicholas: I strongly agree, that would be really nice pedagogically, and force the text to explain every detail of the real format definitions.

icechunk-python/docs/docs/spec.md
TomNicholas: Yes! Otherwise the spec isn't complete!

icechunk-python/docs/docs/spec.md
TomNicholas: Yes, at least the part before the `overwritten` bit.

icechunk-python/docs/docs/spec.md
TomNicholas: It avoids hard-coding fixed-length chunk  grids

TomNicholas (commented):
This is a start, but is quite incomplete. Do you want @ianhi and I to fill it out @paraseba ?

TomNicholas:
> At a high level a few thoughts.
>
> - The flatbuffers are where a lot of the specfiicity really lives. We should try to more tightly incorporate them into this document.

Strongly agree. That would be a much better way to structure this document, and help to ensure we completely describe everything.

> - Could we consider inverting the order of descriptions. i.e. instead of starting at the repo object and drilling down we start at chunks and build up.

Weakly disagree - I think it's helpful to start at the top and work down, because that is the order that files get read when you open the repo.

Also I feel like this spec doc is trying to explain how Icechunk works whilst also describing the format, and only really achieving the former.

paraseba:
Thanks for offering help @TomNicholas. I think it makes sense for you two to pick it up. But I'm not sure we have the time comparing with other priorities. Let's wait on @dcherian to make a decision on this on Fri or next week.

PR #1919, opened 2026-03-26
Add node ID and type to transaction log, clarify move ordering

li-em:
Extend the transaction log to include `node_id` and `node_type` in the move operations. Since `Move` is now defined without `required` fields, we need to check for missing data in functions that assumed it was always present before.

The `ChangeSet` tracks both top-level moves as well as all the children for the top-level moved nodes now, in order to be able to write the transaction log in a consistent order.

Moves in the transaction log are now sorted by their final path. Hypothesis tests validate this invariant.

### DIFF
--- a/icechunk/src/diff.rs
+++ b/icechunk/src/diff.rs
@@ -23,7 +23,7 @@ pub struct DiffBuilder {
 }
 
 impl DiffBuilder {
-    # TODO: Make this SessionResult instead?
+    // TODO: Make this SessionResult instead?
     pub fn add_changes(&mut self, tx: &TransactionLog) -> RepositoryResult<()> {
         self.new_groups.extend(tx.new_groups());
         self.new_arrays.extend(tx.new_arrays());


icechunk-format/flatbuffers/all.fbs
```diff
@@ -1,6 +1,6 @@
 // run this command in the directory icechunk/flatbuffers
 //
-// flatc --rust -o ../src/format/flatbuffers/ --gen-all all.fbs
+// flatc --rust -o ../src/flatbuffers/ --gen-all all.fbs
```
li-em, line 3: This was not updated during the crate split

icechunk-format/src/lib.rs
```diff
@@ -114,6 +114,40 @@ pub type AttributesId = ObjectId<12, AttributesTag>;
...
+    Array,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct Move {
+    pub from: Path,
+    pub to: Path,
+    pub node_id: NodeId,
+    pub node_type: NodeType,
+}
```
li-em, line 129: Moved `Move` here because `NodeId` and `NodeType` are defined here, and not in `icechunk-types`

icechunk-format/src/snapshot.rs
```diff
@@ -140,12 +141,6 @@ pub enum NodeData {
     Group,
 }
 
-#[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub enum NodeType {
-    Group,
-    Array,
-}
-
```
li-em, line 148: Moved to top level, since it is used in the `Move` struct now

icechunk-format/src/transaction_log.rs
```diff
@@ -182,10 +188,22 @@ impl TransactionLog {
             Some(it) => Either::Left(it.iter()),
             None => Either::Right(iter::empty()),
         };
+        #[expect(clippy::expect_used)]
         it.filter_map(|m| {
-            let from = Path::new(m.from()).ok()?;
-            let to = Path::new(m.to()).ok()?;
-            Some(Move { from, to })
+            let from = Path::new(
+                m.from().expect("from is optional in flatbuffers, but required by spec"),
```
li-em: Should probably remove these `expect`, it is annoying because these fields are optional in the flatbuffer now, when they were required the `expect` was not needed

icechunk-format/flatbuffers/transaction_log.fbs
```diff
@@ -15,9 +15,15 @@ table ArrayUpdatedChunks {
   chunks: [ChunkIndices] (required);
 }
 
+enum NodeType: ubyte { Group = 0, Array }
+
+// moves are fully collapsed
```
paraseba: give an example of collapsing, I know what it means now, but probably not in 6 months.

icechunk-format/flatbuffers/transaction_log.fbs
paraseba: Is this sorting still true?

icechunk-format/src/flatbuffers/all_generated.rs
```diff
@@ -1,22 +1,10 @@
 // automatically generated by the FlatBuffers compiler, do not modify
-
 // @generated
-
-use core::cmp::Ordering;
-use core::mem;
-
-extern crate flatbuffers;
-use self::flatbuffers::{EndianScalar, Follow};
+extern crate alloc;
```
paraseba, line 3: are you using the flatc version from the flake.nix @li-em ? Wonder why so many small differences

icechunk-format/src/lib.rs
```diff
@@ -114,6 +114,40 @@ pub type AttributesId = ObjectId<12, AttributesTag>;
...
+    }
+}
+
+impl From<generated::NodeType> for NodeType {
+    fn from(value: generated::NodeType) -> Self {
+        match value {
+            generated::NodeType::Group => NodeType::Group,
+            generated::NodeType::Array => NodeType::Array,
+            #[expect(clippy::panic)]
+            generated::NodeType(v) => panic!("{v} is not a valid node type"),
```
paraseba: Change this to a TryFrom, and return an Icechunk Format Error. One day we'll read icechunk files written by other implementations and we don't want to assume they are correct. For example this could DoS anything using IC on a server.

icechunk-format/src/transaction_log.rs
paraseba: Let's convert to icechunk format errors if possible. See my previous comment.

icechunk/src/change_set.rs
```diff
@@ -920,13 +925,20 @@ pub fn transaction_log_from_change_set(
         updated_groups.into_iter(),
         updated_arrays.into_iter(),
         changed_chunks.into_iter(),
-        cs.moves().cloned(),
+        cs.move_tracker().nodes_by_original.iter().map(
+            |(from, (to, node_id, node_type))| Move {
```
paraseba: Isn't this a bug? Shouldn't we be sorting by order of operation?

icechunk/src/change_set.rs
```diff
@@ -267,7 +270,7 @@ impl MoveTracker {
         // Collect updates first since we can't mutate BTreeMap keys
         // while iterating.
         let mut updates: Vec<(Path, Path, Path)> = Vec::new(); // (original, old_final, new_final)
-        for (orig, current) in self.nodes_by_original.iter_mut() {
+        for (orig, (current, _node_id, _node_type)) in self.nodes_by_original.iter_mut() {
             if let Some(remapped) = Self::remap_path(current, &from, &to) {
                 let old_final = std::mem::replace(current, remapped);
```
ianhi: Pretty sure we need to update this as well in order to preserve the ordering properly.

If i start with these entries
`a->b, a/1 -> b/1, a/2->b/2`

and then  move `b/1 -> d`

then i would want the map to be ordered like this:
`a->b, a/2->b/2,  a/1 -> d,`

So this needs become a shift_insert i think: https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.shift_insert

icechunk/src/change_set.rs
ianhi: If i'm understanding correctly - The idea with indexmap is that it gets sorted as we build it up. so we should not need to do any sorting here.

ianhi:
an alternative - very explcit strategy that avoids all the ordering confusion would be store by node id and for each node save it's full sequence of moves. then everything is completely replayable.

```
Move = {
   from: str,
   to: str
}


NodeMoves{
    nodeid
    moves: list[Move]
}
```
but i think the ordered version we have here is also fully adequate.

icechunk/src/change_set.rs
ianhi: not shift_insnert. i misread the docs. even just a remove then then add would be fine. i think this is the right one: https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.shift_remove then just insert

ianhi: ok https://github.com/earth-mover/icechunk/pull/1919/commits/6eb13f7d79272118fed7b0e309f2ff7ba2bf2372 should fix this

icechunk/src/change_set.rs
```diff
@@ -1470,19 +1537,82 @@ mod tests {
...
+        // Move 3 re-moved /a/x and its child /a/x/z — both should come last,
+        // with parent /a/x before child /a/x/z.
+        let ordered: Vec<_> = mt.nodes_by_original.keys().collect();
+        assert_eq!(
+            ordered,
+            vec![
+                &pathify("/a"),     // carried by move 2
+                &pathify("/b"),     // added by move 2
+                &pathify("/b/y"),   // added by move 2
+                &pathify("/a/x"),   // re-moved by move 3 → last
```
ianhi: At first i was confident. but now I am genuinely unsure what the correct order for the first two elements should be.

Each of the options have a brief ambiguity if you strictlyl follow their move sequences. 


After move 1 (/a → /b/a):
```
/a → /b/a
/a/x → /b/a/x
/a/x/z → /b/a/x/z
```
After move 2 (/b → /c):

Should we have this (current impl):

```
  /a → /c/a
  /b → /c
  /b/y → /c/y
  /a/x → /c/a/x
  /a/x/z → /c/a/x/z
```

or this:
```
  /b → /c
  /a → /c/a
  /b/y → /c/y
  /a/x → /c/a/x
  /a/x/z → /c/a/x/z
```

ianhi: This is solvable with another layer of encapsulation. like a move group. but thats much more annoying to keep track of i guess

icechunk/src/change_set.rs
```diff
@@ -264,31 +267,41 @@ impl MoveTracker {
         // e.g. earlier move deposited /x at /a/x, now /a -> /c:
         //   /x's current path /a/x starts_with /a -> becomes /c/x
         //
-        // Collect updates first since we can't mutate BTreeMap keys
-        // while iterating.
-        let mut updates: Vec<(Path, Path, Path)> = Vec::new(); // (original, old_final, new_final)
-        for (orig, current) in self.nodes_by_original.iter_mut() {
+        // Collect first since we can't mutate the IndexMap while iterating,
+        // then shift_remove + re-insert so re-moved nodes appear at the end
```
dcherian: Please continue the `/x` example from above to illustrate what you mean here.

icechunk/src/change_set.rs
```diff
@@ -264,31 +267,41 @@ impl MoveTracker {
...
+                    orig.clone(),
+                    current.clone(),
+                    remapped,
+                    node_id.clone(),
+                    node_type.clone(),
+                ));
             }
         }
-        for (orig, old_final, new_final) in updates {
+        for (orig, old_final, new_final, node_id, node_type) in to_reinsert {
```
dcherian: Isn't `old_final` the `current` from the previous loop? Would be easier to read if we kept the naming consistent.

icechunk/src/change_set.rs
ianhi: ok no the right thing is hte second example i think. within each group we sort by the final path. that way if you replay them exactly in order there is never a chance that you make a move to a group that does not exist yet, but will later

icechunk/src/change_set.rs
```diff
@@ -1470,19 +1537,82 @@ mod tests {
...
-        // Move 2: /b -> /c (rename /b)
-        // After: /c, /c/a, /c/a/x, /c/y
+        // Move 2: /b -> /c (rename /b — carries along /a, /a/x, /a/x/z)
+        // Move 3: /c/a/x -> /d (individual re-move of subtree child + its child)
+        // After: /c, /c/a, /c/y, /d, /d/z
+        let id_a = NodeId::random();
+        let id_ax = NodeId::random();
+        let id_axz = NodeId::random();
+        let id_b = NodeId::random();
+        let id_by = NodeId::random();
```
dcherian:
```suggestion
        let grp_a = NodeId::random();
        let grp_ax = NodeId::random();
        let arr_axz = NodeId::random();
        let grp_b = NodeId::random();
        let arr_by = NodeId::random();
```

How about this?

icechunk/src/change_set.rs
```diff
@@ -1470,19 +1537,82 @@ mod tests {
...
+            "/d",
+            &[
+                ("/a/x", id_ax.clone(), NodeType::Group),
+                ("/a/x/z", id_axz.clone(), NodeType::Array),
+            ],
+            &id_ax,
+            NodeType::Group,
+        );
 
+        // Original moved_into assertions still hold
```
dcherian: Can we have some nice ASCII art showing the start & end states please?

icechunk/src/change_set.rs
```diff
@@ -1470,19 +1537,82 @@ mod tests {
...
         let result = mt.moved_into(&pathify("/c"));
         assert!(result.contains(&(pathify("/a"), pathify("/c/a"))));
-        assert!(result.contains(&(pathify("/a/x"), pathify("/c/a/x"))));
         assert!(result.contains(&(pathify("/b"), pathify("/c"))));
         assert!(result.contains(&(pathify("/b/y"), pathify("/c/y"))));
+        // /a/x and /a/x/z are no longer under /c — re-moved to /d
+        assert!(!result.iter().any(|(_, f)| *f == pathify("/c/a/x")));
```
dcherian: `starts_with('/c/a')` would be easier to read: https://doc.rust-lang.org/std/primitive.str.html#method.starts_with (if it can be done with Path). Feel free to ignore if you prefer the explicit listing of paths

icechunk/src/strategies.rs
```diff
@@ -651,6 +652,10 @@ fn chunk_payload() -> impl Strategy<Value = ChunkPayload> {
     ]
 }
 
+fn node_type() -> impl Strategy<Value = NodeType> {
```
dcherian:
```suggestion
fn node_types() -> impl Strategy<Value = NodeType> {
```

extremely minor nit: I've been following hypothesis convention of using plurals in this file

ianhi:
added some tests of the final path ordering in both rust and python and they pass. fortunaely path::cmp does the right thing for us

icechunk/src/change_set.rs
```diff
@@ -1470,19 +1536,154 @@ mod tests {
...
+            ],
+            &grp_a,
+            NodeType::Group,
+        );
+        record_move(
+            &mut mt,
+            "/b",
+            "/c",
+            &[
+                ("/b", grp_b.clone(), NodeType::Group),
```
dcherian, line 1581: Why does this not list
```
                ("/b/a", grp_a.clone(), NodeType::Group),
                ("/b/a/x", grp_ax.clone(), NodeType::Group),
                ("/b/a/x/z", arr_axz.clone(), NodeType::Array),
```

Do we provide the subtree as in the snapshot?

dcherian:
I fixed the final state in [`eb78dab` (this PR)](https://github.com/earth-mover/icechunk/pull/1919/commits/eb78dab80852a9adfe5c225bfe75cd44409f95e4)

icechunk/src/change_set.rs
ianhi: ahhh this is a sort weird edge case that we got lucky with. becuase those nodes are all already in the map you don't actually need to include them as subtree nodes. but we should explicitly include them

ianhi: agghh this has expxosed one annoying thing. we search over original nodes in the record move when checking the subtree.

we need an extra branch of logic in the step 2 where we iterate the subtree. and check if we are getting a dirty node via a subtree. or maybe not, becuase it is always handled by the first step. 


I think there is a bug here

icechunk-format/src/flatbuffers/all_generated.rs
li-em: yes, but just to be sure, it is `/nix/store/ynsp1axfm99ifyjf0980zffawabrbvj5-flatbuffers-25.12.19/bin/flatc`

icechunk/src/change_set.rs
ianhi: fix incoming

ianhi:
ok with https://github.com/earth-mover/icechunk/pull/1919/changes/bcc2ad8d77c42df6eab611b714925d0dba856af9 i think we may finally be bug free!

icechunk/src/strategies.rs
ianhi: there are many more to change. should I go throughh and change them all?

icechunk-format/src/lib.rs
li-em: done!

icechunk/src/repository.rs
```diff
@@ -161,6 +161,18 @@ pub type RepositoryError = ICError<RepositoryErrorKind>;
...
+        match e {
+            SessionErrorKind::RepositoryError(e) => e,
+            SessionErrorKind::StorageError(e) => e.into(),
+            SessionErrorKind::FormatError(e) => e.into(),
+            SessionErrorKind::RefError(e) => e.into(),
+            other => RepositoryErrorKind::Other(other.to_string()),
+        }
+    }
+}
+
```
li-em: This feels a bit dangerous, in case the `other` case catches too much, but it is very convenient otherwise...

icechunk/src/strategies.rs
li-em: next PR =]

icechunk-format/flatbuffers/transaction_log.fbs
li-em: nope, updated

icechunk/src/repository.rs
dcherian: think this should be `RepositoryErrorKind::SessionError`?

icechunk/src/repository.rs
```diff
@@ -161,6 +161,18 @@ pub type RepositoryError = ICError<RepositoryErrorKind>;
...
 pub type RepositoryResult<T> = Result<T, RepositoryError>;
 
+impl From<SessionErrorKind> for RepositoryErrorKind {
+    fn from(e: SessionErrorKind) -> Self {
+        match e {
+            SessionErrorKind::RepositoryError(e) => e,
+            SessionErrorKind::StorageError(e) => e.into(),
+            SessionErrorKind::FormatError(e) => e.into(),
+            SessionErrorKind::RefError(e) => e.into(),
+            other => RepositoryErrorKind::Other(other.to_string()),
```
dcherian:
```suggestion
            other => RepositoryErrorKind::SessionError(other),
```

icechunk/src/diff.rs
```diff
@@ -22,14 +22,17 @@ pub struct DiffBuilder {
 }
 
 impl DiffBuilder {
-    pub fn add_changes(&mut self, tx: &TransactionLog) {
+    pub fn add_changes(&mut self, tx: &TransactionLog) -> SessionResult<()> {
```
dcherian: Should we use `RepositoryResult` in this file instead of `SessionResult`. Since it's `Repository.diff`...

icechunk/src/diff.rs
```diff
@@ -22,14 +23,17 @@ pub struct DiffBuilder {
 }
 
 impl DiffBuilder {
-    pub fn add_changes(&mut self, tx: &TransactionLog) {
+    pub fn add_changes(&mut self, tx: &TransactionLog) -> RepositoryResult<()> {
```
dcherian, line 27: this could be `SessionResult` but this way we avoid some conversion to RepositoryResult in `.diff()`

dcherian, line 27:
```suggestion
    # TODO: Make this SessionResult instead?
    pub fn add_changes(&mut self, tx: &TransactionLog) -> RepositoryResult<()> {
```

icechunk/src/diff.rs
```diff
@@ -22,14 +23,18 @@ pub struct DiffBuilder {
 }
 
 impl DiffBuilder {
-    pub fn add_changes(&mut self, tx: &TransactionLog) {
+    # TODO: Make this SessionResult instead?
```
dcherian:
```suggestion
    // TODO: Make this SessionResult instead?
```

PR #1920, opened 2026-03-26
Hypothesis Test Improve + new inspection functions

### LINKED ISSUE #1682: `inspect_snapshot` should return a dict instead of JSON in a string

ianhi:
Closes #1682 

Ok 4 commits here. 3 indepdent improvements that build up to cocmmit 4 that needs all of them.

1. Consolidating test infra a bit to make invariants and models more re-usable
2. add `inspect_repo_info` to allow inspecting transaction logs - this was the missing piece of inspect
3. add `get_node_id` to python. again was sort of a missing piece - and very helpful in python
4. property based move tests with amend - compares two move sequences one as one commit and one with amend(s) and makes sure the final result is the same (currently xfail)

### DIFF
--- a/icechunk-python/tests/test_move.py
+++ b/icechunk-python/tests/test_move.py
@@ -139,6 +139,18 @@ def test_moves(
         sync(model.move(source, dest))
         session.move(f"/{source}", f"/{dest}")
 
+    # All model paths should be reachable as group members in the store
+    model_paths = {path for path, _ in zarr.open_group(model).members(max_depth=None)}
+    store_paths = {
+        path
+        for path, _ in zarr.open_group(session.store, mode="r").members(max_depth=None)
+    }
+    assert model_paths == store_paths, (
+        f"members mismatch after moves:\n"
+        f"  model only: {model_paths - store_paths}\n"
+        f"  store only: {store_paths - model_paths}"
+    )
+
     expected = repr(zarr.open_group(model).tree())
     for label, store in precommit_postcommit_readonly(session, repo):
         actual = repr(zarr.open_group(store, mode="r").tree())


icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -1047,6 +1013,7 @@ def checks(self) -> None:
         self.check_ops_log()
         self.check_repo_info()
         self.check_file_invariants()
+        assert_all_snapshots_have_valid_parent(self.repo)
```
dcherian: this could be under `_assert_ancestry_invariants` or alternativel,y maybe those invariants should be moved out and shared?

ianhi: ahh too many moving parts and I missed that. yes they should be moved out and shared

icechunk-python/python/icechunk/testing/strategies.py
```diff
@@ -121,3 +131,41 @@ def chunk_slicers(
 def chunk_paths(draw: st.DrawFn, numblocks: tuple[int, ...]) -> str:
     blockidx = draw(chunk_coordinates(numblocks))
     return "/".join(map(str, blockidx))
+
+
+def draw_older_than(data: st.DataObject, storage: ic.Storage) -> datetime.datetime:
```
dcherian, line 136:
```suggestion
@composite
def needs_better_name(draw: st.DrawFn, storage: ic.Storage) -> datetime.datetime:
```

?

icechunk-python/python/icechunk/repository.py
```diff
@@ -1980,31 +1980,214 @@ async def total_chunks_storage_async(
         )
         return stats.native_bytes
 
-    def inspect_snapshot(self, snapshot_id: str, *, pretty: bool = True) -> str:
-        return self._repository.inspect_snapshot(snapshot_id, pretty=pretty)
+    def inspect_snapshot(self, snapshot_id: str) -> dict[str, Any]:
```
dcherian, line 1983: Nice this closes https://github.com/earth-mover/icechunk/issues/1682

icechunk-python/python/icechunk/testing/strategies.py
ianhi: I think i had a reason to not do this. but i cannot remember at all why :/ 

will try to figure out

icechunk-python/python/icechunk/testing/trees.py
```diff
@@ -74,62 +74,76 @@ def _write(group: zarr.Group, node: GroupNode) -> None:
...
+        """
+        import json
+
+        from zarr.core.sync import sync
+
+        async def _collect() -> tuple[set[str], set[str]]:
+            arrays: set[str] = set()
+            groups: set[str] = set()
+            proto = zarr.core.buffer.default_buffer_prototype()
+            async for key in store.list_prefix(""):
```
dcherian: We should use `zarr.Group.members` or `groups` and `arrays` instead of parsing keys

icechunk-python/python/icechunk/testing/trees.py
```diff
@@ -74,62 +74,76 @@ def _write(group: zarr.Group, node: GroupNode) -> None:
...
+    def from_paths(cls, arrays: set[str], groups: set[str]) -> GroupNode:
+        """Build a GroupNode from flat sets of array and group paths.
 
-def tree_from_str(spec: str) -> GroupNode:
-    """Build a GroupNode tree from a compact string specification.
+        Example::
 
-    Each line is ``"a: /path"`` for an array or ``"g: /path"`` for a group.
-    Intermediate groups are created implicitly.
+            GroupNode.from_paths(
```
dcherian, line 94: A doctest example here would be v. useful for the next reader

icechunk-python/tests/test_move.py
dcherian, line 154: we can assert that the paths in `moves` are present in `list_prefix`? and in `Group.members()`?

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+)
+@given(
+    tree_moves=tree_and_moves(n_moves=st.integers(min_value=2, max_value=10)),
+    data=st.data(),
+)
+def test_moves_amend(
+    tree_moves: tuple[GroupNode, list[tuple[str, str]]],
+    data: st.DataObject,
+) -> None:
+    """Moves split across commit + amend should produce the same tree
```
dcherian, line 178: This is perfect for now. I'm sure you've realized this is a great general invariant (for post-IC2 testing work). Generate two sequential changes; apply them sequentially on two repos: 2 commits on one; 1 commit + 1 amend on another. Open an issue?

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+    Splits moves into multiple batches with amends between them.
+    Verifies that the transaction log properly collapses chained moves
+    during amend (e.g., a->b + b->c becomes a->c).
+
+    See https://github.com/earth-mover/icechunk/pull/1896
+    """
+    tree, moves = tree_moves
+
+    # Draw 1-3 split points to create multiple amend batches.
+    n_splits = data.draw(st.integers(min_value=1, max_value=3))
```
dcherian: How does `max_value=3` work with `num_moves=2`?

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+    )
+
+    note(f"tree: {tree!r}")
+    note(f"splits at {splits}/{len(moves)}")
+    for i, (s, d) in enumerate(moves):
+        note(f"  move {i}: /{s} -> /{d}")
+
+    # Repo A: all moves in one commit
+    _, session_a, repo_a = tree_to_model_and_icechunk(tree)
+    session_a.commit("init")
```
dcherian, line 206:
```suggestion
    _, _, repo_a = tree_to_model_and_icechunk(tree)
```

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+    session_a = repo_a.rearrange_session("main")
+    for source, dest in moves:
+        session_a.move(f"/{source}", f"/{dest}")
+    session_a.commit("all moves")
+
+    # Repo B: same moves split across initial commit + multiple amends
+    _, session_b, repo_b = tree_to_model_and_icechunk(tree)
+    session_b.commit("init")
```
dcherian, line 214:
```suggestion
    _, _, repo_b = tree_to_model_and_icechunk(tree)
```

dcherian: Seem slike you just need `tree_to_icechunk`

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+    session_b = repo_b.rearrange_session("main")
+    committed = False
+    for i, (source, dest) in enumerate(moves):
+        session_b.move(f"/{source}", f"/{dest}")
+        if i + 1 in splits and not committed:
```
dcherian: very hard to read.

let's do something like group in to n_split lists; commit the first; and then loop over the rest; amending the commit. Some combo of `itertools` `chain`, `repeat`, and `groupby` should do this.

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+        elif i + 1 in splits and committed:
+            session_b.amend(f"amend at {i + 1}")
+            session_b = repo_b.rearrange_session("main")
+    if not committed:
+        session_b.commit("only batch")
+    else:
+        session_b.amend("final amend")
+
+    # Trees must match
+    tree_a = repr(zarr.open_group(repo_a.readonly_session("main").store, mode="r").tree())
```
dcherian, line 223: Does the return type not define `__eq__`? We should probably check node types too (could punt this to later)

icechunk-python/tests/test_move.py
ianhi: I think one of them gets silently ignored. becuasea the check is `if i+1 in splits`

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +150,97 @@ def test_moves(
...
+        f"\nsplits at: {splits}"
+        f"\ntree (one commit):\n{tree_a}"
+        f"\ntree (amended):\n{tree_b}"
+    )
+
+    # Transaction logs must match — amend should collapse chained moves
+    snap_a = repo_a.lookup_branch("main")
+    snap_b = repo_b.lookup_branch("main")
+    tx_a = repo_a.inspect_transaction_log(snap_a)
+    tx_b = repo_b.inspect_transaction_log(snap_b)
```
dcherian, line 236: nice!

icechunk-python/python/icechunk/testing/trees.py
ianhi: im not sure i see the difference from this. what do you mean?

dcherian: If GroupNode has a nice repr, it would be good to show that too; if not please ignore

icechunk-python/tests/test_move.py
ianhi: no :( 

https://github.com/zarr-developers/zarr-python/blob/2dc4036a3086b67804df1bcd53229c9b85c70633/src/zarr/core/_tree.py#L10-L43

the node types are captured by the trees being equal becuase the tree prints out shapes foro arrays but not groups

dcherian: OK good for now! Raise an upstream issue?

ianhi: https://github.com/zarr-developers/zarr-python/issues/3849?reload=1

icechunk-python/python/icechunk/testing/strategies.py
ianhi: ah right. you can decorate with this because a strategy is created once at the start. i.e. we can use the strategy as an `@rule` arugment because then it wont have access to the storage which dynamically udpates. If we turn it into a composite then we could reconstruct the strategy each ttime we need to draw it, but I don't see the advantage over just having a funciton.

icechunk-python/tests/test_move.py
ianhi: should be fixed with the new drawing strategy using st.sets

icechunk-python/python/icechunk/testing/trees.py
ianhi: does not have. should have though honestly

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +164,86 @@ def test_moves(
...
+
+    See https://github.com/earth-mover/icechunk/pull/1896
+    """
+    tree, moves = tree_moves
+
+    # Split moves into 2-4 batches: commit the first, amend the rest.
+    split_points = sorted(
+        {
+            0,
+            *data.draw(st.sets(st.integers(1, len(moves) - 1), min_size=1, max_size=3)),
```
dcherian, line 193: `st.lists(... unique=True)` would also work but this is fine

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +164,86 @@ def test_moves(
...
+    session_a = repo_a.rearrange_session("main")
+    for source, dest in moves:
+        session_a.move(f"/{source}", f"/{dest}")
+    session_a.commit("all moves")
+
+    # Repo B: same moves split across multiple amends
+    _, session_b, repo_b = tree_to_model_and_icechunk(tree)
+    session_b.commit("init")
+
+    for batch in batches:
```
dcherian, line 216: MUCH nicer!

icechunk-python/tests/test_move.py
ianhi: thats cleaner will do

icechunk-python/tests/test_move.py
```diff
@@ -126,6 +164,86 @@ def test_moves(
...
+
+    Splits moves into multiple batches with amends between them.
+    Verifies that the transaction log properly collapses chained moves
+    during amend (e.g., a->b + b->c becomes a->c).
+
+    See https://github.com/earth-mover/icechunk/pull/1896
+    """
+    tree, moves = tree_moves
+
+    # Split moves into 2-4 batches: commit the first, amend the rest.
```
ianhi, line 189:
```suggestion
```

found a better way this is out of date.

PR #1937, opened 2026-03-31
Fix sdist build: symlink README from root

### LINKED ISSUE #1933: Assess files included in sdist
In virtualizarr a CI run tried to build the python package from source and failed https://github.com/zarr-developers/VirtualiZarr/actions/runs/23774477632/job/69273150693

with

```
          Caused by: Failed to parse Cargo.toml into python metadata
          Caused by: Failed to read Readme specified in Cargo.toml, which should be at /home/runner/.cache/rattler/cache/uv-cache/sdists-v9/pypi/icechunk/1.1.21/DUuUdlZ19LWlRVwNHBMKV/src/icechunk-python/README.md
```

this is some setting  at the intersectinoo of cargo.toml, pyproject.toml and maybe a MANIFEST file

dcherian:
## Summary
- Fixes sdist builds failing when installed from PyPI because `readme = "../README.md"` in Cargo.toml/pyproject.toml pointed outside the package directory
- Symlinks `icechunk-python/README.md` → `../README.md` so the root README (with more detail) is included in the sdist
- Changes `readme` paths in both Cargo.toml and pyproject.toml to `"README.md"` (local)
- Moves the old Python-specific README to `CONTRIBUTING.md`

Closes #1933

## Test plan
- [x] Verify `maturin sdist` builds successfully
- [x] Verify the sdist contains the root README content
- [ ] Verify downstream sdist installs (e.g. VirtualiZarr CI) no longer fail


icechunk-python/CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
+# Icechunk Python
```
ianhi, line 1: whats going on with this file? don't we already have contributing docs?

icechunk-python/CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+You can then commit your changes to save progress or share with others:
+
+```python
+snapshot_id = session.commit("Create foo array")
+
+async for parent in repo.ancestry(snapshot_id):
+    print(parent.message)
+```
+
+See [`tests/test_timetravel.py`](tests/test_timetravel.py) for more example usage of the transactional features.
```
ianhi, line 86: is this were we should be pointing?

icechunk-python/CONTRIBUTING.md
dcherian: It's the pre-existing README: https://github.com/earth-mover/icechunk/tree/main/icechunk-python which isn't good for a pypi ad

ianhi:
can we also have the logo be a link to th efile on github instead of a local link? 

then it will work here: https://pypi.org/project/icechunk/

.github/workflows/python-ci.yaml
```diff
@@ -174,7 +174,7 @@ jobs:
         with:
           working-directory: icechunk-python
           target: ${{ matrix.platform.target }}
-          args: --release --out dist -i ${{ env.PYTHON_INTERPRETERS }}
+          args: --release --sdist --out dist -i ${{ env.PYTHON_INTERPRETERS }}
```
li-em, line 177: Make sure we are building from the `sdist` instead of the checked-out repo

TomNicholas:
Tried building the sdist from this branch and hit a new error:

```
warning: readme `../README.md` appears to be a path outside of the package, but there is already a file named `README.md` in the root of the package.
💥 maturin failed
  Caused by: Failed to add path dependency icechunk
  Caused by: File icechunk-2.0.0/icechunk/README.md was already added from /private/tmp/icechunk-sdist-test/icechunk/README.md, can't added it from /private/tmp/icechunk-sdist-test/README.md
```

The workspace `Cargo.toml` sets `readme = "README.md"` (the root README), and `icechunk/Cargo.toml` inherits it via `readme.workspace = true` — which resolves to `../README.md` from the crate's perspective. But `icechunk/` already has its own `README.md`, so maturin gets two files trying to land at the same path in the archive.

TomNicholas:
The same symlink fix probably needs to be applied to `icechunk/` too

PR #1939, opened 2026-03-31
Test format readability with Python libraries

### LINKED ISSUE #1911: Test the universality of the on-disk encoding
What I mean by that is we need a basic test that checks that we or our dependencies are not doing something silly, in a way that would make open IC files with other libraries.

This is what I have in mind:

* write a python test
* it creates a repo with a couple of snaps and a hundred virtual chunks
* read the spec V2 to understand the file format
* using python only, and without IC, open the repo info file. This involves reading a header, verifying some magic numbers and other values, decompressing using zstd, etc
* from the repo info read the metadata using some independent implementation of flatbuffers and flexbuffers (not the ones in rust)
* open the manifest, check header, decompress, read some values, verify we can decompress (zstd) all the virtual chunk locations.

dcherian:
Closes #1911

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -197,6 +197,10 @@ jobs:
               sleep 1
           done
 
+      - name: Install flatc
+        run: |
+          sudo apt-get update && sudo apt-get install -y flatbuffers-compiler
+
       - name: Run pytest (non-hypothesis)
         shell: bash
         working-directory: icechunk-python


paraseba:
Love it, just one question, does it fail if the manifest doesn't have compressed locations? I want it to fail. There are conditions where compression of locations is disabled: few refs, too short, etc. We want to make sure we are testing this location compression.

dcherian:
Yup verified all virtual chunks have compressed locations

Issue #1931, opened 2026-03-30
Fix panic in amend


dcherian:
https://github.com/earth-mover/icechunk/blob/a20e0938dea4ccd50aae6558479bcf047fd87b3b/icechunk/src/session.rs#L3156

is apparently reachable through the magic of hypothesis. After GC we set the parent to None for some commits and then things break. 

We should figure out why we are getting to this branch, and whether some earlier step should have thrown an error

ianhi:
I have a fix here. And test. Sorry was only tracking the failing upstream issue

dcherian:
Yup I remember. Just wanted a "real" issue to track it Let's do it after the move work is done

PR #1932, opened 2026-03-30
Make V2 branch expiration behaviour match V1

dcherian:
Closes #1520 
Closes #1421 

Also add a boatload of V1 quirks to the model

PR #1940, opened 2026-03-31
FIX: orphaned commits + TEST: GC stateful 

ianhi:
I added two versions of stateful tests that catch https://github.com/earth-mover/icechunk/issues/1931

1. A subset of our existing test
pro: no new tests
con: had to add a new commit method and bunch of complicated hypothesis subset machinery

2. a. smaller GC focused statemachine

pro: catches bug very fast
con: adds another stateful test 

we should chose which one we want and then I'll remove the other.

```
Steps:
  Initial -> A -> B          create two commits
  reset main -> Initial      orphan A and B
  expire (A old, B recent)   A released; B retained with parent_id=None
  reset main -> B            point branch at the corrupted snapshot
  amend                      panic
```


dcherian:
Lets do (2)

icechunk/src/ops/gc.rs
```diff
@@ -954,7 +954,9 @@ async fn expire_v2_one_attempt(
                 return Some(new_parent.clone());
             }
         }
-        None
+        // Snapshot not reachable from any ref (e.g. branch was reset away).
+        // Re-parent to the initial snapshot rather than orphaning with None.
+        Some(Snapshot::INITIAL_SNAPSHOT_ID)
```
dcherian, line 959: In #1932 i change the expectation so we assert that ancestry ends at a node with `parent_id == None`.

PR #1944, opened 2026-04-01
PyRepr trait for structured, recursive Python reprs

### LINKED ISSUE #1208: Nicer python reprs (string and HTML)

TomNicholas:
## tl;dr:

Supercedes #1217, following up on #1490 as a way to fully close #1208.

This PR introduces a common trait and utility functions for writing reprs, and uses them on a few key classes. The idea is that once this approach has a +1, I will use it to write reprs for every python class in icechunk-python.

## Summary

- Introduces a `PyRepr` trait in `icechunk/src/display.rs` with `EXECUTABLE`, `cls_name()`, and `fields(executable)` — implementors get `__str__`, `__repr__`, and `_repr_html_` for free
- `__repr__` on executable classes produces multi-line constructor syntax that `eval()` can round-trip
- Non-executable classes (Session, Store, Repository) use `<icechunk.ClassName>` format for both `__repr__` and `__str__`
- Nested reprs recurse correctly: executable parents nest executable child reprs, `__str__`/`_repr_html_` always nest non-executable
- Converts 6 classes to PyRepr: Repository, Session, Store, ManifestFileInfo, CachingConfig, RepositoryConfig
- Moves `format_option_to_string` to `display.rs` as `py_option`
- Adds `py_option_nested_repr` in `icechunk-python/src/display.rs` for `Py<T>` fields

## Test plan

- [x] 16 tests in `test_display.py` covering repr/str/html for all converted classes
- [x] `eval(repr(obj))` round-trip tests for executable classes (CachingConfig)
- [x] Tests for readonly session omitting branch/has_uncommitted_changes
- [x] Tests for nested recursion: executable repr nests executable, str nests non-executable
- [x] cargo fmt, clippy clean

## Follow-up

A second PR will convert the remaining ~24 ad-hoc `__repr__` implementations to use PyRepr, and fix 5 missing-field bugs found during audit. See plan file for details.

Closes #1208

### DIFF
--- a/icechunk-python/src/display.rs
+++ b/icechunk-python/src/display.rs
@@ -58,6 +58,15 @@ pub(crate) trait PyRepr {
     /// so the recursion pattern is consistent.
     fn fields(&self, mode: ReprMode) -> Vec<(&str, String)>;
 
+    /// Dispatch to the appropriate repr method for the given mode.
+    fn render(&self, mode: ReprMode) -> String {
+        match mode {
+            ReprMode::Str => self.__str__(),
+            ReprMode::Repr => self.__repr__(),
+            ReprMode::Html => self._repr_html_(),
+        }
+    }
+
     fn __str__(&self) -> String {
         let fields = self.fields(ReprMode::Str);
         let refs: Vec<(&str, &str)> =
@@ -99,14 +108,7 @@ pub(crate) fn py_option_nested_repr<T: PyRepr + PyClass>(
 ) -> String {
     match opt {
         None => "None".to_string(),
-        Some(py_obj) => Python::attach(|py| {
-            let inner = &*py_obj.borrow(py);
-            match mode {
-                ReprMode::Str => <T as PyRepr>::__str__(inner),
-                ReprMode::Repr => <T as PyRepr>::__repr__(inner),
-                ReprMode::Html => <T as PyRepr>::_repr_html_(inner),
-            }
-        }),
+        Some(py_obj) => Python::attach(|py| py_obj.borrow(py).render(mode)),
     }
 }
 
--- a/icechunk-python/src/repository.rs
+++ b/icechunk-python/src/repository.rs
@@ -1031,12 +1031,7 @@ impl PyRepr for PyRepository {
         let repo = self.0.blocking_read();
         let storage = format!("{}", repo.storage());
         let py_config: PyRepositoryConfig = repo.config().clone().into();
-        let config = match mode {
-            ReprMode::Str => PyRepr::__str__(&py_config),
-            ReprMode::Repr => PyRepr::__repr__(&py_config),
-            ReprMode::Html => PyRepr::_repr_html_(&py_config),
-        };
-        vec![("storage", storage), ("config", config)]
+        vec![("storage", storage), ("config", py_config.render(mode))]
     }
 }
 


icechunk-python/src/config.rs
```diff
@@ -772,6 +766,26 @@ pub struct PyCachingConfig {
     pub num_bytes_chunks: Option<u64>,
 }
 
+// TODO: use `executable` once nested fields implement PyRepr
+#[expect(unused_variables)]
```
TomNicholas: The follow-up PR will implement reprs for every nested class, meaning ~~executable~~ `mode` is always used, and this becomes unnecessary.

icechunk/src/display.rs
```diff
@@ -1,18 +1,78 @@
...
+/// Implementers provide `cls_name`, `fields`, and `EXECUTABLE`.
+/// Default implementations of `__str__`, `__repr__`, and `_repr_html_` are provided.
+///
+/// The `fields(executable)` parameter controls how nested objects render themselves:
+/// - `__repr__` passes `Self::EXECUTABLE`, so an executable class gets executable
+///   nested reprs, producing output that `eval()` can reconstruct.
+/// - `__str__` and `_repr_html_` always pass `false`, because these are meant to be
+///   human-readable — embedding constructor syntax inside a human-readable display
+///   would be confusing, and a half-executable repr is useless for round-tripping.
+pub trait PyRepr {
```
TomNicholas: This is the important class to review

icechunk/src/display.rs
```diff
@@ -1,18 +1,78 @@
...
+
 /// Render a non-executable, human-readable string for a dataclass-like struct.
 ///
-/// `type_name` is the public Python/Rust-facing type name.
-/// `fields` are `(field_name, rendered_value)` pairs, where each value should
-/// already be a string representation appropriate for the nested type.
-pub fn dataclass_str(type_name: &str, fields: &[(&str, String)]) -> String {
+/// `cls_name` is the public Python class name.
+/// `fields` are `(field_name, rendered_value)` pairs.
+pub fn dataclass_str(cls_name: &str, fields: &[(&str, &str)]) -> String {
```
TomNicholas: AFAICT all our reprs are "dataclass-like" - the use of rust structs underneath means that reprs are almost always just a list of fields with values to display.

icechunk/src/display.rs
```diff
@@ -48,11 +117,11 @@ pub fn dataclass_repr(type_name: &str, fields: &[(&str, String)]) -> String {
 /// Uses a `<details><summary>` pattern for nested values that contain newlines.
 /// Nested HTML fragments should be passed in via `fields` values when available; otherwise
 /// plain text will be HTML-escaped minimally.
-pub fn dataclass_html_repr(type_name: &str, fields: &[(&str, String)]) -> String {
+pub fn dataclass_html_repr(cls_name: &str, fields: &[(&str, &str)]) -> String {
```
TomNicholas: We can make this look nicer with CSS or whatever later - but this means there is only one place that needs changing to improve all reprs.

icechunk-python/tests/test_display.py
```diff
@@ -16,125 +13,125 @@ def repo() -> Repository:
...
-
-def test_repository_repr_html(repo: Repository) -> None:
-    """Test Repository._repr_html_ returns valid HTML."""
-    html = repo._repr_html_()
-    assert '<div class="icechunk-repr">' in html
-    assert "<code>&lt;icechunk.Repository&gt;</code>" in html
-    assert (
-        "<strong>storage</strong>: <code>ObjectStorage(backend=InMemoryObjectStoreBackend)</code>"
-        in html
+def assert_repr_roundtrips(obj: object) -> None:
```
TomNicholas, line 16: Most tests are in python because (a) it only matters for python and (b) it makes testing that exectuable reprs are actually executable easier.

icechunk/src/display.rs
TomNicholas: since did that

PR #1950, opened 2026-04-02
feat: Allow changing repo status without creating an Update

paraseba:


icechunk-format/src/repo_info.rs
```diff
@@ -1092,58 +1181,97 @@ impl RepoInfo {
...
-            self.disabled_feature_flags()?,
-            status,
-        )
+                self.config_bytes_raw()?.as_deref(),
+                self.enabled_feature_flags()?,
+                self.disabled_feature_flags()?,
+                status,
+            )
+        } else {
+            Self::from_parts(
```
li-em, line 1259: unfortunate to have to copy most of the same `Self::from_parts` call except for the builder, since it is a closure and can't be put as the return of an `if` (because closures are anonymous, and different types, even if they have same params 😞)

PR #1951, opened 2026-04-02
chore: Change RustFS port to avoid conflicts

paraseba:


icechunk-python/tests/data/test-repo-v1/config.yaml
```diff
@@ -29,7 +29,7 @@ virtual_chunk_containers:
     url_prefix: s3://testbucket
     store: !s3_compatible
       region: us-east-1
-      endpoint_url: http://localhost:9000
+      endpoint_url: http://localhost:4200
```
li-em, line 32: Oh, this ends up inlined, and so need to regenerate all the test data?

PR #1974, opened 2026-04-02
Verify all parent nodes exist when moving a node

### LINKED ISSUE #1949: Error if moving to a group that doesn't exist
`a -> b/a` should fail if `b` doesn't exist already

li-em:
Fix #1949

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -1061,9 +1061,9 @@ impl Session {
         let mut ancestors = path.ancestors();
         // the first element is the `path` itself, which we might be
         // trying to create now; skip it.
-        debug_assert_eq!(ancestors.next().as_ref(), Some(path));
+        let current_path = ancestors.next();
+        debug_assert_eq!(current_path.as_ref(), Some(path));
         for parent in ancestors {
-            dbg!((path, &parent));
             let node = self.get_node(&parent).await;
             if node.is_err() {
                 return Err(SessionError::capture(


icechunk-python/tests/test_move.py
```diff
@@ -264,6 +264,7 @@ def test_doesnt_rebase() -> None:
     session = repo.writable_session("main")
     root = zarr.group(store=session.store, overwrite=True)
     root.create_group("my/old/path", overwrite=True)
+    root.create_group("my/new", overwrite=True)
```
li-em, line 271: This one failed after the check because `/my/new` must exist before we move something under it

dcherian: Nice!

dcherian: Let's add an assertion that we can get the new nodes, presumably that would;'ve failed earlier

icechunk/src/session.rs
```diff
@@ -5027,6 +5047,52 @@ mod tests {
...
+        assert!(matches!(res,
+                ICError { kind, ..} if matches!(&kind,
+                                                SessionErrorKind::AncestorNodeNotFound {prefix, ..}
+                                                if *prefix == dest_path)));
+
+        Ok(())
+    }
+
+    #[tokio_test]
+    async fn implicit_group_creation_in_add() -> Result<(), Box<dyn Error>> {
```
dcherian: Does this fail now?

li-em: aye, removed!

icechunk/src/feature_flags.rs
```diff
@@ -417,6 +418,7 @@ mod tests {
 
         // create a group so we have something to move
         let mut session = repo.writable_session("main").await.unwrap();
+        session.add_group(Path::root(), Bytes::copy_from_slice(b"")).await.unwrap();
```
li-em, line 421: The two tests in this file were not creating the root group

icechunk/src/session.rs
```diff
@@ -5398,6 +5444,7 @@ mod tests {
             nodes.into_iter(),
             [
                 Path::new("/").unwrap(),
+                Path::new("/foo").unwrap(),
```
li-em, line 5459: Another interesting behavior: before it was valid to not have `/foo`, now it fails

icechunk/src/session.rs
```diff
@@ -1049,6 +1052,23 @@ impl Session {
         }))
     }
 
+    #[instrument(skip(self))]
+    async fn check_all_ancestors_exist(&self, path: &Path) -> SessionResult<()> {
+        let mut ancestors = path.ancestors();
+        // the first element is the `path` itself, which we might be
+        // trying to create now; skip it.
+        ancestors.next();
```
```suggestion
        debug_assert_eq!(ancestors.next(), path);
```

icechunk/src/session.rs
dcherian: Yeah that's the same Zarr store issue; the test starts with `/foo/old` without creating `/foo`

dcherian (approved):

icechunk/src/session.rs
li-em: This is a fun one: `debug_assert_eq!` is only executed on the `dev` profile, but we want `ancestors.next()` to always be executed. Doing this ended up triggering heisenbugs because locally the test would pass (dev profile), but in CI we build wheels first (release profile) and so tests were failing!

PR #1979, opened 2026-04-03
Compare GC/expiry behaviour between V1 & V2

### LINKED ISSUE #1925: Verify IC2 library has the same GC/expiration behavior as IC1 lib on IC1 repos

dcherian:
Closes #1925

After many iterations this is the simplest impl I could find.

The goal of this test is to gain confidence that IC2 replicated the behaviour of IC1
w.r.t GC & expiry. Doing so is complicated by bugs in the IC1 impl (due to various side-effects).
So exactly modeling this is not possible/not worth it. Fixing it is also not worth it.

So instead:
1. we run the VersionControlStateMachine as usual
2. on expire and GC:
   a. we copy the repo over to a new temp directory;
   b. run GC & expire with IC1 (`icechunk_v1`)
   c. assert the return values are identical to that with IC2 (`icechunk`)

### DIFF
--- a/icechunk-python/tests/test_stateful_compat.py
+++ b/icechunk-python/tests/test_stateful_compat.py
@@ -217,22 +217,37 @@ class CrossVersionExpireGCStateMachine(VersionControlStateMachine):
         assert self.storage is not None
         self.repo = self.actor.open(self.storage)
         with self._v1_repo_copy() as (v1_repo, _):
-            v2_result = self.repo.expire_snapshots(
-                older_than,
-                delete_expired_branches=delete_expired_branches,
-                delete_expired_tags=delete_expired_tags,
-            )
-            v1_result = v1_repo.expire_snapshots(
-                older_than,
-                delete_expired_branches=delete_expired_branches,
-                delete_expired_tags=delete_expired_tags,
-            )
+            v2_error: Exception | None = None
+            v1_error: Exception | None = None
+            try:
+                v2_result = self.repo.expire_snapshots(
+                    older_than,
+                    delete_expired_branches=delete_expired_branches,
+                    delete_expired_tags=delete_expired_tags,
+                )
+            except (ic.IcechunkError, Exception) as e:
+                v2_error = e
+            try:
+                v1_result = v1_repo.expire_snapshots(
+                    older_than,
+                    delete_expired_branches=delete_expired_branches,
+                    delete_expired_tags=delete_expired_tags,
+                )
+            except (ic_v1.IcechunkError, Exception) as e:
+                v1_error = e
 
-            note(f"expire_snapshots: v2={v2_result!r}, v1={v1_result!r}")
-            assert v2_result == v1_result, (
-                f"expire_snapshots mismatch: v2={v2_result}, v1={v1_result}"
-            )
-            event(f"cross-version expire: {len(v2_result)} snapshots expired")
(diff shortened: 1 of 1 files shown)


icechunk-python/tests/test_stateful_compat.py
```diff
@@ -116,6 +128,227 @@ def test_two_actors_cross_version() -> None:
...
+
+    def checks(self) -> None:
+        assert self.model.spec_version == getattr(self.repo, "spec_version", 1)
+        self.check_list_prefix_from_root()
+        self.check_tags()
+        self.check_branches()
+        # check_ancestry disabled: after expire/GC the model's commits
+        # are synced from on-disk snapshots, not ancestry, so the sets
+        # may diverge from what ancestry traversal returns.
+        # self.check_ancestry()
```
dcherian, line 174: I don't think this si very important; we check ancestry (without GC/expire for V1; with GC/expire for V2) in the other test.

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -116,6 +128,227 @@ def test_two_actors_cross_version() -> None:
...
+            shutil.copytree(self._storage_path, v1_path, dirs_exist_ok=True)
+            v1_storage = ic_v1.local_filesystem_storage(v1_path)
+            yield ic_v1.Repository.open(v1_storage), v1_path
+        finally:
+            shutil.rmtree(v1_path, ignore_errors=True)
+
+    def _sync_model_from_repo(self) -> None:
+        """Sync the dict model's commit/branch/tag state from the actual repo.
+
+        After expire or GC the repo is the source of truth. We sync the
```
ianhi: little confused here

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -116,6 +128,227 @@ def test_two_actors_cross_version() -> None:
...
+        """
+        v1_path = tempfile.mkdtemp()
+        try:
+            shutil.copytree(self._storage_path, v1_path, dirs_exist_ok=True)
+            v1_storage = ic_v1.local_filesystem_storage(v1_path)
+            yield ic_v1.Repository.open(v1_storage), v1_path
+        finally:
+            shutil.rmtree(v1_path, ignore_errors=True)
+
+    def _sync_model_from_repo(self) -> None:
```
ianhi, line 192: can this jsut go on the model? or even `model_from _storage`?

ianhi: the pointn of this is avoid having to model all the bugs right?

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -183,4 +416,7 @@ def test_two_actors_zarr_cross_version() -> None:
     def mk_test_instance_sync() -> CrossVersionTwoActorZarrHierarchyStateMachine:
         return CrossVersionTwoActorZarrHierarchyStateMachine(tempfile.mkdtemp())
 
-    run_state_machine_as_test(mk_test_instance_sync)  # type: ignore[no-untyped-call]
+    run_state_machine_as_test(
+        mk_test_instance_sync,
+        # settings=settings(report_multiple_bugs=False),
+    )  # type: ignore[no-untyped-call]
```
ianhi: ?

PR #1983, opened 2026-04-03
Add structured Storage repr and format_version to Repository

TomNicholas:
## Summary

- Adds `__repr__`, `__str__`, and `_repr_html_` to the `Storage` class, showing backend type and relevant fields (bucket, prefix, region, etc.)
- `Repository` repr now shows `format_version` (spec version, cheap field read) and uses the structured Storage repr instead of the raw Rust `Display` string
- S3 backends show non-default boolean flags (anonymous, allow_http, force_path_style, requester_pays) via `S3Options::info_fields()`

image

### Why `storage_info()` on the `Storage` trait?

`PyStorage` wraps `Arc<dyn Storage>` — a trait object where the concrete backend type and its fields (bucket, region, prefix, etc.) are erased. There's no way to downcast or inspect the concrete type from Python. The existing `Display` impl produces a Rust-internal format string that would need fragile parsing. `storage_info()` provides a structured escape hatch: each backend returns its type name and key-value fields through the trait vtable, so the Python repr can present them cleanly without knowing the concrete type.

### Changes across crates

- **icechunk-storage**: `StorageInfo` struct, `storage_info()` on `Storage` trait, `S3Options::info_fields()`
- **icechunk-arrow-object-store**: `storage_info()` on `ObjectStoreBackend` trait + all 6 backends
- **icechunk-s3**: `storage_info()` on native `S3Storage`
- **icechunk**: `storage_info()` on wrapper storages (latency/logging delegate, redirect returns URL). `StorageInfo` re-export.
- **icechunk-python**: `PyRepr` for `Storage`, updated `Repository` repr, type stubs, 4 new tests

### DIFF
--- a/icechunk-python/src/repository.rs
+++ b/icechunk-python/src/repository.rs
@@ -1139,7 +1139,6 @@ impl PyRepr for PyRepository {
 
     fn fields(&self, mode: ReprMode) -> Vec<(&str, String)> {
         let repo = self.0.blocking_read();
-        let spec_version = format!("{}", repo.spec_version() as u8);
         let info = repo.storage().storage_info();
         let mut storage_fields = vec![("type", info.backend_type.to_string())];
         storage_fields.extend(info.fields);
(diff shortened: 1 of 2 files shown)


icechunk-storage/src/storage.rs
```diff
@@ -442,10 +442,22 @@ impl Display for GetModifiedResult {
 /// Fetch and write the parquet files that represent the repository in object store
 ///
 /// Different implementation can cache the files differently, or not at all.
+/// Structured metadata about a storage backend, for display/repr purposes.
+#[derive(Debug, Clone)]
+pub struct StorageInfo {
+    /// Human-readable backend type, e.g. "S3", "GCS", "in-memory".
+    pub backend_type: &'static str,
+    /// Key-value pairs of relevant configuration (bucket, prefix, path, etc.).
+    pub fields: Vec<(&'static str, String)>,
```
dcherian, line 451: Why not a HashMap?

I guess `&'str` works  because the strings are known at compile time for each Storage type.

TomNicholas: because we want to preserve insertion order

ianhi: indexmap! Already in our deps

PR #1985, opened 2026-04-03
Inline flatbuffer definitions in spec

TomNicholas:
Starting to include snippets of flatbuffers definitions in the spec docs page.


ianhi:
superceded

PR #1989, opened 2026-04-03
Raise error in manifest_files field is inconsistent with actual manifests listed in arrays

### LINKED ISSUE #1926: Raise error in `manifest_files` field is inconsistent with actual manifests listed in arrays
This should be become an error

https://github.com/earth-mover/icechunk/blob/2a6e92df88818bbf7a943d524191ea12739d9be2/icechunk/src/session.rs#L2849-L2858

li-em:
Fix #1926

### DIFF
--- a/icechunk-python/docs/mkdocs.yml
+++ b/icechunk-python/docs/mkdocs.yml
@@ -163,7 +163,6 @@ plugins:
         'icechunk-python/dask/index.md': 'dask.md'
         'icechunk-python/xarray/index.md': 'xarray.md'
         'icechunk-python/virtual/index.md': 'virtual.md'
-        'icechunk-python/concurrency/index.md': 'concurrency.md'
         'icechunk-python/parallel/index.md': 'parallel.md'
         'icechunk-python/performance/index.md': 'performance.md'
         'icechunk-python/cheatsheets/git-users/index.md': 'icechunk-for-git-users.md'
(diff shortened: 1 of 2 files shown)


icechunk/src/session.rs
```diff
@@ -2851,15 +2853,21 @@ async fn do_flush(
...
+    let mrefs = flush_data
+        .manifest_refs
+        .values()
+        .flatten()
+        .map(|x| x.object_id.clone())
+        .collect::<HashSet<_>>();
+    if mfiles != mrefs {
+        return Err(SessionError::capture(
+            SessionErrorKind::ManifestsInconsistencyError(format!(
+                "files contains {mfiles:?}, refs contains {mrefs:?}",
```
paraseba: maybe to make the message nicer we: include both hashes in the Error, and list the size of the diff in the message?

```suggestion
                "Snapshot will reference {mfiles:?} manifest, while array nodes will reference {mrefs:?} manifests",
```

Shall we type as `InconsistentManfiestsError(snapshot: usize, nodes: usize)` or something like that?

li-em: I was going to save both hashsets, but that could become quite large; will update to use `snapshot` and `nodes`, thanks @dcherian!

icechunk/src/session.rs
```diff
@@ -6660,6 +6671,53 @@ mod tests {
...
+        let properties = Default::default();
+        let manifest_config = ManifestConfig::default();
+        let mut flush_data = FlushProcess::new(
+            Arc::clone(&session.asset_manager),
+            &session.change_set,
+            &Snapshot::INITIAL_SNAPSHOT_ID,
+            &manifest_config,
+        );
+
+        // poke at flush_data to make inconsistent
```
dcherian, line 6691: nice test!

dcherian (approved):

PR #1999, opened 2026-04-06
Filter out all checksum headers on R2 (not just CRC32).

dcherian:
There are a number of other checksum
algorithms (https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html). I am occasionally seeing `crc32nvme` locally (which causes an error)

Migrated our python test script to be an integration test.

### DIFF
--- a/icechunk-python/tests/test_move.py
+++ b/icechunk-python/tests/test_move.py
@@ -23,6 +23,8 @@ async def test_basic_move() -> None:
     session = repo.writable_session("main")
     store = session.store
     root = zarr.group(store=store, overwrite=True)
+    root.create_group("my", overwrite=True)
+    root.create_group("my/old", overwrite=True)
     group = root.create_group("my/old/path", overwrite=True)
     group.create_array("array", shape=(10, 10), chunks=(2, 2), dtype="i4", fill_value=42)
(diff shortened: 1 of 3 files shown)


icechunk-s3/src/lib.rs
```diff
@@ -136,7 +136,15 @@ impl Intercept for StripChecksumOn304Interceptor {
...
         if response.status().as_u16() == 304 {
-            response.headers_mut().remove("x-amz-checksum-crc32");
+            let to_remove: Vec<_> = response
+                .headers()
+                .iter()
+                .filter(|(name, _)| name.starts_with("x-amz-checksum-"))
+                .map(|(name, _)| name.to_owned())
+                .collect();
+            for name in to_remove {
+                response.headers_mut().remove(&name);
```
li-em, line 146: Too bad the [HeaderMap](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderMap.html) from reqwest doesn't have a [retain](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain) method, would make this much shorter =]

dcherian: Yup!

PR #2017, opened 2026-04-07
chore(js): Update js dependabot setup

mpiannucci:
change npm to weekly and ignore examples specifically

### DIFF
--- a/icechunk-python/docs/docs/contributing.md
+++ b/icechunk-python/docs/docs/contributing.md
@@ -577,34 +577,21 @@ rustflags = ["-C", "link-arg=-fuse-ld=mold"]
 
 ### Features
 
-- Support more object stores and more of their custom features
+- Bindings to other languages: C, Julia
 - Better Python API and helper functions
-- Bindings to other languages: C, Wasm
-- Better, faster, more secure distributed sessions
+- Branch merge
 - Savepoints and persistent sessions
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -22,3 +22,26 @@ updates:
       actions:
         patterns:
           - "*"
+  - package-ecosystem: "npm"
```
paraseba, line 25: should we include a [cooldown](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-)

mpiannucci: yes!

mpiannucci: added

.github/dependabot.yml
```diff
@@ -22,3 +22,28 @@ updates:
...
+    directory: "/icechunk-js"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "05:00"
+      timezone: "US/Pacific"
+    cooldown:
```
dcherian, line 32: do we need cooldown on the rest?

mpiannucci: you mean on the rust?

mpiannucci: I want to leave this js only for now

PR #2020, opened 2026-04-07
docs: Document redirect storage

### LINKED ISSUE #1663: Document redirect Storage instance
We need to explain when and why to use it

mpiannucci:
Closes #1663

### DIFF
--- a/icechunk-python/docs/docs/howto.md
+++ b/icechunk-python/docs/docs/howto.md
@@ -210,6 +210,23 @@ for snapshot in repo.ancestry(branch="main"):
 session = repo.readonly_session(snapshot_id=snapshot_id)
 ```
 
+### Amend a Snapshot
+
+For more, see [amending](./version-controld.md#amending).
+```python
+session = repo.writable_session("branch_name")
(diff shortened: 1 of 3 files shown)


icechunk-python/docs/docs/storage.md
```diff
@@ -334,6 +334,62 @@ See the [object_store ClientConfigKey documentation](https://docs.rs/object_stor
...
+
+```python
+icechunk.redirect_storage("https://example.com/my-repo-location")
+```
+
+When the repository is first accessed, Icechunk will follow redirects from the given URL until it reaches a recognized storage scheme, and then use that as the underlying backend. The resolved backend is cached for the lifetime of the storage instance.
+
+This is useful when a service controls which bucket or path a repository lives in. For example, a catalog or registry can serve a redirect so that clients don't need to know the final storage location ahead of time — they only need the stable redirect URL. If the data moves to a different bucket or provider, the redirect can be updated without changing any client code.
+
+#### Example redirect server
```
dcherian, line 349: dang nice!

Issue #1958, opened 2026-04-02
Document ops log and transation log


dcherian:
These seem to be missing completely.

PR #2036, opened 2026-04-07
Document ops log

ianhi:
Ops part of https://github.com/earth-mover/icechunk/issues/1958

not sure what about the txn log?


dcherian (approved):
Nice! Yeah i think we skip transaction log for now.

Issue #1345, opened 2025-10-31
Update spec document


paraseba:
Decide how to keep track of different versions of the spec

paraseba:
@ianhi @dcherian @TomNicholas my minor comments on the spec, it looks great in general, only minor details and outdated pieces. Sorry for the chaotic formatting 


- [x] "containing one or more Arrays and Groups": link to the relevant definitions in zarr
- [x] "and list-after-write consistency is expected.": we probably don't need this anymore except for GC but that's a high latency operation so I think we can delete this requirement

- [x] "Write-if-not-exists - For creating new references.": old language from IC1, for creating new objects

- [x] "For the commit process to be safe and consistent,": for the _update_ process

- [x] "it fetches the latest version of the repo info object and applies its changes in memory first": a couple of  things we could clarify to avoid scaring people: the repo object is small so keeping it in mem in fine; the interval of conflict is very small and unrelated to the duration of the write session.

- [x] "the resolution mechanism can be expensive": I'd say something like "retries may be needed"

- [x] "which is used to detect the existence of a valid repository in a given path": no longer the case

- [x] In the explanation about branches we should mention that snapshots have a parent and this link is what constitutes the tree

- [x] "The snapshot file contains pointers to one or more chunk manifest files.": zero or more

The "Node Paths" section renders an empty code block after the first paragraph

- [x] In the file types, let's delete ` 3 = Attributes, `, it was never implemented and it's not needed

- [x] In the file types, 6 = Chunk, I don't think we are using that anywhere

- [x] When we introduce ManifestFileInfoV2 mention that V2 repos write their manifests using that datastructure, and IC1 repos using ManifestFileInfo

- [ ] "Every get operation in Icechunk gets resolved using the requested coordinates to a specific manifest file and": every chunk get, otherwise the snapshot is enough

- [x] There are several code blocks that render empty in icechunk.io

- [x] We are not mentioning the order of the arrays in the flat buffers. Intentional? it would be important to an implementor. I guess it's enough to have them in the flatbuffers right?

- [x] "Open a repository at a specific branch as described above, keeping track of the sequence number and branch name in the session context.": sequence number is very outdated. Let's say the parent snapshot id instead.

- [x] "Changes from spec version 1" section, let's include the new prefix overwritten.

ianhi:
> There are several code blocks that render empty in icechunk.io

hmm i dont see any when I scroll through. what sections were you look at?

e..g the node paths

[...]

PR #2039, opened 2026-04-08
Update spec per feedback

ianhi:
from https://github.com/earth-mover/icechunk/issues/1345


dcherian:
There's a few more to check off it seems like (i made a checklist): https://github.com/earth-mover/icechunk/issues/1345#issuecomment-4202982858

PR #2042, opened 2026-04-08
Reorganize docs directory structure to match sidebar nav

TomNicholas:
## Summary

- Moved doc files from a flat layout into subdirectories that mirror the mkdocs nav sections: `getting-started/`, `understanding/`, `guides/`, and `reference/`
- Updated all internal cross-references (~60 links across 20+ files) and a Python docstring link in `repository.py`
- Added redirects from old flat URLs to new paths so existing links don't break


ianhi:
> Added redirects from old flat URLs to new paths so existing links don't break

❤️

PR #1995, opened 2026-04-06
Split Python namespace into submodules

TomNicholas:
## Summary

Towards #1172, implementing step 2 of the design doc (#1179).

Organizes the flat `icechunk` namespace (~80 items) into focused submodules so that classes are importable from specific paths like `from icechunk.config import RepositoryConfig`. Key design decisions:

- **No breaking changes**: all existing top-level imports (`from icechunk import X`) continue to work unchanged. Existing tests are left as-is to serve as regression tests for this.
- **No deprecation warnings yet**: this PR just establishes the new import paths. Deprecation of top-level imports is a follow-up.
- **`Repository` and exceptions stay top-level**: `ic.Repository` is the main entry point, and `ic.IcechunkError` / `ic.ConflictError` / `ic.RebaseFailedError` are commonly caught by users.
- **`Session` lives under `ic.session`**: users don't construct sessions directly (they come from `repo.writable_session()` etc.), so `Session`, `ForkSession`, and `SessionMode` belong in a submodule.
- **Manifest config classes live in `config`**, not a separate module, since they are all configuration options. `ManifestFileInfo` (repo state metadata) goes in `snapshots`.
- **`store.py` renamed to `zarr.py`** since `IcechunkStore` is a zarr `Store` implementation.
- **`Update`, `UpdateType`, `GCSummary` live in `ops`** — they are operation/event types, not snapshot metadata.
- **Rust `cls_name()` updated** to use fully-qualified submodule paths in `__repr__`/`__str__`/`_repr_html_`.
- **API docs split into one page per submodule** for easier navigation, with `Repository`, exceptions, and utilities on the index page.
- **All doc examples updated** to use `import icechunk as ic` with `ic.submodule.thing` style.

### Submodules

| Module | Contents |
|--------|----------|
| `icechunk.config` | `RepositoryConfig`, `CachingConfig`, `CompressionConfig`, `ManifestSplittingConfig`, `ManifestConfig`, ... |
| `icechunk.conflicts` | `ConflictSolver`, `BasicConflictSolver`, `Conflict`, `ConflictType`, `VersionSelection` |
| `icechunk.credentials` | All credential types, factories, and type aliases |
| `icechunk.ops` | `Update`, `UpdateType`, `GCSummary` |
| `icechunk.session` | `Session`, `ForkSession`, `SessionMode` |
| `icechunk.snapshots` | `SnapshotInfo`, `Diff`, `ManifestFileInfo` |
[...]

PR #2047, opened 2026-04-08
Fix icechunk.store module renamed to icechunk.zarr

TomNicholas:
## Summary

The namespace refactoring PR (#1995) renamed `store.py` to `zarr.py`, breaking the existing `from icechunk.store import IcechunkStore` import path.

This reverts the file rename back to `store.py` while keeping `IcechunkStore` as top-level (`ic.IcechunkStore`). The module is also still accessible as `ic.store.IcechunkStore` for backwards compatibility.


dcherian (approved):

ianhi:
Needs an update to the breaking changes doc. Otherwise looks good

PR #2048, opened 2026-04-08
chore(ci): shard hypothesis at example level for even distribution

samantha-earthmover:
## Summary

Replaces `pytest-shard` (test-node-level sharding) with hypothesis-native example-level sharding. Instead of distributing test *functions* across runners, all 3 shards now run all hypothesis tests but each with 1/3 of the examples and a different seed.

**Changes:**
- `conftest.py`: reads `HYPOTHESIS_NUM_SHARDS` env var, divides `max_examples` by it (200 → 67 per shard)
- Workflow: passes `--hypothesis-seed=${{ matrix.shard-id }}` so each shard explores a different part of the example space
- Removes `pytest-shard` dependency (no longer needed)

## Context

The previous sharding (#2046) used `pytest-shard` which distributes test *nodes* across runners. The problem: `VersionControlStateMachine` is a single test node that takes ~7 min (200 examples × 75 stateful steps), so one shard always gets stuck with it while others finish in 2 min.

With example-level sharding, each shard runs ALL tests but with only 67 examples each. The 3 shards use different `--hypothesis-seed` values so they explore different parts of the property space — combined coverage is equivalent to 201 examples with 3 different starting points.

Expected: ~7m slowest shard → ~2.5m even distribution across all 3.

## Risk level

Low. The `HYPOTHESIS_NUM_SHARDS` env var defaults to 1 (no-op) when unset, so local development and other CI jobs are unaffected. Total hypothesis examples across shards (67×3=201) is roughly equivalent to the previous 200.

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -262,7 +262,7 @@ jobs:
         run: |
           set -e
           source .venv/bin/activate
-          python -m pytest -m hypothesis --hypothesis-seed=${{ matrix.shard-id }}
+          python -m pytest -m hypothesis
 
       - name: Save cached hypothesis directory
         id: save-hypothesis-cache
--- a/icechunk-python/tests/conftest.py
+++ b/icechunk-python/tests/conftest.py
@@ -18,10 +18,11 @@ from icechunk import Repository, SpecVersion, in_memory_storage, local_filesyste
 # Select with: pytest --hypothesis-profile=nightly
 #
 # When HYPOTHESIS_NUM_SHARDS > 1, max_examples is divided across shards so
-# each parallel runner does a fraction of the work. Use --hypothesis-seed=<N>
-# per shard to ensure different shards explore different example spaces.
+# each parallel runner does a fraction of the work. derandomize is disabled
+# so each shard naturally gets a unique random seed — no fixed seeds needed.
 # ---------------------------------------------------------------------------
 _num_shards = int(os.environ.get("HYPOTHESIS_NUM_SHARDS", "1"))
+_sharding = _num_shards > 1
 
 settings.register_profile(
     "default",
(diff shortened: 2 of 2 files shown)


.github/workflows/python-check.yaml
```diff
@@ -251,18 +251,18 @@ jobs:
...
         working-directory: icechunk-python
         env:
           HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY: "1"
+          HYPOTHESIS_NUM_SHARDS: "3"
           NOCOLOR: 1
         run: |
           set -e
           source .venv/bin/activate
-          python -m pytest -m hypothesis --num-shards=3 --shard-id=${{ matrix.shard-id }}
+          python -m pytest -m hypothesis --hypothesis-seed=${{ matrix.shard-id }}
```
dcherian:
```suggestion
          python -m pytest -m hypothesis
```

we shouldn't be fixing the seed; that would just get us repeated runs of rthe same inputs

icechunk-python/tests/conftest.py
```diff
@@ -24,7 +32,7 @@
 settings.register_profile(
     "ci",
     parent=settings.get_profile("ci"),
-    max_examples=200,
+    max_examples=math.ceil(200 / _num_shards),
```
dcherian, line 36: hmm.. this maaaaay be ok but not sure at all.

.github/workflows/python-check.yaml
samantha-earthmover: so non-deterministic runs are ok? sorry, haven't interacted with hypothesis before!

icechunk-python/tests/conftest.py
samantha-earthmover: if the seed is random this should be fine, we'd still be running the exact same number of examples total. assuming collisions odds are low enough

PR #2049, opened 2026-04-08
Add set_virtual_refs_arr with numpy arrays

### LINKED ISSUE #1574: Method to allow setting array of virtual refs in a zero-copy way

TomNicholas:
## Summary

- Closes #1574
- Adds `set_virtual_refs_arr()` and `set_virtual_refs_arr_async()` to `IcechunkStore`, accepting a `list[str]` of locations plus numpy `uint64` arrays for offsets and lengths
- Avoids creating per-chunk `VirtualChunkSpec` Python objects — locations are borrowed directly from the Python list in Rust, numeric arrays are zero-copy via PyO3 numpy bindings
- Adds `numpy = "0.27"` dependency to `icechunk-python` only (not the core Rust crate)
- Helpers live in a dedicated `virtualrefs` Rust module
- Includes end-to-end benchmarks comparing old vs new API

### Benchmark results (100k refs)

| Method | Without #2045 | With #2045 | 
|--------|--------------|------------|
| `set_virtual_refs` (VirtualChunkSpec) | ~237ms | ~120ms |
| `set_virtual_refs_arr` (list + numpy) | ~204ms | ~77ms |

Combined with the `optimize-set-virtual-refs` base branch: **~3x faster** than the original baseline.

Depends on #2045.

### DIFF
--- a/icechunk-python/benchmarks/test_benchmark_writes.py
+++ b/icechunk-python/benchmarks/test_benchmark_writes.py
@@ -316,9 +316,7 @@ def test_write_virtual_refs_e2e(benchmark, repo) -> None:
             )
             for i in range(NUM_VIRTUAL_CHUNK_REFS)
         ]
-        session.store.set_virtual_refs(
-            "array", chunks, validate_containers=False
-        )
+        session.store.set_virtual_refs("array", chunks, validate_containers=False)
 
 
 @pytest.mark.benchmark(group="refs-write-e2e")
@@ -350,8 +348,8 @@ def test_write_virtual_refs_arr_e2e(benchmark, repo) -> None:
         [f"{FAKE_URL_PREFIX}-{i}.nc" for i in range(NUM_VIRTUAL_CHUNK_REFS)],
         dtype=np.dtypes.StringDType(),
     )
-    offsets_arr = (np.arange(NUM_VIRTUAL_CHUNK_REFS, dtype=np.uint64) * 10)
-    lengths_arr = (np.arange(NUM_VIRTUAL_CHUNK_REFS, dtype=np.uint64) + 5)
+    offsets_arr = np.arange(NUM_VIRTUAL_CHUNK_REFS, dtype=np.uint64) * 10
+    lengths_arr = np.arange(NUM_VIRTUAL_CHUNK_REFS, dtype=np.uint64) + 5
 
     @benchmark
     def write():


Cargo.toml
```diff
@@ -55,6 +55,7 @@ futures = "0.3.32"
 itertools = "0.14.0"
 miette = { version = "7.6.0", features = ["fancy"] }
 noxious-client = "1.0"
+numpy = "0.27"
```
li-em, line 58: Prefer to keep version and as much of dep config in the workspace Cargo.toml, so it is easier to maintain same dependency config across workspace members

TomNicholas: done in https://github.com/earth-mover/icechunk/pull/2049/changes/1acc0fdb67d1ab02a9bc8b85eac0b55175271bc7

PR #2065, opened 2026-04-10
Enable icechunk compilation for wasm32-unknown-emscripten

mpiannucci:
Patch backon to allow TokioSleeper on emscripten targets instead of gloo-timers (which requires browser JS APIs unavailable in emscripten). The root issue is that emscripten defaults to panic=unwind, causing wasm-bindgen's MaybeUnwindSafe trait to require UnwindSafe bounds that gloo-timers closures don't satisfy.


dcherian:
Is there a different library we could use instead? I just picked that one because it looked good. Clearly we have different requirements now.

mpiannucci:
@dcherian Yeah we can revisit this when i rev on the cache interface

PR #2078, opened 2026-04-15
don't allow non-regular chunk grids on format 1

ianhi:
Technically not allowing non regular chunk grids in v1 data becuase the current zarr-python implementation uses a different name for non-regular grids than the zep (rectilinear vs rectangular) https://zarr.dev/zeps/draft/ZEP0003.html

### DIFF
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -1484,7 +1484,7 @@ impl Session {
         Ok(())
     }
 
-    fn spec_version(&self) -> SpecVersionBin {
+    pub(crate) fn spec_version(&self) -> SpecVersionBin {
         self.asset_manager.spec_version()
     }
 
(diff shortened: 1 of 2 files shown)


dcherian (approved):

ianhi:
evidentally not zep but extension: https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear

dcherian:
Shall we backport this?

ianhi:
On icechunk<2 it does already fail actually. just with a confusing error:

```
    return await self._store.set(key, value.to_bytes())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
icechunk.IcechunkError:   × bad metadata
  │
  │ context:
  │    0: icechunk::store::set
  │            with key="a/zarr.json"
  │              at icechunk/src/store.rs:287
  │
  ├─▶ bad metadata
  ╰─▶ invalid value: string "array", expected the word 'group' at line 33 column 22
```

Issue #1934, opened 2026-03-31
Add Zizmor


dcherian:
[zizmorcore/zizmor](https://github.com/zizmorcore/zizmor) is doing the rounds as a security focused linter for github actions: [docs.zizmor.sh/audits](https://docs.zizmor.sh/audits/) . Seems prudent to add it.

li-em:
#2104 will get very close to fixing this, only three workflows still ignored (`detect-unused-dependencies,js-ci,python-ci`)

li-em:
Finished in #2294

PR #2079, opened 2026-04-15
ci: move more commands used in CI into Just tasks

li-em:
Migrate a subset of the CI workflows to use `just` for task orchestration, `pixi` for environment management.

Since I was touching these workflows I also cleaned `zizmor --pedantic` warnings, and added a pre-commit check so these workflows don't introduce new warnings in the future (related to #1934)

I also verified the new `just` tasks work with `nix`, so `pixi` is another choice of env setup (and not a requirement).
And for CI I used pixi to setup the environment, bumping the pixi config to use 3.14 by default and adding the `osx-64` platform so we can setup `macos-intel`.

Where relevant (the upstream workflows) I split the jobs to narrow GitHub Actions permissions (separating `contents` from `issues` permissions, for example).

Remove the maturin import hook from  `just develop` task, and create a new `import-hook` task for it. I was having issues building docs and slowdown when running `pytest` in parallel, and prefer to run `maturin develop` when there are rust changes instead of auto-running on import. But the option to set the hook is still available for those who want it.

## Rust caching

Maximize Rust cache sharing across jobs (where it makes sense), so we can lower our cache usage. Or, rather, use it better, there is a bit of churn when many PRs are ongoing because each job has a separate cache, but after #1625 and `profile=ci` the dependencies should be mostly the same, and shareable between job runs.

Following zizmor advice, only save the cache in runs on the `main` branch, to avoid [cache poisoning](https://docs.zizmor.sh/audits/#cache-poisoning). Although it only refers to release jobs, since we are sharing caches in PRs it makes sense to limit them too, because an attacker can potentially poison the cache in a PR and mess with other jobs.

Add Python version to the rust-cache key for correct cache invalidation, since we do build `icechunk-python` and that has a Python version dependency.

## Skipped on purpose

- `js-ci`: using dependencies not available on pixi yet
- `python-check`: great candidate for conversion, but need to move quite a bit into `Justfile`
- `python-ci`: comment on top saying it was autogenerated with `maturin`, so avoiding changing it without checking what newer versions of `maturin` do. Also has to work on `main` and `v1.x` branches (release workflow).
- `detect-unused-dependencies`: currently disabled, too many false positives.

## Other

Fix AI usage policy link, it was broken after docs where moved.

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -46,7 +46,7 @@ jobs:
           # Currently: 3.14
           prefix-key: "rust-py314"
           shared-key: "rust"
-          #save-if: ${{ github.ref == 'refs/heads/main' }}
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - uses: prefix-dev/setup-pixi@a0af7a228712d6121d37aba47adf55c1332c9c2e # v0.9.4
         with:
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -46,7 +46,7 @@ jobs:
           # Currently: 3.14
           prefix-key: "rust-py314"
           shared-key: "rust"
-          save-if: ${{ github.ref == 'refs/heads/main' }}
+          #save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - uses: prefix-dev/setup-pixi@a0af7a228712d6121d37aba47adf55c1332c9c2e # v0.9.4
         with:
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -13,6 +13,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  content: read
+
 env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
@@ -33,6 +36,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
 
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
-  content: read
+  contents: read
 
 env:
   CARGO_INCREMENTAL: 0
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -41,7 +41,12 @@ jobs:
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
         with:
-          add-job-id-key: "false"
+          # The rust cache includes a dep on the Python version (when building icechunk-python)
+          # so we need to bump the version here whenever Python version is updated
+          # Currently: 3.14
+          prefix-key: "rust-py314"
+          shared-key: "rust"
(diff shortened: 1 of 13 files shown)


icechunk-python/pyproject.toml
```diff
@@ -212,6 +212,14 @@ cargo-deny = ">=0.19.0,<0.20"
 prek = ">=0.3.3,<0.4"
 netcdf4 = ">=1.7.4,<2"
 cargo-nextest = ">=0.9.106,<0.10"
+flatbuffers = "*"
+samply = ">=0.13.1,<0.14"
+
+[tool.pixi.feature.build.target.linux-64.dependencies]
+sysroot_linux-64 = ">=2.34"
```
li-em, line 219: `samply` depends on the `gettid` syscall, introduced in `glibc 2.30`. This is the closest sysroot containing it.

.pre-commit-config.yaml
```diff
@@ -30,6 +30,15 @@ repos:
     hooks:
       - id: codespell
 
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
+    rev: v1.24.1
+    hooks:
+    - id: zizmor
+      files: \.github\/workflows\/(code-quality|dependency-check|python-upstream|rust-ci|rust-upstream|windows-check)\.ya?ml
```
li-em, line 37: We should update this as the remaining workflows are `zizmor`-compliant

.github/workflows/code-quality.yaml
```diff
@@ -33,11 +36,17 @@ jobs:
...
 
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
         with:
-          add-job-id-key: "false"
+          # The rust cache includes a dep on the Python version (when building icechunk-python)
+          # so we need to bump the version here whenever Python version is updated
+          # Currently: 3.14
+          prefix-key: "rust-py314"
+          shared-key: "rust"
```
li-em, line 48: This config generates names like `rust-py314-rust-Windows_NT-x64-38199bb0-ef60912a`, one cache per platform (OS + arch). Weird to have `rust` listed twice, but 🤷

.github/workflows/rust-ci.yaml
```diff
@@ -51,13 +51,14 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
 
       - name: Check if MSRV is up to date
         run: |
           set -e
-          grep 'rust-version = "${{ env.RUST_CHANNEL }}"' Cargo.toml
+          grep "rust-version = \"${MSRV}\"" Cargo.toml
```
li-em, line 59: An example of the [template injection](https://docs.zizmor.sh/audits/#template-injection) audit from zizmor

.github/workflows/rust-upstream.yaml
```diff
@@ -17,21 +17,22 @@ concurrency:
 
 permissions:
   contents: read
-  issues: write
```
li-em, line 20: Example of defaulting to `contents: read` as much as possible, and only use more permissions when needed. In this case it was necessary to split job in two, since permissions are at most job-level, not step-level.

.github/workflows/publish-rust-library.yml
```diff
@@ -6,77 +6,62 @@ env:
...
-          # workspaces: "rust -> target"
-          key: ${{ env.RUST_CHANNEL }}
-          cache-on-failure: true
+          pixi-version: ${{ env.PIXI_VERSION }}
+          cache: true
+          activate-environment: true
+          manifest-path: icechunk-python/pyproject.toml
 
       # FIXME: uncomment
-      # - name: Install cargo-deny
```
dcherian, line 64: Needs uncommenting?

dcherian (approved):
Seems ok to me. Noticed one FIXME that I called out (`cargo-deny`)

.github/workflows/publish-rust-library.yml
li-em: This was already commented out (in https://github.com/earth-mover/icechunk/pull/1897), to avoid running the tests (again) on release

PR #2080, opened 2026-04-15
fix: skip empty paths in set_virtual_refs_arr

TomNicholas:
## Summary

This basically moves a python `for...if` in virtualizarr ([here](https://github.com/zarr-developers/VirtualiZarr/blob/3065bcd7b35079d70daf2e7ba969d3c54f843b46/virtualizarr/manifests/manifest.py#L463), called from [here](https://github.com/zarr-developers/VirtualiZarr/blob/3065bcd7b35079d70daf2e7ba969d3c54f843b46/virtualizarr/writers/icechunk.py#L578)) into `icechunk-python` rust.

- Skip empty location strings in `set_virtual_refs_arr` instead of raising a URL parse error
- Empty paths represent missing chunks in VirtualiZarr manifests — they should not be written as virtual refs
- Adds a test verifying that only non-empty chunks are written

Found while testing zarr-developers/VirtualiZarr#967.

### DIFF
--- a/icechunk-python/python/icechunk/store.py
+++ b/icechunk-python/python/icechunk/store.py
@@ -392,7 +392,8 @@ class IcechunkStore(Store, SyncMixin):
             The product must equal the length of the arrays.
             Arrays are assumed to be flattened in C (row-major) order.
         locations : list[str]
-            URLs to external files containing chunk data.
+            URLs to external files containing chunk data. Empty strings
+            represent missing chunks and are silently skipped.
             Example: ["s3://bucket/file1.nc", "s3://bucket/file2.nc"]
         offsets : np.ndarray
             1-D uint64 array of byte offsets within each file.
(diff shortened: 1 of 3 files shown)


icechunk-python/python/icechunk/store.py
```diff
@@ -392,7 +392,8 @@ def set_virtual_refs_arr(
             The product must equal the length of the arrays.
             Arrays are assumed to be flattened in C (row-major) order.
         locations : list[str]
-            URLs to external files containing chunk data.
+            URLs to external files containing chunk data. Empty strings
```
dcherian, line 395: Can we use `None` instead please? & `Vec<Option<String>>` on the rust side?

TomNicholas: I would rather not as then VirtualiZarr has to do a loop in python before I can pass to this method. I added a comment to explain this in https://github.com/earth-mover/icechunk/pull/2080/changes/9b6b97305b8475aea716512c273cc6fe1e5e948b

dcherian (approved):

PR #2083, opened 2026-04-16
test: extract tree model from hypothesis importing file

ianhi:
This allows you to import the surprisingly useful tree model without having hypothesis installed. This is helpful for example in subsync verification withoutu needing hypothesis.

We could go even further and have the trees strategies live inside strategies.py so it's super clear which import paths require hypothesis.

### DIFF
--- a/icechunk-python/python/icechunk/testing/models.py
+++ b/icechunk-python/python/icechunk/testing/models.py
@@ -1,10 +1,25 @@
-"""Model stores for comparison testing."""
+"""Models for comparison testing.
+
+ModelStore is a superclass of the zarr memory store with added icechunk
+methods shift_array, move, and utility copy.
+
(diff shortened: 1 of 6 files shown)


dcherian:
Shall we upstream to Zarr instead?

ianhi:
> Shall we upstream to Zarr instead?



i dont love having to use minimum zarr version for core test infra because then how do we run our min version tests?

ianhi:
So maybe we upstream + keep this for a bit? so our min tests still work

dcherian:
Yep makes sense

dcherian (approved):

ianhi:
> So maybe we upstream + keep this for a bit? so our min tests still work

we will still ahve to keep our move strategies as those don't make sense inside zarr-python

PR #2092, opened 2026-04-23
docs: add blog posts page linking to Icechunk posts

TomNicholas:


icechunk-python/docs/docs/assets/blog/carbonplan-producing-ocr-data.png
dcherian, line 1: Should we save this? Does the social card not render nicely?

TomNicholas: This post is a special case - there is no image in their social card, and their [original post](https://carbonplan.org/blog/producing-ocr-data) doesn't seem to actually have an image file anywhere - I think their graphics renders in your browser client-side somehow?

So to give us a nice image the easiest thing to do was just to save the (small) screenshot 🤷 

I can remove it if you prefer though.

dcherian (approved):

PR #2097, opened 2026-04-27
Re-enable `shuttle` tests.

### LINKED ISSUE #2088: re-enable `shuttle`
There is a shuttle and shuttle-tokio release, so we can re-enable those tests now.

xref https://github.com/awslabs/shuttle/issues/254

dcherian:
Closes #2088

Needed for our next icechunk testing blogpost

### DIFF
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -176,7 +176,12 @@ jobs:
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
         with:
-          key: shuttle-${{ env.MSRV }}
+          # The rust cache includes a dep on the Python version (when building icechunk-python)
+          # so we need to bump the version here whenever Python version is updated
+          # Currently: 3.14
+          prefix-key: "rust-py314"
+          shared-key: "rust"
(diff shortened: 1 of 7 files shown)


.github/workflows/rust-ci.yaml
```diff
@@ -157,32 +157,32 @@ jobs:
...
+      - name: Cache Dependencies
+        uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
+        with:
+          key: shuttle-${{ env.MSRV }}
+
+      - name: Compile shuttle tests
+        run: cargo test -p icechunk --features shuttle --test test_shuttle --no-run
+
+      - name: Run shuttle tests
+        run: cargo test -p icechunk --features shuttle --test test_shuttle -- --nocapture
```
li-em: I will look into making this a Just task later =]

li-em: I think we can also reuse the shared cache here, since all the deps don't depend on `shuttle`, only the final IC compilation

li-em (approved):

.github/workflows/rust-ci.yaml
dcherian: I pushed a commit that I think will share the cache. Leaving the Just task to you :) Thanks for the review!

PR #2102, opened 2026-04-29
FIX/TEST: Invalid moves

ianhi:
Fixed two bugs with move.

1. You could move a group inside of itself - including the root group!
2. you could move a node to have an array as it's parent

Added a descriptive error message for each of these to help the user.

Also we didn't have any hypothesis tests exercising invalid move paths so I've added those - which would have caught the above bugs.


icechunk-python/python/icechunk/testing/models.py
```diff
@@ -11,6 +11,7 @@
 from __future__ import annotations
 
 import itertools
+import posixpath
```
li-em: can we use `pathlib.PurePosixPath` instead?
https://docs.python.org/3/library/pathlib.html#pathlib.PurePosixPath
(which is likely overkill, and `posixpath` is already better than f-strings from before, so just a suggestion, not a strong requirement)

icechunk/src/session.rs
```diff
@@ -796,15 +816,43 @@ impl Session {
...
                 to.to_string(),
             )));
         }
 
-        // verify all parent nodes in "to" path exist
-        self.check_all_ancestors_exist(&to).await?;
+        // destination's immediate parent must exist and be a group. Deeper
+        // ancestors are guaranteed groups when the immediate parent is one
+        // (the tree is well-formed).
+        if let Some(parent) = to.ancestors().nth(1) {
```
paraseba: why not just `parent()` here?

paraseba (approved):

icechunk/src/session.rs
ianhi: that's a method we don't have on Path

https://github.com/earth-mover/icechunk/blob/v2.0.4/icechunk-types/src/lib.rs#L52-L89

butu we could total add that

ianhi: ok I added that method to our Path

PR #2122, opened 2026-05-12
feat(store): add Store.array_chunk_iterator for the VirtualiZarr parser

TomNicholas:
## Summary

Adds `IcechunkStore.array_chunk_iterator(array_path, batch_size)` — a per-array async generator yielding **columnar batches** of chunk references, designed for the new `IcechunkParser` in VirtualiZarr (companion PR: zarr-developers/VirtualiZarr#991).

Each yielded batch is a 6-tuple:

```
(coords, kinds, paths, offsets, lengths, inlined)

  coords   : np.ndarray[uint32, (n, ndim)]   — chunk grid coordinates
  kinds    : np.ndarray[uint8]               — 1=virtual, 2=native, 3=inline
  paths    : list[str]                        — URL (virtual) | bare chunk_id (native) | "" (inline)
  offsets  : np.ndarray[uint64]
  lengths  : np.ndarray[uint64]
  inlined  : dict[int, bytes]                 — keyed by index within this batch
```

`vcc://` virtual locations are resolved against the session's registered containers before the batch crosses FFI. Missing chunks are not yielded (the iterator is sparse).

## Design notes

This API was designed to **keep icechunk free of VirtualiZarr vocabulary**. The earlier sketch (a `Store.array_chunk_refs` method returning a dense `ArrayChunkRefs` struct sized to the chunk grid, with a `__inlined__` sentinel string in the path column and a `native_chunks_prefix` parameter) leaked VZ's manifest shape into icechunk's API. The iterator form:

- Doesn't know the array's chunk grid shape — the caller derives that from zarr metadata.
- Doesn't render native chunk URLs — returns bare `chunk_id` strings, caller decides the prefix.
- Doesn't emit any VZ sentinels — inline chunks live in a separate `dict` keyed by index, and `kinds` tells callers what each row is.
- Memory-bounded by `batch_size`, not by total chunk count.

The rust core is unchanged. The pyo3 binding wraps the existing `Session::array_chunk_iterator` (which already yields `Stream<Item = SessionResult<ChunkInfo>>`) with a `.chunks(batch_size)` adapter and per-batch columnar conversion. One minor addition on `Session`: a public `resolve_virtual_location` accessor so the binding can expand `vcc://` URLs without exposing the internal resolver Arc.

Adds `bytes` as a direct dep of `icechunk-python` (already a workspace dep).

## Test plan

- [x] 4 new Python tests in `icechunk-python/tests/test_array_chunk_iterator.py`:
  - columnar shapes + dtypes
  - kind tags + path values for virtual/inline mix
  - bare `chunk_id` (no URL scheme) for native chunks
  - batch splitting at `batch_size`
- [x] Existing rust test suite (210 tests) still passes.
- [x] stubtest clean for the new method.
- [ ] Bench at scale (preliminary local run: ~1 μs/chunk parse, ~64 B/chunk of Python-side memory; extrapolates to ~16 min and ~62 GiB peak for 1B chunks).

## Companion PR

VirtualiZarr-side parser: zarr-developers/VirtualiZarr#991.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2405,6 +2405,7 @@ version = "2.0.4"
 dependencies = [
  "async-stream",
  "async-trait",
+ "bytes",
  "chrono",
  "clap",
  "futures",
--- a/icechunk-python/Cargo.toml
+++ b/icechunk-python/Cargo.toml
@@ -23,6 +23,7 @@ icechunk = { workspace = true, features = ["logs"] }
 
 async-stream.workspace = true
 async-trait.workspace = true
+bytes.workspace = true
 chrono.workspace = true
 futures.workspace = true
 itertools.workspace = true
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -2546,6 +2546,20 @@ class PyStore:
         checksum: str | datetime.datetime | None,
         validate_container: bool,
     ) -> None: ...
+    def array_chunk_iterator(
+        self,
+        array_path: str,
+        batch_size: int,
+    ) -> AsyncIterator[
+        tuple[
+            np.ndarray[Any, np.dtype[np.uint32]],
+            np.ndarray[Any, np.dtype[np.uint8]],
+            list[str],
+            np.ndarray[Any, np.dtype[np.uint64]],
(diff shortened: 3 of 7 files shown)


TomNicholas:
Superseded by #2124 — moved to an upstream branch since I have push access to this repo and it'll let CI run without the fork-permission tweaks.

PR #2124, opened 2026-05-13
feat(store): add Store.array_chunk_iterator for the VirtualiZarr parser

TomNicholas:
## Summary

Adds `IcechunkStore.array_chunk_iterator(array_path, batch_size)` — a per-array async generator yielding **columnar batches** of chunk references, designed for the new `IcechunkParser` in VirtualiZarr (companion PR: zarr-developers/VirtualiZarr#991).

Each yielded batch is a 6-tuple:

```
(coords, kinds, paths, offsets, lengths, inlined)

  coords   : np.ndarray[uint32, (n, ndim)]   — chunk grid coordinates
  kinds    : np.ndarray[uint8]               — 1=virtual, 2=native, 3=inline
  paths    : list[str]                        — URL (virtual) | bare chunk_id (native) | "" (inline)
  offsets  : np.ndarray[uint64]
  lengths  : np.ndarray[uint64]
  inlined  : dict[int, bytes]                 — keyed by index within this batch
```

`vcc://` virtual locations are resolved against the session's registered containers before the batch crosses FFI. Missing chunks are not yielded (the iterator is sparse).

## Design notes

This API was designed to **keep icechunk free of VirtualiZarr vocabulary**. The earlier sketch (a `Store.array_chunk_refs` method returning a dense `ArrayChunkRefs` struct sized to the chunk grid, with a `__inlined__` sentinel string in the path column and a `native_chunks_prefix` parameter) leaked VZ's manifest shape into icechunk's API. The iterator form:

- Doesn't know the array's chunk grid shape — the caller derives that from zarr metadata.
- Doesn't render native chunk URLs — returns bare `chunk_id` strings, caller decides the prefix.
- Doesn't emit any VZ sentinels — inline chunks live in a separate `dict` keyed by index, and `kinds` tells callers what each row is.
- Memory-bounded by `batch_size`, not by total chunk count.

The rust core is unchanged. The pyo3 binding wraps the existing `Session::array_chunk_iterator` (which already yields `Stream<Item = SessionResult<ChunkInfo>>`) with a `.chunks(batch_size)` adapter and per-batch columnar conversion. One minor addition on `Session`: a public `resolve_virtual_location` accessor so the binding can expand `vcc://` URLs without exposing the internal resolver Arc.

Adds `bytes` as a direct dep of `icechunk-python` (already a workspace dep).

## Test plan

- [x] 4 new Python tests in `icechunk-python/tests/test_array_chunk_iterator.py`:
  - columnar shapes + dtypes
  - kind tags + path values for virtual/inline mix
  - bare `chunk_id` (no URL scheme) for native chunks
  - batch splitting at `batch_size`
- [x] Existing rust test suite (210 tests) still passes.
- [x] stubtest clean for the new method.
- [ ] Bench at scale (preliminary local run: ~1 μs/chunk parse, ~64 B/chunk of Python-side memory; extrapolates to ~16 min and ~62 GiB peak for 1B chunks).

## Companion PR

VirtualiZarr-side parser: zarr-developers/VirtualiZarr#991.

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -2546,6 +2546,20 @@ class PyStore:
         checksum: str | datetime.datetime | None,
         validate_container: bool,
     ) -> None: ...
+    def array_chunk_iterator(
+        self,
+        array_path: str,
+        batch_size: int,
+    ) -> AsyncIterator[
+        tuple[
+            np.ndarray[Any, np.dtype[np.uint32]],
+            np.ndarray[Any, np.dtype[np.uint8]],
+            list[str],
+            np.ndarray[Any, np.dtype[np.uint64]],
+            np.ndarray[Any, np.dtype[np.uint64]],
+            dict[int, bytes],
+        ]
+    ]: ...
     async def set_virtual_ref_async(
         self,
         key: str,
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -2553,11 +2553,11 @@ class PyStore:
     ) -> AsyncIterator[
         tuple[
             np.ndarray[tuple[int, int], np.dtype[np.uint32]],  # coords (n, ndim)
-            np.ndarray[tuple[int], np.dtype[np.uint8]],        # kinds (n,)
-            list[str],                                         # paths
-            np.ndarray[tuple[int], np.dtype[np.uint64]],       # offsets (n,)
-            np.ndarray[tuple[int], np.dtype[np.uint64]],       # lengths (n,)
-            dict[int, bytes],                                  # inlined
+            np.ndarray[tuple[int], np.dtype[np.uint8]],  # kinds (n,)
+            list[str],  # paths
+            np.ndarray[tuple[int], np.dtype[np.uint64]],  # offsets (n,)
+            np.ndarray[tuple[int], np.dtype[np.uint64]],  # lengths (n,)
+            dict[int, bytes],  # inlined
         ]
     ]: ...
     async def set_virtual_ref_async(
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -2552,12 +2552,12 @@ class PyStore:
         batch_size: int,
     ) -> AsyncIterator[
         tuple[
-            np.ndarray[Any, np.dtype[np.uint32]],
-            np.ndarray[Any, np.dtype[np.uint8]],
-            list[str],
-            np.ndarray[Any, np.dtype[np.uint64]],
-            np.ndarray[Any, np.dtype[np.uint64]],
-            dict[int, bytes],
+            np.ndarray[tuple[int, int], np.dtype[np.uint32]],  # coords (n, ndim)
+            np.ndarray[tuple[int], np.dtype[np.uint8]],        # kinds (n,)
+            list[str],                                         # paths
+            np.ndarray[tuple[int], np.dtype[np.uint64]],       # offsets (n,)
+            np.ndarray[tuple[int], np.dtype[np.uint64]],       # lengths (n,)
+            dict[int, bytes],                                  # inlined
         ]
     ]: ...
     async def set_virtual_ref_async(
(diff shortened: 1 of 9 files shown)


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -2546,6 +2546,20 @@ class PyStore:
         checksum: str | datetime.datetime | None,
         validate_container: bool,
     ) -> None: ...
+    def array_chunk_iterator(
+        self,
+        array_path: str,
+        batch_size: int,
+    ) -> AsyncIterator[
+        tuple[
+            np.ndarray[Any, np.dtype[np.uint32]],
```
paraseba: isn't this a 2D array?

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -2546,6 +2546,20 @@ class PyStore:
...
+            np.ndarray[Any, np.dtype[np.uint8]],
```
paraseba: why a list[str] here? is it because it's hard to get an ndarray of variable length strings?

icechunk-python/python/icechunk/store.py
```diff
@@ -274,6 +274,47 @@ def set_virtual_ref(
...
+        ]
+    ]:
+        """Async generator of columnar chunk-reference batches for one array.
+
+        Each yielded batch is a 6-tuple::
+
+            (coords, kinds, paths, offsets, lengths, inlined)
+
+            coords:   np.ndarray[uint32, (n, ndim)]  — chunk grid coordinates
+            kinds:    np.ndarray[uint8]              — 1=virtual, 2=native, 3=inline
```
paraseba: we have a ChunkType Enum in this crate (and it's mapped to PY iirc), we should reuse it

icechunk-python/python/icechunk/store.py
```diff
@@ -274,6 +274,47 @@ def set_virtual_ref(
...
+            paths:    list[str]                      — URL (virtual) | bare chunk_id (native) | "" (inline)
+            offsets:  np.ndarray[uint64]
+            lengths:  np.ndarray[uint64]
+            inlined:  dict[int, bytes]               — keyed by index within this batch
```
paraseba: so, all non inline chunks are not present in the dict? Let's clarify in the docstring.

icechunk-python/python/icechunk/store.py
paraseba: I wonder if you need `kinds` or you should just add a prefix to the `paths`: `u:`, `n:`, `i:`. I guess having kinds is cleaner?

icechunk-python/src/store.rs
```diff
@@ -34,6 +40,12 @@ use crate::{
 
 type KeyRanges = Vec<(String, (Option<ChunkOffset>, Option<ChunkOffset>))>;
 
+/// Per-slot tag values for [`PyStore::array_chunk_iterator`] batches.
+/// Kept in sync with the docstring on that method.
+const KIND_VIRTUAL: u8 = 1;
+const KIND_NATIVE: u8 = 2;
+const KIND_INLINE: u8 = 3;
```
paraseba: Let's use `ChunkType`

icechunk-python/src/store.rs
```diff
@@ -606,6 +618,136 @@ impl PyStore {
...
+    fn array_chunk_iterator(
+        &self,
+        array_path: String,
+        batch_size: u32,
+    ) -> PyResult<PyAsyncGenerator> {
+        let store = Arc::clone(&self.0);
+        let res = try_stream! {
+            let session_lock = store.session();
+            let session = session_lock.read_owned().await;
+            let array_path = if array_path.starts_with('/') {
```
paraseba: we may have some function that does this

icechunk-python/python/icechunk/_icechunk_python.pyi
TomNicholas: the rust-numpy crate doesn't yet support numpy's (relatively new) variable-length string dtype 😞

TomNicholas: I say relatively new - it's almost as old as VirtualiZarr is! 🙃

icechunk-python/python/icechunk/store.py
TomNicholas: I went with `kinds` as it seemed easier to use as a mask python-side.

icechunk-python/python/icechunk/store.py
TomNicholas: done

paraseba (approved):

PR #2133, opened 2026-05-15
fix(xarray): defer icechunk.dask import in write_lazy

### LINKED ISSUE #2128: to_icechunk() unconditionally imports icechunk.dask, breaks on numpy-only writes when dask isn't installed
## Summary

``icechunk.xarray.to_icechunk()`` always calls ``writer.write_lazy()`` after ``writer.write_eager()``, and ``write_lazy()`` imports ``icechunk.dask`` (which imports ``dask``) as its **first** statement — before checking whether there is any lazy work to do. As a result, calling ``to_icechunk()`` on a fully numpy-backed Dataset fails with ``ModuleNotFoundError: No module named 'dask'`` in any environment that doesn't have ``dask`` installed, even though the actual write is entirely eager and could complete without ever touching a chunk manager.

This forces every downstream library that uses ``to_icechunk()`` to declare ``dask`` as a hard runtime dependency, even when its actual data is numpy-only — purely so that an unreachable ``session_merge_reduction`` call site can be imported.

## Reproduction

```python
import tempfile
from pathlib import Path

import numpy as np
import xarray as xr
import icechunk
from icechunk.xarray import to_icechunk

with tempfile.TemporaryDirectory() as td:
    storage = icechunk.local_filesystem_storage(str(Path(td) / "repo"))
    repo = icechunk.Repository.create(storage)

    ds = xr.Dataset(
        {"x": (("t",), np.arange(10, dtype=np.float32))},
        coords={"t": np.arange(10)},
    )

    # No dask anywhere; data is plain numpy ndarray:
    assert type(ds["x"].data).__module__ == "numpy"

    to_icechunk(ds, session, group="g")  # raises if dask is not installed
    session.commit("ok")
```

In a clean env without ``dask``:

```
uv run --no-project --with "icechunk==1.1.21" --with xarray --with numpy python repro.py
```

```
  File "/tmp/icechunk_repro.py", line 31, in <module>
    to_icechunk(ds, session, group="g")
  File ".../icechunk/xarray.py", line 340, in to_icechunk
    maybe_fork_session = writer.write_lazy(
  File ".../icechunk/xarray.py", line 187, in write_lazy
    from icechunk.dask import session_merge_reduction
  File ".../icechunk/dask.py", line 8, in <module>
    import dask
ModuleNotFoundError: No module named 'dask'
```

## Root cause

In ``icechunk/xarray.py`` (1.1.21), ``_XarrayDatasetWriter.write_lazy`` is:

```python
def write_lazy(
    self,
    chunkmanager_store_kwargs: MutableMapping[Any, Any] | None = None,
    split_every: int | None = None,
) -> ForkSession | None:
    """
    Write lazy arrays (e.g. dask) to store.
    """
    from icechunk.dask import session_merge_reduction   # ← always runs

    if not self._initialized:
        raise ValueError("Please call `write_metadata` first.")

    if not self.writer.sources:                          # ← the actual lazy-check, too late
        return None
    ...
```

The dask import sits **before** the early-return guard. Re-ordering them, or moving the import inside the post-guard branch, makes ``to_icechunk()`` work on numpy-only Datasets in environments without ``dask`` installed.

## Suggested fix

```python
def write_lazy(
    self,
    chunkmanager_store_kwargs: MutableMapping[Any, Any] | None = None,
    split_every: int | None = None,
) -> ForkSession | None:
    """
    Write lazy arrays (e.g. dask) to store.
    """
    if not self._initialized:
        raise ValueError("Please call `write_metadata` first.")

    if not self.writer.sources:
        return None

    # Only import the dask integration when there is actually lazy work.
    from icechunk.dask import session_merge_reduction

    ...
```

This keeps the dask path available when needed and lets dask-free callers use ``to_icechunk()`` without installing dask.

## Why this matters downstream

Downstream libraries that have moved off Dask but still use ``to_icechunk()`` for numpy-backed datasets currently have to keep ``dask`` pinned as a runtime dependency purely because of this import. At the call site the data is provably numpy-only, the write is fully eager, and yet ``write_lazy`` still fires and still imports ``icechunk.dask``. The only reason it works in production today is that ``dask`` happens to be installed.

## Environment

- icechunk: 1.1.21
- xarray: 2026.4.0
- numpy: 2.4.4
- Python: 3.12.3

Happy to open a PR with the trivial reorder if it's helpful.

dfinlayson-cti:
## Summary

`_XarrayDatasetWriter.write_lazy` imported `icechunk.dask` (which imports `dask`) as its very first statement, before the guards that short-circuit when there is no lazy work. Move the import past both guards so it only fires when we're actually about to call `session_merge_reduction`.

This is the trivial reorder discussed in #2128 — `write_lazy` should not require `dask` to be importable on the early-exit paths.

## Test plan

- [x] Reorder confirmed via local diff (3 lines moved, no logic change).
- [x] Verified `write_lazy` still imports `session_merge_reduction` immediately before the only call site, so the dask path is unchanged.

Closes #2128

### DIFF
--- a/icechunk-python/python/icechunk/xarray.py
+++ b/icechunk-python/python/icechunk/xarray.py
@@ -184,8 +184,6 @@ class _XarrayDatasetWriter:
         """
         Write lazy arrays (e.g. dask) to store.
         """
-        from icechunk.dask import session_merge_reduction
-
         if not self._initialized:
             raise ValueError("Please call `write_metadata` first.")
 
@@ -193,6 +191,8 @@ class _XarrayDatasetWriter:
             raise TypeError(
                 "Attempting to execute a write with dask but no dask arrays were detected."
             )
+
+        from icechunk.dask import session_merge_reduction
 
         chunkmanager_store_kwargs = chunkmanager_store_kwargs or {}
         chunkmanager_store_kwargs["load_stored"] = False


paraseba (approved):

paraseba:
Thank you @dfinlayson-cti !

PR #2143, opened 2026-05-20
Add static request headers support to HTTP storage

### LINKED ISSUE #997: Support basic and bearer auth for Http Storage
### Problem

HTTP Support was added for virtual chunks but there is no way to specify auth currently. 

### Proposed Solution

Add auth credentials to the implementation, for example bearer and basic auth

betolink:
Icechunk's HTTP storage backend had no way to pass authentication or custom headers when accessing protected endpoints (ike NASA data behind EDL auth). This PR adds an optional headers parameter to `HttpConfig` (Rust), http_storage / http_store (Python), and the equivalent JS bindings, allowing users to supply static key-value headers (e.g. `Authorization: Bearer <token>`) that are attached to every request via object_store's `ClientOptions::with_default_headers`. 

Header values are redacted in Display/Debug output to avoid leaking secrets in logs.  A new test (`test_http_storage_with_auth_header`) spins up a local warp server that requires a valid Authorization header, verifying both the success and rejection paths. The change is fully backward-compatible, existing serialized `HttpConfig` configs deserialize without modification.

I reviewed the code as I used Opus 4.6 to implement the changes, first by looking at why headers are not exposed in the underlying http client config https://docs.rs/object_store/latest/object_store/client/enum.ClientConfigKey.html that icechunk uses.  If this PR is considered and merged, it will close #997


mpiannucci:
cc @TomNicholas

icechunk-arrow-object-store/Cargo.toml
```diff
@@ -21,6 +21,7 @@ async-trait.workspace = true
 bytes.workspace = true
 chrono.workspace = true
 futures.workspace = true
+http = { workspace = true, optional = true }
```
TomNicholas: None of the rest of these deps are optional - what's the implication of that?

icechunk-arrow-object-store/src/lib.rs
```diff
@@ -758,23 +764,34 @@ impl ObjectStoreBackend for LocalFileSystemObjectStoreBackend {
 pub struct HttpObjectStoreBackend {
     pub url: String,
     pub config: Option<HashMap<ClientConfigKey, String>>,
+    /// Static HTTP headers injected into every request.
+    /// Values are redacted in `Display` output to avoid leaking credentials in logs.
```
mpiannucci: These comments arent necessary here

icechunk-arrow-object-store/src/lib.rs
```diff
@@ -808,6 +825,31 @@ impl ObjectStoreBackend for HttpObjectStoreBackend {
...
+                let name = HeaderName::from_bytes(k.as_bytes()).map_err(|e| {
+                    other_error(format!("invalid HTTP header name {k:?}: {e}"))
+                })?;
+                let value = HeaderValue::from_str(v).map_err(|e| {
+                    other_error(format!("invalid HTTP header value for {k:?}: {e}"))
+                })?;
+                header_map.insert(name, value);
+            }
+            let mut client_opts = ClientOptions::new().with_default_headers(header_map);
+            if !config.contains_key(&ClientConfigKey::AllowHttp)
```
mpiannucci: This is too limiting, users can specify any number of config currently and you need to make sure that all config passed in (not just allow_http) are propagated

Changelog.python.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Python Icechunk Library 2.0.6
+
+### Features
+
+- Add `headers` parameter to `http_storage` and `http_store` to support static HTTP request headers (e.g. `Authorization`).
+
```
mpiannucci, line 31: This will be done when we cut the release,

icechunk-arrow-object-store/src/lib.rs
```diff
@@ -808,6 +825,31 @@ impl ObjectStoreBackend for HttpObjectStoreBackend {
             builder = builder.with_config(ClientConfigKey::AllowHttp, "true");
         }
 
+        // Inject static headers via ClientOptions so they are sent on every request.
+        // Note: if the URL is plain HTTP we must also carry allow_http=true into the
+        // ClientOptions to avoid overriding the AllowHttp flag set above.
```
mpiannucci: Comment not necessary

icechunk-arrow-object-store/src/lib.rs
```diff
@@ -758,23 +764,34 @@ impl ObjectStoreBackend for LocalFileSystemObjectStoreBackend {
...
+    /// Static HTTP headers injected into every request.
+    /// Values are redacted in `Display` output to avoid leaking credentials in logs.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub headers: Option<HashMap<String, String>>,
 }
 
 #[cfg(feature = "http")]
 impl Display for HttpObjectStoreBackend {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let config_str = self
```
mpiannucci, line 789: Is there a way with object_store to pass in headers to the config directly? Hard for me to tell, would be good to get solid answer about it

mpiannucci (commented):
Some cleanup needed, overall this is super useful knowing it enables EDL

Thanks for contributing

TomNicholas (commented):
Nice! I _think_ that once this is in we can rebase https://github.com/earth-mover/icechunk/pull/2085 on top and that will give us some docs 🙂

icechunk-arrow-object-store/src/lib.rs
```diff
@@ -808,6 +825,31 @@ impl ObjectStoreBackend for HttpObjectStoreBackend {
...
 
+        // Inject static headers via ClientOptions so they are sent on every request.
+        // Note: if the URL is plain HTTP we must also carry allow_http=true into the
+        // ClientOptions to avoid overriding the AllowHttp flag set above.
+        if let Some(hdrs) = &self.headers
+            && !hdrs.is_empty()
+        {
+            let mut header_map = HeaderMap::new();
+            for (k, v) in hdrs {
+                let name = HeaderName::from_bytes(k.as_bytes()).map_err(|e| {
```
paraseba: is there a from_str method?

icechunk/src/config.rs
```diff
@@ -47,7 +68,7 @@ pub enum ObjectStoreConfig {
     #[cfg(feature = "object-store-fs")]
     LocalFileSystem(PathBuf),
     #[cfg(feature = "object-store-http")]
-    Http(HashMap<String, String>),
+    Http(HttpConfig),
```
paraseba, line 71: I think this breaks compatibility. Icechunk optionally stores configuration on disk. Repositories that have Virtual Chunk Containers pointing to an HTTP store will break with this change. I think...

We need to introduce this new feature without breaking stored configurations. This is what I recommend to test: generate two repositories with the existing version of icechunk, write them to the filesystem. Make sure they have HTTP virtual chunk credentials and that you are storing the on disk configuration. Then apply your change and verify the configuration can still be read

paraseba (changes requested):
Blocking because I think this will break compatibility of the on disk format for configuration. But it's very close, we just need to fix that.

Changelog.python.md
betolink: Oh yeah I didn't see an "Unreleased" section. Will work on the rest of the comments ASAP.

icechunk-arrow-object-store/Cargo.toml
betolink: Back at this PR today! I think it should not be optional? since it's already a transitive dependency.

icechunk-arrow-object-store/src/lib.rs
betolink: Refactoring this to pass all the options.

icechunk/src/config.rs
betolink: @paraseba do you think this grants a unit test? testing a round trip with the hashmap to ensure backward compatibility besides doing the full store generation.

paraseba: Yes, we should probably include all types of stores in a new repo in icechunk-python/tests/data

icechunk-arrow-object-store/src/lib.rs
betolink: I don't think so, this comes from the  original issue on why it cannot be done natively. but there was a bug in the previous implementation that erased extra opts if a user had the auth headers enabled.

icechunk-arrow-object-store/src/lib.rs
betolink: comments were removed.

icechunk-arrow-object-store/src/lib.rs
betolink: same as above

icechunk-arrow-object-store/Cargo.toml
betolink: Done!

icechunk/src/config.rs
betolink: There is a couple new repos in the test fixtures and a new set of unit tests for when this new `HttpConfig` is present, I think it covers this case of being able to read old v1 stores. Let me know if this is what you were after.

icechunk-arrow-object-store/src/lib.rs
betolink: Now we accumulate all the options and headers, not just `AllowHttp`

TomNicholas:
@betolink note we merged some [changes](https://github.com/earth-mover/icechunk/pull/2154) from `main` into this branch to make some CI tests pass.

But the failure in the WASM build seems real btw - can you look into that?

betolink:
Just pushed a fix @TomNicholas

icechunk/src/config.rs
```diff
@@ -960,4 +993,103 @@ virtual_chunk_containers:
...
     }
+
+    /// Verify that repositories written by icechunk 2.0.5 (which used
+    /// `ObjectStoreConfig::Http(HashMap<String,String>)`) can still be read
+    /// after this PR changed the type to `Http(HttpConfig)`.
+    /// This test pins that behavior so a serde or dependency
+    /// change cannot silently break existing repositories.
+    #[icechunk_macros::test]
+    #[cfg(feature = "object-store-http")]
+    fn test_http_vcc_backward_compat_yaml() {
```
paraseba, line 1004: great test

paraseba:
This was very close @betolink and it's great. I made some minor changes, and merged  main. The only significant change I made was dropping the stuff related to test_can_read_old. It wasn't exactly right, but I think we have good enough coverage with your other tests.

Thank you very much for the great contribution!

paraseba (approved):

PR #2144, opened 2026-05-20
v1: fix wheel building

### LINKED ISSUE #2064: Linux nightly builds seem to fail for v1 support branch
https://github.com/earth-mover/icechunk/actions/runs/24227233770/job/70730883120

li-em:
Fix #2064

Wheel building on the `support/v1.x` branch was broken, and once that was fixed, several
other CI checks turned out to have rotted since the branch's last green run. This PR
fixes both.

## Wheel building

- Bump the minimum maturin to `>=1.13.1`. Older maturin versions fail to build the
  current source layout.
- Stop pointing `readme` at `../README.md` in `icechunk-python/pyproject.toml` and both
  `Cargo.toml` files. Paths outside the package directory break wheel/crate packaging;
  `icechunk-python` gets its own `README.md` and the `icechunk` crate uses a symlink to
  the repository README.
- Restore the Read the Docs build steps in `.readthedocs.yaml` (the `build.html` command
  had been lost, so RTD had nothing to run).
- Pin all GitHub Actions to SHAs and sync them with `main`; drop the unmaintained
  pre-commit action in favor of running `prek` directly.
- Update `RUST_CHANNEL` to 1.91.1 and refresh Rust dependencies.

## CI repairs

Everything below is drift accumulated while the branch was idle, uncovered as each layer
of CI came back to life:

- **mypy (Code Quality)**: run under Python 3.11 instead of 3.12. With 3.12, uv installs
  the latest zarr (3.2.x), which uses PEP 695 `type` statements that mypy rejects under
  this branch's `python_version = "3.11"`. Under 3.11, uv resolves a zarr without them.
- **check-xarray-docs**: backport the `to_icechunk` `mode` docstring from `main`;
  upstream xarray reformatted the corresponding `to_zarr` docstring.
- **docs**: replace `Group.create_dataset` (removed in current zarr) with
  `Group.create_array` in `version-control.md`. This was failing the RTD build in
  mkdocs strict mode.
- **zarr floor**: align with `main` — `zarr>=3.1.0` in `pyproject.toml` and matching
  minimum matrix pins (zarr 3.1.0, xarray 2025.07.1, dask/distributed 2025.2.0).
  xarray 2025.01.2 is incompatible with zarr 3.1's dtype classes.
- **cargo-deny**: update `anyhow` (RUSTSEC-2026-0190) and `crossbeam-epoch`
  (RUSTSEC-2026-0204). Ignore the quick-xml DoS advisories (RUSTSEC-2026-0194/0195, same
  rationale as `main`: object_store pins quick-xml below the fixed version) and the pyo3
  advisories (RUSTSEC-2026-0176/0177: the fix requires pyo3 0.29, too disruptive for a
  support branch, and neither affected API is used).
- **pytest-mypy-plugins**: pin `<4` in the xarray-backends jobs; 4.x removed the
  `--mypy-only-local-stub` flag that xarray 2025.07.1's pytest config passes.
- **zarr store test suite**: backport `main`'s overrides of
  `test_get_bytes`/`test_get_json` (and sync variants). zarr >=3.1.6 added these to the
  inherited `StoreTests`, writing arbitrary JSON to `zarr.json`, which icechunk rejects
  because it validates metadata on write.
- **test_r2[public-url]**: skipped. The public dataset at `v1/era5_weatherbench2` now
  reports Icechunk v2 format, which this branch cannot open. If a v1-format copy is
  restored at that prefix, the skip can be reverted.


icechunk/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "icechunk"
 version = "0.3.24"
 description = "Transactional storage engine for Zarr designed for use on cloud object storage"
-readme = "../README.md"
+readme = "README.md"
```
ianhi, line 5: will this end up rendering as `../Readme.md` which is the literal content of `icechunk/readme.md`? this is what gets displayed in the cargo/pypi page right?

ianhi (approved):
small q about the readmes. otherwise looks good

icechunk/Cargo.toml
li-em: This one is a symlink now, which works for cargo. I had to copy the Python package README because symlinks don't work properly in windows, and there was a change that PyPI would pick the metadata from the windows package and render it wrong...

I checked both crates.io and pypi.org READMEs match previous versions with the current config

PR #2147, opened 2026-05-22
Design doc for expiration bug

paraseba:


ianhi:
Would an alternative no format change workaround to the size limit be to limit the behavior of GC/Expiration? So we just refuse to compact things beyond the 2 GB size limit and compact to 2 commits instead of 1 when necessary?

design-docs/016-expired-transaction-logs.md
```diff
@@ -0,0 +1,177 @@
...
+contribution to fetch its own transaction log **plus** every id listed
+in its `expired_tx_logs`. Order within a single snapshot's
+contribution follows `expired_tx_logs` order (oldest first), then
+the snapshot's own log last.
+
+#### Amend
+
+Amend fetches `previous_log` and merges it with `this_tx_log` to produce
+the tx_log of the amended snapshot D'. We do **not** absorb
+`expired_tx_logs` into this merge, instead the chain is copied.
```
ianhi, line 107: I don't follow the distinction of absorbed vs copied here.

design-docs/016-expired-transaction-logs.md
```diff
@@ -0,0 +1,177 @@
...
+```flatbuffers
+table SnapshotInfo {
+    id: ObjectId12 (required);
+    parent_offset: int32;
+    flushed_at: uint64;
+    message: string (required);
+    metadata: [MetadataItem];
+
+    // NEW FIELD
+    expired_tx_logs: [ObjectId12];
```
ianhi, line 58: I'm struggling a bit with the name expired here becuase it doesn't fully describe that these aren't expired, they are more so included here. But I also dont have a better name to suggest :/

design-docs/016-expired-transaction-logs.md
```diff
@@ -0,0 +1,177 @@
...
+   solver compares the rebasing session's change set against each commit's
+   transaction log. If between the session's `snapshot_id` and the current
+   branch tip there are now expired-and-gone commits whose transaction logs
+   would have flagged a conflict, those conflicts become invisible.
+
+4. **`inspect::transaction_log_json`**. The tx log file is incomplete.
+
+### Merge transaction logs at expiration time?
+
+- **Flatbuffer size limits.** The buffer cannot exceed ~2 GB. Tx log merges would accumulate until they exceed the limit.
```
ianhi, line 41: Just for a sense of scale how much would I have to commit to exceed this. Also, in theory I can exceed this in a single commit right? What happens then?

ianhi (approved):
Overall this seems like the most elegant long term solution available to us here. Left a few inline comments but the overall shape seems reasonable to me. 

We should think a little about spec versioning here. This is pretty clearly not 2.0 anymore. Do we have a way to track sub versions (additive changes?)


> The launch design ships only the list. Compaction is a follow-up if and
when growth becomes a real problem.

this feels reasonable. But I would ask that we consider the same question for scale that I raised in my prior comment. How near are we to having people hit the 2 GB flatbuffer limit. Could we just do compaction of tx logs and punt this for a few months? Not saying thats better but if we could last until an IC3 was released that could be nice.

design-docs/016-expired-transaction-logs.md
ianhi: I looked into this a bit and I think this might be pretty difficult to hit? Size here scales as

`16 + 4*D` bytes 

with `D` being array dimensionality means even at 5D you would need to have modified 60 million chunks in a commit to hit the limit. Is that reasonably achievable? You can push a little  further with moves that store arbitrary stings and creating/deleting nodes. but still seems like a very difficult to hit limit

design-docs/016-expired-transaction-logs.md
dcherian: It's saying that if the parent snaphost has `expired_tx_logs` we aren't merging the whole thing together in one amend; Instead we merge the "two most recent tx logs" (basically) and keep `expired_tx_logs` as is.

design-docs/016-expired-transaction-logs.md
```diff
@@ -0,0 +1,177 @@
...
+No data is corrupted; reads and writes continue to function.
+
+An old icechunk binary running **GC** against a
+new repo doesn't know about `expired_tx_logs` and will delete
+the transaction log files those ids reference, regressing the fix on
+that repo. We need to be careful and deal with the case where an `expired_tx_log` is not there.
+
+We accept this risk. The fix is shipped as a normal release; users on
+the new release benefit; users mixing old and new icechunk binaries
+have some risk.
```
dcherian, line 157: 👍🏾

design-docs/016-expired-transaction-logs.md
paraseba: yeah, I'll try to think of a better name

design-docs/016-expired-transaction-logs.md
paraseba: Real world users have hit this limit before. It's particularly troublesome for virtual chunks, where adding millions in a single commit is "cheap".

PR #2156, opened 2026-05-28
Account for lost successful responses in conditional PUT 

### LINKED ISSUE #2099: Icechunk does not account for lost successful responses when issuing conditonal PUTs
## Summary  
Icechunk's conditional-PUT logic on S3-compatible backends is **not idempotent under retry**. The AWS SDK's transparent retry layer re-issues the same `If-None-Match: *` (or `If-Match: <etag>`) request on transient errors. If the *original* request actually wrote the object on the server but the response was lost, the retry's precondition correctly fails → icechunk surfaces a fake `NotOnLatestVersion` / `ConflictError`.

  This hits **Tigris** disproportionately because Tigris seems to have highest number of "response interruption" type errors (499 being a major one). Downstream we see this as a steady drip of errors that have the appears of consistency errors caused by concurrency:

  - `branch update conflict: (None) != (Some(1CECHNKREP0F1RSTCMT0))`   
  - `config was updated by other session` with `previous_version=VersionInfo { etag: None, generation: None }`
  - `repo info object was updated after this session started`
  - `Failed to commit, expected parent: Some("1CECHNKREP0F1RSTCMT0"), actual parent: Some(<random>)`
  - `tag already exists, tags are immutable: v1.0`  

I don't believe this is just limited to tigris!

## Fix?

The code that seems to verify my hypothesis is here https://github.com/earth-mover/icechunk/blob/50d99abe/icechunk-s3/src/lib.rs#L415-L448

Potential fixes could be
-  Issue a HEAD request and check the returned eTag matches what we were trying to PUT when we get conditional failures
- Retry but with `If-Match` instead of `If-None-Match`

li-em:
Fix #2099

Design doc: [design-docs/017-conditional-put-lost-response.md](https://github.com/earth-mover/icechunk/pull/2156/changes#diff-2a8484dd22c7c331b057ddbd969eb791f06d5dd91854ba2aa549023607a4eca8)

### DIFF
--- a/icechunk-s3/src/lib.rs
+++ b/icechunk-s3/src/lib.rs
@@ -53,8 +53,9 @@ use icechunk_types::ICResultExt as _;
 use serde::{Deserialize, Serialize};
 use tokio::sync::OnceCell;
 use tokio_util::io::StreamReader;
-use tracing::{error, instrument, trace};
+use tracing::{error, instrument, trace, warn};
 use typed_path::Utf8UnixPath;
+use uuid::Uuid;
 
 /// How object keys are laid out inside the bucket for a given repository.
 ///
@@ -367,6 +368,26 @@ fn stream2stream(
     Box::pin(res)
 }
 
+/// Per-PUT token written into `x-amz-meta-icechunk-write-id` so the read-back
+/// in `resolve_precondition_conflict` can recognise our own retried write.
+const WRITE_ID_METADATA_KEY: &str = "icechunk-write-id";
+
+static CONDITIONAL_WITHOUT_METADATA_WARNED: std::sync::OnceLock<()> =
+    std::sync::OnceLock::new();
+
+fn warn_conditional_without_metadata_once() {
+    CONDITIONAL_WITHOUT_METADATA_WARNED.get_or_init(|| {
+        warn!(
+            "conditional PUT is enabled but `unsafe_use_metadata` is \
+             disabled — lost-response recovery for conditional writes \
+             requires user metadata to stamp write-ids; without it, \
+             transient PUT failures may surface as spurious conflicts \
+             even when the write actually landed. See \
+             icechunk_storage::Settings::unsafe_use_metadata."
+        );
+    });
+}
+
 impl S3Storage {
     /// Build an [`S3Storage`].
     ///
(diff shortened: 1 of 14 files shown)


icechunk-storage/src/readback.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// write-id, or no object at all. All map identically per-caller.
+    NotOurs,
+    /// Our id matched but the object has no usable version identity.
+    MissingVersion,
+}
+
+/// Classify a successful read-back of a found object. A failed read-back is
+/// inconclusive and never reaches here — the backend returns the HEAD error
+/// instead; an absent object is conclusively `NotOurs` at the backend.
+pub fn classify_readback(
```
paraseba: nitpicking: this could be a method in `ReadbackOutcome` impl

paraseba: absolutely ignorable comment, of course

icechunk-storage/src/readback.rs
```diff
@@ -0,0 +1,226 @@
...
+pub fn resolve_precondition(
+    readback: StorageResult<ReadbackOutcome>,
+    key: &str,
+) -> StorageResult<VersionedUpdateResult> {
+    match readback? {
+        ReadbackOutcome::OurWrite(new_version) => {
+            warn!(
+                key,
+                "precondition failed but our write-id is stored; retried PUT, success"
+            );
```
paraseba, line 98: I don't think we should warn here. After all this is a recovered transient issue right?

icechunk-storage/src/readback.rs
```diff
@@ -0,0 +1,226 @@
...
+/// `NoSuchUpload`/404, or an `object_store` generic transport error). Anything
+/// but `OurWrite` propagates `original`; a failed read-back propagates its error.
+pub fn resolve_lost_response(
+    readback: StorageResult<ReadbackOutcome>,
+    key: &str,
+    original: StorageError,
+) -> StorageResult<VersionedUpdateResult> {
+    match readback? {
+        ReadbackOutcome::OurWrite(new_version) => {
+            warn!(
```
paraseba: same, we probably don't want to warn, if anything maybe a trace

icechunk-storage/src/storage.rs
```diff
@@ -233,6 +233,14 @@ pub struct Settings {
     #[serde(default)]
     pub unsafe_use_conditional_create: Option<bool>,
 
+    /// Whether to stamp object writes with user metadata. Defaults to `true`.
```
paraseba:
```suggestion
    /// Whether to stamp objects with user metadata. Defaults to `true`.
```

icechunk-s3/src/lib.rs
```diff
@@ -367,6 +373,66 @@ fn stream2stream(
...
+    }
+    Uuid::new_v4().to_string()
+}
+
+/// Test-only write-id injection. Thread-local, consumed once.
+#[cfg(feature = "test-util")]
+pub mod test_util {
+    use std::cell::RefCell;
+
+    thread_local! {
```
paraseba, line 395: Can you document this mechanism a bit more?

paraseba (approved):
Absolutely awesome, only minor comments

PR #2159, opened 2026-06-01
docs: document default values in config class docstrings

TomNicholas:
## Summary

The reprs for icechunk's config classes render effective defaults dynamically (e.g. `num_snapshot_nodes: 500000 (default)`), but the `__new__` docstrings in `_icechunk_python.pyi` only documented defaults for a subset of classes and were inconsistent in format.

This adds an explicit `Default: <value>` line to every Optional parameter across the config-related classes: `S3Options`, `CompressionConfig`, `CachingConfig`, `ManifestPreloadConfig`, `ManifestSplittingConfig`, `ManifestVirtualChunkLocationCompressionConfig`, `ManifestConfig`, `StorageRetriesSettings`, `StorageTimeoutSettings`, `StorageConcurrencySettings`, `StorageSettings`, `RepoUpdateRetryConfig`, `RepositoryConfig`.

Also corrects a stale value: `S3Options.network_stream_timeout_seconds` was documented as `Default: 60`, but the underlying Rust default is `None` — when unset, the AWS SDK applies its own default.

Values were taken from the `*::default()` / accessor methods in `icechunk/src/config.rs` and `icechunk-storage/src/storage.rs`, matching what the existing reprs render.

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -93,6 +93,8 @@ class S3Options:
         """
         Optional region to use for the storage backend.
 
+        Default: None
+
         Returns
         -------
         str | None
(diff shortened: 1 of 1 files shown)


paraseba (approved):
Thank you @TomNicholas !

PR #2164, opened 2026-06-02
Implement decompression and flatbuffer verification on spawn_blocking tasks

### LINKED ISSUE #2158: `AssetManager::fetch_manifest` performance investigation
## Summary

Calling `AssetManager::fetch_manifest` concurrently from many tasks (84 in our reproducer) is **~25× slower** than an inline implementation that does the exact same work — same `Storage::get_object`, same zstd decompression, same `Manifest::from_buffer` parse — but does **not** go through the `AssetManager` call frame.

Per-fetch throughput collapses from ~120 MB/s in the inline path to ~5 MB/s through `AssetManager`. The cost shows up only under concurrency: sequential calls through `AssetManager` are fast (per-fetch latency matches the inline path).

Tested on icechunk **2.0.4**, AWS EC2 **m7g.16xlarge** in **us-east-1**, fetching 84 manifests (~30 MB decompressed each) from `s3://ems-earthmover-public-0f0978f1/awgubnv4_goes_16/manifests/`.

## Numbers

| Scenario | wall (ms) | p50 / fetch (ms) | per-stream throughput |
|---|---:|---:|---:|
| **SLOW** — 84 concurrent `asset_manager.fetch_manifest_unknown_size` | **9,104** | 6,450 | **~5 MB/s** |
| **FAST** — 84 concurrent inline `storage.get_object` + decompress + parse | **352** | 314 | **~95 MB/s** |
| Sequential `asset_manager.fetch_manifest_unknown_size` (reference) | 28,768 | 332 | ~90 MB/s |

The two concurrent scenarios do logically identical work and end at the same parsed `Manifest`. The fast version is essentially what `AssetManager::fetch_manifest` itself runs on a cache miss, just inlined:

```rust
// FAST — 84× concurrent, ~300 ms wall total
let (mut read, _) = storage.get_object(&settings, &path, None).await?;
let mut compressed = Vec::new();
read.read_to_end(&mut compressed).await?;
// 38-byte icechunk header (magic + impl-name + spec + file-type), then zstd
let mut decoder = zstd::stream::read::Decoder::new(&compressed[38..])?;
let mut decompressed = Vec::new();
std::io::Read::read_to_end(&mut decoder, &mut decompressed)?;
let _ = Manifest::from_buffer(decompressed);
```

vs.

```rust
// SLOW — 84× concurrent, ~8,000 ms wall total
asset_manager.fetch_manifest_unknown_size(&manifest_id).await?;
```

## What is NOT the cause

We patched the vendored icechunk to bisect `AssetManager::fetch_manifest`. Each of the following, individually or cumulatively, leaves the wall time unchanged (still ~7,400–8,500 ms):

- Skipping `manifest_cache.get_value_or_guard_async` (so quick_cache lookup is
  bypassed).
- Skipping `manifest_cache.insert(...)` on the success path.
- Skipping `request_semaphore.acquire()` (default 256 permits, 84 in flight).
- Replacing streaming-decode (`ZstdDecoder` over `AsyncBufRead`) with
  drain-then-sync-decode (`body.collect()` then `zstd::stream::read::Decoder`
  on the `Vec<u8>`).
- Replacing the shared `self.storage: Arc<dyn Storage>` with a freshly
  constructed `new_s3_storage(...)` instance (process-wide, init'd once via
  `OnceCell`, shared across calls).

The bottleneck localizes precisely to "going through `AssetManager::fetch_manifest`'s call frame" but is not any of the obvious internal lines. Remaining (unverified) candidates we couldn't test from outside icechunk:

- Some shared state on `AssetManager` we haven't traced (`repo_cache: RwLock<...>`,
  `manifest_cache_size_warned: AtomicBool`, or cross-cache infrastructure in
  `quick_cache`).
- Tokio scheduling behavior for futures with deeper `.await` nesting (Phase 7
  has one `.await` per task; the slow path has 4–5).

## Reproducer

A self-contained Rust script is attached. It opens a public GOES-16 icechunk repo, walks the snapshot to find the 84 target manifests, then runs:

1. Sequential baseline through `AssetManager` (reference).
2. 84 concurrent `asset_manager.fetch_manifest_unknown_size` calls (SLOW).
3. 84 concurrent inline `storage.get_object` + decompress + parse (FAST).

It prints wall time and per-fetch latency distribution for each, plus a slow/fast ratio.

## Repro source

<details>
<summary><code>Cargo.toml</code></summary>

```toml
[package]
name = "repro-icechunk-manifest-contention"
version = "0.1.0"
edition = "2021"

[dependencies]
icechunk = "2.0.4"
tokio = { version = "1", features = ["full"] }
futures = "0.3"
anyhow = "1"
zstd = "0.13"
```
</details>

<details>
<summary><code>src/main.rs</code> (~300 lines)</summary>

```rust
//! Minimal reproducer: `AssetManager::fetch_manifest` is 25× slower than an
//! inline implementation that does the same work.
//!
//! Two scenarios, both fetching the same 84 manifests:
//!
//!   SLOW: `asset_manager.fetch_manifest_unknown_size(id)` — the icechunk path
//!   FAST: inline `storage.get_object` + zstd decode + `Manifest::from_buffer`
//!
//! Both end at the same logical result (a parsed `Manifest`). The fast version
//! is literally what `AssetManager::fetch_manifest` does on a cache miss, just
//! without going through the `AssetManager` call frame.
//!
//! Required env vars (Arraylake-delegated S3 credentials, ~1h validity):
//!     AWS_ACCESS_KEY_ID
//!     AWS_SECRET_ACCESS_KEY
//!     AWS_SESSION_TOKEN
//!
//! Recommended environment: AWS EC2 `m7g.16xlarge` in us-east-1 (same region
//! as the source bucket).
//!
//! Usage:
//!     cargo run --release -p zax-bench --bin manifest_contention_repro

use std::collections::HashMap;
use std::sync::Arc;
use std::time::{Duration, Instant};

use anyhow::Context;
use futures::future::join_all;
use icechunk::config::{
    CachingConfig, Credentials, S3Credentials, S3Options, S3StaticCredentials,
};
use icechunk::format::manifest::Manifest;
use icechunk::format::ManifestId;
use icechunk::repository::VersionInfo;
use icechunk::virtual_chunks::VirtualChunkContainer;
use icechunk::{new_s3_storage, ObjectStoreConfig, Repository, RepositoryConfig};
use tokio::io::AsyncReadExt;

// --- Configuration ---------------------------------------------------------

const SRC_BUCKET: &str = "ems-earthmover-public-0f0978f1";
const SRC_PREFIX: &str = "awgubnv4_goes_16";
const NOAA_PREFIX: &str = "s3://noaa-goes16/";
const GROUP: &str = "/ABI-L2-MCMIPF/pre-2023-04-19";
const ARRAY: &str = "CMI_C03";

// B9-style slice: t=[0, 200_000), x_chunk=4, y_chunk=4.
// With manifest split (t=2400, x=12, y=12) this covers 84 manifests.
const SLICE_T_HI: u32 = 200_000;
const X_CHUNK: u32 = 4;
const Y_CHUNK: u32 = 4;
const MANIFESTS_PREFIX: &str = "manifests";

// --- Repository construction -----------------------------------------------

async fn open_repo() -> anyhow::Result<Arc<Repository>> {
    let creds = static_creds()?;
    let s3_config = s3_options();
    let storage = new_s3_storage(
        s3_config,
        SRC_BUCKET.to_string(),
        Some(SRC_PREFIX.to_string()),
        Some(S3Credentials::Static(creds)),
    )?;

    let noaa = VirtualChunkContainer::new(
        NOAA_PREFIX.to_string(),
        ObjectStoreConfig::S3(S3Options {
            anonymous: true,
            ..s3_options()
        }),
    )
    .map_err(anyhow::Error::msg)?;

    let mut vcc = HashMap::new();
    vcc.insert(NOAA_PREFIX.to_string(), noaa);

    // Rule out manifest cache eviction as a confounder.
    let caching = CachingConfig {
        num_chunk_refs: Some(100_000_000),
        ..Default::default()
    };
    let config = RepositoryConfig {
        virtual_chunk_containers: Some(vcc),
        caching: Some(caching),
        ..Default::default()
    };

    let mut authorize: HashMap<String, Option<Credentials>> = HashMap::new();
    authorize.insert(
        NOAA_PREFIX.to_string(),
        Some(Credentials::S3(S3Credentials::Anonymous)),
    );

    Ok(Arc::new(Repository::open(Some(config), storage, authorize).await?))
}

fn static_creds() -> anyhow::Result<S3StaticCredentials> {
    Ok(S3StaticCredentials {
        access_key_id: std::env::var("AWS_ACCESS_KEY_ID")
            .context("AWS_ACCESS_KEY_ID not set")?,
        secret_access_key: std::env::var("AWS_SECRET_ACCESS_KEY")
            .context("AWS_SECRET_ACCESS_KEY not set")?,
        session_token: std::env::var("AWS_SESSION_TOKEN").ok(),
        expires_after: None,
    })
}

fn s3_options() -> S3Options {
    S3Options {
        region: Some("us-east-1".to_string()),
        endpoint_url: None,
        anonymous: false,
        allow_http: false,
        force_path_style: false,
        network_stream_timeout_seconds: None,
        requester_pays: false,
    }
}

/// Walk the snapshot to find the manifest IDs that cover the B9 slice.
async fn find_target_manifests(repo: &Arc<Repository>) -> anyhow::Result<Vec<ManifestId>> {
    use icechunk::format::snapshot::NodeData;
    use icechunk::format::Path as IPath;

    let session = repo
        .readonly_session(&VersionInfo::BranchTipRef("main".to_string()))
        .await?;
    let group: IPath = GROUP.try_into()?;
    let nodes: Vec<_> = session.list_nodes(&group).await?.collect::<Result<_, _>>()?;

    for node in &nodes {
        if node.path.to_string() != format!("{GROUP}/{ARRAY}") {
            continue;
        }
        let NodeData::Array { manifests, .. } = &node.node_data else { continue };
        let mut targets = Vec::new();
        for m in manifests {
            let extents: Vec<_> = m.extents.iter().cloned().collect();
            if extents.len() != 3 {
                continue;
            }
            let t = &extents[0];
            let x = &extents[1];
            let y = &extents[2];
            if t.start < SLICE_T_HI && t.end > 0 && x.contains(&X_CHUNK) && y.contains(&Y_CHUNK) {
                targets.push(m.object_id.clone());
            }
        }
        return Ok(targets);
    }
    anyhow::bail!("array {GROUP}/{ARRAY} not found")
}

// --- Scenarios -------------------------------------------------------------

/// SEQUENTIAL baseline: fetch one manifest at a time through `AssetManager`.
/// Establishes per-fetch cost with no concurrency.
async fn run_sequential(
    repo: &Arc<Repository>,
    mids: &[ManifestId],
) -> (Duration, Vec<Duration>) {
    let am = repo.asset_manager().clone();
    let start = Instant::now();
    let mut durs = Vec::with_capacity(mids.len());
    for mid in mids {
        let t0 = Instant::now();
        let _ = am.fetch_manifest_unknown_size(mid).await;
        durs.push(t0.elapsed());
    }
    (start.elapsed(), durs)
}

/// SLOW: 84 concurrent `AssetManager::fetch_manifest_unknown_size` calls.
/// This is the actual icechunk manifest-loading path that a chunk read takes
/// (via `session.get_chunk_ref` → `get_old_chunk` → `fetch_manifest`).
async fn run_slow_concurrent(
    repo: &Arc<Repository>,
    mids: &[ManifestId],
) -> (Duration, Vec<Duration>) {
    let am = repo.asset_manager().clone();
    let start = Instant::now();
    let futs = mids.iter().map(|mid| {
        let am = am.clone();
        let mid = mid.clone();
        async move {
            let t0 = Instant::now();
            let _ = am.fetch_manifest_unknown_size(&mid).await;
            t0.elapsed()
        }
    });
    let durs = join_all(futs).await;
    (start.elapsed(), durs)
}

/// FAST: 84 concurrent inline calls that do the exact same work
/// `AssetManager::fetch_manifest` does on a cache miss, but without going
/// through the `AssetManager` call frame:
///
///   1. `storage.get_object(settings, path, None)`        — same as icechunk
///   2. drain `AsyncBufRead` to a `Vec<u8>`               — same bytes
///   3. zstd-decompress the body (after the 38-byte header)
///   4. `Manifest::from_buffer(decompressed)`             — same parse
async fn run_fast_concurrent(mids: &[ManifestId]) -> anyhow::Result<(Duration, Vec<Duration>)> {
    use icechunk::storage::Storage;
    let storage: Arc<dyn Storage + Send + Sync> = new_s3_storage(
        s3_options(),
        SRC_BUCKET.to_string(),
        Some(SRC_PREFIX.to_string()),
        Some(S3Credentials::Static(static_creds()?)),
    )?;
    let settings = storage.default_settings().await?;

    let paths: Vec<String> = mids
        .iter()
        .map(|m| format!("{MANIFESTS_PREFIX}/{m}"))
        .collect();

    let start = Instant::now();
    let futs = paths.iter().map(|path| {
        let storage = storage.clone();
        let settings = settings.clone();
        let path = path.clone();
        async move {
            let t0 = Instant::now();
            if let Ok((mut read, _)) = storage.get_object(&settings, &path, None).await {
                let mut compressed = Vec::new();
                let _ = read.read_to_end(&mut compressed).await;
                // Skip the 38-byte icechunk header (12 magic + 24 impl-name + 1 spec + 1 file-type),
                // then zstd-decode the rest.
                if compressed.len() > 38 {
                    if let Ok(mut decoder) =
                        zstd::stream::read::Decoder::new(&compressed[38..])
                    {
                        let mut decompressed = Vec::new();
                        let _ = std::io::Read::read_to_end(&mut decoder, &mut decompressed);
                        let _ = Manifest::from_buffer(decompressed);
                    }
                }
            }
            t0.elapsed()
        }
    });
    let durs = join_all(futs).await;
    Ok((start.elapsed(), durs))
}

// --- Reporting -------------------------------------------------------------

fn summary(label: &str, total: Duration, mut durs: Vec<Duration>) {
    durs.sort();
    let to_ms = |d: Duration| d.as_secs_f64() * 1000.0;
    let n = durs.len();
    println!("=== {label} ===");
    println!("  wall:         {:>8.0} ms", to_ms(total));
    println!(
        "  per-fetch ms: min={:>5.0}  p50={:>5.0}  p95={:>5.0}  max={:>5.0}",
        to_ms(durs[0]),
        to_ms(durs[n / 2]),
        to_ms(durs[(n as f64 * 0.95) as usize]),
        to_ms(durs[n - 1]),
    );
    println!();
}

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    eprintln!("Opening repository...");
    let repo = open_repo().await?;
    let mids = find_target_manifests(&repo).await?;
    eprintln!("Found {} target manifests\n", mids.len());

    eprintln!("Scenario 1/3: sequential baseline (via AssetManager)...");
    let (total_seq, durs_seq) = run_sequential(&repo, &mids).await;
    summary(
        &format!("SEQUENTIAL through AssetManager ({} manifests)", mids.len()),
        total_seq,
        durs_seq,
    );

    eprintln!("Scenario 2/3: concurrent SLOW (via AssetManager)...");
    // Open a fresh repo so the manifest cache is cold (matches the sequential
    // baseline's starting state).
    drop(repo);
    let repo = open_repo().await?;
    let (total_slow, durs_slow) = run_slow_concurrent(&repo, &mids).await;
    summary(
        &format!(
            "CONCURRENT {} through asset_manager.fetch_manifest",
            mids.len()
        ),
        total_slow,
        durs_slow,
    );

    eprintln!("Scenario 3/3: concurrent FAST (inline equivalent)...");
    let (total_fast, durs_fast) = run_fast_concurrent(&mids).await?;
    summary(
        &format!(
            "CONCURRENT {} inline (storage + decompress + Manifest::from_buffer)",
            mids.len()
        ),
        total_fast,
        durs_fast,
    );

    let slow_ms = total_slow.as_secs_f64() * 1000.0;
    let fast_ms = total_fast.as_secs_f64() * 1000.0;
    println!("=== Result ===");
    println!("  SLOW (AssetManager) wall:  {slow_ms:.0} ms");
    println!("  FAST (inline)       wall:  {fast_ms:.0} ms");
    println!("  ratio:                     {:.1}×", slow_ms / fast_ms);

    Ok(())
}
```
</details>

<details>
<summary>Get credentials and run</summary>

```bash
eval "$(python3 -c '
import asyncio
from arraylake import Client
c = asyncio.run(Client().aclient._get_s3_delegated_credentials_from_bucket(
    org="earthmover-public", nickname="earthmover"))
print(f"export AWS_ACCESS_KEY_ID={c.aws_access_key_id}")
print(f"export AWS_SECRET_ACCESS_KEY={c.aws_secret_access_key}")
print(f"export AWS_SESSION_TOKEN={c.aws_session_token}")
')"

cargo run --release
```
</details>

### Expected output (m7g.16xlarge, us-east-1)

```
=== SEQUENTIAL through AssetManager (84 manifests) ===
  wall:           28768 ms
  per-fetch ms: min=  194  p50=  332  p95=  473  max=  585

=== CONCURRENT 84 through asset_manager.fetch_manifest ===
  wall:            9104 ms
  per-fetch ms: min=  273  p50= 6450  p95= 8833  max= 9100

=== CONCURRENT 84 inline (storage + decompress + Manifest::from_buffer) ===
  wall:             352 ms
  per-fetch ms: min=  164  p50=  314  p95=  333  max=  347

=== Result ===
  SLOW (AssetManager) wall:   9104 ms
  FAST (inline)       wall:    352 ms
  ratio:                       25.9×
```

li-em:
Fix #2158

While part of #2158 was due to `join_all` behavior (and mostly fixed by using `tokio::task::JoinSet`), there is an issue when decompressing and validating flatbuffers: they are CPU-bound tasks, and being executed in a tokio task, which blocks progress for other futures.

This PR uses a `spawn_blocking` and a semaphore to run CPU-bound tasks without blocking the IO-bound tasks. This still runs on the tokio threadpool, so in the future we can improve even more by making a separate threadpool (and ideally expose it to clients, so they can provide their own)

### DIFF
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -1048,8 +1048,9 @@ fn decode_gate() -> &'static Semaphore {
     &GATE
 }
 
-/// Cap preallocation so a forged size hint can't OOM us.
-const MAX_DECODE_PREALLOC: usize = 64 * 1024 * 1024;
+/// Starting buffer capacity for objects without a known size, to avoid a few
+/// initial reallocations while `read_to_end` grows the buffer.
+const DEFAULT_PREALLOC_HINT: usize = 8192;
 
 /// Sync header validation; returns (spec version, compression). Body at `ICECHUNK_FILE_HEADER_LEN`.
 fn check_header(
@@ -1057,9 +1058,16 @@ fn check_header(
     file_type: FileTypeBin,
 ) -> RepositoryResult<(SpecVersionBin, CompressionAlgorithmBin)> {
     use format_constants::*;
-    if buf.len() < ICECHUNK_FILE_HEADER_LEN
-        || !buf.starts_with(ICECHUNK_FORMAT_MAGIC_BYTES)
-    {
+    if buf.len() < ICECHUNK_FILE_HEADER_LEN {
+        return Err(RepositoryErrorKind::FormatError(
+            IcechunkFormatErrorKind::InvalidIcechunkHeaderSize {
+                found: buf.len(),
+                expected: ICECHUNK_FILE_HEADER_LEN,
+            },
+        ))
+        .capture();
+    }
+    if !buf.starts_with(ICECHUNK_FORMAT_MAGIC_BYTES) {
         return Err(RepositoryErrorKind::FormatError(
             IcechunkFormatErrorKind::InvalidMagicNumbers,
         ))
@@ -1208,9 +1216,7 @@ where
     T: Send + 'static,
     F: FnOnce(SpecVersionBin, Vec<u8>) -> Result<T, IcechunkFormatError> + Send + 'static,
 {
-    // hint can be unvalidated snapshot metadata; cap the preallocation
-    let mut compressed =
-        Vec::with_capacity((compressed_size_hint as usize).min(MAX_DECODE_PREALLOC));
+    let mut compressed = Vec::with_capacity(compressed_size_hint as usize);
     read.read_to_end(&mut compressed).await.capture()?;
 
     // on the async task: fail fast (no permit) and keep span ancestry on header errors
@@ -1322,9 +1328,12 @@ async fn fetch_snapshot(
     let path = format!("{SNAPSHOTS_FILE_PATH}/{snapshot_id}");
     let (read, _) =
         storage.get_object(storage_settings, path.as_str(), None).await.inject()?;
-    fetch_and_decode(read, 0, FileTypeBin::Snapshot, |spec_version, buffer| {
-        deserialize_snapshot(spec_version, buffer).map(Arc::new)
-    })
+    fetch_and_decode(
+        read,
+        DEFAULT_PREALLOC_HINT as u64,
+        FileTypeBin::Snapshot,
+        |spec_version, buffer| deserialize_snapshot(spec_version, buffer).map(Arc::new),
+    )
     .await
 }
 
@@ -1398,9 +1407,14 @@ async fn fetch_transaction_log(
     let _permit = semaphore.acquire().await.capture()?;
     let (read, _) =
         storage.get_object(storage_settings, path.as_str(), None).await.inject()?;
-    fetch_and_decode(read, 0, FileTypeBin::TransactionLog, |spec_version, buffer| {
-        deserialize_transaction_log(spec_version, buffer).map(Arc::new)
-    })
+    fetch_and_decode(
+        read,
+        DEFAULT_PREALLOC_HINT as u64,
+        FileTypeBin::TransactionLog,
+        |spec_version, buffer| {
+            deserialize_transaction_log(spec_version, buffer).map(Arc::new)
+        },
+    )
     .await
 }
 
@@ -1582,7 +1596,7 @@ pub async fn fetch_repo_info_from_path(
         Ok(GetModifiedResult::Modified { data, new_version }) => {
             let repo_info = fetch_and_decode(
                 data,
-                0,
+                DEFAULT_PREALLOC_HINT as u64,
                 FileTypeBin::RepoInfo,
                 |spec_version, buffer| {
                     deserialize_repo_info(spec_version, buffer).map(Arc::new)
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -1025,8 +1025,9 @@ fn binary_file_header(
     // magic numbers
     buffer.extend_from_slice(ICECHUNK_FORMAT_MAGIC_BYTES);
     // implementation name
-    let implementation = format!("{:<24}", &*ICECHUNK_CLIENT_NAME);
-    buffer.extend_from_slice(&implementation.as_bytes()[..24]);
+    let implementation =
+        format!("{:<width$}", &*ICECHUNK_CLIENT_NAME, width = ICECHUNK_IMPL_NAME_LEN);
+    buffer.extend_from_slice(&implementation.as_bytes()[..ICECHUNK_IMPL_NAME_LEN]);
     // spec version
     buffer.push(spec_version as u8);
     buffer.push(file_type as u8);
@@ -1036,6 +1037,9 @@ fn binary_file_header(
 }
 
 /// Caps concurrent CPU decodes so the blocking pool can't oversubscribe the cores.
+/// Absent under shuttle: a process-global semaphore would leak across shuttle
+/// executions, and the gate's core-oversubscription job is meaningless there.
+#[cfg(not(feature = "shuttle"))]
 fn decode_gate() -> &'static Semaphore {
     static GATE: LazyLock<Semaphore> = LazyLock::new(|| {
         let n = std::thread::available_parallelism().map_or(8, |n| n.get());
@@ -1044,29 +1048,24 @@ fn decode_gate() -> &'static Semaphore {
     &GATE
 }
 
-/// Binary header: magic(12) + implementation name(24) + spec(1) + file type(1) + compression(1).
-const FILE_HEADER_LEN: usize = 12 + 24 + 1 + 1 + 1;
-// check_header's buf[36/37/38] offsets assume a 12-byte magic
-const _: () = assert!(format_constants::ICECHUNK_FORMAT_MAGIC_BYTES.len() == 12);
-
 /// Cap preallocation so a forged size hint can't OOM us.
 const MAX_DECODE_PREALLOC: usize = 64 * 1024 * 1024;
 
-/// Sync header validation; returns (spec version, compression). Body at `FILE_HEADER_LEN`.
+/// Sync header validation; returns (spec version, compression). Body at `ICECHUNK_FILE_HEADER_LEN`.
 fn check_header(
     buf: &[u8],
     file_type: FileTypeBin,
 ) -> RepositoryResult<(SpecVersionBin, CompressionAlgorithmBin)> {
-    if buf.len() < FILE_HEADER_LEN
-        || !buf.starts_with(format_constants::ICECHUNK_FORMAT_MAGIC_BYTES)
+    use format_constants::*;
+    if buf.len() < ICECHUNK_FILE_HEADER_LEN
+        || !buf.starts_with(ICECHUNK_FORMAT_MAGIC_BYTES)
     {
         return Err(RepositoryErrorKind::FormatError(
             IcechunkFormatErrorKind::InvalidMagicNumbers,
         ))
         .capture();
     }
-    // 12..36 = impl name (ignored)
-    let spec_version_byte = buf[36];
+    let spec_version_byte = buf[ICECHUNK_SPEC_VERSION_OFFSET];
     let spec_version: SpecVersionBin = spec_version_byte
         .try_into()
         .map_err(|_| {
@@ -1079,7 +1078,7 @@ fn check_header(
         })
         .capture()?;
 
-    let file_type_byte = buf[37];
+    let file_type_byte = buf[ICECHUNK_FILE_TYPE_OFFSET];
     let actual_file_type: FileTypeBin = file_type_byte
         .try_into()
         .map_err(|_| {
@@ -1099,7 +1098,7 @@ fn check_header(
         .capture();
     }
 
-    let compression: CompressionAlgorithmBin = buf[38]
+    let compression: CompressionAlgorithmBin = buf[ICECHUNK_COMPRESSION_OFFSET]
         .try_into()
         .map_err(|_| {
             RepositoryErrorKind::FormatError(
@@ -1218,15 +1217,19 @@ where
     let (spec_version, compression) = check_header(&compressed, file_type)?;
     debug_assert_eq!(compression, CompressionAlgorithmBin::Zstd);
 
+    // gate concurrent CPU decodes; absent under shuttle (see decode_gate)
+    #[cfg(not(feature = "shuttle"))]
     let decode_permit = decode_gate().acquire().await.capture()?;
     // keep error span ancestry on the blocking thread
     let span = tracing::Span::current();
     tokio::task::spawn_blocking(move || -> RepositoryResult<T> {
         let _entered = span.entered();
         // release on decode completion, not on cancelled-future drop
+        #[cfg(not(feature = "shuttle"))]
         let _decode_permit = decode_permit;
         let mut decompressed =
-            zstd::decode_all(&compressed[FILE_HEADER_LEN..]).capture()?;
+            zstd::decode_all(&compressed[format_constants::ICECHUNK_FILE_HEADER_LEN..])
+                .capture()?;
         // trim decode_all's ≤2x slack before it's cached
         decompressed.shrink_to_fit();
         deserialize(spec_version, decompressed).inject()
--- a/icechunk/src/asset_manager.rs
+++ b/icechunk/src/asset_manager.rs
@@ -28,10 +28,7 @@ static RETRYABLE_ERROR: LazyLock<regex::Regex> = LazyLock::new(|| {
     )
     .unwrap()
 });
-use async_compression::{
-    Level,
-    tokio::bufread::{ZstdDecoder, ZstdEncoder},
-};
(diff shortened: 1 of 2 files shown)


icechunk/src/asset_manager.rs
```diff
@@ -1038,66 +1036,69 @@ fn binary_file_header(
...
+fn check_header(
+    buf: &[u8],
     file_type: FileTypeBin,
 ) -> RepositoryResult<(SpecVersionBin, CompressionAlgorithmBin)> {
-    let mut buf = [0; 12];
-    read.read_exact(&mut buf).await.capture()?;
-    // Magic numbers
-    if format_constants::ICECHUNK_FORMAT_MAGIC_BYTES != buf {
+    use format_constants::*;
+    if buf.len() < ICECHUNK_FILE_HEADER_LEN
```
paraseba: Can we separate this first condition into a more precise error? `InvalidIcechunkHeaderSize(usize)` or something

icechunk/src/asset_manager.rs
```diff
@@ -1187,22 +1188,54 @@ async fn fetch_manifest(
...
-    let (spec_version, compression) = check_header(&mut read, file_type).await?;
+    deserialize: F,
+) -> RepositoryResult<T>
+where
+    T: Send + 'static,
+    F: FnOnce(SpecVersionBin, Vec<u8>) -> Result<T, IcechunkFormatError> + Send + 'static,
+{
+    // hint can be unvalidated snapshot metadata; cap the preallocation
+    let mut compressed =
+        Vec::with_capacity((compressed_size_hint as usize).min(MAX_DECODE_PREALLOC));
```
paraseba: This doesn't protect us from OOM as the comment in MAX_DECODE_PREALLOC claims, because read_to_end will expand the buffer. So, I'm not sure we need this protection? Having real protection would be great, but I don't think we need to solve that now.

icechunk/src/asset_manager.rs
```diff
@@ -1289,9 +1322,10 @@ async fn fetch_snapshot(
     let path = format!("{SNAPSHOTS_FILE_PATH}/{snapshot_id}");
     let (read, _) =
         storage.get_object(storage_settings, path.as_str(), None).await.inject()?;
-    let (spec_version, buffer) =
-        check_and_decompress(read, FileTypeBin::Snapshot).await?;
-    deserialize_snapshot(spec_version, buffer).map(Arc::new).inject()
+    fetch_and_decode(read, 0, FileTypeBin::Snapshot, |spec_version, buffer| {
```
paraseba: 0 is probably not a good hint. Maybe something like 4k

icechunk/src/asset_manager.rs
```diff
@@ -1364,9 +1398,10 @@ async fn fetch_transaction_log(
     let _permit = semaphore.acquire().await.capture()?;
     let (read, _) =
         storage.get_object(storage_settings, path.as_str(), None).await.inject()?;
-    let (spec_version, buffer) =
-        check_and_decompress(read, FileTypeBin::TransactionLog).await?;
-    deserialize_transaction_log(spec_version, buffer).map(Arc::new).inject()
+    fetch_and_decode(read, 0, FileTypeBin::TransactionLog, |spec_version, buffer| {
```
paraseba: Maybe 8k as hint?

icechunk/src/asset_manager.rs
```diff
@@ -1545,11 +1580,16 @@ pub async fn fetch_repo_info_from_path(
     debug!("Downloading repo info");
     match storage.get_object_conditional(storage_settings, path, previous_version).await {
         Ok(GetModifiedResult::Modified { data, new_version }) => {
-            let (spec_version, buffer) =
-                check_and_decompress(data, FileTypeBin::RepoInfo).await?;
-            deserialize_repo_info(spec_version, buffer)
-                .map(|ri| Some((Arc::new(ri), new_version)))
-                .inject()
+            let repo_info = fetch_and_decode(
+                data,
```
paraseba, line 1598: maybe 4k as hint

paraseba (approved):

icechunk/src/asset_manager.rs
li-em: I made a constant and use a shared hint of 8k for all three cases

PR #2166, opened 2026-06-02
Bump rustfs compose image to 1.0.0-beta.6

paraseba:


li-em (approved):
Can likely also bump `rc` to `0.1.19` in the `init` container

PR #2168, opened 2026-06-03
Add anonymous credential support for Azure Blob Storage

mpiannucci:
## Summary

Adds anonymous credential support for Azure Blob Storage, mirroring the existing GCS anonymous support. This lets users read **public Azure containers without any credentials**.

Anonymous access maps to the underlying `object_store` `MicrosoftAzureBuilder::with_skip_signature(true)` — the same mechanism the GCS backend uses for `GcsCredentials::Anonymous`.

## Usage

```python
import icechunk as ic

storage = ic.azure_storage(
    account="myaccount",
    container="public-container",
    prefix="data",
    anonymous=True,
)
```

Also available as `ic.azure_anonymous_credentials()` and via `ic.azure_credentials(anonymous=True)`.


paraseba (approved):
can we add an integration test to some public bucket to make sure it works against real Azure?

mpiannucci:
> can we add an integration test to some public bucket to make sure it works against real Azure? 

I currently dont know any stable enough to integration test

I will add an issue for the future tho

Issue #2151, opened 2026-05-26
Bug: `shift_array` with rectilinear chunks


ianhi:
We need change the current behavior when we have rectilinear chunks. Right now you shift array on a non-uniform chunk grid you can commit without error and then end up with an corrupted data (confined to the commit, you can still time travel).

```python
import icechunk as ic
import zarr
from zarr.core.chunk_grids import RectilinearChunkGrid

zarr.config.set({"array.rectilinear_chunks": True})

repo = ic.Repository.create(ic.in_memory_storage())
session = repo.writable_session("main")

root = zarr.group(store=session.store)
arr = root.create_array(
    name="a",
    shape=(3,),
    dtype="int32",
    chunks=RectilinearChunkGrid(chunk_shapes=((1, 2),)),
)
arr[:] = [10, 20, 30]
session.commit("write")

session = repo.writable_session("main")
session.shift_array("/a", [1])
session.commit("shift")

session = repo.readonly_session("main")
arr = zarr.open_array(session.store, path="a")
print(arr[:])  # ValueError: cannot reshape array of size 1 into shape (2,)
```

## Solution

I think we have three options

1. Re-write teh chunk grid. 

in my example it would be transformed to `2,1`. This is metadata only and keeps the shape and overall chunking. This would be my preference. 

2. Reject this shift

easiest, but a real bummer to lose this feature.  

3.  decode and reslice chunks

no longer metadata only, potentially very costly. Not super into this option.

PR #2179, opened 2026-06-04
design doc: rectilinear reindex

ianhi:
Design doc for fixing: https://github.com/earth-mover/icechunk/issues/2151

There were far more options than I considered at first!

All prose is hand written. I used AI only for grammar/spelling and a small amount of making the func signatures.

### DIFF
--- a/design-docs/017-shift-array-rectilinear.md
+++ b/design-docs/017-shift-array-rectilinear.md
@@ -1,0 +1,114 @@
+<!-- markdownlint-disable MD013 -->
+<!-- markdownlint-disable MD012 -->
+# `shift_array` and `reindex_array` with rectilinear chunks
+
+## Motivation
+
+Zarr-python added support for rectilinear chunk grids behind a flag in version 3.2.0. These follow the [rectilinear zarr-extension](https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear), not [ZEP 3](https://zarr.dev/zeps/draft/ZEP0003.html). The plan is to move this support from experimental to stable in the next minor version (<https://github.com/zarr-developers/zarr-python/issues/4025>).
+
+These grids allow for variable chunk sizes along a given dimension. So you can have a 7-element 1-D array with chunk sizes `[1, 1, 2, 3]`.
+
+### What this breaks today
+
+Icechunk is agnostic to chunk size information except for the `shift_array` and `reindex_array` functions which modify the ordering of chunks. If you apply a shift to a rectilinear chunk array today then you can commit it without issue. But when anyone tries to actually read the chunk data they will get an error `cannot reshape array of size 1 into shape (2,)` because we change the mapping to array index to chunk, but do not update the chunk grid that zarr uses to understand what size chunk it should be looking for.
+
+See <https://github.com/earth-mover/icechunk/issues/2151>
+
+
+## Design
+
+### Chunk Size Cycling
+
+The core of the bug is that the declared chunk sizes in the zarr metadata end up out of sync with the on-disk data sizes. So to fix this we need to either refuse to do chunk reindexing on rectilinear arrays, or update the chunk grid as part of the operation.
+
+Disallowing `shift_array` on rectilinear chunks is not a good solution, so we need to decide how to fill in the chunk space of the vacated slots.
+
+#### Shift Array
+
+For a chunk grid of `[1, 1, 2, 3]` shifted right by 1 chunk, there are four options that preserve the array shape.
+
+1. Periodic boundary - `[3, 1, 1, 2]`
+2. One fill chunk that takes up all empty space `[3, 1, 1, 2]`
+3. Fill with size 1 - `[1,1,1,1,1,2]`
+4. Rechunk to include the empty part in the existing chunk. So the first chunk would become (3+1) `4,1,2`
+
+
+Option 3 will not work because a chunk with size 1 is far too small for real-world applications. Option 4 has the potential to be computationally intensive as it might end up rechunking real data to combine it with empty chunks.
+
+That leaves Options 1 and 2, which are always equivalent for a shift of magnitude 1. They differ with a shift of magnitude 2 or greater. For a shift by 2 with our example grid of `[1, 1, 2, 3]` they give:
+
+1. `[2, 3, 1, 1]`
+2. `[5, 1, 1]`
+
+
+For any periodic data, like chunks of months, Option 1 is more sensible than creating a large empty chunk that may not fit well with writing workflows.
+
+That leaves Option 1, which does have the downside that it may be slightly confusing that chunk sizes are shifted around the periodic boundary while values are **not** shifted to the other side of the boundary. However, it seems to be the only strategy that gives generally usable results.
+
+**Decision:** periodic boundary for chunk sizes but not chunk values.
+
+#### reindex_array
+
+`reindex_array` is a lower-level, much more powerful tool than `shift_array`. It allows a user to arbitrarily reorder chunks in an array. Since a user can define both forward and optionally backward mappings, and not every chunk needs to be mapped somewhere new, there is no easy general solution to how to restructure the chunk grid shapes. So `reindex_array` needs to gain a way for the user to indicate how to transform the chunk grid. We also need to consider if we want any validation of the new chunk grid as the user can very easily construct a broken grid that will prevent data reads.
+
+##### New Grid
+
+The simplest approach here is to have the user provide a function that determines the chunk size of the vacated chunk.
+
+1. Change signature of `forward`/`backward` to allow this.
+
+   Today:
+
+   ```python
+   forward(idx: list[int])  -> list[int] | None
+   backward(idx: list[int]) -> list[int] | None
+   ```
+
+   Proposed:
+
+   ```python
+   forward(idx: list[int])  -> tuple[list[int] | None, list[int] | None]
+   backward(idx: list[int]) -> tuple[list[int] | None, list[int] | None]
+   ```
+
+   `forward(idx)` returns `(new_idx_or_None, chunk_shape_at_idx)` where `chunk_shape_at_idx` is the chunk shape at source position `idx`. `chunk_shape_at_idx` can also be `None` to indicate to leave the shape the same as a convenience for regular grids.
+2. Require two extra functions for chunk shapes (one per grid direction), leaving `forward`/`backward` unchanged.
+
+   ```python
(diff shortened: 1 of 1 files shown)


design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+`shift_array` is a useful feature, as are rectilinear chunks. They also play well in combination as an exemplar use case of rectilinear chunks is months of the year havign different shapes. So for something like weather predictions over a year we might expect rectilinear chunks. Predictions rolling over a year is a also a good use case of `shift_array`. Therefore simply disallowing `shift_array` on rectilinear chunks is not a good solution. We need to decide on the strategy how to correctly and consistently fill in the chunk sizes of the vacated slots.
+
+#### Shift Array
+
+For a chunk grid of `[1, 1, 2, 3]` shifted right by 1 chunk, there are four options that preserve the array shape.
+
+1. Periodic boundary - `[3, 1, 1, 2]`
+2. One fill chunk that takes up all empty space `[3, 1, 1, 2]`
```
paraseba: in would be most helpful to start with a shift > 1 to see the difference between 1 and 2.

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+3. Fill with size 1 - `[1,1,1,1,1,2]`
+4. Rechunk to include the empty part in the existing chunk. So the first chunk would become (3+1) `4,1,2`
+
+
+Option 3 will not work because a chunk with size 1 is far too small for real-world applications. Option 4 has the potential to be computationally intensive as it might end up rechunking real data to combine it with empty chunks.
```
paraseba: I think rejecting option 3 for this reason is right, but only because Zarr doesn't allow to modify the grid. Icechunk could allow you. Right?

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+
+That leaves Options 1 and 2, which are always equivalent for a shift of magnitude 1. They differ with a shift of magnitude 2 or greater. For a shift by 2 with our example grid of `[1, 1, 2, 3]` they give:
```
paraseba: let's remind the reader that the shift is in chunk space.

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+
+1. `[2, 3, 1, 1]`
+2. `[5, 1, 1]`
+
+
+For any periodic data, like chunks of months, Option 1 is more sensible than creating a large empty chunk that may not fit well with writing workflows.
+
+That leaves Option 1, which does have the downside that it may be slightly confusing that chunk sizes are shifted around the periodic boundary while values are **not** shifted to the other side of the boundary. However, it seems to be the only strategy that gives generally usable results.
```
paraseba: How about the lowest level option: ask the user, "hey, after shifting I have to fill holes of size `[Nx, Ny, ...]` array elements,  how do you want to chunk those holes?". Here `Ni` is the sum of the chunk sizes shifted in that dimension . This could be, for example, in the shape of a function argument:

```
  // the sizes across one dimension of the chunks that overflow
type DimensionOverFlow = Vec<NonZero<usize>>

fn shift_array(....,  rechunk_hole: &[DimensionOverflow] -> Vec<DimensionOverflow>) 
```

`rechunk_hole` indicates, for each dimension the new grid for the holes.

Of course, we need to place some constraints and checks on the returned values (input and output size must be same size, sum of elements for each dim must be equal, etc.

Then we can offer "pre-made" functions that implement rotation, single chunk, `fixed_grid_of_size(n)`, etc.

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+   forward_chunk_shape(idx: list[int])  -> list[int]
+   backward_chunk_shape(idx: list[int]) -> list[int]
+   ```
+
+   `forward_chunk_shape(idx)` returns the chunk shape at `idx`.
+   `backward_chunk_shape(idx)` returns the chunk shape at location that was mapped to `idx`.
+
+Option 2 splits the computation of chunk grid and new location and would require 2 calls across the Rust/Python boundary for each chunk in the array.
+
+So Option 1 is cleaner and easier for the user. However, it has the downside that it is an API change to `reindex_array`. This is an acceptable risk because there are likely very few users of `reindex_array` today.
```
paraseba: Is there an option here to not break the API by allowing the function return 

```
list[int] | 
tuple[list[int] | None, list[int] | None] |
None
```
?

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+
+Without any validation it would be possible for a `reindex_array` user to corrupt their array by writing invalid chunk sizes. Options
+
+1. Accept this risk and warn in documentation.
+   - Users of `reindex_array` are using an advanced tool and have both the power and responsibility that comes with that.
+2. Perform basic validation
+   - Per-axis rectilinear consistency (chunks sharing an axis index agree on that axis's size)
+   - Per-axis sums equal dimension length
+
+Option 2, doing validation, has a lot of potential for complexity and we may accidentally miss some edge cases. So an incomplete validation is likely worse than none at all as it would give false confidence. Given that users of `reindex_array` are doing an advanced operation it is acceptable for them to accept the risk of writing an incorrect chunk grid. I.e. they assume the responsibility for carefully validating their final chunk grid. Additionally this is safe because they can always roll back the operation using Icechunk time travel.
```
paraseba, line 253: I think I  disagree here:
* Are there really many edge cases?
* If it that's so, I think an effort at validation is better than no validation. We don't need to advertise full validation exists.

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+### Array Metadata implementation location
+
+Any fix for the chunk grid necessarily will need access to `ArrayMetadata` and related structs which currently live in `store.rs`. However `shift_array` and `reindex_array` are in `session.rs`. So to fix the chunk grid issues we will need to import from store into session which inverts their current import relationship. While Rust can handle this, it's a sign that a new abstraction around metadata would be helpful.
+
+Options:
+
+1. Accept things as they are and just import both directions.
```
paraseba, line 277: 👎

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+2. Perform basic validation
+   - Per-axis rectilinear consistency (chunks sharing an axis index agree on that axis's size)
+   - Per-axis sums equal dimension length
+
+Option 2, doing validation, has a lot of potential for complexity and we may accidentally miss some edge cases. So an incomplete validation is likely worse than none at all as it would give false confidence. Given that users of `reindex_array` are doing an advanced operation it is acceptable for them to accept the risk of writing an incorrect chunk grid. I.e. they assume the responsibility for carefully validating their final chunk grid. Additionally this is safe because they can always roll back the operation using Icechunk time travel.
+
+
+### Array Metadata implementation location
+
+Any fix for the chunk grid necessarily will need access to `ArrayMetadata` and related structs which currently live in `store.rs`. However `shift_array` and `reindex_array` are in `session.rs`. So to fix the chunk grid issues we will need to import from store into session which inverts their current import relationship. While Rust can handle this, it's a sign that a new abstraction around metadata would be helpful.
```
paraseba, line 273: Can we make the session functions lower level and have store call them, and _then_ update the metadata based on output from those functions?

design-docs/017-shift-array-rectilinear.md
```diff
@@ -0,0 +1,115 @@
...
+   ```
+
+   `forward(idx)` returns `(new_idx_or_None, chunk_shape_at_idx)` where `chunk_shape_at_idx` is the chunk shape at source position `idx`. `chunk_shape_at_idx` can also be `None` to indicate to leave the shape the same as a convenience for regular grids.
+2. Require two extra functions for chunk shapes (one per grid direction), leaving `forward`/`backward` unchanged.
+
+   ```python
+   forward(idx: list[int])              -> list[int] | None
+   backward(idx: list[int])             -> list[int] | None
+   forward_chunk_shape(idx: list[int])  -> list[int]
+   backward_chunk_shape(idx: list[int]) -> list[int]
```
paraseba: Help me think about this, do we also need two functions here? Can we have only `regrid_vacuumed_chunk` that is called on every chunk that was cleared, doesn't matter if it was cleared because of forward or backward?

design-docs/017-shift-array-rectilinear.md
ianhi: > could allow you

Could, but does not currently. Would be a ncie feature to be able to re-chunk empty chunks as that would be a very cheap metadata op.

design-docs/017-shift-array-rectilinear.md
ianhi: ah very good point. and I like the term "vacuumed"

design-docs/017-shift-array-rectilinear.md
ianhi: yes that is possible, just kinda drifts farther into one python's most menacing features. variable return sizing. but I guess explicitly requiring a tuple or even a custom object could get around that. Might be the cleanest. Though I do think (no evidence) that reindex_array has zero real world user today.

design-docs/017-shift-array-rectilinear.md
ianhi: this comment sent me into a deep rabbit hole. very impactful :) there's actually several restrictions about what chunks we even need user input for. in many cases the chunk size si fully constrained by other chunks. putting in draft till i can sort it all out a bit better.

ianhi:
Ok after considering the impact of vacuuming I spent awhile considering how to state the problem more precisely and it turns out there are both more constraints and more ambiguities than I had considered at first.

I've re-written a big portion to better state the problem. Be curious to have your take on re-reading the first few sections @paraseba prior to pushing forward on the API design. I want to make sure I fully captured the actual problem space first.

ianhi:
Another thing I hadn't thought that I suspect is also broken. When you have a regular chunk grid where the array elements aren't evenly divisible by the number of chunks the final chunk along an axis is truncated. I'd be that that is broken today as well

---

edit: tried this out, zarr-python ends up handling this by figuring out the fall off behavior for the final chunk correctly. but maybe that is somehow in conflict with the other semantics of shift. this will require more thought

PR #2184, opened 2026-06-05
Track pruned ancestor transaction logs across expiration

paraseba:
Expiration re-parents a branch's oldest surviving snapshot to a new root and drops the expired ancestors, but the edited snapshot's tx log still only describes the delta from its original (now-deleted) parent. Consumers that fold tx logs along the RepoInfo ancestry (diff, amend, rebase, inspect) then silently produce incomplete results.

Add an additive pruned_ancestor_tx_logs field to SnapshotInfo recording the tx-log ids dropped from a snapshot's path, populate it at expiration, and have GC retain and the consumers fold those logs back in.

Add cross-version (1.1.21 / 2.0.5 / current) post-expiration can_read_old style fixtures.

See design-docs/016-expired-transaction-logs.md for the full design.


icechunk/src/inspect.rs
```diff
@@ -442,6 +444,18 @@ struct UpdatedChunksInspect {
...
+/// Marks a [`TransactionLogInspect`] that is not a single on-disk log but a
+/// synthetic merge, because the snapshot's ancestry was collapsed by expiration.
+#[derive(Debug, Serialize, Deserialize)]
+struct SyntheticCompositeInspect {
+    note: String,
+    /// Pruned-ancestor transaction logs merged into this view, oldest first.
+    merged_pruned_ancestor_tx_logs: Vec<String>,
+    /// Referenced pruned-ancestor logs absent from storage, so skipped. Expected
+    /// only when an older GC (predating pruned-log retention) deleted them.
+    missing_tx_logs: Vec<String>,
```
mpiannucci, line 456: Good solution 👍

mpiannucci (approved):
This is a good solution to a complex problem, great work thinking this all through

icechunk/src/inspect.rs
```diff
@@ -442,6 +444,18 @@ struct UpdatedChunksInspect {
     chunks: Vec<Vec<u32>>,
 }
 
+/// Marks a [`TransactionLogInspect`] that is not a single on-disk log but a
+/// synthetic merge, because the snapshot's ancestry was collapsed by expiration.
+#[derive(Debug, Serialize, Deserialize)]
+struct SyntheticCompositeInspect {
+    note: String,
```
dcherian, line 451: what is `note`?

icechunk/src/inspect.rs
```diff
@@ -452,8 +466,11 @@ struct TransactionLogInspect {
     updated_arrays: Vec<String>,
     updated_chunks: Vec<UpdatedChunksInspect>,
     moved_nodes: Vec<MoveInspect>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
dcherian, line 469: Is this a back-compat thing?

icechunk/tests/test_gc.rs
```diff
@@ -552,13 +552,640 @@ pub async fn test_expire_and_garbage_collect_deleting_expired_refs()
         garbage_collect(Arc::clone(&asset_manager), &gc_config, None, 100).await?;
 
     assert_eq!(summary.snapshots_deleted, 7);
-    assert_eq!(summary.transaction_logs_deleted, 7);
+    // The 7 released snapshots' files are deleted, but their transaction logs
+    // are all retained: each is referenced by some edited snapshot's
+    // pruned_ancestor_tx_logs
+    assert_eq!(summary.transaction_logs_deleted, 0);
```
dcherian, line 558: testing-wise, it would be nice if this was at least `1`

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -1204,26 +1263,65 @@ def check_file_invariants(self) -> None:
...
+            # is still in the repo info. `inspect_transaction_log` surfaces, per
+            # such snapshot, any referenced pruned logs that are wrongly absent.
+            missing_pruned: set[str] = set()
+            for sid in snapshots:
+                try:
+                    tx_log = self.repo.inspect_transaction_log(sid)
+                except IcechunkError:
+                    # A snapshot with no own tx log (e.g. the initial commit of
+                    # a pre-upgrade V1 repo) carries no pruned logs.
+                    continue
```
dcherian, line 1304: Just to be sure of when we hit this error 

```suggestion
                    assert self.model.initial_spec_version == 1
                    assert self.mode.migrated
                    continue
```

dcherian (approved):
LGTM. Was more elegant than expected :)

icechunk/src/inspect.rs
paraseba: We want the JSON to show you an explanation of what the hell is going on, this is the easiest way I found :shrug:

icechunk/src/inspect.rs
paraseba: Not really no, this type only exists to generate some nice looking json. I put this because almost every tx log is not a composite, so I don't want to bother the poor user with this information, better for them to not even see this field.

icechunk/tests/test_gc.rs
paraseba: yes, but I added some other test that has the "mixed" case: test_gc_deletes_only_unreferenced_expired_tx_logs

icechunk/src/repository.rs
```diff
@@ -1828,28 +1829,60 @@ impl Repository {
...
+                                    StorageErrorKind::ObjectNotFound,
+                                ),
+                            ..
+                        }) if is_pruned => {
+                            tracing::warn!(
+                                snapshot_id = %id,
+                                "Transaction log of an expiration-pruned ancestor is \
+                                 missing (likely deleted by an older Icechunk GC); the \
+                                 diff will be incomplete"
+                            );
```
ianhi, line 1870: "a/the" transaction log?

I wonder if it's worth linking to a docs page/issue or maybe a design doc here. Otherwise this (and equivalent errors) might leave a user feeling a bit lost.

A little more actionable and clear might be somethign like:


"A Transaction log of an expiration-pruned ancestor is missing, likely deleted by an Icechunk older than 2.1.0 running garbage collection. This diff will be incomplete. Upgrade all clients that run garbage_collect to Icechunk 2.1.0 or  later. Already-deleted logs cannot be recovered."

icechunk-format/flatbuffers/repo.fbs
```diff
@@ -33,6 +33,12 @@ table SnapshotInfo {
 
     // Snapshot-level metadata.
     metadata: [MetadataItem];
+
+    // Tx-log ids of ancestor commits expiration removed from the repo,
```
ianhi:
```suggestion
    // Tx-log ids of ancestor commits that the expiration process removed from the repo,
```

sligthly more explicit

icechunk-python/tests/data_generation/generate_expired_repos.py
```diff
@@ -0,0 +1,320 @@
...
+        ("expire-repo-v2-by-2.0.5", icechunk_v2, 2, False),
+        ("expire-repo-v1-by-2.0.5-with-ancestor-tx-log-tracking", ic_wc, 1, True),
+        ("expire-repo-v2-by-2.0.5-with-ancestor-tx-log-tracking", ic_wc, 2, False),
+    ]
+
+    for dirname, module, spec_version, is_v1 in fixtures:
+        path = str(data_dir / dirname)
+        if module is None:
+            print(f"SKIP {dirname}: writer module not importable")
+            continue
```
ianhi: I think we should fail here? e.g. if we yank 2.0.5 from pypi accidentally and then for some reason regen the data files it would be silently skipped and not actually updated.

icechunk-python/tests/data_generation/generate_expired_repos.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def build_expired_repo(ic: Any, path: str, spec_version: int | None) -> None:
+    """Build the shared DAG and run two rounds of expiration + GC.
+
+    DAG (main line, oldest first)::
+
+        init -> root -> a -> b -> c -> d -> e -> g -> h -> i -> j(tip)
+                             ^tag      ^feat ^doomed1   ^doomed2
+
```
ianhi: Can we use the actual Icechunk ancestry graph here? Or alternatively I used claude to construct this list I found a little bit clearer

```
  init → root → a → b → c → d → e → g → h → i → j (main)
                    │   │   │           │
                    │   │   │           └─ doomed2_work   (doomed2 — deleted in round 2)
                    │   │   │
                    │   │   └─ doomed1_work               (doomed1 — deleted in round 1)
                    │   │
                    │   └─ feat_pre → feat_s1 → feat_s2   (feature)
                    │
                    └─ (tag) protect-b
```

icechunk-python/tests/data_generation/generate_expired_repos.py
```diff
@@ -0,0 +1,320 @@
...
+        return __import__(name)
+    except ImportError:
+        return None
+
+
+def describe(ic_wc: Any, dirname: str, path: str) -> None:
+    """Print the observed post-expiration state, read by the working copy."""
+    repo = ic_wc.Repository.open(storage=ic_wc.local_filesystem_storage(path))
+    main_tip = repo.lookup_branch("main")
+    state = read_state(ic_wc, path)
```
ianhi, line 302: should we have some asserts in here to make sure we generated the values correctly?

icechunk-python/tests/test_can_read_old.py
```diff
@@ -444,6 +445,137 @@ async def test_icechunk_can_read_old_repo_with_manifest_splitting_v2_migrated()
...
+
+async def do_read_expired_repo(path: str, *, tracks_tx_logs: bool) -> None:
+    """Read a repo that went through two rounds of expiration + GC.
+
+    ``tracks_tx_logs`` selects the expectation: the v2 fixture written by a
+    library that populates ``pruned_ancestor_tx_logs`` keeps its expired-ancestor
+    logs (so the boundary commit's transaction log inspects as a complete
+    synthetic composite and a root-to-tip diff is complete); every other fixture
+    lost those logs to GC (degraded but non-crashing).
+    """
```
ianhi, line 474: I think for completeness we need a version of the test that has 2.0.5 read a properly expired repo by the working version. Otherwise we aren't actually testing the forward compat. We need this to confirm our claim of "additive" in terms of the spec.

ianhi: This might also go into the generate script cross_cehck check

icechunk-python/tests/test_can_read_old.py
```diff
@@ -444,6 +445,137 @@ async def test_icechunk_can_read_old_repo_with_manifest_splitting_v2_migrated()
...
+        "./tests/data/expire-repo-v1-by-2.0.5-with-ancestor-tx-log-tracking",
+        tracks_tx_logs=False,
+    )
+
+
+@pytest.mark.filterwarnings("ignore:datetime.datetime.utcnow")
+async def test_read_expired_repo_v2_by_working_copy() -> None:
+    await do_read_expired_repo(
+        "./tests/data/expire-repo-v2-by-2.0.5-with-ancestor-tx-log-tracking",
+        tracks_tx_logs=True,
```
ianhi, line 575: confused by the naming here. if working copy wrote this and there is tracking why is 2.0.5 in the name?

icechunk/src/inspect.rs
ianhi: can we document this in the python docstring so people know what it is/that it exists https://github.com/earth-mover/icechunk/blob/7151b6e5bb6d05cce16c3ee60cfabd3c2b0ed82b/icechunk-python/python/icechunk/repository.py#L2056-L2059

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -462,6 +472,27 @@ def expire_snapshots(
             if id not in ref_pointees:
                 self.commits.pop(id, None)
 
+        # A retained snapshot whose parent was released is
+        # re-parented to the root, dropping every ancestor between it and the
```
ianhi, line 476: root here means init snapshot right? That's inconsistent with the generated repos where we use init and then `root` as the first user commit. 
```suggestion
        # re-parented to the initial snapshot, dropping every ancestor between it and the
```

ianhi: or we should change in the generate_expired_script.

or just leave it. Idk how often someone will be reading these two files back to back like I am right now.

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -519,12 +550,37 @@ def garbage_collect(
...
+                anc: str | None = c.parent_id
+                while anc is not None and anc in deleted:
+                    collapsed.append(anc)
+                    ancestor = self.ondisk_snaps.get(anc)
+                    anc = ancestor.parent_id if ancestor is not None else None
+                harvested: list[str] = []
+                for x in reversed(collapsed):
+                    harvested.extend(self.pruned_ancestor_tx_logs.get(x, []))
+                    harvested.append(x)
+                self.pruned_ancestor_tx_logs[cid] = harvested
```
ianhi, line 578: I think there is a very subtle possible future bug here. 

 Specifically we are both reading and writing to this dict in a loop. we might suffer from the comment simplified reparenting we perform right below this (outside the diff)

```python
            # V2's delete_snapshots_from_repo_info rewrites parent pointers
            # for kept snapshots whose parent was GC'd.
            for c in self.commits.values():
                if c.parent_id is not None and c.parent_id in deleted:
                    c.parent_id = self.initial_snapshot_id
```

which reparents to the intial snapshot always.

when the actual rust code doing the GC reparents to the nearest ancestor. which you commented in gc.rs

```
   Then:
///      - `reset_branch`("main", snap1) -> main now points at snap1; snap[2..5] are unreachable.
///      - GC with cutoff before = `snap3.flushed_at`.
///      - snap3, 4, and 5 are too new for GC so they are retained
///      - snap0 and 1 are pointed, so they are retained too
///      - snap2 is garbage collected (old and unreachable)
///      - We are left with a dropped interior gap with a kept ancestor below it:
///        snap1  (kept: reachable)
///        snap2  (DROPPED)
///        snap3  (kept: too new)
///      - In this case the function will assign snap 1 as parent of snap 3
///
```

Actually getting there is a little bit baroque but technically possible

so we can end up with a main branch reset to snap 5 giving us: `snap1, snap3, snap4, snap5` 

then tag snap3 

now if we run expire again at the start we will have 
`pruned_ancestor_tx_logs['snap3'] = 2`

then in our first loop we will set 
`pruned_ancestor_tx_logs['snap3'] = ['snap1']`

and then when computing the pruned list for snap5 we will ultimately not have snap2 in there.

The solution is to use a frozen `self.pruned_ancestor_tx_logs`

also I think we need to fix the modeling behavior of reparenting to the inti snapshot only as otherwise we can't catch the bug I mention in my overall comment

--- 

sorry if this comment is a little hard to follow, i've thoroughly cooked my brain trying to think this through.

ianhi (commented):
I think I have found a niche bug in the new pruned collection logic

When you have a snapshot that has `pruned_ancestor_tx_logs` populated and has an ancestor that gets expired then the pruned log will be overwritten and not contain the originally filled values.

I'm pretty sure that the fix is to access the snapshot's current log and append it to the `pruned` here: https://github.com/earth-mover/icechunk/blob/001bb869215ea508aaf197830449927c71e2c52c/icechunk/src/ops/gc.rs#L978-L982

It's wild to get here but you can technically do it.

Given a repo with this commit sequence where I mark possible expire/gc points with parens

`init, (1) A, (2) B, (3) C (4)`
`reset_branch('main', 'A')`
`garbage_collect(3)`           - `pruned_ancestor_tx_logs['C'] = 'B'`
`reset_branch('main', 'C')`
`expire(2)` - `pruned_ancestor_tx_logs['C'] = 'A'`

will result in the incorrect `pruned_ancestor_tx_logs` for `C` of only `A` not `[A, B]`

Running the demo script below gives

```
plain expire:    C.pruned_ancestor_tx_logs ['A', 'B']  diff(init → C) [[0], [1], [2]]
------------
after reset+GC:  C.pruned_ancestor_tx_logs ['B']     diff(init → C) [[0], [1], [2]]
then re-expire:  C.pruned_ancestor_tx_logs ['A']     diff(init → C) [[0], [2]]
```

<details>

<summary>repro script</summary>

```python
"""Expire/GC bug: re-expiring a snapshot overwrites its pruned_ancestor_tx_logs
instead of accumulating, so a diff silently loses an intermediate commit.

All three lines build the same history (init → A → B → C, each commit writes one
chunk of `data`) and end with C reparented onto the root, so C.pruned_ancestor_tx_logs
*should* be [A, B] and diff(init → C) *should* be [[0], [1], [2]] every time.

  1. plain expire        — correct:  C.pruned = [A, B],  diff = [[0],[1],[2]]
  2. reset + GC          — C ends up with parent A and pruned = [B]
  3. then re-expire       — BUG:      C.pruned = [A]    (B dropped),  diff = [[0],[2]]
"""

import time
from datetime import UTC, datetime

import zarr

import icechunk as ic


def write_and_commit(repo, row, value, message):
    zarr.open_array(session.store, path="group1/data", mode="a")[row] = value
    return session.commit(message)


def moment():
    """A timestamp strictly between the commits surrounding the call."""
    time.sleep(0.05)
    t = datetime.now(UTC)
    time.sleep(0.05)
    return t


def fresh_repo():
    """A new repo at init → A → B (A writes data[0], B writes data[1])."""
    config = ic.RepositoryConfig.default()
    config.inline_chunk_threshold_bytes = 0  # one file per chunk
    repo = ic.Repository.create(
        storage=ic.in_memory_storage(), config=config, spec_version=2
    )
    init = repo.lookup_branch("main")
    zarr.group(store=session.store, overwrite=True).create_group("group1").create_array(
        "data", shape=(4,), chunks=(1,), dtype="int32", fill_value=0
    )[0] = 10
    A = session.commit("A: write data[0]")
    B = write_and_commit(repo, 1, 20, "B: write data[1]")
    return repo, init, A, B


def report(tag, repo, labels):
    history = list(repo.ancestry(branch="main"))
    init, C = history[-1].id, history[0].id  # the root and main's tip
    composite = repo.inspect_transaction_log(C).get("synthetic_composite")
    pruned = (
        [labels.get(i, i) for i in composite["merged_pruned_ancestor_tx_logs"]]
        if composite
        else None
    )
    chunks = sorted(
        repo.diff(from_snapshot_id=init, to_snapshot_id=C).updated_chunks[
            "/group1/data"
        ]
    )
    print(
        f"{tag:16} C.pruned_ancestor_tx_logs {str(pruned):8}  diff(init → C) {chunks}"
    )


# 1. Plain expiration: A and B collapse straight into C. Correct -> [A, B].
repo, init, A, B = fresh_repo()
C = write_and_commit(repo, 2, 30, "C: write data[2]")
repo.expire_snapshots(moment())
report("plain expire:", repo, {init: "init", A: "A", B: "B", C: "C"})

# 2/3. reset + GC gives C a real parent (A) and pruned = [B]; a later expire then
#      overwrites pruned, dropping B.
repo, init, A, B = fresh_repo()
cutoff = moment()
C = write_and_commit(repo, 2, 30, "C: write data[2]")

labels = {init: "init", A: "A", B: "B", C: "C"}
repo.reset_branch("main", A)
repo.garbage_collect(cutoff)
repo.reset_branch("main", C)
print("------------")
report("after reset+GC:", repo, labels)
repo.expire_snapshots(moment())
report("then re-expire:", repo, labels)
```
</details>

ianhi:
> will result in the incorrect pruned_ancestor_tx_logs for C of only A not [A, B]

Thinking on it the really nice invariant here is to again use icechunk as a test for itself. We can have a test that makes a commit sequence. And then do deepak's trick of copying the local store into two indepdent copies. and on one set we do a simple expire, and on the other a bunch of crazy history reducing operations in arbitrary order - then at the end the final result needs to be identical.

icechunk/src/inspect.rs
paraseba: Good catch!

icechunk/src/repository.rs
paraseba: Thank you

icechunk-python/tests/data_generation/generate_expired_repos.py
paraseba: great!

icechunk-python/tests/data_generation/generate_expired_repos.py
paraseba: Generative ASCII Art to the rescue 
```
    Main line, oldest -> newest. The two ``|`` are the expiration boundaries:
    ``t1`` sits between ``e`` and ``g``, ``t2`` between ``i`` and ``j``. The first
    node past each boundary (``g``, then ``j``) is the survivor tip that gets
    re-parented over the snapshots expired just before it. Forks hang below the
    node they branch from::

                                              t1              t2
                                               |               |
        init -> root -> a -> b -> c -> d -> e  |  g -> h -> i  |  j  (main tip)
                             |    |    |               |
                             |    |    |               +---- doomed2  : branch off h, +doomed2_work  -> DELETED round 2
                             |    |    +---- doomed1  : branch off d, +doomed1_work  -> DELETED round 1
                             |    +---- feature  : branch off c, +feat_pre/feat_s1/feat_s2  (survives)
                             +---- protect-b: TAG on b  (keeps b reachable past expiry)
```

icechunk-python/tests/data_generation/generate_expired_repos.py
paraseba: Most assertions are the job of test_can_read_old.py, but I added a few important ones here.

icechunk-python/tests/test_can_read_old.py
paraseba: Great catch, missed this. Adding to generation script.

icechunk-python/tests/data_generation/generate_expired_repos.py
ianhi: that is excellent art. showing the t1/t2 is a great addition

icechunk-python/tests/test_can_read_old.py
paraseba: names are crap, that actually means 2.0.5 + ancestor tracking, so, working copy. I'll improve.

icechunk-python/tests/test_stateful_repo_ops.py
paraseba: It's not necessarily always the initial snapshot. There are scenarios in which you can end with a parent that is not the initial commit. For example, if as GC is walking down the ancestry it finds a snapshot "protected" by a branch or tag, that one cannot be expired.

icechunk/src/repository.rs
paraseba: Done, used your snippet

icechunk-format/flatbuffers/repo.fbs
paraseba: done

icechunk-python/tests/data_generation/generate_expired_repos.py
paraseba: done

icechunk-python/tests/data_generation/generate_expired_repos.py
paraseba: done

icechunk-python/tests/data_generation/generate_expired_repos.py
paraseba: done

icechunk-python/tests/test_can_read_old.py
paraseba: done

icechunk-python/tests/test_can_read_old.py
paraseba: done

paraseba:
Amazing bug report @ianhi ! That was in fact a bug and it was trivial to fix with your comment and reproducer. Thank you.

icechunk-format/src/snapshot.rs
```diff
@@ -381,24 +381,32 @@ pub struct SnapshotInfo {
...
+    /// Notice: icechunk V1 and older versions of icechunk V2
+    /// didn't generate this field, so those repositories will
+    /// have [] until expiration is ran using a version > 2.0.6
+    pub pruned_ancestor_tx_logs: Vec<SnapshotId>,
 }
 
-impl TryFrom<&Snapshot> for SnapshotInfo {
-    type Error = IcechunkFormatError;
-
-    fn try_from(value: &Snapshot) -> Result<Self, Self::Error> {
```
li-em, line 389: I highly doubt anyone was using this outside IC, but since it is public API maybe worth keeping it?
```rust
impl TryFrom<&Snapshot> for SnapshotInfo {
    type Error = IcechunkFormatError;

    fn try_from(value: &Snapshot) -> Result<Self, Self::Error> {
        SnapshotInfo::from_snapshot_file(value)
    }
}
```

icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -930,8 +989,14 @@ def discard_changes(self) -> None:
     @rule(branch=branches, commit=commits)
     def reset_branch(self, branch: str, commit: str) -> None:
         # V1 expired snapshots stay on disk so reset_branch would succeed,
-        # but modelling that divergence isn't worthwhile — just skip.
-        assume(not (self.model.spec_version == 1 and commit not in self.model.commits))
```
paraseba, line 934: @ianhi I doubled down on this weakening of the test to make it pass.

ianhi: The weakening for v1? That seems fine

ianhi (approved):
This is great seba! Very thorough 👏 for wrangling this.


---

The one thing I would consider addding, but woudl recommend it in a follow up, is to have a regular run of the that 2.0.5 can read things written by newer versions. 

But that is a more general problem that we don't do any regular forward compat testing. We should have a nightly test that compares our oldest support 2.x version against main - like we do with the 1.x vs 2.x

li-em (approved):

paraseba:
> The one thing I would consider addding, but woudl recommend it in a follow up, is to have a regular run of the that 2.0.5 can read things written by newer versions.

yeah, that's an issue. The only thing I did was adding more checks and assertion in the fixture generation scripts.

Would you mind writing an issue with this gap @ianhi ?

icechunk-format/src/snapshot.rs
paraseba: I had this in my first version, but I decided to drop it intentionally, because it loses information. Basically, the Snapshot doesn't have the new pruned tx logs field, and this conversion could be used inadvertently. So I replaced it by an explicitly named function.

PR #2187, opened 2026-06-05
SPEC FIX: correct the description of sort order + add a test

### LINKED ISSUE #2183: Spec: invalid statement about component-wise sort order in common.fbs
https://github.com/earth-mover/icechunk/commit/c2f38ef2ebaed27c6ad40fa44ce776fc57aecfd5 added this mentions:
```
// When comparing paths for sort order, comparison is component-wise
// lexicographic: split on `/`, then compare segments left to right.
// For example: /a < /a/b < /ab < /b.
```

and https://github.com/earth-mover/icechunk/commit/3eb19e8c8d32bbe605c899850416882d4461aedd:
```
// Note: this differs from raw byte comparison. For instance,
// /a/b < /a-b in component-wise order, but /a-b < /a/b in byte order.
```

and the Snapshot definition
```
  // All nodes in this snapshot.
  // MUST be sorted by path (see node path sort order in common.fbs).
  nodes: [NodeSnapshot] (required);
```

Is that addition about a component-splitting comparison actually correct ?

Experimentation with icechunk python shows not:
```
import icechunk
import zarr

storage = icechunk.local_filesystem_storage("test_path_order")
repo = icechunk.Repository.create(storage)

with repo.transaction("main", message="first commit") as store:
    root = zarr.group(store)

    a_minus_b = root.create(
        "a-b",
        shape=(1,),
        dtype="uint8",
        compressors=None,
    )
    a_minus_b[...] = 2
    group_a = root.create_group("a")
    a_slash_b = group_a.create(
        "b",
        shape=(1,),
        dtype="uint8",
        compressors=None,
    )
    a_slash_b[...] = 1
```

then
```
dd if=test_path_order/snapshots/01Y374EFH250KVHK4ZX0 bs=1 skip=39 | zstdcat > x.bin
flatc --strict-json --raw-binary --json snapshot.fbs -- x.bin
cat x.json  | jq ".nodes[].path"
```
outputs:
```
"/"
"/a"
"/a-b"
"/a/b"
```

So the nodes are sorted in byte order not component-wise order

ianhi:
Write the correct description of field sort order into the spec. I also added a test of this to make sure we don't break it at some point in the future. We might consider adding more for other things we assert as MUST in the spec, but that can be a follow up.

fixes: https://github.com/earth-mover/icechunk/issues/2183

attn @rouault @paraseba 

Is this a format change? On the surface yes because we are changing the written spec. But in practice no, this is updating the description in the spec to match the reality of every IC2 in existence as there currently re no other implementors of `write` for icechunk


icechunk-python/tests/test_format_universality.py
```diff
@@ -1008,3 +1006,52 @@ def test_extra_fields_forward_compatibility(
             new_tag_session.store, path="group1/big_chunks", mode="r"
         )[:]
         np.testing.assert_array_equal(new_tag_big, orig_tag_big)
+
+
+def test_snapshot_nodes_sorted_by_bytes(tmp_path: Path, flatbuf_mod: ModuleType) -> None:
```
paraseba, line 1011: could you verify this test would fail under the previous spec description? And add a comment in the test code explaining how it would fail

paraseba (approved):

icechunk-python/tests/test_format_universality.py
ianhi: How do you mean? I think it does that already in that the asserted order does not match what the prior spec claimed

paraseba: I just want to see an `assert` showing me that. As proof that the test case is complex enough to trigger it (the /, the -, etc).

ianhi: Sorry im still confused. is this assert 
` assert [n["path"] for n in repo.inspect_snapshot(snapshot_id)["nodes"]] == expected` 

not that?

ianhi: or that i give the wrong order and it will fail i guess

paraseba: Basically, I want to see that with the node names you selected, the wrong order and the right order are different.

ianhi: I think I'm still not properly understanding what you want here :/

paraseba: Whatever I wanted, who cares, let's merge

PR #2190, opened 2026-06-07
fix: build object-store-s3 without the s3 (aws-sdk) feature

mrauhala:
## Problem

Building the Rust crate with the lighter `object_store` S3 backend but **without** the `s3` (aws-sdk) backend fails to compile:

```
cargo build -p icechunk --no-default-features --features object-store-s3
```

```
error[E0277]: `?` couldn't convert the error to `ICError<VirtualReferenceErrorKind>`
   --> icechunk/src/virtual_chunks.rs:562
562 |     urlencoding::decode(host)?.into_owned()
```

`VirtualChunkResolver` decodes the S3 virtual-chunk host with a bare `?`. The `From<FromUtf8Error>` that `?` relies on for `VirtualReferenceError` is only in scope under the `s3` feature, so the `object-store-s3`-only configuration doesn't build.

## Fix

Use `.capture()?`, matching the **identical** `urlencoding::decode(...)` call a few lines above (line ~425) and the surrounding error handling:

```rust
-                    urlencoding::decode(host)?.into_owned()
+                    urlencoding::decode(host).capture()?.into_owned()
```

No behavioural change — it just makes the existing error handling consistent and feature-independent.

## Why

Downstreams that only need the `object_store` S3 backend currently can't avoid pulling the full `aws-sdk-s3` tree. With this fix (plus a `default-features = false` option in `zarrs_icechunk`), an `object-store-s3 + object-store-fs` build drops the AWS SDK entirely and reads anonymous public Icechunk repos fine over `object_store` — verified end-to-end against the dynamical.org AIFS/GFS/ICON-EU datasets.


paraseba:
Great catch @mrauhala. Thanks for your contribution!

paraseba (approved):

Issue #2194, opened 2026-06-09
Deprecate `None` for virtual chunk authorization


TomNicholas:
### Problem

Currently when authorizing access to virtual chunk containers for a private bucket during `Repository.open()` these two kwargs values are equivalent:

```python
authorize_virtual_chunk_access={"s3://my-private-bucket": None}
```

```python
authorize_virtual_chunk_access={"s3://my-private-bucket/": icechunk.S3Credentials(from_env=True)}
```

but only the latter actually indicates clearly what you are doing: allowing Icechunk to look for local S3 credentials (which may of course be sensitive), which can then be used to read anything from the entirety of `s3://my-private-bucket`.

Given that the purpose of the `authorize_virtual_chunk_access` arg is to force users to explicitly opt-in to accessing potentially-sensitive storage locations for safety reasons, the auto-magicness of `None` seems undesirable.

It's also not ideal that `None` could be interpreted to mean "false" instead of the intended meaning here of "default". So a user might reasonably read the code above as meaning "do **not** authorize access to `s3://my-private-bucket`" when it actually means the opposite.

### Consistency with other explicit API

We already have some existing API for being much more explicit about this. As well as `icechunk.S3Credentials(from_env=True)` we also have `s3_anonymous_credentials()`, allowing the (much safer) anon-only case to be clearly distinguished:

```python
authorize_virtual_chunk_access={"s3://my-company-bucket": s3_anonymous_credentials()}
```

This kind of explicit sentinel is preferable, but not all our API forces you to use this explicit option, and some paths do not have a corresponding sentinel defined.

### Proposed changes

I want the only way to authorize virtual chunk access to be to use an explicit sentinel. So I want `None` to never be an allowed value.

That would mean that all types of virtual chunk access required use of the correct sentinel object:

```python
# anonymous S3 bucket
authorize_virtual_chunk_access={"s3://my-private-bucket/": icechunk.s3_anonymous_credentials()}

# private S3 bucket - credentials from env
authorize_virtual_chunk_access={"s3://my-private-bucket/": icechunk.S3Credentials(from_env=True)}

# private S3 bucket - explicit credentials
authorize_virtual_chunk_access={"s3://my-private-bucket/": icechunk.S3Credentials(access_key=...)}

# local filesystem - explicitly no credentials required
authorize_virtual_chunk_access={"file:///my-home-dir/": icechunk.credentials.LocalFilesystemAccess}  # a new sentinel type

# local filesystem - explicitly no credentials required
[...]

PR #2197, opened 2026-06-09
Deprecate `None` for virtual chunk authorization (phase 1)

TomNicholas:
Phase 1 of #2194. Adds explicit, per-backend credential sentinels and deprecates (warns on, but still accepts) `None` in `authorize_virtual_chunk_access`.

## Why

`None` is overloaded and silently permissive. For S3 it means "look for credentials in the environment (or anonymous)", so `{"s3://bucket": None}` quietly authorizes Icechunk to use ambient/private credentials — the opposite of the explicit opt-in this argument is meant to provide. `None` also reads like "false" (do *not* authorize), when it means the reverse.

## What

- **New no-auth sentinels** for backends that need no credentials: `icechunk.credentials.LocalFileSystemAccess` (`file://`) and `icechunk.credentials.HttpAccess` (`http(s)://`) — backed by `Credentials::LocalFileSystemAccess` / `Credentials::HttpAccess` in the Rust core and pyo3. (Available under `icechunk.credentials`, not the top-level namespace.)
- **`None` now warns** at repo-open (Python `DeprecationWarning`; Rust `tracing::warn!`), naming the explicit replacement and including a copy-pasteable snippet, e.g.:
  ```
  authorize_virtual_chunk_access={"file:///data/": ic.credentials.LocalFileSystemAccess}
  ```
  For S3 the message also flags the private-credential exposure.
- **Behaviour is unchanged** — `None` still works exactly as before this release. Phase 2 (a separate PR) will make it raise.
- Docs, docstrings, and a changelog entry updated to model the explicit sentinels.

## Migration

| backend | replace `None` with |
|---|---|
| S3 / Tigris | `ic.S3Credentials.FromEnv()` or `ic.s3_anonymous_credentials()` (wrapped, or via `containers_credentials`) |
| GCS | `ic.GcsCredentials.Anonymous()` |
| file:// | `ic.credentials.LocalFileSystemAccess` |
| http(s):// | `ic.credentials.HttpAccess` |

## Notes

- A pre-existing ergonomic gap (the kwarg requires top-level `Credentials` wrappers, so unwrapped `S3Credentials.FromEnv()` is rejected while the new sentinels are accepted directly) is **intentionally left out of scope** here; tracking separately.


icechunk/src/repository.rs
```diff
@@ -2135,17 +2135,53 @@ fn validate_credentials(
...
     Ok(())
 }
 
+/// Emit a phase-1 deprecation warning for a `None` virtual-chunk credential, naming the
+/// explicit replacement for the container's backend (inferred from the url prefix scheme).
+fn warn_deprecated_none_credential(url_prefix: &str) {
+    let scheme = url_prefix.split_once("://").map(|(s, _)| s).unwrap_or(url_prefix);
+    let suggestion = match scheme {
+        "s3" | "tigris" => {
+            "Passing `None` here silently authorizes Icechunk to read credentials from \
```
paraseba, line 2196: I think you left the context/explanation here, that is later repeated in the `warn!`. I think you want only the suggestion here?

icechunk/src/repository.rs
```diff
@@ -2135,17 +2135,53 @@ fn validate_credentials(
...
+             the environment (or use anonymous access), which can expose private \
+             credentials. Pass an explicit `Credentials::S3(S3Credentials::FromEnv)` \
+             (or `S3Credentials::Anonymous`) instead."
```
paraseba:
```suggestion
             or `S3Credentials::Anonymous` instead."
```

icechunk-python/python/icechunk/repository.py
```diff
@@ -23,6 +23,64 @@
...
+    if scheme in ("gs", "gcs"):
+        return "ic.Credentials.Gcs(ic.GcsCredentials.Anonymous())", ""
+    if scheme == "file":
+        return "ic.credentials.LocalFileSystemAccess", ""
+    if scheme in ("http", "https"):
+        return "ic.credentials.HttpAccess", ""
+    return None, ""
+
+
+def _warn_on_none_virtual_chunk_credentials(
```
paraseba, line 54: Don't we get this from Rust? Do we need add it here too?

TomNicholas: The Rust side emits a `tracing::warn!`, not a Python warning. At the default log level a Python user sees nothing.

Also the pythonic behaviour is to emit a Python `DeprecationWarning` - people might scan for that. The rust warning is different.

So we kinda need both.

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -3264,7 +3264,26 @@ class Credentials:
...
+    class LocalFileSystemAccess:
+        """Sentinel authorizing access to a local-filesystem (``file://``) virtual
+        chunk container, which requires no credentials. Surfaced as
+        ``LocalFileSystemAccess``."""
+
+        def __new__(cls) -> Credentials.LocalFileSystemAccess: ...
+
+    class HttpAccess:
+        """Sentinel authorizing access to an HTTP(S) virtual chunk container, which
+        requires no credentials. Surfaced as ``HttpAccess``."""
```
TomNicholas:
```suggestion
        requires no credentials."""
```

icechunk/src/repository.rs
TomNicholas: The reason there is more context here is that the statement "Passing `None` here silently authorizes Icechunk to read credentials from the environment" only applies if you were trying to read from S3. The auto-cred-grabbing behaviour is not present for all other storage types.

icechunk-python/python/icechunk/repository.py
paraseba: Cool. I _think_ default log level should be warning though.

paraseba (approved):

PR #2201, opened 2026-06-11
Enable nix dev env on macOS

li-em:
Reorg nix flake a bit so it can be used on macOS too.

Arrange `just develop` and `just pytest` so they can execute under both Nix and pixi dev environments.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -43,7 +43,14 @@ gen-flatbuffers:
 
 [doc("Prepare environment for development")]
 develop *args:
-  cd icechunk-python && maturin develop --uv --profile {{profile}} "$@"
+  #!/usr/bin/env bash
+  set -euo pipefail
+  cd icechunk-python
+  if [[ -n "${CONDA_PREFIX:-}" ]]; then
+    export VIRTUAL_ENV="$CONDA_PREFIX"
+    export UV_NO_SYNC=1
+  fi
+  exec uv run --active maturin develop --uv --profile {{profile}} "$@"
 
 [doc("Install maturin import hook for more convenient development flow")]
 import-hook:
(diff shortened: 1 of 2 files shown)


Justfile
```diff
@@ -147,7 +154,14 @@ py-pre-commit $SKIP="rust-pre-commit-fast,rust-pre-commit,rust-pre-commit-ci" *a
 
 [doc("Run Python tests via pytest")]
 pytest *args:
-  cd icechunk-python && uv run pytest "$@"
```
li-em, line 150: This ended up breaking `pixi` in #2184, the fix is forcing `uv` to re-use the pixi env instead of creating a new one.

Justfile
```diff
@@ -43,7 +43,14 @@ gen-flatbuffers:
...
 develop *args:
-  cd icechunk-python && maturin develop --uv --profile {{profile}} "$@"
+  #!/usr/bin/env bash
+  set -euo pipefail
+  cd icechunk-python
+  if [[ -n "${CONDA_PREFIX:-}" ]]; then
+    export VIRTUAL_ENV="$CONDA_PREFIX"
+    export UV_NO_SYNC=1
+  fi
+  uv run --active maturin develop --uv --profile {{profile}} "$@"
```
li-em, line 53: mostly changed to match what `just pytest` is doing

flake.nix
```diff
@@ -28,135 +28,151 @@
...
+    forAllSystems = lib.genAttrs systems;
+  in {
+    packages = forAllSystems (system: {
+      default = fenix.packages.${system}.stable.toolchain;
+    });
+
+    # This example provides two different modes of development:
+    # - Impurely using uv to manage virtual environments
+    # - Pure development using uv2nix to manage virtual environments
+    devShells = forAllSystems (
```
li-em, line 87: This ended up reformatting a lot of the file, but boils down to using `forAllSystems` to support `x86_64-linux` and `aarch64-darwin`. We might add `aarch64-linux` in the future too.

paraseba (approved):

Issue #2123, opened 2026-05-13
Expiration doesn't update `diff`


paraseba:
### What happened?

When expiration is run on a linear history repo, the result is a repo in which the oldest commit contains all the data from any previous commits, but the `diff` for that commit doesn't reflect the contents, it contains only the changes applied in the original commit, not all "merged" ancestry.

### What did you expect to happen?

The `diff` (and transaction log) should reflect the fact that the new commit is "updated" to produce a change equivalent to all its ancestry "merged"

### Minimal Complete Verifiable Example

```Python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "icechunk",
#     "zarr",
# ]
# ///
import icechunk as ic
import zarr

repo = ic.Repository.create(storage=ic.in_memory_storage())

session = repo.writable_session("main")
zarr.group(store=session.store, overwrite=True)
session.commit("initial commit")

for i in range(1, 10):
    session = repo.writable_session("main")
    root = zarr.open_group(store=session.store, mode="r+")
    root.create_group(f"group_{i}")
    session.commit(f"add group_{i}")

tip = repo.lookup_snapshot(next(repo.ancestry(branch="main")).id)
repo.expire_snapshots(tip.written_at)

ro = repo.readonly_session(branch="main")
root = zarr.open_group(store=ro.store, mode="r")
print(f"groups in repo: {sorted(root.group_keys())}")

tip_snap, root_snap = list(repo.ancestry(branch="main"))
diff = repo.diff(from_snapshot_id=root_snap.id, to_snapshot_id=tip_snap.id)
print(diff)

expected = {f"/group_{i}" for i in range(1, 10)}
assert diff.new_groups == expected, f"got {diff.new_groups}, expected {expected}"
```

### MVCE confirmation

- [x] Minimal example — the example is as focused as reasonably possible to demonstrate the underlying issue in icechunk.
- [x] Complete example — the example is self-contained, including all data and the text of any traceback.
- [x] Verifiable example — the example copy & pastes into an IPython prompt, returning the result.
- [x] New issue — a search of GitHub Issues suggests this is not a duplicate.
- [x] Recent environment — the issue occurs with the latest version of icechunk and its dependencies.

### Relevant log output

```Python
groups in repo: ['group_1', 'group_2', 'group_3', 'group_4', 'group_5', 'group_6', 'group_7', 'group_8', 'group_9']
Groups created:
    /group_9


Traceback (most recent call last):
[...]

PR #2215, opened 2026-06-16
Update spec and version policy for Icechunk spec 2.1

paraseba:
* I'm embedding the old flatbuffer files verbatim so we can use them in documenting spec 2.
* Starting the changelog for the unreleased version
* Renaming the now old V2 spec file, with minimal updates to:
   * indicate it's no longer the current version
   * Link to the old versions of flatbuffer files


icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -36,16 +42,18 @@ Users should always feel safe upgrading to the latest patch release.
...
+
+Across major versions, Icechunk libraries do **not** support forward compatibility. For example, Icechunk 1.x.y cannot read or write Icechunk format version 2 or later. Users must upgrade to a library version whose major version matches or exceeds the desired format.
 
-Icechunk libraries do **not** support forward-compatible reading or writing. For example, Icechunk 1.x.y cannot read or write Icechunk format version 2 or later. Users must upgrade to a library version whose major version matches or exceeds the desired format.
+Compatibility across minor versions: within a single major version, all minor spec versions are fully forward and backward compatible. Any 2.x library can read and write any spec 2.y repository. When a library operates on a repository written under a newer minor spec than it implements, reads and writes remain safe, but behavior is degraded for the features it does not recognize. For example, an Icechunk 2.0.x library can read and write a spec 2.1 repository, ignoring the fields introduced in 2.1 and behaving as it did before that feature existed.
```
ianhi: > ignoring the fields introduced in 2.1 and behaving as it did before that feature existed.

We should make it explicit that 2.0 will inadvertently remove the fields. At least thats how i understand it.

Is there any way we can forward proof that for all future versions some like leave untouched if not understood?

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -36,16 +42,18 @@ Users should always feel safe upgrading to the latest patch release.
 
 When a new major version is released, the previous major version will continue to receive **bug fixes and security patches for at least one year**. For example, if Icechunk 2.0 is released on April 1, 2026, Icechunk 1.x will receive bug and security fixes until at least April 1, 2027.
 
-## Migration  and Compatibility
+## Migration and Compatibility
 
-### Migratioon
+### Migration
```
ianhi, line 47: my favorite typo gone :(

icechunk-format/src/flatbuffers/README.md
```diff
@@ -46,7 +46,7 @@ All flatbuffer files (except chunks) have a 39-byte header:
 
 Bytes 0-11:  Magic bytes "ICE🧊CHUNK" (12 bytes: "ICE"=3, 🧊=4 UTF-8, "CHUNK"=5)
 Bytes 12-35: Implementation name, padded to 24 bytes (e.g., "ic-0.1.0")
-Byte 36:     Spec version (1=V1.0, 2=V2.0)
+Byte 36:     Spec version (1=V1.x, 2=V2.x)
```
ianhi: huh so in this one place we were doing fractional versioning?

To be consistent with the value in the binary file format of the spec needs this change:

```suggestion
Byte 36:     Spec version (1=V1, 2=V2.x)
```

icechunk-python/docs/docs/reference/spec.md
```diff
@@ -62,7 +62,7 @@ The storage system is not required to support random-access writes. Once written
 ### Consistency and Optimistic Concurrency
 
 Icechunk achieves transactional consistency using only the limited consistency guarantees offered by object storage.
-Icechunk V2 does this entirely via careful management of creation and conditional updating of the `RepoInfo` object.
+Icechunk V2+ does this entirely via careful management of creation and conditional updating of the `RepoInfo` object.
```
ianhi: Not super important but this implies we will not change this for IC3, which is probably correct but not necessarily so. V2.x maybe?

icechunk-python/docs/docs/reference/spec.md
```diff
@@ -245,6 +245,10 @@ Each snapshot in the repository has a `SnapshotInfo` entry in the repo info file
 --8<-- "icechunk-format/flatbuffers/repo.fbs:snapshot_info"
 ```
 
+Version 2.1 of the spec added the `pruned_ancestor_tx_logs: [ObjectId12]` field. Expiration uses this field to keep track of the full ancestry of transaction logs that generated this snapshot but were later expired. The list is ordered by ancestry, oldest first. Implementations MUST treat the concatenation of all these transaction logs, followed by the snapshot's own transaction log, as the full transaction history for the commit.
```
ianhi: Is the right terminology the history of the commit or the history of the snapshot?

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -36,16 +42,18 @@ Users should always feel safe upgrading to the latest patch release.
...
 
 Migrating an Icechunk repository from one major format version to the next is a **metadata-only operation** — it does not require copying or duplicating your data. Only Icechunk's internal metadata is updated; your chunk data remains in place.
 
 We maintain a documented and tested upgrade path from all major versions starting with Icechunk 1 to the latest version. Upgrading may be a multi-step process across major versions (e.g. upgrade 1 → 2, then 2 → 3), and we cover each step with integration tests.
 
 ### Compatibility
 
-An Icechunk library of major version N can read and write Icechunk Spec versions N and N-1. We make best effort to support reading older spec versions, but do not guarantee it. For example, Icechunk 3.x can read and write Spec versions 2 and 3, and will make best effort to read Spec version 1 data.
+Compatibility across major versions: an Icechunk library of major version N can read and write Icechunk Spec major versions N and N-1. We make best effort to support reading older spec versions, but do not guarantee it. For example, Icechunk 3.x can read and write Spec versions 2 and 3, and will make best effort to read Spec version 1 data.
```
ianhi: Should be consistent with minor versioning here. With the above description of how minor versions are wiritten (ommited minor = `.0`) this reads as 3.x can read and write 2.0 but not necessarily 2.x

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -36,16 +42,18 @@ Users should always feel safe upgrading to the latest patch release.
...
+
+Across major versions, Icechunk libraries do **not** support forward compatibility. For example, Icechunk 1.x.y cannot read or write Icechunk format version 2 or later. Users must upgrade to a library version whose major version matches or exceeds the desired format.
```
ianhi: This reads better to me, but I'm very possibly wrong on the correct grammar here. 

```suggestion
Icechunk libraries do **not** support forward compatibility across major versions. For example, Icechunk 1.x.y cannot read or write Icechunk format version 2 or later. Users must upgrade to a library version whose major version matches or exceeds the desired format.
```

ianhi: Also should maybe join the above paragraph to be part of the "Compatibility across major versions:" heading

Changelog.python.md
```diff
@@ -1,5 +1,22 @@
 # Changelog
 
+## Python Icechunk Library 2.1.0 [unreleased]
+
+This release introduces on-disk spec version 2.1, an additive change that is forward- and backward-compatible with spec 2.0.
```
ianhi: link to spec here I think

icechunk-python/docs/mkdocs.yml
```diff
@@ -257,7 +257,8 @@ nav:
   - Blog: blog-posts.md
   - Reference:
     - Spec:
-      - v2 (current): reference/spec.md
+      - v2.1 (current): reference/spec.md
```
ianhi: We should have this page be a redirect to the recent version so that we can directly link spec 2.1, currently you can permalink to v1 and v2 but there is no way to link to v2.x

ianhi (approved):
Some minor comments that would be good to address but overall looks good

icechunk-python/docs/docs/reference/spec.md
```diff
@@ -217,7 +217,7 @@ With the exception of chunk files, each type of file is encoded using [flatbuffe
 
 #### Repo Info File
 
-The repo info file is the single entry point for an Icechunk repository and the only mutable object in a V2 repo. It MUST be stored at `$ROOT/repo`. Every read operation starts by fetching this file. Every update to the repository (commits, tag creation, configuration changes) is a conditional write on this file.
+The repo info file is the single entry point for an Icechunk repository and the only mutable object in a V2+ repo. It MUST be stored at `$ROOT/repo`. Every read operation starts by fetching this file. Every update to the repository (commits, tag creation, configuration changes) is a conditional write on this file.
```
TomNicholas: Otherwise we are unecessarily making claims about how IC3+ will work?
```suggestion
The repo info file is the single entry point for an Icechunk repository and the only mutable object in a V2.x repo. It MUST be stored at `$ROOT/repo`. Every read operation starts by fetching this file. Every update to the repository (commits, tag creation, configuration changes) is a conditional write on this file.
```

icechunk-python/docs/docs/reference/spec.md
TomNicholas: Do we not want to link to some explanation of _why_ we made this change?

icechunk-python/docs/docs/reference/spec.md
```diff
@@ -347,7 +351,7 @@ Each `ManifestRef` identifies a manifest file and the region of the array's chun
 
 ##### `ManifestFileInfo`
 
-The snapshot's `manifest_files` / `manifest_files_v2` lists provide summary metadata about every manifest file referenced by the snapshot. These are separate from the per-array `ManifestRef` pointers — they describe the manifest files themselves rather than which chunks they cover. V2 repositories write their manifests using `ManifestFileInfoV2`; IC1 repositories use `ManifestFileInfo`.
+The snapshot's `manifest_files` / `manifest_files_v2` lists provide summary metadata about every manifest file referenced by the snapshot. These are separate from the per-array `ManifestRef` pointers — they describe the manifest files themselves rather than which chunks they cover. V2+ repositories write their manifests using `ManifestFileInfoV2`; IC1 repositories use `ManifestFileInfo`.
```
TomNicholas:
```suggestion
The snapshot's `manifest_files` / `manifest_files_v2` lists provide summary metadata about every manifest file referenced by the snapshot. These are separate from the per-array `ManifestRef` pointers — they describe the manifest files themselves rather than which chunks they cover. V2.x repositories write their manifests using `ManifestFileInfoV2`; IC1 repositories use `ManifestFileInfo`.
```

icechunk-python/docs/docs/reference/spec-v2-1.md
```diff
@@ -505,7 +509,14 @@ A tag can be created from any snapshot.
     1. If unsuccessful because there was some previous update, read the file again and retry
     1. If unsuccessful because the snapshot doesn't exist, fail the update
 
-## Changes from spec version 1
+## Changes from spec version 2 to 2.1
+
+- Added the optional `pruned_ancestor_tx_logs` field to the `SnapshotInfo` table in the `repo` flatbuffer file.
```
TomNicholas, line 514: It's optional from the perspective of the flatbuffers definition, but is it an optional field from the perspective of this spec?

TomNicholas: I guess it is optional in both senses

icechunk-python/docs/docs/reference/spec-v2-1.md
```diff
@@ -505,7 +509,14 @@ A tag can be created from any snapshot.
...
+- `pruned_ancestor_tx_logs` is populated during expiration, recording the transaction logs of ancestor commits that were removed from the repository.
+- The change is backwards and forwards compatible. `pruned_ancestor_tx_logs` is an optional flatbuffers field, so a 2.0.x reader simply ignores it and a 2.1 reader treats its absence as "never expired". Because of this, Icechunk libraries in the 2.0.x series can still read and write repositories that use spec version 2.1; in that case expiration is degraded to the 2.0.x behavior.
```
TomNicholas, line 516: > expiration is degraded to the 2.0.x behavior

Is there a writeup anywhere public of the actual bug (i.e. the degraded behaviour)?

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -2,7 +2,13 @@
 
 ## On-Disk Format
 
-The Icechunk on-disk format is defined by the format specification ("Icechunk Spec"), which is versioned by a single integer. The spec version increments if and only if there is an on-disk incompatible change. The current spec is [version 2](./spec.md) (used by Icechunk 2.x); the previous format is documented in [version 1](./spec-v1.md) (used by Icechunk 1.x).
+The Icechunk on-disk format is defined by the format specification ("Icechunk Spec"). A spec version is written as `<major>` with an optional `.<minor>` suffix (for example `1`, `2`, or `2.1`); an omitted minor is equivalent to `.0`.
+
+The major version increments only on an on-disk incompatible change, one that older libraries cannot safely read or write. A major bump changes the on-disk format version (described below) and may require a migration to adopt. The minor version increments on an on-disk compatible change. A minor bump does not change the on-disk format version and requires no migration. Spec versions 1 and 2 had no minor component (equivalently `.0`); 2.1 is the first minor release.
```
TomNicholas: > requires no migration

But there may be scenarios where a migration can optionally be performed, to get access to some new feature right? I'm thinking of manifest-level statistics.

So maybe this should be

> A minor version bump never requires a migration, but in some cases a optional migration may be available, generally to access a new format-level feature.

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -2,7 +2,13 @@
...
+
+Every metadata file records a single-byte on-disk format version in its header. This byte equals the spec major version only, and is `2` for both spec 2.0 and spec 2.1. Minor spec versions are not encoded in the header; instead, minor-version features are detected at read time.
```
TomNicholas, line 9: We will want to change this for IC3

icechunk-python/docs/docs/reference/version-policy.md
TomNicholas: are these two things consistent?

> fully forward and backward compatible

> behavior is degraded

icechunk-format/src/flatbuffers/README.md
paraseba: good catch

icechunk-python/docs/docs/reference/spec.md
paraseba: changing it

paraseba: I think the spec is the state of the world not the rationale for it. As an hypothetical implementor I don't care about the why, only about how to interpret the field.

icechunk-python/docs/docs/reference/spec-v2-1.md
paraseba: yeah, compatible 2.0.x implementations won't generate it.

icechunk-python/docs/docs/reference/spec-v2-1.md
paraseba: Only [this](https://github.com/earth-mover/icechunk/issues/2123) and my mental record of when I dreamed about this bug.

icechunk-python/docs/docs/reference/version-policy.md
paraseba: great, thanks

icechunk-python/docs/docs/reference/version-policy.md
paraseba: Probably, maybe.

icechunk-python/docs/docs/reference/version-policy.md
paraseba: changing it

icechunk-python/docs/docs/reference/version-policy.md
paraseba: I'll take your word on grammar

icechunk-python/docs/docs/reference/version-policy.md
paraseba: hm ... excellent question. I guess "compatible" is lifting weight here. I'll try to clarify.

paraseba: Dropping fully and qualifying compatible, feel free to make other suggestions:

> forward and backward read/write compatible

icechunk-python/docs/mkdocs.yml
paraseba: Good idea, trying it

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -2,7 +2,13 @@
 
 ## On-Disk Format
 
-The Icechunk on-disk format is defined by the format specification ("Icechunk Spec"), which is versioned by a single integer. The spec version increments if and only if there is an on-disk incompatible change. The current spec is [version 2](./spec.md) (used by Icechunk 2.x); the previous format is documented in [version 1](./spec-v1.md) (used by Icechunk 1.x).
+The Icechunk on-disk format is defined by the format specification ("Icechunk Spec"). A spec version is written as `<major>` with an optional `.<minor>` suffix (for example `1`, `2`, or `2.1`); an omitted minor is equivalent to `.0`.
+
+The major version increments only on an on-disk incompatible change, one that older libraries cannot safely read or write. A major bump changes the on-disk format version (described below) and may require a migration to adopt. The minor version increments on an on-disk compatible change. A minor version bump never requires a migration, but in some cases an optional migration may be available, generally to access a new format-level feature. Spec versions 1 and 2 had no minor component (equivalently `.0`); 2.1 is the first minor release.
```
dcherian, line 7: I suggest bolding : `A minor version bump never requires a migration` and `A major version may require a migration to adopt.`

icechunk-python/docs/docs/reference/version-policy.md
```diff
@@ -36,16 +42,16 @@ Users should always feel safe upgrading to the latest patch release.
...
 
-Icechunk libraries do **not** support forward-compatible reading or writing. For example, Icechunk 1.x.y cannot read or write Icechunk format version 2 or later. Users must upgrade to a library version whose major version matches or exceeds the desired format.
+Compatibility across minor versions: within a single major version, all minor spec versions are forward and backward read/write compatible. Any 2.x library can read and write any spec 2.y repository. When a library operates on a repository written under a newer minor spec than it implements, reads and writes remain safe, but behavior is degraded for the features it does not recognize. For example, an Icechunk 2.0.x library can read and write a spec 2.1 repository, ignoring the `pruned_ancestor_tx_logs` field introduced in 2.1 and behaving as it did before that feature existed.
```
dcherian, line 57: Would also bold `Any 2.x library can read and write any spec 2.y repository.`

PR #2219, opened 2026-06-18
Preserve all URL parts in virtual chunk locations

### LINKED ISSUE #2218: BUG: Virtual Refs drop several parts of a URL
Virtual chunk refs drop several parts of a URL that might be reasonable to include (discoverred via port) but true for several other valid constructions:

output of below:

```
wrote : https://user:pass@host.com:8443/a//b/../c.bin?versionId=42#frag
stored: https://host.com/a/c.bin
```

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "icechunk",
#     "zarr",
# ]
# ///
import icechunk as ic
import zarr

LOC = "https://user:pass@host.com:8443/a//b/../c.bin?versionId=42#frag"

repo = ic.Repository.create(storage=ic.in_memory_storage())
s = repo.writable_session("main")
a = zarr.open_group(s.store, mode="a").create_array(
    name="x", shape=(10,), dtype="float32", chunks=(10,)
)
s.store.set_virtual_refs(
    array_path=a.path,
    chunks=[ic.VirtualChunkSpec(index=[0], location=LOC, offset=0, length=40)],
    validate_containers=False,
)

stored = next(iter(s.all_virtual_chunk_locations()))
print("wrote :", LOC)
print("stored:", stored)

```

paraseba:
VirtualChunkLocation::from_url dropped userinfo, port, query and fragment, and collapsed empty path segments. Serialize the normalized URL instead so these are preserved.

Fixes #2218


ianhi (dismissed):
I think there's also a problem here. We probably need to reject some virtual chunks at write time. the ones that object-store can't handle

```python
import tempfile

import icechunk as ic
import zarr

d = tempfile.mkdtemp()
open(f"{d}/chunk.bin", "wb").write(b"x" * 40)
prefix = f"file://{d}/"

cfg = ic.RepositoryConfig.default()
cfg.set_virtual_chunk_container(ic.VirtualChunkContainer(prefix, ic.local_filesystem_store(d)))
repo = ic.Repository.create(
    storage=ic.in_memory_storage(),
    config=cfg,
    authorize_virtual_chunk_access=ic.containers_credentials({prefix: None}),
)
s = repo.writable_session("main")
zarr.open_group(s.store, mode="a").create_array(name="x", shape=(10,), dtype="f4", chunks=(10,))

s.store.set_virtual_refs(  # note the "//"
    array_path="/x",
    chunks=[ic.VirtualChunkSpec(index=[0], location=f"file://{d}//chunk.bin", offset=0, length=40)],
    validate_containers=False,
)
s.commit("ref")
zarr.open_group(repo.readonly_session("main").store, mode="r")["x"][0:1]

```


gives a super long error that ultimately ends in:

```
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
icechunk.IcechunkError:   × unknown error
  │
  │ context:
  │    0: icechunk::store::get
  │            with key="x/c/0" byte_range=From(0)
  │              at icechunk/src/store.rs:183
  │
  ├─▶ unknown error
  ╰─▶ Path "/var/folders/tc/fkgp35zn7z913f9cmsxcl6pc0000gn/T/tmpwu81uzcb//chunk.bin" contained empty path segment
```

icechunk-format/src/manifest.rs
```diff
@@ -345,38 +345,32 @@ impl VirtualChunkLocation {
...
+                _ => {
+                    return Err(VirtualReferenceError::capture(
+                        VirtualReferenceErrorKind::CannotParseBucketName(path.into()),
+                    ));
+                }
+            }
         }
 
-        Ok(VirtualChunkLocation(result))
+        // The `url` crate already normalizes the URL (collapsing `/../` and `/./`
```
ianhi: is this ok? are we silently changing a location. e.g. if this is an s3 key would we then poitn to a differetn opbject?

paraseba: This is wrong. 

But I'm struggling  with what is the right behavior. I suppose there is no way around having to split and re join for s3:// etc ... because we want to preserve everything in there. On the other hand file... we want to be smart about security

paraseba:
> I think there's also a problem here. We probably need to reject some virtual chunks at write time. the ones that object-store can't handle

Can we juts invoke user error in that situation? Maybe improving the error message? Writing virtual locations is performance critical, we don't want to do a lot there.

PR #2224, opened 2026-06-19
Honor port in virtual chunk URLs when reading from HTTP stores

### LINKED ISSUE #2223: BUG: virtual chunk reads ignore extra url info (e.g. port)
With https://github.com/earth-mover/icechunk/pull/2219 we now save info like port in a virtual chunk ref. But we don't yet use that in the read path

see https://github.com/earth-mover/icechunk/blob/1f932ce667cf9547d1f86ec4046c4c24707d2400/icechunk/src/virtual_chunks.rs#L708-L718

paraseba:
Closes #2223


li-em (approved):

ianhi:
Does this only add the port, what about the other url parts? I can totally see someone (perhaps inadvisably) versioning a dataset with `?version=3`

paraseba:
> Does this only add the port, what about the other url parts? I can totally see someone (perhaps inadvisably) versioning a dataset with `?version=3`

Those should be preserved by the full url that we now maintain. There is a test that uses this URL:
```
    location = "https://user:pass@host.com:8443/a//b/../c.bin?versionId=42#frag"
```

PR #2236, opened 2026-06-23
Sync changelog from release branch

paraseba:


paraseba:
This is purely to make it easier to add things, and later rebase release branch

PR #2237, opened 2026-06-24
Send logs to stderr instead of stdout

paraseba:
This prevents our own logs from corrupting the user's program output.

As a escape hatch, user can set ICECHUNK_LOG_TO_STDOUT to recover the previous behavior.

This is a braking change, but I think it's more correct and most users will be OK with it.

### DIFF
--- a/icechunk-python/python/icechunk/_icechunk_python.pyi
+++ b/icechunk-python/python/icechunk/_icechunk_python.pyi
@@ -3714,6 +3714,9 @@ def initialize_logs() -> None:
 
     Reads the value of the environment variable ICECHUNK_LOG to obtain the filters.
     This is autamtically called on `import icechunk`.
+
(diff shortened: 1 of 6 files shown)


li-em (approved):

icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -3714,6 +3714,9 @@ def initialize_logs() -> None:
 
     Reads the value of the environment variable ICECHUNK_LOG to obtain the filters.
     This is autamtically called on `import icechunk`.
```
dcherian, line 3716:
```suggestion
    This is automatically called on `import icechunk`.
```

hehe spotted small typo. we should run a spellchecker I guess

dcherian (commented):
LGTM

PR #2240, opened 2026-06-24
rename generator -> iterator to better reflect reality

ianhi:
The typing of these methods was inaccurately declaring them as generators when they were in fact iterators. See the bottom of this table https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes. generators have `asend`, `athrow` and `aclose`.

This is pretty niche, but it did trick an agent into writing some broken code when I was using `array_chunk_iterator` to do some clever stuff with reading a virtualized file more effficietnyl. Specifically the `aclose` is useful in that situation to close the stream. Otherwise the iterator holds a session lock until the iterator is fully exhausted, then garbage collected. That is slower because we have to wait for the whole stream, and also leads to non-determinisitic behavior in the completion of the stream. Or it's also possible to accidentally deadlock yourself if you `break` out of the iter loop without finishing it. With `aclose` you'd have an escape hatch against that. 

So this PR does three related things:

1. Add `aclose` to our iterator to allow deterministic behavior on closing these streams.
    - this code is the part I am most unsure about
    - notably they are still iterators, but iterators with an extra method not full generators
3.  Renamed `PyAsyncGenerator` to be an  `PyAsyncCloseableIterator` 
4. Unified the typing across all files to be accurate to what we actually return. 
    -This is also where the `__class_getitem__` comes from. It is necessary to allow subscripting for type annotations. We didn't need this previously because we were using the built in `collections.AsyncIterator` in our annotations which is not strictly accurate.


Changelog.python.md
```diff
@@ -21,6 +21,7 @@ This release introduces on-disk [spec version 2.1](https://icechunk.io/en/stable
 - Expiration now records the full ancestry of pruned transaction logs in a new optional `pruned_ancestor_tx_logs` field on `SnapshotInfo`, so `diff`, `amend`, `rebase`, and `inspect` produce correct results after expiration ([#2184](https://github.com/earth-mover/icechunk/pull/2184)).
 - Preserve all URL parts (userinfo, port, query, fragment, and empty path segments) when storing virtual chunk locations ([#2219](https://github.com/earth-mover/icechunk/pull/2219)).
 - Honor the port when reading virtual chunks from HTTP stores, so a reference like `http://host:8080/...` is fetched from the correct port ([#2223](https://github.com/earth-mover/icechunk/issues/2223)).
+- The objects returned by the async iteration APIs (`Store.list`/`list_prefix`/`list_dir`/`array_chunk_iterator`, `Repository.async_ancestry`/`ops_log_async`) are now correctly typed and named as async iterators (`AsyncCloseableIterator`) rather than the misleading `PyAsyncGenerator`, and gained an `aclose()` method so `contextlib.aclosing(...)` and deterministic early-exit cleanup work ([#XXXX](https://github.com/earth-mover/icechunk/pull/XXXX)).
```
ianhi:
```suggestion
- The objects returned by the async iteration APIs (`Store.list`/`list_prefix`/`list_dir`/`array_chunk_iterator`, `Repository.async_ancestry`/`ops_log_async`) are now correctly typed and named as async iterators (`AsyncCloseableIterator`) rather than the misleading `PyAsyncGenerator`, and gained an `aclose()` method so `contextlib.aclosing(...)` and deterministic early-exit cleanup work ([#2240](https://github.com/earth-mover/icechunk/pull/2240)).
```

paraseba (approved):

PR #2242, opened 2026-06-26
Fix S3 empty-prefix key layout and add layout auto-detection

### LINKED ISSUE #2239: Empty storage `prefix` produces object keys with a leading slash (`/chunks/...`)
## Summary

When an S3 `Storage` is created with an empty `prefix` (the Python default,
`icechunk.s3_storage(prefix="")`), every object is written under a key with a
**leading slash** — `/chunks/<id>`, `/refs/...`, `/snapshots/...`, etc. (an empty
top-level "folder" named `/` in S3 browsers, URI `s3://<bucket>//chunks/<id>`).

icechunk reads these back consistently, so repos open and round-trip fine *within*
icechunk. I believe the leading slash is unintended, and it breaks any
external tool that reconstructs object keys from the repo URL — concretely,
VirtualiZarr's `IcechunkParser` 404s when fetching native chunks.

## Where it comes from

`icechunk-s3/src/lib.rs`:

```rust
fn prefixed_path(&self, path: &str) -> String {
    format!("{}/{path}", self.prefix)   // prefix="" -> "/{path}"
}
```

`prefixed_path` builds the S3 key for `get`/`put`/`copy`/`head` (e.g.
`get_object_range_conditional` does `let key = self.prefixed_path(path);`). With an
empty `self.prefix`, `format!("{}/{path}", "")` yields a leading slash. The prefix is
genuinely empty here — the constructor preserves it:

```rust
let prefix = prefix.unwrap_or_default();                          // None/"" -> ""
let prefix = prefix.strip_suffix("/").unwrap_or(prefix.as_str()).to_string();  // "" stays ""
```

## Why this looks like a bug

The same file builds keys two other ways that do **not** produce a leading slash for
an empty prefix:

- `get_path_str` (used by `delete_batch` / listing):
  ```rust
  let path = PathBuf::from_iter([self.prefix.as_str(), file_prefix, id]);
  // PathBuf::from_iter(["", "chunks", "<id>"]) -> "chunks/<id>"   (no leading slash)
  ```
- `list_objects` even normalizes the double/leading slash explicitly:
  ```rust
  let prefix = format!("{}/{}", self.prefix, prefix).replace("//", "/");
  ```

## Reproduce

```python
import icechunk as ic, zarr
storage = ic.s3_storage(
    bucket="my-bucket", prefix="",            # the default
    region="...", endpoint_url="...",          # S3-compatible endpoint
    access_key_id="...", secret_access_key="...", force_path_style=True,
)
repo = ic.Repository.create(storage)
s = repo.writable_session("main")
g = zarr.group(store=s.store)
a = g.create_array("a", shape=(2,), chunks=(1,), dtype="i4", compressors=None)
a[0] = 1
s.commit("init")
```

Then list the bucket (e.g. boto3 `list_objects_v2`): the keys come back as
`'/chunks/<id>'`, `'/refs/...'`, `'/snapshots/...'` — all with a leading slash.

### Downstream symptom (how it surfaced)

VirtualiZarr's `IcechunkParser` reconstructs native-chunk locations as
`f"{url}/chunks/{id}"` (no leading slash) and `ManifestStore.get` then `lstrip("/")`s
the key, so it requests `s3://my-bucket/chunks/<id>` and gets a 404 — the real object
is at `s3://my-bucket//chunks/<id>`. (This is not a VirtualiZarr defect; it
reasonably assumes well-formed keys.)

## Environment

- icechunk 2.0.5, Python `s3_storage(prefix="")`, S3-compatible endpoint (Ceph,
  path-style).
- Surfaced via `virtualizarr==2.6.2` `IcechunkParser`.

## Expected

`s3_storage(prefix="")` should write keys without a leading slash (`chunks/<id>`),
matching `get_path_str`/`list_objects`, i.e. `prefixed_path` should skip the separator
when `prefix` is empty.

### Minimal Complete Verifiable Example

```Python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "icechunk<2",
#   "zarr",
# ]
#
# [[tool.uv.index]]
#
# [tool.uv.sources]
# zarr = { index = "scientific-python-nightly-wheels" }
#
# [tool.uv]
# prerelease = "allow"
# ///
#
# This script automatically imports the development branch of icechunk to check for issues.
# Please delete this header if you have _not_ tested this script with `uv run`!

import icechunk
icechunk.print_debug_info()
# your reproducer code ...
```

### MVCE confirmation

- [ ] Minimal example — the example is as focused as reasonably possible to demonstrate the underlying issue in icechunk.
- [ ] Complete example — the example is self-contained, including all data and the text of any traceback.
- [ ] Verifiable example — the example copy & pastes into an IPython prompt, returning the result.
- [ ] New issue — a search of GitHub Issues suggests this is not a duplicate.
- [ ] Recent environment — the issue occurs with the latest version of icechunk and its dependencies.

### Relevant log output

```Python

```

### Anything else we need to know?

_No response_

### Environment

<details>

platform:  Windows-11-10.0.26200-SP0
python:  3.13.5
icechunk:  2.0.5
zarr:  3.2.1
numcodecs:  0.15.1
xarray:  2026.4.0
virtualizarr:  10000.dev598+gfd5f7059e.d20260624

</details>

paraseba:
Closes: #2239

- We autodetect the layout (initial / which we call "rooted" vs. clean no-bug keys)
- We still offer a escape hatch if you really want to skip auto detection and create a specific layout (for example for tests).
- Autodetection has a minimum performance price:
  - Exclusively in repos with an empty prefix, 4 concurrent HEAD requests are executed the first time the storage instance is used.
  - Nothing changes for repos with non-empty prefix, or repos that are using the ObjectStore based Storage.

Main test changes:
- New `icechunk-python/tests/test_key_layout.py`: clean-key writes, round-trip, and rejecting `legacy_rooted_keys` with a non-empty prefix.
- `test_gc.rs`: fix a real-store flake in `do_test_gc`. GC deletes by `created_at` (S3 `LastModified`, truncated to whole seconds) while the cutoff came from `flushed_at` (microseconds), so a sub-second gap near the cutoff was unreliable.

CI (.github/workflows/rust-ci.yaml):
- Run the real-object-store integration tests on `workflow_dispatch` and on any PR labeled `run-integration-tests`, not only the nightly cron.
- Fix integration tests that were not running at all on the cron.
- Run the prefix-unique `--ignored` tests across the whole OS matrix, but run the bucket-root "rooted" tests on a single runner so concurrent jobs don't race on the same keys.
- Raise the job timeout 20 -> 45 min for the slow `macos-15-intel` runner now that integration tests run on the matrix.

Tooling and docs:
- compose.yaml: add MinIO container to test with an object store that collapses objects with an initial /.

### DIFF
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -10,6 +10,7 @@ on:
   schedule:
     # three times a day to run the integration tests that take a long time
     - cron:  '33 3,10,15 * * *'
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -31,7 +32,8 @@ env:
 jobs:
   rust:
     name: Rust CI
-    timeout-minutes: 20
+    # Headroom for the slow macos-15-intel runner
+    timeout-minutes: 45
     runs-on: ${{ matrix.os }}
     defaults:
       run:
@@ -76,6 +78,12 @@ jobs:
           done
           for _ in {1..60}; do
               if curl --silent --fail "http://localhost:10000/devstoreaccount1/testcontainer?sv=2023-01-03&ss=btqf&srt=sco&spr=https%2Chttp&st=2025-01-06T14%3A53%3A30Z&se=2035-01-07T14%3A53%3A00Z&sp=rwdftlacup&sig=jclETGilOzONYp4Y0iK9SpVRLGyehaS5lg5booJ9VYA%3D&restype=container"; then
+              break
+              fi
+              sleep 1
+          done
+          for _ in {1..60}; do
+              if curl --silent --fail "http://localhost:4202/minio/health/live"; then
               break
               fi
               sleep 1
@@ -136,8 +144,13 @@ jobs:
         run: |
           just profile=ci unit-test
 
+      # Runs nightly, on manual dispatch, or on any PR labeled
+      # `run-integration-tests`
       - name: Run integration tests against object stores
-        if: github.event_name == 'cron'
+        if: >-
+          github.event_name == 'schedule'
+          || github.event_name == 'workflow_dispatch'
+          || contains(github.event.pull_request.labels.*.name, 'run-integration-tests')
         env:
           R2_BUCKET: ${{ secrets.R2_BUCKET }}
           R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
@@ -154,8 +167,26 @@ jobs:
           TIGRIS_ACCESS_KEY_ID: ${{ secrets.TIGRIS_ACCESS_KEY_ID }}
           TIGRIS_SECRET_ACCESS_KEY: ${{ secrets.TIGRIS_SECRET_ACCESS_KEY }}
 
+          # passed via env rather than interpolated into the shell below, to
+          # satisfy zizmor's template-injection audit
+          MATRIX_OS: ${{ matrix.os }}
+
         run: |
-          cargo test --profile ci --workspace --all-targets -- --ignored
+          # Don't let a failure (or flake) in one group hide the other.
+          set +e
+          rc=0
+          # Most real-store ignored tests use unique prefixes, so they're safe to
+          # run on every matrix OS in parallel.
+          # Exclude icechunk-python: it has no Rust tests, but its PyO3 lib-test
+          # binary aborts at launch trying to load libpython
+          cargo test --profile ci --workspace --exclude icechunk-python --tests -- --ignored --skip rooted_roundtrip || rc=1
+          # The rooted-layout tests operate at the bucket root (empty prefix, no
+          # unique prefix), so concurrent matrix jobs would race on the same keys:
+          # run them on a single runner.
+          if [ "$MATRIX_OS" = "ubuntu-latest" ]; then
+            cargo test --profile ci -p icechunk --test main rooted_roundtrip -- --ignored || rc=1
+          fi
+          exit $rc
 
   shuttle:
     name: Shuttle concurrency tests
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -19,6 +19,7 @@ This release introduces on-disk [spec version 2.1](https://icechunk.io/en/stable
 ### Fixes
 
 - Expiration now records the full ancestry of pruned transaction logs in a new optional `pruned_ancestor_tx_logs` field on `SnapshotInfo`, so `diff`, `amend`, `rebase`, and `inspect` produce correct results after expiration ([#2184](https://github.com/earth-mover/icechunk/pull/2184)).
+- The native S3 backend used to write every object under a leading slash (`/chunks/...`) when `prefix` was empty, which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects. New empty-prefix repositories now use clean keys, and pre-existing "rooted" repositories are detected and handled automatically ([#2239](https://github.com/earth-mover/icechunk/issues/2239)).
 - Preserve all URL parts (userinfo, port, query, fragment, and empty path segments) when storing virtual chunk locations ([#2219](https://github.com/earth-mover/icechunk/pull/2219)).
 - Honor the port when reading virtual chunks from HTTP stores, so a reference like `http://host:8080/...` is fetched from the correct port ([#2223](https://github.com/earth-mover/icechunk/issues/2223)).
 
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -19,7 +19,8 @@ This release introduces on-disk [spec version 2.1](https://icechunk.io/en/stable
 ### Fixes
 
 - Expiration now records the full ancestry of pruned transaction logs in a new optional `pruned_ancestor_tx_logs` field on `SnapshotInfo`, so `diff`, `amend`, `rebase`, and `inspect` produce correct results after expiration ([#2184](https://github.com/earth-mover/icechunk/pull/2184)).
-- The native S3 backend used to write every object under a leading slash (`/chunks/...`) when `prefix` was empty, which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects. New empty-prefix repositories now use clean keys, and pre-existing "rooted" repositories are detected and handled automatically ([#2239](https://github.com/earth-mover/icechunk/issues/2239)).
+- The native S3 backend used to write every object under a leading slash (`/chunks/...`) when `prefix` was empty, which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects. New empty-prefix repositories now use clean keys, and pre-existing "rooted" repositories are detected and handled automatically. Rooted repositories are
+not migrated into a new format, so external tools may still have some issues with those repos, but Icechunk 2.1.0 knows how to read and write to rooted and non-rooted repos ([#2239](https://github.com/earth-mover/icechunk/issues/2239)).
 - Preserve all URL parts (userinfo, port, query, fragment, and empty path segments) when storing virtual chunk locations ([#2219](https://github.com/earth-mover/icechunk/pull/2219)).
 - Honor the port when reading virtual chunks from HTTP stores, so a reference like `http://host:8080/...` is fetched from the correct port ([#2223](https://github.com/earth-mover/icechunk/issues/2223)).
 
--- a/icechunk-python/docs/docs/guides/storage.md
+++ b/icechunk-python/docs/docs/guides/storage.md
@@ -73,6 +73,24 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
     )
     ```
 
+!!! note "Empty prefix and legacy repositories"
+
+    When `prefix` is empty (`None` or `""`) the repository lives at the bucket
+    root and objects are stored with clean keys (`chunks/...`). Repositories
+    created before [#2239](https://github.com/earth-mover/icechunk/issues/2239)
+    instead stored empty-prefix objects under a leading slash (`/chunks/...`);
+    Icechunk detects that layout automatically when opening such a repository, so
+    no action is needed to keep reading and writing it.
+
+    The `legacy_rooted_keys` parameter overrides that detection:
+
+    - `None` (default) — auto-detect the layout by probing storage. Use this
+      unless you have a specific reason not to.
+    - `True` — force the legacy leading-slash layout, skipping the probe (for
+      example, for a write-only worker that cannot probe storage). Only valid
+      with an empty prefix.
+    - `False` — force the standard layout, skipping the probe.
+
 #### Tigris
 
 [Tigris](https://www.tigrisdata.com/) is available as a storage backend for Icechunk. Icechunk provides a helper function specifically for [creating Tigris storage configurations](../reference/storage.md#icechunk.storage.tigris_storage).
--- a/icechunk-python/docs/docs/guides/storage.md
+++ b/icechunk-python/docs/docs/guides/storage.md
@@ -73,24 +73,6 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
     )
     ```
 
-!!! note "Empty prefix and legacy repositories"
-
-    When `prefix` is empty (`None` or `""`) the repository lives at the bucket
-    root and objects are stored with clean keys (`chunks/...`). Repositories
-    created before [#2239](https://github.com/earth-mover/icechunk/issues/2239)
-    instead stored empty-prefix objects under a leading slash (`/chunks/...`);
-    Icechunk detects that layout automatically when opening such a repository, so
-    no action is needed to keep reading and writing it.
-
-    The `legacy_rooted_keys` parameter overrides that detection:
-
-    - `None` (default) — auto-detect the layout by probing storage. Use this
-      unless you have a specific reason not to.
-    - `True` — force the legacy leading-slash layout, skipping the probe (for
-      example, for a write-only worker that cannot probe storage). Only valid
-      with an empty prefix.
-    - `False` — force the standard layout, skipping the probe.
-
 #### Tigris
 
 [Tigris](https://www.tigrisdata.com/) is available as a storage backend for Icechunk. Icechunk provides a helper function specifically for [creating Tigris storage configurations](../reference/storage.md#icechunk.storage.tigris_storage).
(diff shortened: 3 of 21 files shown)


Changelog.python.md
```diff
@@ -19,6 +19,7 @@ This release introduces on-disk [spec version 2.1](https://icechunk.io/en/stable
 ### Fixes
 
 - Expiration now records the full ancestry of pruned transaction logs in a new optional `pruned_ancestor_tx_logs` field on `SnapshotInfo`, so `diff`, `amend`, `rebase`, and `inspect` produce correct results after expiration ([#2184](https://github.com/earth-mover/icechunk/pull/2184)).
+- The native S3 backend used to write every object under a leading slash (`/chunks/...`) when `prefix` was empty, which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects. New empty-prefix repositories now use clean keys, and pre-existing "rooted" repositories are detected and handled automatically ([#2239](https://github.com/earth-mover/icechunk/issues/2239)).
```
ianhi: How do these interact with the extant external readers? is that something we need to worry about now?

.github/workflows/rust-ci.yaml
```diff
@@ -154,8 +167,26 @@ jobs:
...
+          # Most real-store ignored tests use unique prefixes, so they're safe to
+          # run on every matrix OS in parallel.
+          # Exclude icechunk-python: it has no Rust tests, but its PyO3 lib-test
+          # binary aborts at launch trying to load libpython
+          cargo test --profile ci --workspace --exclude icechunk-python --tests -- --ignored --skip rooted_roundtrip || rc=1
+          # The rooted-layout tests operate at the bucket root (empty prefix, no
+          # unique prefix), so concurrent matrix jobs would race on the same keys:
+          # run them on a single runner.
+          if [ "$MATRIX_OS" = "ubuntu-latest" ]; then
+            cargo test --profile ci -p icechunk --test main rooted_roundtrip -- --ignored || rc=1
```
ianhi, line 187: what happens if two of these tests run at the same time on different PRs? or if PR run coincides with a scheduled one? 

I guess that's just unlikely enough. I don't see any obvious way around it

Changelog.python.md
ianhi: Also I think handled is doing a little too much work here. On a first read I thought that meant that the repos are cured. But I think what you mean is that we handle reading them correctly, is that right?

icechunk-s3/src/lib.rs
```diff
@@ -56,6 +56,52 @@ use tokio_util::io::StreamReader;
...
+    LegacyRoot,
+}
+
+/// Well-known object names probed by [`S3Storage::probe_layout`] to auto-detect
+/// the key layout of a pre-existing repository. These are fixed keys that exist
+/// in every repository.
+///
+/// It's awful to have to have this tacit dependency here, this is icechunk-format stuff.
+/// But... reality hits you hard. I don't want an explicit dependency between the crates
+/// but there are some tests that verify these strings are "right".
```
ianhi: Can we explicitly mention where these are. I think you mean the tests in `icechunk/src/refs.rs`?

icechunk-s3/src/lib.rs
```diff
@@ -743,7 +887,13 @@ impl Storage for S3Storage {
...
         prefix: &str,
     ) -> StorageResult<BoxStream<'a, StorageResult<ListInfo<String>>>> {
-        let prefix = format!("{}/{}", self.prefix, prefix).replace("//", "/");
+        let layout = self.layout(settings).await?;
+        // The caller-supplied prefix can carry a leading `/` — e.g. someone
+        // inspecting a legacy rooted repo (whose raw keys look like `/chunks/...`)
+        // through the public `list_objects` API. `key_for` then doubles it against
+        // LegacyRoot's synthesized `/` or Standard's `{prefix}/{relpath}` join, and
+        // S3 matches `//` literally (so the listing would hit nothing). Collapse it.
+        let prefix = self.key_for(layout, prefix).replace("//", "/");
```
ianhi: Won't this break the prefix if the user has a `//` deep inside the prefix?

like `/some/awful//path/chunks`

I guess this used to be the case as well, but maybe worth fixing if it is a bug

icechunk-s3/src/lib.rs
```diff
@@ -979,6 +1139,63 @@ impl S3Storage {
...
+}
+
+/// `bucket` is only used to build the mixed-layout error.
+fn decide_layout(seen: &AnchorProbe<bool>, bucket: &str) -> StorageResult<KeyLayout> {
+    match seen {
+        // Both layouts present: ambiguous and unsafe
+        AnchorProbe { clean: true, rooted: true } => Err(other_error(format!(
+            "repository in bucket {bucket} has objects under both the standard and the \
+             legacy leading-slash key layouts; this is ambiguous and unsafe. \
+             See https://github.com/earth-mover/icechunk/issues/2239"
```
ianhi, line 1192: I really like the pointing to an issue at least. But I'd love to be able to go further and point to actionable advice here. 

1. how could a user even end up in this situation?
2. if they do how do they recover?

icechunk/tests/test_key_layout.rs
```diff
@@ -0,0 +1,476 @@
...
+//!
+//! Before the fix, an empty `prefix` made the native-S3 backend write every
+//! object under a leading slash (`/chunks/...`); external tools 404'd and GC
+//! silently orphaned objects (delete used a different join than write). These
+//! tests verify the fix against local rustfs.
+//!
+//! ## Why no live `LegacyRoot` round-trip here
+//!
+//! Empty-prefix repositories live at the bucket root, so each test uses its own
+//! freshly created bucket (via rustfs root credentials) for isolation.
```
ianhi: I dont fully understand this. Maybe I'm missing what a `live` round trip is.

icechunk-s3/src/lib.rs
```diff
@@ -56,6 +56,52 @@ use tokio_util::io::StreamReader;
...
+}
+
+/// Well-known object names probed by [`S3Storage::probe_layout`] to auto-detect
+/// the key layout of a pre-existing repository. These are fixed keys that exist
+/// in every repository.
+///
+/// It's awful to have to have this tacit dependency here, this is icechunk-format stuff.
+/// But... reality hits you hard. I don't want an explicit dependency between the crates
+/// but there are some tests that verify these strings are "right".
+pub const DEFAULT_LAYOUT_ANCHORS: &[&str] = &["repo", "refs/branch.main/ref.json"];
```
ianhi, line 83: Will we incorrectly not detect a collision with two v1 repos both on the first commit which is the only reasonable case for snapshot collision? I think for v2 there the repostatus has a time stamp in it so that should be safe, but v1 ref.json only has the snapshot

icechunk/tests/test_key_layout.rs
```diff
@@ -0,0 +1,476 @@
...
+        false,
+    );
+    garbage_collect(Arc::clone(repo.asset_manager()), &gc_config, None, 100).await?;
+    Ok(())
+}
+
+#[tokio_test]
+#[ignore = "needs credentials from env"]
+async fn rooted_roundtrip_in_aws() -> Result<(), Box<dyn std::error::Error>> {
+    if let Some(store) = common::aws_real_store() {
```
ianhi: Will this pass without actually running the test if the keys are not present (or changed) because the aws_real_store returns None in that case, then we skip the test

icechunk/tests/test_key_layout.rs
```diff
@@ -0,0 +1,476 @@
...
+    let client = mk_client(
+        &rustfs_options(),
+        root_credentials(),
+        vec![],
+        vec![],
+        &Settings::default(),
+    )
+    .await;
+    let resp =
+        client.list_objects_v2().bucket(bucket).send().await.expect("list_objects_v2");
```
ianhi, line 137: per https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html

this returns up to 1000 keys. So not technically every key in a bucket. Doesn't matter here, but is there a world where we would be using this method on a repo with more than 1000 keys?

ianhi: I think the action item if any would be a comment not a behavior change

icechunk/tests/test_key_layout.rs
```diff
@@ -0,0 +1,476 @@
...
+        1,
+    )
+    .await?;
+
+    let err = Repository::create(
+        None,
+        root_storage(&bucket, Some(""), false),
+        HashMap::new(),
+        Some(SpecVersionBin::V2),
+        true,
```
ianhi: Here you do the combo (v2, v2) but can we also test whole set, especially the mixed cases?

(v1, v2)
(v2, v1)
(v1, v1)

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -73,6 +73,24 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
     )
     ```
 
+!!! note "Empty prefix and legacy repositories"
+
+    When `prefix` is empty (`None` or `""`) the repository lives at the bucket
+    root and objects are stored with clean keys (`chunks/...`). Repositories
+    created before [#2239](https://github.com/earth-mover/icechunk/issues/2239)
```
ianhi: Let's reference the version number, that's much easier for people to utilize. Otherwise they have to go the issue -> PR -> github releases to figure out what they wrote with.

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -73,6 +73,24 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
...
+    no action is needed to keep reading and writing it.
+
+    The `legacy_rooted_keys` parameter overrides that detection:
+
+    - `None` (default) — auto-detect the layout by probing storage. Use this
+      unless you have a specific reason not to.
+    - `True` — force the legacy leading-slash layout, skipping the probe (for
+      example, for a write-only worker that cannot probe storage). Only valid
+      with an empty prefix.
+    - `False` — force the standard layout, skipping the probe.
```
ianhi: What is teh consequence of doing this for a user?

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -73,6 +73,24 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
...
+    Icechunk detects that layout automatically when opening such a repository, so
+    no action is needed to keep reading and writing it.
+
+    The `legacy_rooted_keys` parameter overrides that detection:
+
+    - `None` (default) — auto-detect the layout by probing storage. Use this
+      unless you have a specific reason not to.
+    - `True` — force the legacy leading-slash layout, skipping the probe (for
+      example, for a write-only worker that cannot probe storage). Only valid
+      with an empty prefix.
```
ianhi: can a write only worker properly use icechunk?

I thought we do some checks with GETs on opening the repo.

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -73,6 +73,24 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
     )
     ```
 
+!!! note "Empty prefix and legacy repositories"
```
ianhi: Can this maybe go further down the page. and become a `##`  or similar heading so we can permalink it from the error messages?

icechunk-s3/src/lib.rs
```diff
@@ -362,18 +436,86 @@ impl S3Storage {
...
+        if self.config.requester_pays {
+            req = req.request_payer(aws_sdk_s3::types::RequestPayer::Requester);
+        }
+        match req.send().await {
+            Ok(out) => Ok(Some(out.e_tag().unwrap_or_default().to_string())),
+            Err(sdk_err) => {
+                let absent = sdk_err.as_service_error().is_some_and(|e| e.is_not_found())
+                    || sdk_err
+                        .raw_response()
+                        .is_some_and(|r| matches!(r.status().as_u16(), 404 | 400));
```
ianhi, line 522: Worth considering teh 403 case? or would we not have reached here if we had credential issues?

ianhi (approved):
Seems good to me. Approval modulo some of the comments with suggestions.

Also overall this was a joy to review, I spent the majority of my time reading your excellent detailed comments, they really let me understand the behavior and choices 👏

Changelog.python.md
paraseba: Good point, I'll update. No, we are not "fixing" the repos, only maintaining read/write compatibility

.github/workflows/rust-ci.yaml
paraseba: yeah, in that case tests will break, that's the only reason I'm adding this `if` as a bad workaround. The solution would be to be able to create and destroy new buckets in tests, but that's not a great practice either. This is one of the reasons why in a follow up I'm completely disabling the ability to create rooted repos.

icechunk-s3/src/lib.rs
paraseba: yes, that, adding

icechunk-s3/src/lib.rs
paraseba: great catch

icechunk-s3/src/lib.rs
paraseba: I'm not sure what to say. This should never happen if they only used IC to write (and there are no bugs)

icechunk-s3/src/lib.rs
paraseba: This is only to separate the cases of `/refs/...` and `refs/...`. In IC1, creating that object is the atomic "repo created" event. So there is a V1 repo there iif that object exists (with or without the /)

icechunk/tests/test_key_layout.rs
paraseba: good catch

icechunk/tests/test_key_layout.rs
paraseba: commenting

icechunk-s3/src/lib.rs
ianhi: Maybe the ask is "this should not have been possible. Please open an issue at ..."

icechunk-python/docs/docs/guides/storage.md
paraseba: good idea

icechunk-python/docs/docs/guides/storage.md
paraseba: you could set up things in that way in  weird contexts, mostly using rust

icechunk-s3/src/lib.rs
paraseba: No, that should follow the store error branch

li-em (approved):

PR #2245, opened 2026-06-26
Disallow creating empty-prefix repositories on object stores

paraseba:


Changelog.python.md
```diff
@@ -19,7 +19,7 @@ This release introduces on-disk [spec version 2.1](https://icechunk.io/en/stable
...
-- The native S3 backend used to write every object under a leading slash (`/chunks/...`) when `prefix` was empty, which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects. New empty-prefix repositories now use clean keys, and pre-existing "rooted" repositories are detected and handled automatically ([#2239](https://github.com/earth-mover/icechunk/issues/2239)).
+- The native S3 backend used to write every object under a leading slash (`/chunks/...`) when `prefix` was empty, which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects. Pre-existing "rooted" repositories are detected and handled automatically. Creating new repositories at an empty prefix on an object store is no longer allowed. Existing empty-prefix repositories can still be opened and updated ([#2239](https://github.com/earth-mover/icechunk/issues/2239)).
```
TomNicholas, line 22: > which made repositories unreadable by external tools and could cause garbage collection to silently orphan objects

So does this PR fix that for pre-existing "rooted" repositories?

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -75,14 +75,16 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
...
-    When `prefix` is empty (`None` or `""`) the repository lives at the bucket
-    root and objects are stored with clean keys (`chunks/...`). Repositories
-    created before [#2239](https://github.com/earth-mover/icechunk/issues/2239)
-    instead stored empty-prefix objects under a leading slash (`/chunks/...`);
-    Icechunk detects that layout automatically when opening such a repository, so
-    no action is needed to keep reading and writing it.
-
-    The `legacy_rooted_keys` parameter overrides that detection:
+    Creating a repository at an empty `prefix` (`None` or `""`) — i.e. at the
+    bucket root — is not supported on object stores in modern Icechunk versions
```
TomNicholas:
```suggestion
    bucket root — is not supported on object stores since Icechunk v2.1.0
```

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -75,14 +75,16 @@ When using Icechunk with s3 compatible storage systems, credentials must be prov
...
+    and raises an error; use a non-empty `prefix`. In-memory and local-filesystem
+    storage are unaffected.
+
+    Existing empty-prefix repositories can still be **opened and updated**:
+    Icechunk detects their on-disk layout automatically, so no action is
+    needed to keep using them.
+
+    When opening, the `legacy_rooted_keys` parameter overrides that detection:
 
     - `None` (default) — auto-detect the layout by probing storage. Use this
```
TomNicholas: So presumably (I haven't looked at the implementation yet) in order to probe storage you make two concurrent requests for the two possible keys for the repo_info object.

TomNicholas (commented):
What happens if someone creates a repo at the root of the bucket manually (e.g. by bulk-copying objects over from another location)? Presumably that is treated the same as any other pre-existing "rooted" store, because there is no on-disk difference, and no dependency on library versions.

li-em (approved):

icechunk/src/repository.rs
```diff
@@ -129,6 +129,10 @@ pub enum RepositoryErrorKind {
     Tag(String),
     #[error("repositories can only be created in clean prefixes")]
     ParentDirectoryNotClean,
+    #[error(
+        "creating a repository at an empty prefix (the object store bucket root) is not supported in modern Icechunk versions; use a non-empty prefix. Existing empty-prefix repositories can still be opened and updated."
```
ianhi: Cant make an inline suggestion here for some reason. But another spot for a specific version.

Changelog.python.md
ianhi: no. sadly doing that would require moving all the `/chunks` objects. Also its a bit confusing so heads up that this is a PR merging into the other PR, not main

icechunk/tests/test_storage.rs
```diff
@@ -419,6 +419,100 @@ async fn check_clean_repo() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+/// Creating a repository at an empty prefix (the bucket/container root) is refused on
+/// every cloud object-store backend — native S3 and the `object_store`-based S3,
+/// Azure, and GCS — but allowed on in-memory and local-filesystem storage, and via
```
ianhi, line 424: Should we just be consistent everywhere?

ianhi (commented):
seems good. especially for merging into the other pr

Changelog.python.md
paraseba: yeah, forgot to mention this needs to go on top of the previous PR to make sense

icechunk-python/docs/docs/guides/storage.md
paraseba: yes, we probe well-known objects in the v1 and v2 version formats, lazily on first access to the storage.

icechunk/tests/test_storage.rs
paraseba: no, because having to pass a prefix for memory storage would be a pain without any reason, what does the prefix even mean for in memory (and a breaking change). For filesystem, i think it makes sense to support empty prefix, and it doesn't have the difficulties we have for testing in the case of object stores.

PR #2257, opened 2026-07-02
Reject shift_array/reindex_array on non-regular chunk grids

jhamman:
## Real human summary

#2151 documents a bug with `shift_array` and `reindex_array` for non-regular chunk grids. We've decided the complete fix is necessary but likely to take longer than we're willing to wait before addressing the risk of this bug hitting a real use case. So the stop-gap we've decided on is to reject shift_array/reindex_array for non-regular chunk grids. Then we'll come back to implement the more complete fix in a later PR.

👇 below is Claude's summary of the PR.

----

Part of #2151 (intentionally not closing it — this is the stopgap, not the fix).

## Problem

`shift_array` / `reindex_array` relabel chunk indices in the manifest without ever looking at chunk sizes. That's safe on regular grids, where every chunk is encoded at the full uniform chunk shape, but on a rectilinear grid slot `i` expects exactly `chunk_shapes[i]` elements. A shift commits cleanly and leaves payloads in slots that decode to different sizes — the array can no longer be read back (`ValueError: cannot reshape array of size 1 into shape (2,)`). `reindex_array` is exposed to Python with arbitrary index mappings, so it has the same failure mode.

## Change

- `reindex_array` (and therefore `shift_array`) now fails up front with a new `UnsupportedChunkGridForReindex` error unless the node's `user_data` parses as zarr array metadata positively declaring a `regular` chunk grid. The check fails closed: metadata we can't parse is rejected too. Parsing is intentionally minimal — only the chunk grid name is read.
- The rejected operation stages nothing in the change set.
- Docs: warning in the moving-chunks guide, notes on both docstrings, and `design-docs/018-shift-array-rectilinear-guard.md` recording the decision.

## Relationship to #2179

#2179 designs the real fix (rewriting the chunk grid during reindex, gap ambiguity, settled/open edges in N dimensions). This PR is the immediate data-corruption stopper; the design doc here defers to #2179 for the follow-up and takes number 018 since #2179 claims 017. For plain `shift_array` the follow-up is a metadata-only right-rotation of `chunk_shapes`; arbitrary reindex mappings need the user-facing grid API from #2179.

## Tests

- Rust: rejection of both operations on both spec versions, plus the regular-grid happy path.
- Python: repro from the issue asserting the error and that the session stays clean. Works on zarr 3.2.x (tuple-of-tuples `chunks`) as well as zarr main (`RectilinearChunkGrid`).
- Stateful suite: the `shift_array` rule now expects the rejection when hypothesis draws a rectilinear array — it was previously exposed to this corruption since the package conftest enables `array.rectilinear_chunks`. The grid-type check matches both the `RegularChunkGrid` (zarr main) and `RegularChunkGridMetadata` (zarr 3.2.x) class names.

Changelog entry deliberately omitted per `RELEASE.md` (written at release time).


icechunk-python/python/icechunk/_icechunk_python.pyi
```diff
@@ -2725,6 +2725,13 @@ class PySession:
             position is checked to determine whether it should be cleared — if
             ``backward`` returns ``None`` (out of bounds) or points to a position
             with no chunk, that position is reset to the fill value.
+
+        Raises
+        ------
+        IcechunkError
```
paraseba: I think this may be misleading because there are many other conditions that raise errors. Users may think this is a comprehensive list. Maybe we just continue to skip the Raises section?

icechunk-python/python/icechunk/session.py
```diff
@@ -275,6 +275,13 @@ def reindex_array(
             position is checked to determine whether it should be cleared — if
             ``backward`` returns ``None`` (out of bounds) or points to a position
             with no chunk, that position is reset to the fill value.
+
+        Raises
```
paraseba: See my comment on the .pyi file.

icechunk/src/session.rs
```diff
@@ -187,6 +187,12 @@ pub enum SessionErrorKind {
         "invalid chunk index: coordinates {coords:?} are not valid for array at {path}"
     )]
     InvalidIndex { coords: ChunkIndices, path: Path },
+    #[error(
+        "cannot move chunks of the array at `{path}`: it uses a `{grid_type}` chunk grid, and only `regular` chunk grids are supported. \
+         Reindexing chunks on a non-regular grid would corrupt the array, because chunk payloads no longer match \
+         the sizes of their new grid positions. See https://github.com/earth-mover/icechunk/issues/2151"
+    )]
```
paraseba, line 194: "move chunks" could be misinterpreted as an actual change in object store. Maybe "reindex chunks"?

icechunk/src/session.rs
```diff
@@ -222,6 +228,32 @@ pub enum ReindexMapping<'a> {
     },
 }
 
+/// Returns the chunk grid name if `user_data` holds zarr array metadata with a
+/// non-`regular` chunk grid.
+///
+/// The session treats `user_data` as opaque bytes, but when it holds zarr array
```
paraseba: Let's skip the second paragraph of this command. I'd replace it by mentioning that the parsing is as minimal as possible to defend against upstream changes. We are only parsing the grid name

icechunk/src/session.rs
```diff
@@ -222,6 +228,32 @@ pub enum ReindexMapping<'a> {
...
+        name: String,
+    }
+    #[derive(Deserialize)]
+    struct PartialArrayMetadata {
+        chunk_grid: ChunkGridName,
+    }
+
+    match serde_json::from_slice::<PartialArrayMetadata>(user_data) {
+        Ok(meta) if meta.chunk_grid.name != "regular" => Some(meta.chunk_grid.name),
+        _ => None,
```
paraseba: Returning None (which means regular) when we can't parse seems risky. I think the safe logic is: we only allow reindexing when we know for a fact that it's is a regular grid. We fail safely by disallowing.

jhamman:
Addressed the review in 81f1f404:

- **Fail closed** — reindexing is now allowed only when the metadata positively declares a `regular` chunk grid; unparseable metadata is rejected (`grid_type: "unknown"`). Added a test for that case; the existing Rust shift test now creates its array with real regular-grid metadata since empty `user_data` is no longer allowed through.
- Error message says "cannot reindex chunks" instead of "cannot move chunks".
- Helper doc comment now explains the parsing is intentionally minimal (only the grid name) to defend against upstream metadata changes.
- Dropped the `Raises` sections from both docstrings; kept the regular-grid-only constraint as a note.

paraseba (approved):
Thank you @jhamman . 

Could you add a line to the Fixes section of the changelog?

icechunk-python/docs/docs/guides/moving-chunks.md
```diff
@@ -18,6 +18,12 @@ This enables **rolling time windows**—continuously updating datasets like fore
 | [`shift_array`][icechunk.session.Session.shift_array] | Uniform shifts | Simple—just specify offset |
 | [`reindex_array`][icechunk.session.Session.reindex_array] | Custom transformations | Maximum—you control every chunk |
 
+!!! warning "Regular chunk grids only"
+    Both methods require the array to use a regular chunk grid, and raise an
+    `IcechunkError` otherwise. On a rectilinear grid each chunk's size depends
+    on its position, so relabeling chunk indices would leave chunk data that no
+    longer matches the expected size at its new position.
+
```
ianhi, line 30: Can we link to what a rectilinear grid is? A naive reader might think that they have rectilinear if they have chunks of different sizes on diff dims

icechunk/src/session.rs
ianhi: This is also the terminology I adopted before: https://icechunk.io/en/stable/guides/moving-chunks/

:(

maybe we should change that docs as well?

ianhi: The other note here. I suspect most people encourter this via shift_array - not reindex so this may be a confusing error message to receive.
can we include `shift_array` in the error message?

ianhi (commented):
a few small comments

jhamman:
Addressed @ianhi's comments in d86b6480:

- The error message now names both entry points: "cannot shift or reindex chunks of the array at `...`: `shift_array` and `reindex_array` require an array whose metadata declares a `regular` chunk grid, found `rectilinear`."
- The docs warning links the [rectilinear chunk grid extension spec](https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear) and spells out that per-dimension chunk sizes are still regular — a grid is rectilinear only when sizes vary *within* a dimension.

On the "moving chunks" terminology in the guide: left as is for this PR. The guide's intro already stresses that the data never moves ("chunks are addressed through a manifest... A 1TB shift completes in milliseconds"), and retitling the page/URL is a bigger editorial change — happy to do it as a follow-up if you both want it.

ianhi (approved):

PR #2267, opened 2026-07-10
Python: build an exception tree rooted by `IcechunkError`

### LINKED ISSUE #2098: Typed exception class for "ref not found" errors
Today, calling `Repository.readonly_session_async(branch=...)` (or `tag=...`) with a missing ref raises a generic `icechunk.IcechunkError` whose only signal that the ref is missing is the message text (`"ref not found \`<name>\`"`).

This forces callers that want to distinguish "ref doesn't exist" from "storage/manifest/IO failure" to do brittle message-substring matching.

Could we get a dedicated subclass — e.g. `icechunk.RefNotFoundError(IcechunkError)` — raised specifically for the missing-ref case? That'd let downstreams `except RefNotFoundError` cleanly and let other `IcechunkError`s continue to surface as real failures.

Same ask would apply to `lookup_branch_v2` / `lookup_tag` paths.

Concrete motivator: arraylake's booth tile/discovery service wants to return HTTP 404 for unknown branches/tags but 500 for genuine icechunk errors, and currently can't tell them apart without parsing the message.

li-em:
Fixes #2098 

All exceptions raised by Icechunk derive from `IcechunkError`, grouped into categories by what a handler would do about them. Every instance carries:
- `message`: the chain of causes followed by the operation context (also what `str(e)` returns). The full diagnostic report is attached as a PEP 678 note, visible in tracebacks.
- `kind`: a stable machine-readable code identifying the precise failure without needing one class per failure.

The tree:
```
    IcechunkError
    ├── ConflictError
    │   └── RebaseFailedError
    ├── NotFoundError                 (also a KeyError)
    │   ├── NodeNotFoundError
    │   ├── SnapshotNotFoundError
    │   ├── RefNotFoundError
    │   └── RepositoryNotFoundError
    ├── AlreadyExistsError
    ├── ReadOnlyError
    ├── InvalidInputError             (also a ValueError)
    ├── SessionStateError
    ├── StorageError
    ├── FormatError
    └── InternalError
```

## Compatibility

Existing handlers keep working:
- No string matching needs to change.
- everything that raised `IcechunkError` is still catchable with the same error, but can opt-in for more specific ones.
- the paths that raised bare `KeyError`/`ValueError` now raise subclasses of `IcechunkError` (`NotFoundError`, `InvalidInputError`), but can still be caught with `KeyError`/`ValueError`.
- `ConflictError`/`RebaseFailedError` keep their payloads and messages, but are also an `IcechunkError` now.

Important: `ConflictError` and `RebaseFailedError` are now also `IcechunkError` (they were disjoint before), so in code with both `except IcechunkError:` and `except ConflictError:` clauses, whichever clause comes first now wins — an `except IcechunkError:` listed first will swallow conflicts that previously fell through to the `except ConflictError:` clause.

## Uncaught exception output

Similar — same information, organized differently. Before, `str(e)` was the full miette graphical report (the ``x ... | context ... `->`` block), printed as the exception message. Now the traceback shows a specific class (`icechunk.RefNotFoundError` instead of `icechunk.IcechunkError`), `str(e)` is the plain cause chain plus the same operation-context spans, and the miette report still appears below the traceback as a PEP 678 note. Code that parses `str(e)` will see the framing change; code that prints it loses nothing. Example output in the [error handling guide](https://github.com/earth-mover/icechunk/blob/li-em/py_exc_tree/icechunk-python/docs/docs/guides/error-handling.md).


paraseba:
@li-em can you add in the PR description information about compatibility? What happens to user code if we merge? Can they miss exceptions they were catching? Do they need to change any string matching they were doing?

icechunk-python/python/icechunk/_exceptions.py
```diff
@@ -0,0 +1,236 @@
...
+    DUPLICATE_SNAPSHOT_ID = "duplicate-snapshot-id"
+    # internal
+    CONCURRENCY = "concurrency"
+    FLUSH = "flush"
+    PARTIAL_VALUES_PANIC = "partial-values-panic"
+    CONFLICTING_PATH_NOT_FOUND = "conflicting-path-not-found"
+    INCONSISTENT_MANIFESTS = "inconsistent-manifests"
+    POISON_LOCK = "poison-lock"
+    PICKLE = "pickle"
+    UNKNOWN = "unknown"
```
paraseba, line 158: How do we keep this in sync with the Rust code?

icechunk-python/python/icechunk/_exceptions.py
```diff
@@ -0,0 +1,236 @@
...
+    ConflictError,
+    IcechunkError,
+    RebaseFailedError,
+)
+
+
+class ErrorKind(StrEnum):
+    """Stable machine-readable codes for the `kind` attribute of [`IcechunkError`][icechunk.IcechunkError].
+
+    Codes are append-only: new ones may appear in any release, but existing
```
paraseba, line 43: Can you add an explanation of why we must not rename/remove?

paraseba:
Also, please add a few lines to the description explaining how this affects a normal python program with an uncaught exception. Will the printed error be the same? similar? different?

Changelog.python.md
```diff
@@ -2,6 +2,10 @@
 
 ## Python Icechunk Library [unreleased]
 
+### Features
+
+- Add an exception tree: all Icechunk errors now derive from `IcechunkError` and are grouped into catchable categories. Every exception carries a stable machine-readable `kind` code (see the new `icechunk.ErrorKind` enum), and the full diagnostic report is attached as a PEP 678 note ([#2267](https://github.com/earth-mover/icechunk/issues/2151)).
```
paraseba, line 7: Could we add a section to the documentation with the following:
* info about the exception tree,
* examples of catching the different types of exceptions
* example output for uncaught exceptions, `str` and `repr` of errors
* maybe the full error kind list?

icechunk-python/python/icechunk/_exceptions.py
li-em: There is [a test](https://github.com/earth-mover/icechunk/pull/2267/changes#diff-6505c39e8e785a9b4b190ff22452fcd93bde1c00828f902d7e178ba88cd1cdcaR50-R53) to check if a new one was added on the Rust side, and missing here. So at least the test is going to fail.

This could technically be defined completely in the Rust side and construct the Python enum programatically, but then we would need to move this block to `.pyi` for the type annotations (which PyO3 doesn't create automatically yet), which kind of defeats the purpose 😞 

The `UNKNOWN` catch-all is annoying, but dropping `#[non_exhaustive]` on the Rust enum is worse because then we have a semver-hazard.

icechunk-python/python/icechunk/_exceptions.py
li-em: Mostly because people will depend on it (example: `if error.kind == ErrorKind.CHUNK_NOT_FOUND:`), and so might miss it if we rename/remove

li-em:
> @li-em can you add in the PR description information about compatibility? What happens to user code if we merge? Can they miss exceptions they were catching? Do they need to change any string matching they were doing?

They will not miss any exceptions they were catching, but now they can opt to catch a more specific one.

With one caveat: now `ConflictError` and `RebaseFailedError` are also an `IcechunkError` (they were disjoint before), so code like this:
```
try:
  ...
except IcechunkError:
  # will catch this one now
except ConflictError:
  # will not be caught, because `IcechunkError` was caught first
```
will catch the more general `IcechunkError` first.

Can still keep the three top-level errors that exist now (`IcechunkError`, `ConflictError`, `RebaseFailedError`) and make the full change on IC 3.x too, since this is technically a breaking change.

paraseba (approved):

paraseba:
Fantastic work @li-em !!! Thank you!

Issue #2220, opened 2026-06-18
manifest.fbs documents a raw compressed_location encoding that isn't implemented


Shane98c:
Ok super minor but thought I'd report: 

The schema comment says `compression_algorithm = 0` means `compressed_location` holds **raw bytes**, but the Rust writer/reader never do that: uncompressed locations go in the plain `location` field, and `compressed_location` is only ever zstd-dict-compressed.

(Permalinks pinned to `02d6afac`.)

**Schema** — says `compressed_location` holds raw bytes for `0`:

https://github.com/earth-mover/icechunk/blob/02d6afac7ff1ace75422810f114b0b097887e8bf/icechunk-format/flatbuffers/manifest.fbs#L80-L81

**Writer** — sets one or the other, never raw bytes in `compressed_location`:

https://github.com/earth-mover/icechunk/blob/02d6afac7ff1ace75422810f114b0b097887e8bf/icechunk-format/src/manifest.rs#L954-L959

**Reader** — builds no decompressor unless `compression_algorithm == ZSTD_DICT`:

https://github.com/earth-mover/icechunk/blob/02d6afac7ff1ace75422810f114b0b097887e8bf/icechunk-format/src/manifest.rs#L464

…then errors (`MissingLocationCompressionDictionary`) on a `compressed_location` without one:

https://github.com/earth-mover/icechunk/blob/02d6afac7ff1ace75422810f114b0b097887e8bf/icechunk-format/src/manifest.rs#L733-L735

So `compression_algorithm = 0` + `compressed_location` would error, not decode raw bytes. The writer/reader behavior was introduced in #1776 and the current schema comment appears to have been added/updated in #2027.

**Suggested fix:** correct the comment to match the implementation, e.g.

```
// 0 = none: locations are stored uncompressed in ChunkRef.location.
// 1 = zstd dictionary compression (use location_dictionary).
```

Happy to send that one-liner. (Alternatively, implement the raw-bytes path in the writer/reader.)

Came across this while implementing this in https://github.com/EarthyScience/icechunk-js/pull/22

TomNicholas:
Thanks for spotting this! I think we would accept a fix to the schema comment.

paraseba:
Great catch @Shane98c ! We cannot update the reader/writers because we want to maintain backwards compatibility, but the comment is definitely wrong.

PR #2277, opened 2026-07-17
Fix manifest.fbs comment for compression_algorithm = 0

Shane98c:
Fixes the incorrect schema comment reported in https://github.com/earth-mover/icechunk/issues/2220: the comment on `compression_algorithm` claimed `0` means `compressed_location` holds raw bytes, but the implementation never does that. Uncompressed locations go in the plain `ChunkRef.location` field. This just updates the comment to match the implementation.


paraseba (approved):

paraseba:
Thank you @Shane98c

PR #2292, opened 2026-07-22
Select appropriate reduction for dask-array

### LINKED ISSUE #2291: Installing dask-array breaks icechunk's dask write path (extract_session: 'function' object has no attribute 'session')
### What happened?

Hit while testing https://matthewrocklin.com/frisky-xarray/#try-it-out on a real workload — an icechunk-era5 daily append, rechunking 24x12x12 to 8736x12x12 along `valid_time`. Reduced to the MVCE below.

`to_icechunk` on a dask-backed Dataset raises `AttributeError: 'function' object has no attribute 'session'` from `icechunk.distributed.extract_session` whenever [`dask-array`](https://github.com/mrocklin/dask-array) is merely **installed** in the environment. The script below never imports it, and `dask_array.xarray.register()` is not called — installing the package is enough, because it ships an `xarray.chunkmanagers` entry point named `dask` that xarray's entry-point scan picks up on its own.

`icechunk.dask.session_merge_reduction` calls `da.store(..., return_stored=True, compute=False)` and then `da.reduction(..., chunk=extract_session)`, reading `.session` off each stored `zarr.Array` block. With `dask-array` present the block that arrives instead has a *callable* `.store`, i.e. it's an array collection rather than the `zarr.Array` that `return_stored=True` is documented to yield — consistent with the `from_array which is already a Dask collection` warnings that precede the traceback.

Not sure whether the fix belongs here or in `dask-array`, so crossposted as https://github.com/mrocklin/dask-array/issues/11. Filing here because icechunk is the only consumer I know of that depends on the identity of `da.store(return_stored=True)` blocks.

### What did you expect to happen?

`to_icechunk` writes and commits, as it does when `dask-array` is not installed.

### Minimal Complete Verifiable Example

```Python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "icechunk",
#   "zarr",
#   "dask",
#   "xarray",
#   "dask-array",   # <-- installed but never imported; delete this line and the script passes
# ]
#
# [[tool.uv.index]]
#
# [tool.uv.sources]
# dask = { git = "https://github.com/dask/dask", branch = "main" }
# xarray = { git = "https://github.com/pydata/xarray", branch = "main" }
#
# [tool.uv]
# prerelease = "allow"
# ///
#
# dask/xarray from main per https://matthewrocklin.com/frisky-xarray/#try-it-out
# Fails identically on released icechunk==2.1.1, dask==2026.7.0, xarray==2026.4.0.

import tempfile

import numpy as np
import xarray as xr

import icechunk
from icechunk import Repository, local_filesystem_storage
from icechunk.xarray import to_icechunk

icechunk.print_debug_info()

repo = Repository.create(local_filesystem_storage(tempfile.mkdtemp()))
ds = xr.Dataset({"a": ("x", np.arange(100.0))}).chunk({"x": 10})

to_icechunk(ds, session, mode="w")  # AttributeError: 'function' object has no attribute 'session'
print("committed", session.commit("write"))
```

### MVCE confirmation

- [x] Minimal example — the example is as focused as reasonably possible to demonstrate the underlying issue in icechunk.
- [x] Complete example — the example is self-contained, including all data and the text of any traceback.
- [x] Verifiable example — the example copy & pastes into an IPython prompt, returning the result.
- [x] New issue — a search of GitHub Issues suggests this is not a duplicate.
- [x] Recent environment — the issue occurs with the latest version of icechunk and its dependencies.

### Relevant log output

```Python
.../dask/array/core.py:3641: UserWarning: Passing an object to dask.array.from_array which is already a Dask collection. This can lead to unexpected behavior.
  warnings.warn(
[... x6 ...]
  File "issue.py", line 34, in <module>
    to_icechunk(ds, session, mode="w")
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File ".../icechunk/xarray.py", line 354, in to_icechunk
    forked_session = writer.write_lazy(
        chunkmanager_store_kwargs=chunkmanager_store_kwargs,
        split_every=split_every,
    )
  File ".../icechunk/xarray.py", line 206, in write_lazy
    return session_merge_reduction(stored_arrays, split_every=split_every)
  File ".../icechunk/dask.py", line 130, in session_merge_reduction
    *da.compute(*merged_sessions)  # type: ignore[no-untyped-call]
     ~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File ".../dask/base.py", line 769, in compute
    results = schedule(expr, keys, **kwargs)
  File ".../icechunk/dask.py", line 41, in wrapper
    return func(*args, **kwargs)
  File ".../icechunk/distributed.py", line 53, in extract_session
    return store.session
           ^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'session'
```

### Anything else we need to know?

A/B in a single venv, only `dask-array` changing (`uv run issue.py` with and without the `dask-array` dependency line):

| env | icechunk 2.1.1-dev17+g8a42c350 | icechunk 2.1.1 |
| --- | --- | --- |
| `dask-array` installed, no `register()` | fails | fails |
| `dask-array` installed + `dask_array.xarray.register()` | fails | fails |
| `dask-array` not installed | commits | commits |
| `dask-array` not installed, `frisky==0.7.0` `hijack(Client())` | commits | commits |

Same in both dask/xarray combinations (`main` and released `dask==2026.7.0` / `xarray==2026.4.0`), so it is not a released-vs-`main` regression. `frisky` on its own is fine — only `dask-array` triggers it.

One asymmetry that may help: on `xarray@main`, `guess_chunkmanager("dask")` returns `dask_array._xarray.DaskArrayExprManager`, so the entry point clearly wins. On released `xarray==2026.4.0` the same call still returns the builtin `xarray.namedarray.daskmanager.DaskManager`, yet `to_icechunk` fails identically — suggesting an import-time side effect in addition to the manager swap.

Happy to test a patch.

### Environment

<details>

```
platform:  macOS-26.5.2-arm64-arm-64bit-Mach-O
python:  3.13.13
icechunk:  2.1.1-dev17+g8a42c350
zarr:  3.2.2.dev114+gecc2d777
numcodecs:  0.16.5
xarray:  2026.7.1.dev6+gadc8005a2
dask:  2026.7.0.post6+ga54329156
dask-array:  0.7.0
```

</details>

li-em:
Fix #2291

### DIFF
--- a/icechunk-python/python/icechunk/dask.py
+++ b/icechunk-python/python/icechunk/dask.py
@@ -1,4 +1,5 @@
 import functools
+import sys
 from collections.abc import Callable, Mapping
 from typing import TYPE_CHECKING, Any
 
@@ -103,16 +104,26 @@ def store_dask(
     return session_merge_reduction(stored_arrays, split_every=split_every, **store_kwargs)
 
 
+def _reduction_for(arr: Any) -> Any:
+    # dask-array's Array isn't a da.Array; stock da.reduction re-wraps it via from_array.
+    if isinstance(arr, da.Array):
+        return da.reduction
+    dask_array = sys.modules.get("dask_array")
+    if dask_array is not None and isinstance(arr, dask_array.Array):
+        return dask_array.reduction
+    raise TypeError(f"Expected a dask array, received {type(arr)!r}")
+
+
 def session_merge_reduction(
     arrays: Array | list[Array], *, split_every: int | None, **store_kwargs: Any
 ) -> ForkSession:
-    if isinstance(arrays, da.Array):
+    if not isinstance(arrays, (list, tuple)):
         arrays = [arrays]
     # Now we tree-reduce all changesets
     # reduce the individual arrays since concatenation isn't always trivial due
     # to different shapes
     merged_sessions = [
-        da.reduction(  # type: ignore[no-untyped-call]
+        _reduction_for(arr)(
             arr,
             name="ice-changeset",
             chunk=computing_meta(extract_session),
--- a/icechunk-python/tests/test_dask.py
+++ b/icechunk-python/tests/test_dask.py
@@ -5,12 +5,14 @@ pytest.importorskip("distributed")
 
 import tempfile
 
+import numpy as np
+
 import dask
 import distributed
 import xarray as xr
 import zarr
 from icechunk import Repository, local_filesystem_storage
-from icechunk.dask import store_dask
+from icechunk.dask import session_merge_reduction, store_dask
 from icechunk.xarray import to_icechunk
 from tests.test_xarray import create_test_data, roundtrip
 from xarray.testing import assert_identical
@@ -42,6 +44,11 @@ def test_store_dask(any_spec_version: int | None) -> None:
         zarray = zarr.open_array(fork.store, path="array")
         with dask.config.set(scheduler="processes"):
             store_dask(sources=[dask_array], targets=[zarray])
+
+
+def test_session_merge_reduction_rejects_unknown_arrays() -> None:
+    with pytest.raises(TypeError, match="Expected a dask array"):
+        session_merge_reduction(np.arange(10), split_every=None)  # type: ignore[arg-type]
 
 
 def test_distributed(any_spec_version: int | None) -> None:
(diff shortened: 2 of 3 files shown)


aaronspring:
thank you. works for me as far as I can see

dcherian (approved):
LGTM. thanks!

PR #2293, opened 2026-07-24
Remove unused Sphinx dependencies and config

VeckoTheGecko:
There were a bunch of Sphinx-only dependencies and config being included for no reason. This PR removes them and regenerates the `pixi.lock`.

Did some spelunking to find out how they were introduced since the project has never used Sphinx - and from what I can tell they're from https://github.com/earth-mover/icechunk/pull/832 (likely due to an LLM halucination when reading the old `poetry.lock` file, since they're mentioned there but weren't being pulled in).

Doing a `pixi ls | grep sphinx` is quick validation.

(🚀 my first icechunk contrib :) )

### DIFF
--- a/icechunk-python/.gitignore
+++ b/icechunk-python/.gitignore
@@ -60,9 +60,6 @@ coverage.xml
 
 .DS_Store
 
-# Sphinx documentation
-docs/_build/
-
 # PyCharm
 .idea/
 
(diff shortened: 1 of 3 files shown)


li-em (approved):

li-em:
Thanks @VeckoTheGecko !

(🚀 the first of many contribs, I'm sure 😁 )

PR #2294, opened 2026-07-24
Fix upstream-dev CI job: zarr test overrides, stateful testing, upstream GHA

li-em:
Couple of different problems detected in #1921 fixed on this PR:

- [scientific-python/issue-from-pytest-log-action](https://github.com/scientific-python/issue-from-pytest-log-action/pull/74) was not uploading the issue body anymore, fixed upstream, bumped version here.

- #2175 removed our zarr tests overrides, but that leaves us in an inconsistent state while zarr 3.3 (or the current `main`) is not released, and we bump our minimum supported zarr. Opted to checking if the `_get_*` exist instead of a version range check, and we can actually drop the overrides once we move newer zarr releases

## stateful testing fixes

- `test_zarr/test_stateful.py`: use the storage-key chunk grid in `shift_array`. The rule used `cdata_shape`, which counts *inner* chunks for sharded arrays, while store keys and `Session.shift_array` operate on the outer (shard) grid. zarr main's hypothesis strategies now generate sharded arrays, exposing this: the model shuffled keys over a grid that doesn't exist (e.g. producing `0/c/1` for a single-shard array). A new `storage_chunk_sizes` helper uses zarr's `write_chunk_sizes` when available and derives the grid from `shards or chunks` on older zarr. Includes deterministic regression tests, verified against both zarr 3.1.6 and zarr nightly.

- `test_stateful_repo_ops.py`: match the model's GC re-parenting to Rust. The model re-parented kept snapshots whose parent was GC'd to the *initial* snapshot, but `delete_snapshots_from_repo_info` re-parents over the deleted run to the *nearest surviving ancestor*. After a `reset_branch` back onto such a snapshot, model and repo disagreed on reachability, and the next `garbage_collect` predicted the wrong `snapshots_deleted` count. Triggered only when a GC deletes a middle run while older (ref-pointed) and newer (`flushed_at`-protected) snapshots both survive — hence the intermittent nightly appearances.

- `test_stateful_repo_ops.py`: guard `create_branch`/`create_tag` against V1 expired-ancestor chains. V1 expiration rewrites only ref-reachable chains, so an unreachable commit's on-disk ancestry can still pass through expired snapshots the model dropped; creating a ref there resurrects history the model no longer tracks and fails `check_ancestry`. Applies the same full-ancestry `assume()` that `reset_branch` already uses.

### DIFF
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -930,8 +930,14 @@ class VersionControlStateMachine(RuleBasedStateMachine):
         note(f"Creating branch {name!r} for commit {commit!r}")
 
         # V1 expired snapshots stay on disk so create_branch succeeds, but
-        # the model no longer tracks their store contents.
-        assume(not (self.model.spec_version == 1 and commit not in self.model.commits))
+        # the model no longer tracks their store contents. This also covers a
+        # commit whose ancestry *reaches* an expired-but-on-disk snapshot: the
+        # real ancestry resurrects it while the model has popped it, so skip
+        # unless the whole chain is still modelled.
+        if self.model.spec_version == 1:
+            assume(commit in self.model.commits)
+            ancestry = {s.id for s in self.repo.ancestry(snapshot_id=commit)}
+            assume(ancestry <= set(self.model.commits))
 
         # we can create a tag and branch with the same name
         if name not in self.model.branch_heads and commit in self.model.commits:
@@ -950,8 +956,14 @@ class VersionControlStateMachine(RuleBasedStateMachine):
     def create_tag(self, name: str, commit_id: str) -> str:
         note(f"Creating tag {name!r} for commit {commit_id!r}")
         # V1 expired snapshots stay on disk so create_tag succeeds, but
-        # the model no longer tracks their store contents.
-        assume(not (self.model.spec_version == 1 and commit_id not in self.model.commits))
+        # the model no longer tracks their store contents. This also covers a
+        # commit whose ancestry *reaches* an expired-but-on-disk snapshot: the
+        # real ancestry resurrects it while the model has popped it, so skip
+        # unless the whole chain is still modelled.
+        if self.model.spec_version == 1:
+            assume(commit_id in self.model.commits)
+            ancestry = {s.id for s in self.repo.ancestry(snapshot_id=commit_id)}
+            assume(ancestry <= set(self.model.commits))
         if (
             name in self.model.created_tags
             or name in self.model.tags
--- a/icechunk-python/tests/test_stateful_repo_ops.py
+++ b/icechunk-python/tests/test_stateful_repo_ops.py
@@ -556,6 +556,7 @@ class Model:
             if created_at_by_id[k] < older_than
         }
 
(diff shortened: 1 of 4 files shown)


icechunk-python/tests/test_stateful_repo_ops.py
```diff
@@ -989,14 +1012,10 @@ def discard_changes(self) -> None:
     @rule(branch=branches, commit=commits)
     def reset_branch(self, branch: str, commit: str) -> None:
         # V1 expired snapshots stay on disk so reset_branch would succeed,
-        # but modelling that divergence isn't worthwhile — just skip. This also
-        # covers resetting to a commit whose ancestry *reaches* an expired-but-
-        # on-disk snapshot: the real ancestry resurrects it while the model has
-        # popped it, so skip unless the whole chain is still modelled.
+        # but modelling that divergence isn't worthwhile — just skip.
```
ianhi, line 1015: is this comment no longer true?

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -28,12 +29,31 @@
...
+def storage_chunk_sizes(arr: "Array[Any]") -> tuple[tuple[int, ...], ...]:
+    """Per-dimension sizes of the storage-key chunk grid (shards when sharded).
+
+    Store keys and icechunk's shift_array operate on this grid, not on the
+    inner (read) chunks that cdata_shape/read_chunk_sizes count for sharded
+    arrays. Older zarr lacks write_chunk_sizes but also lacks rectilinear
+    grids, so cells there are regular and can be computed directly.
+    """
+    if hasattr(arr, "write_chunk_sizes"):
+        return arr.write_chunk_sizes  # type: ignore[no-any-return]
```
ianhi, line 52: Can we leave a comment about why we are usign write (outer) instead of read (inner) chunk sizes. I got pretty tripped up by this at first

icechunk-python/tests/test_zarr/test_stateful.py
```diff
@@ -398,3 +411,59 @@ def test_zarr_store() -> None:
...
+        model, name="0", shape=(4,), shards=(4,), chunks=(2,), dtype="i1", fill_value=1
+    )
+    arr_store = zarr.create_array(
+        store, name="0", shape=(4,), chunks=(4,), dtype="i1", fill_value=1
+    )
+    data = np.zeros((4,), dtype="i1")
+    arr_model[:] = data
+    arr_store[:] = data
+
+    model_keys = sorted(collect_aiterator(model.list_prefix("")))
```
ianhi: if you asyncify this test then we can drop the `collect_aiterator` import i think

ianhi (approved):
little headspinning to thnk about GC once more - but all seems correct to me!

icechunk-python/tests/test_stateful_repo_ops.py
li-em: it got moved to the `_v1_chain_fully_modelled` docstring

icechunk-python/tests/test_zarr/test_stateful.py
li-em: Expanded the docstring, it is indeed confusing 😵

PR #2295, opened 2026-07-24
More GHA caching, clean remaining zizmor warnings

li-em:
Another round of CI cleanup:

- fix zizmor warnings for the workflows that were still ignored. Now they are all passing
- Tweak GHA caching, avoid cancelling runs on main because they would not save updated ones, and churn due to env vars from `setup-pixi`
- when running upstream tests in a PR (triggered by using the `test-upstream` label) use the `ci` hypothesis profile instead of `nightly` (which takes hours to run)


paraseba (approved):
I have no idea what this does, but I trust you to double check it works

PR #2299, opened 2026-07-24
CLI improvements (`branch`, `tag`, `ancestry` subcommands)

DahnJ:
Continuation of #1544  from @jleben

This only implements a subset of the commands from #1544 to keep the PR smaller. I will later open a subsequent PR with the rest.

```
branch list <repo
branch create <repo> <branch> <from ref>
branch delete <repo> <branch>
tag list <repo>
tag create <repo> <tag> <target ref>
tag delete <repo> <tag>
ancestry <repo> [<ref>]
```

Changes
- Extends and reuses `inspect.rs` ([comment](https://github.com/earth-mover/icechunk/pull/1554#discussion_r2688555371))
- Add tests

### DIFF
--- a/icechunk/src/cli/interface.rs
+++ b/icechunk/src/cli/interface.rs
@@ -2,7 +2,7 @@
 
 use crate::format::ChunkIndices;
 use crate::format::snapshot::{DimensionName, NodeData};
-use crate::repository::VersionInfo;
+use crate::repository::{RepositoryErrorKind, VersionInfo};
 use chrono::Local;
 use clap::{Args, Parser, Subcommand};
 use dialoguer::{Input, Select};
@@ -68,8 +68,7 @@ struct HistoryArgs {
     repo: RepositoryAlias,
     #[arg(
         name = "reference",
-        default_value = "main",
-        help = "Branch, tag or snapshot to show history for."
+        help = "ID of snapshot to show history for"
     )]
     reference: String,
     #[arg(short = 'n', default_value_t = 10, help = "Number of snapshots to list")]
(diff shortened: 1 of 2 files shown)


icechunk/src/cli/interface.rs
```diff
@@ -213,14 +281,45 @@ async fn get_storage(
...
+    }
+    if let Some(snapshot_id) = repository.lookup_branch(reference).await.ok() {
+        return Ok(snapshot_id);
+    }
+    if let Some(snapshot_id) = repository.lookup_tag(reference).await.ok() {
+        return Ok(snapshot_id);
+    }
+    Err(anyhow::anyhow!("`{reference}` is not a valid snapshot id, branch, or tag"))
+}
+
```
DahnJ, line 314: Right, I just read [this comment](https://github.com/earth-mover/icechunk/pull/1554#discussion_r2688553000)

> for now I'd prefer to keep the CLI as logic-free as possible [..] Then we can add this logic to the main rust code and extend support in the python library and CLI

Happy to roll this back to just accepting snapshot id for now, but my immediate reaction to testing the CLI out was "can I just use `main`". Would be interested in adding this functionality into the core library to reuse here.

icechunk/src/cli/interface.rs
```diff
@@ -229,29 +328,130 @@ async fn repo_create(init_cmd: &CreateCommand, config: &CliConfig) -> Result<()>
...
+async fn list_tags(
+    args: &TagListArgs,
+    config: &CliConfig,
+    mut writer: impl std::io::Write,
+) -> Result<()> {
+    let repository = open_repository(&args.repo, config).await?;
+    let tags = repository.list_tags().await.context("Failed to list tags")?;
+    for tag in tags {
```
paraseba, line 381: would be nice to run these in parallel, for repos with many tags

paraseba (approved):
Looks great!

PR #2305, opened 2026-07-28
Reduce ReadTheDocs build time, CI cache sizes

li-em:
Couple independent changes, but too small for a PR for each one.

RTD was failing lately because it was hitting the build time limit. Using `opt-level = 0` there which should be much faster, and we don't really need any optimization for docs. Should be ~65% faster and stay inside limits.

Set `incremental = false` and `strip = "symbols"` to reduce cache size, they should be ~30% smaller now. Incremental doesn't help much in CI (better for local dev), and `strip` doesn't mess with coverage or any other checks we do in CI.

Partition the hypothesis shards explicitly, instead of using `pytest-shard`. It was too imbalanced, and hypothesis was the longest running part of CI. A fallback catches any new hypothesis tests, and we can rebalance again later.

Bring the `README` pre-commit check from #2144 so `icechunk-python/README.md` and `README.md` stay in sync. They can't be symlinks without generating issues, so at least this catches inconsistencies.

### DIFF
--- a/.github/workflows/python-check.yaml
+++ b/.github/workflows/python-check.yaml
@@ -224,7 +224,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shard-id: [0, 1, 2]
+        # Explicit partition keyed on test cost: each heavy stateful machine
+        # anchors a shard, cheap property tests ride along on the second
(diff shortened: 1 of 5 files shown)


ianhi (approved):
sweet

paraseba (approved):

PR #2307, opened 2026-07-28
Stop mkdocs-jupyter from probing plain markdown pages

li-em:
mkdocs-jupyter was using pandoc (not installed locally in pixi/nix dev envs, but available in RTD) to process all the markdown files in the docs, which caused longer running times in RTD.

There is only one notebook (`glad-ingest.ipynb`) that needs `mkdocs-jupyter`, and nothing in the markdown files, so ignore them.


ianhi (approved):
nice. we could also just fully convert this notebook to markdownexec so we are consistent all over. but this is a great callout to speed thing up

PR #2310, opened 2026-07-31
Use storage `created_at` instead of host `flushed_at` during GC

li-em:
First noticed this when using macOS for development, now it also appeared on a nightly stateful test (VersionControlTest, run [30410732472](https://github.com/earth-mover/icechunk/actions/runs/30410732472/job/90445972311)).

On macOS it was caused by drift in the VM clock where the rustfs container was running (mostly from closing the laptop lid and opening again later). #2167 fixed the failing test, but not the underlying problem.

Add a test mirroring the hypothesis example that triggered the bug.

---

`garbage_collect_one_attempt` does repo-info retention of unreachable snapshots on the host-stamped `flushed_at`, while file deletion uses the storage `created_at` (LIST mtime). A snapshot whose timestamps straddle the cutoff was half-deleted: dropped from the repo info (losing its `pruned_ancestor_tx_logs` references, whose transaction logs were then deleted), while its snapshot file survived on disk.

### DIFF
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Garbage collection judged repo-info retention of an unreachable snapshot by its host-stamped `flushed_at`, but file deletion by the storage `created_at`. A snapshot falling between the two timestamps was dropped from the repo info — deleting the transaction logs its `pruned_ancestor_tx_logs` still referenced — while its snapshot file survived. Both decisions now use the storage `created_at` ([#2310](https://github.com/earth-mover/icechunk/pull/2310)).
+- Fix garbage collection deleting still-referenced transaction logs when the host and object-store clocks are skewed ([#2310](https://github.com/earth-mover/icechunk/pull/2310)).
 
 ## Python Icechunk Library 2.1.2
 
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Python Icechunk Library [unreleased]
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


icechunk/src/ops/gc.rs
```diff
@@ -411,12 +411,35 @@ async fn garbage_collect_one_attempt(
...
+        // half-deletes a snapshot in the (flushed_at, created_at] window: it is
+        // dropped from the repo info (losing its pruned_ancestor_tx_logs
+        // references) while its file survives.
+        let created_at_by_id: HashMap<SnapshotId, DateTime<Utc>> =
+            if config.deletes_snapshots() {
+                asset_manager
+                    .list_snapshots()
+                    .await?
+                    .inspect_err(|e| error!("Listing snapshots: {e}"))
+                    .filter_map(|snap| ready(snap.ok().map(|s| (s.id, s.created_at))))
```
paraseba: I think this is swallowing errors. Using `try_collect()?` would be safer.

li-em: 🤦 yup

Changelog.python.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Python Icechunk Library [unreleased]
+
+### Fixes
+
+- Garbage collection judged repo-info retention of an unreachable snapshot by its host-stamped `flushed_at`, but file deletion by the storage `created_at`. A snapshot falling between the two timestamps was dropped from the repo info — deleting the transaction logs its `pruned_ancestor_tx_logs` still referenced — while its snapshot file survived. Both decisions now use the storage `created_at` ([#2310](https://github.com/earth-mover/icechunk/pull/2310)).
```
paraseba: I'd make this description much shorter, just something along the lines of fix GC in systems with heavily skewed clocks, or whatever.

paraseba (approved):

PR #2311, opened 2026-07-31
Limit what github actions are cached

li-em:
- set `cache: false` for all `setup-pixi` uses, it is actually faster to let pixi rebuild them than restoring the cache from GHA...
- disable rust cache in `zarrs-upstream`, since the icechunk target wasn't being reused anyway


paraseba (approved):
thank you @li-em !!!!!

PR #2316, opened 2026-08-04
Reproduce upstream test failure

li-em:
Test reproducing the [failure](https://github.com/earth-mover/icechunk/actions/runs/30865286687/job/91855674463) in #1921


li-em:
Fixed in #2349

PR #2317, opened 2026-08-05
Design document for better I/O control

paraseba:


design-docs/019-better-io-control.md
```diff
@@ -0,0 +1,131 @@
+# Better I/O Control
+
+Scope: changes to Icechunk to more easily control its I/O behavior and
+improve its I/O performance when it doesn't fully own the machine.
```
rabernat, line 4: The meta question for me is: which layer of the stack should implement this type of I/O logic? 

A big problem with the Python stack is ambiguous delegation of responsibility around concurrency.

In zax-prototype, I found it very useful to _have the compute engine manage the I/O_. This meant turning off all of Icechunk's concurrency limits and orchestrating its own I/O logic. This makes sense to me; the higher up the stack, the more awareness is possible of the full context in which a request is happening.

design-docs/019-better-io-control.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Problems with the current architecture
+
+We are thinking about server-side use cases for Icechunk. Notice these are
+not the only problems, these are only the problems we are attacking here.
```
rabernat: I don't understand this sentece.

design-docs/019-better-io-control.md
```diff
@@ -0,0 +1,131 @@
...
+- The only repo-wide gate to I/O is `AssetManager::request_semaphore`
+  (default 256), fixed at construction, counting *logical* fetches. Each
+  logical fetch can fan out to `max_concurrent_requests_for_object` GETs.
+- Virtual chunk reads (`VirtualChunkResolver` → `ChunkFetcher`) bypass the
+  semaphore entirely.
+- No ability to change limits at runtime: concurrency is fixed when the
+  Repository is built, so an external scheduler cannot grow or shrink a
+  running workload's I/O share. Even more, `reopen` doesn't allow to change.
+- No priorities: manifest fetches are prioritized together with chunks, even
+  if they arguably can block much more work.
```
rabernat: 👍  This is quite important I think.

design-docs/019-better-io-control.md
```diff
@@ -0,0 +1,131 @@
...
+        if let Some(s) = self.inner.take() { s.abort() }
+    }
+}
+```
+
+Acquisition happens per HTTP request, *after* range-splitting: a large
+object fetch acquires once per part, not once per logical fetch,
+and the virtual-chunk fetchers consult the same governor.
+
+The governor is injected at `Repository::open` time, and carried in
```
rabernat: Why should the governor be attached to the Repo? Wouldn't a dedicated context (potentially shareable across many repos) be more useful?

rabernat (commented):
Overall I like the direction of this. It seems compatible with the AWS transfer manager "bandwidth-weighted token bucket" that I implemented to great effect in zax-prototype (see [transfer manager research](https://github.com/earth-mover/zax-prototype/blob/main/performance_analysis/2026-06-13-transfer-manager-architecture.md), [token_bucket.rs](https://github.com/earth-mover/zax-prototype/blob/main/zax-io-engine/src/token_bucket.rs)). Clearly this type of logic is needed _somewhere in the stack_ in order to achieve optimal I/O performance.

I am missing a bit of context, however, about where exactly this fits in to the bigger picture. Without this context, I have many questions that may or may not be relevant.

My main question is structural: why should this be attached to a Repo? If the use case is to optimize I/O on the zax-backend, it seems like what we really want is a dedicated I/O governor for _the entire server_, since everything running there has to share a network connection. Furthermore, we might want to do I/O against non-icechunk things, like vanilla Zarr stores (well supported in zax-prototype). 

So my recommendation is to add some context about how we will use this specifically for zax server.

design-docs/019-better-io-control.md
paraseba: The goal is pluggable governors. Icechunk should have a default one that is great and able to work without extra information or context. Server side users my build their own because they know more about the context, for example, that they are running concurrent Icechunk requests for multiple repos.

Similar pluggability needs to be applied at other layers, for example, asset caches.

design-docs/019-better-io-control.md
paraseba: Yes, you could feed the same governor to many repos. Imagine a main governor from which you can `split` sub-governors.

design-docs/019-better-io-control.md
dcherian: >  This makes sense to me; the higher up the stack, the more awareness is possible of the full context in which a request is happening.

It has some context yes, but is missing context like prioritizing manifests.

PR #2318, opened 2026-08-06
Reuse manifests written by failed commit attempts when retrying after rebase

aldenks:
This is a PR as a proposal, I'm going to close it right away, but am using it to float the idea. The current implementation exists only to give an idea of what an implementation would look like.

**Goal: make uncoordinated but semi-cooperative writes work with higher writer concurrency and/or many arrays per store.**

Uncooperative writes have infrastructure and code simplicity benefits for writers. These uncooperative writers don't directly communicate, _but it is often relatively simple to distribute work between them so they each write to disjoint regions_ which guarantees that `commit("...", rebase_with=ConflictDetector())` will succeed given enough rebase attempts.

Right now this works well up until the point where the time to rebase and retry a commit exceeds the total attempted rate of commits across writers. More concurrent writers increase contention, more arrays per commit increase the time a rebase+retry takes. When this threshold is crossed writers get stuck in a loop rewriting manifests and never winning the race to land their commit.

By reducing the amount of time it takes to rebase and retry a commit that lost the compare-and-set race to another writer, we make this semi-cooperative writer case scale to much higher concurrent writers and/or more updated arrays per commit.

- Right now: a retried commit remerges, recompresses, and reuploads all manifests in the commit, increasing the time to rebase+retry
- Proposal: only re merge, compress, and upload manifests changed by other writers. reuse manifests that don't need to change. 

**Question: is the added complexity of this worth it?** If so, I'm happy to refine this PR or open an issue.

One anecdotal example: in a workflow where each commit touches ~150 small (< 1MB) manifests with ~5 minutes of work on each writer between commits, if I go any higher than 10 concurrent writers they all get stuck in rebase loops and never succeed. This isn't terribly impactful right now, 10 writers is ok.

Because pickling ForkSession serializes static creds, using that would require another communication mechanism outside of a publicly accessible bucket.

### DIFF
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -2,6 +2,10 @@
 
 ## Python Icechunk Library [unreleased]
 
+### Features
+
+- Retrying a commit that lost the compare-and-set race against a concurrent committer (via `commit(rebase_with=...)` or `rebase()` followed by a new `commit()`) no longer regenerates every manifest. Manifests written by the failed attempt are reused for arrays the concurrent commit didn't touch and whose session changes the rebase didn't patch, so losing a commit race is now proportionally cheaper for uncoordinated writers working on disjoint arrays ([#2318](https://github.com/earth-mover/icechunk/pull/2318)).
+
 ### Fixes
 
 - Fix garbage collection deleting still-referenced transaction logs when the host and object-store clocks are skewed ([#2310](https://github.com/earth-mover/icechunk/pull/2310)).
--- a/icechunk/src/session.rs
+++ b/icechunk/src/session.rs
@@ -16,8 +16,9 @@ use itertools::{Itertools as _, enumerate, repeat_n};
 use regex::bytes::Regex;
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, hash_map::DefaultHasher},
     future::{Future, ready},
+    hash::Hasher,
     ops::Range,
     pin::Pin,
     sync::Arc,
@@ -497,6 +498,11 @@ pub struct Session {
     snapshot_id: SnapshotId,
     change_set: ChangeSet,
     default_commit_metadata: SnapshotProperties,
+    // Manifests written by a failed commit attempt of this session, reusable
+    // by the next attempt. Only meaningful within the process that wrote
+    // them, so it's not serialized with the session.
+    #[serde(skip)]
+    flushed_manifests: FlushedManifestCache,
 }
 
 impl Session {
@@ -522,6 +528,7 @@ impl Session {
             snapshot_id,
             change_set: ChangeSet::for_edits(),
             default_commit_metadata: SnapshotProperties::default(),
+            flushed_manifests: FlushedManifestCache::default(),
         }
     }
 
@@ -555,6 +562,7 @@ impl Session {
             snapshot_id,
             change_set: ChangeSet::for_edits(),
             default_commit_metadata,
+            flushed_manifests: FlushedManifestCache::default(),
         }
     }
 
@@ -583,6 +591,7 @@ impl Session {
             snapshot_id,
             change_set: ChangeSet::for_rearranging(),
             default_commit_metadata,
+            flushed_manifests: FlushedManifestCache::default(),
         }
     }
 
@@ -1621,6 +1630,10 @@ impl Session {
             self.snapshot_id(),
             self.config.manifest(),
         );
+        // Anonymous flushes don't race on a branch, so they are never
+        // retried; there is no reason to keep the manifests they write
+        // available for reuse.
+        let mut flushed_manifests = FlushedManifestCache::default();
         let new_snap = do_flush(
             flush_data,
             message,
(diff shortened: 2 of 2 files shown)


aldenks:
AI written details:



## Problem

When a commit loses the compare-and-set race on a branch, the flush that preceded the CAS has already fetched, merged, compressed and uploaded a new manifest for **every** node with changes in the session. Retrying the commit — via `commit().rebase(...)` or a manual `rebase()` + `commit()` — re-does all of that manifest work from scratch, even for nodes the concurrent committer never touched (the previously uploaded manifests just become garbage, since manifest ids are random, not content-addressed).

For semi-coordinated writers — processes intentionally writing to disjoint regions/arrays without serializing their commits (the "uncooperative writes" pattern from the [parallel writes guide](https://icechunk.io/en/latest/understanding/parallel/)) — the cost of losing the race is therefore proportional to the full size of the session's changes rather than to the actual overlap. This is painful for stores with hundreds of arrays (hundreds of manifests) per commit.

## Approach

Manifests are immutable objects with random ids: the ones written by a failed attempt are still in storage, merely unreferenced. The retried flush can therefore reference them instead of regenerating them, **whenever it can prove they'd be identical**.

The proof is a per-node fingerprint of every input that determines the node's manifests:

- the manifest refs the node carries in the parent snapshot — these change if the concurrent commit touched the node;
- the manifest split extents — these cover shape/dimension-name changes (splitting config is immutable for the life of a session);
- the session's chunk changes for the node, iterated in deterministic `BTreeMap` order — these change if the rebase conflict solver patched them (e.g. `UseTheirs` dropping chunks), or if the user wrote more chunks after a failed commit.

The fingerprint is two independent 64-bit hashes plus total serialized length, computed by streaming the msgpack encoding of those inputs through the hashers (no buffering, no clone of the change set). Because the comparison is per-node between two attempts of the *same* session — not adversarial, not many-to-many — an accidental collision (~2⁻¹²⁸ per changed node per retry, with a length guard on top) is not a practical concern. Any mismatch falls back to the existing full merge path, so a false *miss* only costs the work we do today.

Mechanics:

- `Session` gains a `#[serde(skip)]` `FlushedManifestCache` (node id → fingerprint + manifest refs/file infos). It is replaced wholesale by each flush with what that flush produced, and cleared on successful commit.
- On the next flush attempt, `flush_existing_node` / `flush_new_node` recompute the fingerprint from the (possibly new) parent snapshot and the (possibly patched) change set; on a hit they return the cached refs without any manifest fetches or writes.
- Reuse is disabled for `rewrite_manifests` (user explicitly wants fresh manifests) and anonymous flushes (they never race on a branch).
- Since the cache lives in the session, both the `commit().rebase(...)` retry loop and the manual `commit()`-fails → `rebase()` → `commit()` flow benefit, with no public API changes. Python/JS bindings pick it up automatically.

One caveat worth noting: reuse assumes the failed attempt's manifests are still in storage when the retry runs (seconds later). This is the same exposure an in-flight commit already has to a concurrently running GC, and it is covered by GC's documented age cutoff: per [design doc 011](https://github.com/earth-mover/icechunk/blob/main/design-docs/011-IC2-gc-and-expiration-consistency.md) and the expiration guide, GC only deletes objects created strictly before the user-supplied cutoff timestamp.

## Commit structure

1. **Tests first** (`Add test coverage for commit-with-rebase manifest and data behavior`): fills coverage gaps in the exact areas the optimization touches, all passing on the unmodified implementation — disjoint-array writers through `commit().rebase()` with manifest-identity assertions, chunk conflicts (`UseOurs`/`UseTheirs`) through the retry loop, session usability + extra writes after a failed commit, and data correctness across multiple racing rebase attempts. CI on this commit demonstrates the pinned behavior predates the optimization.
2. **Optimization** (`Reuse manifests written by failed commit attempts when retrying after rebase`): the change above, plus tests that count manifest objects in storage to prove reuse happens (exactly one manifest per unchanged array across any number of retries, in both the builder and manual flows) and that it is correctly disabled when inputs change (concurrent writes to the same array get merged; chunks dropped by `UseTheirs` don't resurrect the stale manifest). All tests run over both spec versions.
3. **Changelog** (`Add changelog entry for manifest reuse on commit retry`): entry under `[unreleased]` in `Changelog.python.md`, per repo convention. The entry currently links this fork PR; swap the link for the upstream PR number if/when this is opened upstream.

## Testing

- `cargo test -p icechunk --lib`: 247 passed.
- `cargo test -p icechunk --test main concurrency` and in-memory `gc` integration tests pass (MinIO-dependent tests fail in my environment for lack of a local MinIO — connection refused on `127.0.0.1:4200` — unrelated to this change).
- `cargo clippy -p icechunk --lib --tests` clean, `cargo fmt` applied, `cargo check --workspace --tests` clean.

aldenks:
@li-em @paraseba Running this by you as an idea (see description). Going to close it to not clog up your PRs. My question: is the added complexity of this worth it in your eyes? If so, I'm happy to refine this PR or open an issue.

paraseba:
This sounds extremely interesting @aldenks ! Any chance you could write a short design document in the `design-docs` dir? Either way I'll review and have some feedback before Monday

aldenks:
@paraseba definitely will do, its probably going to be next week by the time i'll get to the design doc

Issue #2325, opened 2026-08-11
Add configurable chunk directory layout


mschou1306:
### Problem

Icechunk puts all chunks in a single `chunks/` folder and, as has also been described in #1867, some file systems run into issues (or even hard limits) with larger number of files in one folder. 
Object storage backends generally do not impose per-directory file-count limits because `chunks/` is a key prefix rather than a real directory. However, local filesystem storage remains important for deployments that cannot use object storage, so the layout should remain configurable at the repository level.

### Proposed Solution

Add an **optional** config setting to control the chunk directory layout. This is set at the repository level, and is used whenever we access chunks (read, write, list, delete etc.). Any existing repositories will continue using the existing flat layout structure.

A simple alternative to the flat structure is using an ID prefix (so chunk `ABCD` would be placed in `chunks/AB/ABCD`). 
Alternatively, we could possibly leverage the [Zarr array](https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#stored-representation) as a semantic path.

One implementation question may be whether this can only be set on `create` in order to avoid multiple chunk layouts in one repository and/or having to migrate chunk layouts.

### Additional context

I am willing to contribute the feature myself, but if you have any preferences, wishes, or insights w.r.t. the implementation, let me know 😊

li-em:
Hi @mschou1306! Thanks for opening the issue, it is indeed a problem (and not only for local filesystems, there are S3-compatible object stores with the same issue 😅 )

Some suggestions on how to proceed:

- write up a design document under `design-docs/`, and open a PR. It doesn't need to be the final design or complete, but it is useful to track progress and document findings as you implement it.
- Should we follow the [git repo layout for objects](https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-objects0-9a-f0-9a-f)? A suggestion to consider in the design doc, not mandatory
- Currently the spec is not strongly worded, but [it does point](https://icechunk.io/en/stable/reference/spec-v2-1/#overview) to `$ROOT/chunks/{chunk_id}` as the path for chunks. For example, [VirtualiZarr][0] assumes the chunks have only this schema. Of course that is fixable, but we don't know how other software is interacting with Icechunk, so we want to be careful with any larger change.
[...]

PR #2330, opened 2026-08-17
Add design doc for configurable chunk directory layout

mschou1306:
See #2325

### DIFF
--- a/design-docs/019-native-chunk-object-layout.md
+++ b/design-docs/019-native-chunk-object-layout.md
@@ -1,0 +1,160 @@
+# Chunk object layout
+
+## Problem and scope
+
+Icechunk currently stores native chunks at:
+
+```text
+$ROOT/chunks/{chunk_id}
+```
+
+This creates one very large filesystem directory or object-store prefix, which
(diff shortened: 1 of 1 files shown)


design-docs/019-native-chunk-object-layout.md
```diff
@@ -0,0 +1,160 @@
...
+those readers. They must either implement the repository metadata and resolver
+or explicitly support flat repositories only.
+
+Readers that do not understand the selected layout must reject the repository
+during `open`, before any native chunk lookup. Because an older reader would
+assume the flat path, a non-flat layout requires an incompatible spec major
+version that older readers reject. Clients must reject unsupported versions
+rather than guessing. An unrecognized layout field that older clients silently
+ignore is insufficient, because they could assume the flat path and report a
+misleading missing-chunk error.
```
TomNicholas, line 97: Yes I think VirtualiZarr's `IcechunkParser` would deal with this quite poorly - it would confidently return as if the store contained arrays but no chunks whatsoever (which would then effectively populate all those arrays with the zarr fill value).

PR #2336, opened 2026-08-17
Reject chunk references with length zero

### LINKED ISSUE #2328: Disallow setting chunk refs with length zero
I think Icechunk has the same validation problem as VirtualiZarr does, see https://github.com/zarr-developers/VirtualiZarr/issues/1088

```python
sess.store.set_virtual_ref("a/c/0/1", url, offset=0, length=0)
# -> accepted at write
sess.commit(...)                      # -> accepted, snapshot DT5K4MTTQAP2
# read -> InvalidInputError: byte range From(0) is out of bounds for chunk of length 0
```

It's native refs too

```python
await sess.store.set("a/c/0/1", Buffer.from_bytes(b""))   # accepted
# read -> InvalidInputError: byte range From(0) is out of bounds for chunk of length 0
```

TomNicholas:
Length-zero chunks are nonsensical, cannot be read, and therefore should never be written. This adds validation to all chunk setting methods to raise a dedicated error if the user attempts to set a zero-length chunk.

I won't bother with any kind of migration - I don't know of any existing invalid stores, such invalid chunks already raise a clear error at read-time, and the only possible migration would be to delete the bad chunk refs, which is pretty straightforward.

Closes #2328.


Changelog.python.md
```diff
@@ -7,6 +7,7 @@
...
 - `to_icechunk` no longer passes `synchronizer` and `zarr_version` to xarray's `ZarrStore.open_group`; xarray removed both parameters and passing them made `to_icechunk` fail with a `TypeError` on xarray development versions ([#2312](https://github.com/earth-mover/icechunk/pull/2312)).
+- Writing a chunk with length 0 is now rejected instead of being committed. A chunk must decode to the full chunk shape, so no valid chunk is ever zero bytes long, and such a chunk could only fail once it was read back — long after the commit that introduced it. This is how a sparse GeoTIFF's unstored tiles (`offset = 0, byteCount = 0`) used to reach a repository. Applies to inline, virtual and materialized chunks alike, which means Icechunk deliberately rejects an empty write at a chunk key where a plain key-value store would accept it, in the same way it already rejects invalid zarr keys and invalid metadata. To record that a chunk is not stored at all, delete it rather than writing a zero-length one; it then reads back as the array's fill value ([#2328](https://github.com/earth-mover/icechunk/issues/2328)).
```
paraseba, line 10: This is a bit longer that we usually do for such a simple change.

paraseba (approved):

PR #2339, opened 2026-08-19
CI: Speed up python check 

li-em:
Build CI test wheels with a `ci-wheel` profile (opt-level 2, no LTO, parallel codegen) instead of release. `LTO + codegen-units=1` was ~3 minutes of every build.

Skip the hypothesis shards in the merge queue, since PRs and main already run them on the same commits.


dcherian (approved):
Wizard

PR #2346, opened 2026-08-25
Feature: add a new `hf_storage` function for accessing Hugging Face buckets

li-em:
Add a new preset for Hugging Face buckets, and docs on how to use it.

Fix: Hugging Face `endpoint_url` needs the trailing slash for compatibility with the AWS SDK, add a new `endpoint_with_bucket_separator` function to deal with it.


icechunk-python/docs/docs/guides/storage.md
```diff
@@ -114,6 +114,40 @@ icechunk.r2_storage(
...
+    prefix="quickstart-demo-1",
+    namespace="my-username",
+    access_key_id="HFAK...",
+    secret_access_key="my-secret-key",
+)
+```
+
+`namespace` is the owner of the bucket: your username or an organization name. The gateway scopes every request to the namespace in the endpoint path. `hf_storage` therefore builds the endpoint as `https://s3.hf.co/<namespace>` and sends the bare bucket name. It also sets the region to `us-east-1` and forces path-style URLs. The gateway requires both.
+
+One Hugging Face behavior limits Icechunk. The gateway discards user metadata. Icechunk stamps a write id in that metadata on every conditional write. It reads the id back after a suspicious failure. The id separates a lost success response from a real conflict. On Hugging Face that read always fails. A conditional write whose response is lost in transit can therefore report a conflict that never happened. Retry the commit when that happens.
```
paraseba: We can probably shorten this paragraph a lot, I doubt any users will be interested in this level of detail. 

```suggestion
The Hugging Face gateway discards user metadata which imposes some limitations on Icechunk: A conditional write whose response is lost in transit can report a conflict that never happened. Retry the commit if that happens.
```

icechunk-python/docs/docs/guides/storage.md
```diff
@@ -114,6 +114,40 @@ icechunk.r2_storage(
...
+
+You can stop Icechunk from sending metadata the gateway discards:
```
paraseba: I'd drop this for part two reasons:
* it doesn't buy the user anything
* it's more future proof, some day HF may start supporting metadata

icechunk-python/python/icechunk/storage.py
```diff
@@ -574,6 +575,101 @@ def r2_storage(
...
+    """Create a Storage instance that saves data in a Hugging Face Storage Bucket.
+
+    Data goes through Hugging Face's S3-compatible gateway. The credentials are
+    the S3 access key and secret generated from a Hugging Face access token. The
+    token itself is not an S3 credential. See
+    https://huggingface.co/docs/hub/storage-buckets-s3 for how to generate them.
+
+    The gateway discards user metadata. Icechunk therefore cannot recover from a
+    lost response to a conditional write. Such a write can report a spurious
+    conflict on the retry.
```
paraseba: let's make it clear that this is in a very unlikely scenario of a request lost in the way back to the client

icechunk-python/python/icechunk/storage.py
```diff
@@ -574,6 +575,101 @@ def r2_storage(
...
+        Extra HTTP headers to attach to both read and write requests. They are
+        merged with ``read_headers``/``write_headers``, which take precedence per
+        role on a key conflict.
+    """
+    return s3_storage(
+        bucket=bucket,
+        prefix=prefix,
+        # the gateway serves one region and only path-style URLs
+        region="us-east-1",
+        endpoint_url=f"{endpoint_url.rstrip('/')}/{namespace}",
```
paraseba: I know it's almost no logic but, I think we should put this function in Rust as the rest of the storage implementations

paraseba (commented):
Would it be too hard to add the secrets in GHA and have a HF backend in `test_storage.rs` ? That's would raise our confidence that things actually work.

li-em:
> Would it be too hard to add the secrets in GHA and have a HF backend in `test_storage.rs` ? That's would raise our confidence that things actually work.

Done, they are scoped to just one bucket under our org in HF.

paraseba (approved):
Amazing @li-em ! Just make sure to check the logs and verify the e2e test is running against HF

PR #2349, opened 2026-08-26
Fix the two failing hypothesis stateful tests

paraseba:
These errors were found by the nightly upstream-dev job.

zarr >= 3.3's `add_array` recreates the drawn array in the store under test without the sharding codec the model store got, so an all-fill-value write elides a chunk object on one side only. Override it with the fix from zarr-developers/zarr-python#4288 until that reaches a release.

`check_file_invariants` required orphan tx logs to equal the pruned-ancestor refs of every snapshot whose *file* is on disk. This is no longer true now that we keep expired tx logs.

### DIFF
--- a/.github/workflows/python-upstream.yaml
+++ b/.github/workflows/python-upstream.yaml
@@ -21,6 +21,9 @@ permissions:
 
 env:
   PIXI_VERSION: "v0.72.0"
+  # Title of the issue the pytest reporter keeps up to date. Pinned here (it is
+  # also the action's default) so the comment step below can find that issue.
+  UPSTREAM_ISSUE_TITLE: "⚠️ Nightly upstream-dev CI failed ⚠️"
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/python-upstream.yaml
```diff
@@ -221,6 +224,34 @@ jobs:
...
+        run: |
+          comment_on() {
+            number=$(gh issue list --state open --limit 200 --json number,title |
+              jq -r --arg t "$1" 'map(select(.title == $t)) | .[0].number // empty')
+            if [ -z "$number" ]; then
+              echo "no open issue titled '$1'"
+              return
+            fi
+            gh issue comment "$number" --body \
+              "Failed again in tonight's run: $RUN_URL — the issue body has the current failures."
```
li-em, line 250: Nice!

li-em (approved):

PR #2352, opened 2026-08-27
dev: sync nix and pixi dev env versions

li-em:
- `nix flake update`, keep ruff 0.15.x because there are formatting changes in 0.16.x
- sync pixi and nix versions, using nix as the reference
- drop Ian's hypothesis fork, there were many changes to hypothesis since it was forked
- convert `js-ci` to use pixi + just tasks


icechunk-python/pyproject.toml
```diff
@@ -211,6 +213,11 @@ platforms = [
   "osx-64",
 ]
 
+# zarr 3.2 changed the type hints that the zarr test overrides suppress.
+# mypy then reports unused suppressions. A separate change adds zarr 3.2 support.
+[tool.pixi.feature.dev.pypi-dependencies]
+zarr = "<3.2"
```
paraseba, line 219: This line scares me a bit? What are the consequences?

paraseba (approved):

icechunk-python/pyproject.toml
li-em: It is mostly because our `zarr` baseline is still `3.1.0`. We have this in the `pyproject` dep table:
```
dependencies = ["zarr>=3.1.0"]
```

What you pointed sets up `<3.2` for pixi, especially for the code quality checks (mypy). But we still test `latest` and `upstream` for zarr, so we are covering the latest and greatest (but not `mypy` checking it, because then we break older zarr versions...)

li-em: More concretely: we can't make it `mypy`-compliant over a range of zarr versions (3.1 - 3.3)...

PR #2368, opened 2026-09-09
Fix(tests): Add whole-second rounding for the GC cutoff

li-em:
Tigris lists `LastModified` rounded to whole seconds. GC reads that value as `created_at`, so an object written after the cutoff but inside the same second compares as older and gets deleted.

If the timestamp has sub-second precision, keep working as is. If not, then cutoff is seconds-based now (mostly to avoid per-store configuration, although we will need to change if stores with wildly different ways of handling timestamps show up).

### DIFF
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -2,6 +2,10 @@
 
 ## Python Icechunk Library [unreleased]
 
+### Fixes
+
+- Garbage collection no longer deletes objects that a caller wrote after its cutoff on Tigris. Tigris lists whole-second timestamps. The floored value can precede the write by up to one second. GC now deletes an object only after that whole second precedes the cutoff.
+
 ## Python Icechunk Library 2.2.0
 
 ### Features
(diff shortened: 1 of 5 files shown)


Changelog.python.md
```diff
@@ -2,6 +2,10 @@
 
 ## Python Icechunk Library [unreleased]
 
+### Fixes
+
+- Garbage collection no longer deletes objects written after its cutoff on stores that list whole-second timestamps, such as Tigris. Such a timestamp is floored, so the write can be up to one second later than the listing reports; GC now deletes only once that whole second precedes the cutoff.
```
paraseba: Maybe we shouldn't scare people with this changelog, it's mostly a theoretical, test-only scenario

paraseba (approved):

Changelog.python.md
li-em: aye, will fix and add the PR link now that I have a number

Issue #2364, opened 2026-09-08
Storage tiering of native chunks by array (per-array storage class or key prefix)


elyall:
### Problem

We store multiscale pyramids. The full-resolution level is by far the largest and the least frequently read — most reads hit the downsampled levels. We'd like full-res chunks in a cheaper, less-available tier and everything else in the default tier.

Today this isn't expressible, for two independent reasons:

1. **Tier is repo-wide.** `StorageSettings.chunks_storage_class` (#904) applies to every native chunk in the repo.
2. **No per-array prefix.** All native chunks land at `chunks/{chunk_id}`, so lifecycle rules — which every tiered provider scopes by key prefix — can't target a subset of arrays.

There's also a portability gap: `chunks_storage_class` is only applied by the native S3 backend (`icechunk-s3`). The `object_store` backend builds `PutOptions` without `Attribute::StorageClass`, so on `gcs_storage`, `azure_storage`, and `s3_object_store_storage` the setting is silently ignored — even though `object_store` maps that attribute to the correct header on all three clouds.

Here's how the tiered backends in the [storage guide](https://icechunk.io/en/stable/guides/storage/) compare:

| Backend (constructor) | Rust backend | Tier set on PUT via | `chunks_storage_class` honored today | Lifecycle transition filters | Offline tier (restore required) |
|---|---|---|---|---|---|
| AWS S3 (`s3_storage`) | `icechunk-s3` | `x-amz-storage-class` | ✅ | prefix, tags, size | Glacier Flexible / Deep Archive → 403 `InvalidObjectState` |
| Tigris (`tigris_storage`) | `icechunk-s3` | `x-amz-storage-class` (`STANDARD_IA`, `GLACIER`, `GLACIER_IR`) | ✅ | prefix | Archive (`GLACIER`) → 403 `InvalidObjectState` |
| Cloudflare R2 (`r2_storage`) | `icechunk-s3` | `x-amz-storage-class` (`STANDARD_IA`) | ✅ | prefix | none |
| GCS (`gcs_storage`) | `object_store` | `x-goog-storage-class` (`NEARLINE`, `COLDLINE`, `ARCHIVE`) | ❌ ignored | prefix / suffix (no tags) | none — `ARCHIVE` is online, priced on retrieval |
[...]

PR #2370, opened 2026-09-10
Honor storage_class in the object_store backends

### LINKED ISSUE #904: Support setting storage class on object_store based Storage instances

elyall:
Option (0) from #2364. Closes #904.

## What

`StorageSettings.storage_class` / `metadata_storage_class` / `chunks_storage_class` were only applied by the native S3 backend (`s3_storage`, `tigris_storage`, `r2_storage`). The `object_store` backend (`gcs_storage`, `azure_storage`, `s3_object_store_storage`) built its `PutOptions.attributes` entirely inside the `unsafe_use_metadata` guard and never set `Attribute::StorageClass`, so the setting was silently ignored there.

`put_object` now inserts `Attribute::StorageClass` from `settings.storage_class()` outside that guard (the class is not user metadata). `object_store` maps it to `x-amz-storage-class` / `x-goog-storage-class` / `x-ms-access-tier`.

## Local filesystem

`object_store`'s `LocalFileSystem` returns `NotImplemented` for any non-empty put attributes, so an unconditional insert would fail every write on a local repo whose config sets a class. I added `ObjectStoreBackend::supports_storage_class()` (default `true`) and the local filesystem backend returns `false`, so the setting is ignored there as it is today. Happy to switch this to an explicit error if you'd rather not ignore it silently.

## Tests

- Unit (`icechunk-arrow-object-store`, in-memory / tempdir): the class reaches the store even with `unsafe_use_metadata = false` and nothing else leaks past the guard; no attribute when unset; local filesystem still writes with a class configured.
- Integration: `test_storage_classes_object_store` mirrors the existing AWS-gated `test_storage_classes` through `new_s3_object_store_storage`, sharing a helper. Only runs when `AWS_BUCKET` is set.
- Not yet verified against real GCS or Azure.

## Docs

- `_icechunk_python.pyi`: dropped "Currently not supported in GCS", noted provider-specific class names and the local filesystem exception.
- Changelog entry under Fixes.

`cargo fmt --check`, `just lint` (`-D warnings`, Rust 1.96) and the test suites pass locally.

### DIFF
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Update tests to deal with Tigris and other stores listing whole-second timestamps ([#2368](https://github.com/earth-mover/icechunk/pull/2368)).
+- `storage_class`, `metadata_storage_class` and `chunks_storage_class` now apply to the `object_store` backends — `gcs_storage`, `azure_storage` and `s3_object_store_storage`. They were silently ignored there; only `s3_storage`, `tigris_storage` and `r2_storage` honored them. The value is passed to the provider unchanged, so use its own names (`NEARLINE` on GCS, `Cool` on Azure). The local filesystem backend has no storage classes and keeps ignoring the setting ([#904](https://github.com/earth-mover/icechunk/issues/904), [#2364](https://github.com/earth-mover/icechunk/issues/2364)).
 
 ## Python Icechunk Library 2.2.0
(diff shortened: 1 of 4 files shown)


paraseba:
Thank you @elyall, this is great. I made some small changes and I'm going to merge it.

paraseba (approved):

PR #2371, opened 2026-09-11
GC: list object-id prefixes concurrently on S3

samantha-earthmover:
>  This is claude but it mostly makes sense and I don't want to edit out the deep rusty bits i dont quite get.

GC lists `chunks/`, `manifests/`, `snapshots/` and `transactions/` as one serial chain of 1000-key ListObjectsV2 pages. For a repository with 56 million chunks that is about 56,000 sequential requests. At roughly 0.15 s per page, the chunk listing alone takes more than two hours. #2358 made the snapshot reads concurrent; this listing is the remaining serial part.

Object ids are Crockford base32, so an id starts with one of 32 characters. This PR lists each of them concurrently:

- `Storage::list_objects_with_id_prefixes(settings, prefix, id_prefixes)` returns the objects under `prefix` whose ids start with one of `id_prefixes`, unordered. The default filters a single `list_objects`, so every backend returns the same set.
- `S3Storage` overrides it with one ListObjectsV2 per `<prefix>/<id_prefix>`, merged with `select_all`. Ids stay relative to `prefix`.
- The latency, logging and redirect wrappers forward the call.
- `OBJECT_ID_FIRST_CHARS` in `icechunk-format`, with a test that it is exactly the set of possible first characters of an `ObjectId`.
- `AssetManager::list_{chunks,manifests,snapshots,transaction_logs}` use it, so every GC listing pass is split. No caller depends on listing order.

Not in this PR: object_store backends (GCS, Azure, local) keep one listing. object_store cannot list a partial path segment. `list_with_offset` could split them, but only on backends that return keys in order.

Tests: `test_list_objects_with_id_prefixes` on every `with_storage` backend, the GC tests (including minio), and the new format test.

### DIFF
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -2,6 +2,10 @@
 
 ## Python Icechunk Library [unreleased]
 
+### Performance
+
+- On S3-compatible storage, garbage collection now lists chunks, manifests, snapshots and transaction logs with 32 concurrent listings, one per possible first character of the object id. Before, each prefix was one serial chain of 1000-key pages: about 56,000 sequential requests for a repository with 56 million chunks. GCS, Azure and local storage still use one listing ([#TBD](https://github.com/earth-mover/icechunk/pull/TBD)).
+
 ### Fixes
 
 - Update tests to deal with Tigris and other stores listing whole-second timestamps ([#2368](https://github.com/earth-mover/icechunk/pull/2368)).
--- a/Changelog.python.md
+++ b/Changelog.python.md
@@ -4,7 +4,7 @@
 
 ### Performance
 
-- On S3-compatible storage, garbage collection now lists chunks, manifests, snapshots and transaction logs with 32 concurrent listings, one per possible first character of the object id. Before, each prefix was one serial chain of 1000-key pages: about 56,000 sequential requests for a repository with 56 million chunks. GCS, Azure and local storage still use one listing ([#TBD](https://github.com/earth-mover/icechunk/pull/TBD)).
+- On S3-compatible storage, garbage collection now lists chunks, manifests, snapshots and transaction logs with 32 concurrent listings, one per possible first character of the object id. Before, each prefix was one serial chain of 1000-key pages: about 56,000 sequential requests for a repository with 56 million chunks. GCS, Azure and local storage still use one listing ([#2371](https://github.com/earth-mover/icechunk/pull/2371)).
 
 ### Fixes
 
--- a/icechunk-format/src/lib.rs
+++ b/icechunk-format/src/lib.rs
@@ -204,6 +204,13 @@ impl<const SIZE: usize, T: FileTypeTag> From<&ObjectId<SIZE, T>> for String {
     }
 }
 
+/// The Crockford base32 alphabet. The string form of every [`ObjectId`] starts with
+/// one of these characters, so a listing split on them covers every object id.
+pub const OBJECT_ID_FIRST_CHARS: [&str; 32] = [
+    "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "A", "B", "C", "D", "E", "F", "G",
+    "H", "J", "K", "M", "N", "P", "Q", "R", "S", "T", "V", "W", "X", "Y", "Z",
+];
+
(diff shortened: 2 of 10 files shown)


icechunk-format/src/lib.rs
```diff
@@ -204,6 +204,13 @@ impl<const SIZE: usize, T: FileTypeTag> From<&ObjectId<SIZE, T>> for String {
     }
 }
 
+/// The Crockford base32 alphabet. The string form of every [`ObjectId`] starts with
+/// one of these characters, so a listing split on them covers every object id.
+pub const OBJECT_ID_FIRST_CHARS: [&str; 32] = [
```
paraseba: This is available from the `base32` crate we already depend on.

icechunk-format/src/lib.rs
```diff
@@ -717,6 +724,21 @@ mod tests {
         );
     }
 
+    #[icechunk_macros::test]
+    fn test_object_id_first_chars_are_every_possible_first_char() {
```
paraseba, line 728: I think this test can disappear once you use the characters from the base32 crate, but otherwise, would be good to turn it into a property test.

icechunk-storage/src/storage.rs
```diff
@@ -562,6 +563,25 @@ pub trait Storage: fmt::Debug + Display + sealed::Sealed + Sync + Send {
...
+        prefix: &str,
+        id_prefixes: &[&str],
+    ) -> StorageResult<BoxStream<'a, StorageResult<ListInfo<String>>>> {
+        let id_prefixes: Vec<String> =
+            id_prefixes.iter().map(|id_prefix| (*id_prefix).to_string()).collect();
+        Ok(self
+            .list_objects(settings, prefix)
+            .await?
+            .try_filter(move |info| {
+                ready(id_prefixes.iter().any(|id_prefix| info.id.starts_with(id_prefix)))
```
paraseba: I think a better way would be to put the `id_prefixes` into a HashSet (you can keep one around in a `static`) and then check the first character of each `info.id` for presence in the set.

icechunk/src/asset_manager.rs
```diff
@@ -1015,48 +1016,45 @@ impl AssetManager {
...
 
     #[instrument(skip(self))]
     pub async fn list_transaction_logs(
         &self,
     ) -> RepositoryResult<BoxStream<'_, RepositoryResult<ListInfo<SnapshotId>>>> {
+        self.list_object_ids(TRANSACTION_LOGS_FILE_PATH).await
+    }
+
+    /// Unordered: storage can list each possible first character of the id concurrently.
+    async fn list_object_ids<'a, Id>(
```
paraseba, line 1044: lovely

Changelog.python.md
```diff
@@ -2,6 +2,10 @@
 
 ## Python Icechunk Library [unreleased]
 
+### Performance
+
+- On S3-compatible storage, garbage collection now lists chunks, manifests, snapshots and transaction logs with 32 concurrent listings, one per possible first character of the object id. Before, each prefix was one serial chain of 1000-key pages: about 56,000 sequential requests for a repository with 56 million chunks. GCS, Azure and local storage still use one listing ([#2371](https://github.com/earth-mover/icechunk/pull/2371)).
```
paraseba: Let's make this shorter, highlighting the performance win

Changelog.python.md
```diff
@@ -2,14 +2,15 @@
...
-- Update tests to deal with Tigris and other stores listing whole-second timestamps ([#2368](https://github.com/earth-mover/icechunk/pull/2368)).
-
 ### Performance
 
+- Garbage collection lists 32 ways in parallel on S3, one listing per possible first character of an object id. Listing a 2.8 million object repository drops from 299 s to 15 s. Other backends keep a single listing ([#2371](https://github.com/earth-mover/icechunk/pull/2371)).
 - Speed up manifest metadata lookups during commits, reads, and manifest preloading in repositories with many manifests by using binary search instead of linear scans.
 
+### Fixes
+
+- Update tests to deal with Tigris and other stores listing whole-second timestamps ([#2368](https://github.com/earth-mover/icechunk/pull/2368)).
```
paraseba, line 7: bad merge maybe?

paraseba (approved):

PR #2372, opened 2026-09-11
Tests: Check ancestry file completeness in the `VersionControlTest` model

li-em:
Address latest batch of #1921 

Since #2184 we've been having hypothesis stateful testing failures because snapshots are left on disk until GC is performed for the pruned tx logs, and the model is tracking snapshot presence instead of verifying history is as expected, with one snapshot file and tx log for every snapshot in ancestry.


dcherian:
Perhaps we should switch the checked invariant to "history is as expected, there is a snapshot file & tx log for every snapshot in ancestry".

icechunk/tests/test_gc.rs
```diff
@@ -879,6 +879,130 @@ async fn test_gc_retains_snapshot_between_flushed_and_created_at()
...
+        100,
+    )
+    .await?;
+    assert_eq!(result.released_snapshots.len(), 1);
+    assert!(result.edited_snapshots.contains(&c));
+    let (repo_info, _) = am.fetch_repo_info().await?;
+    assert_eq!(
+        repo_info.find_snapshot(&c)?.pruned_ancestor_tx_logs,
+        vec![a.clone(), b.clone()]
+    );
```
paraseba, line 921: Can we assert here that even if we run GC at this point, the tx logs for `a` and `b` are not dropped?

icechunk/tests/test_gc.rs
```diff
@@ -879,6 +879,130 @@ async fn test_gc_retains_snapshot_between_flushed_and_created_at()
     Ok(())
 }
 
+/// A snapshot that `expire` releases from the repo info keeps its file on
+/// disk: expire never deletes files. The release destroys the snapshot's
+/// `pruned_ancestor_tx_logs`, which live only on the repo info. GC then
+/// deletes the tx logs those refs protected. The `created_at` gate can
+/// still keep the released snapshot's own file. That stranded file is
```
paraseba: The test looks correct but the description scared me. It's only true because there are branches involved. TX logs should only be deleted if they are not part of any `pruned` lists.

paraseba (approved):
Please see my comments before merging

PR #2381, opened 2026-09-15
Binary search over manifest files

aldenks:
Speed up manifest metadata lookup during commits, reads, and manifest preloading in repositories with many manifests by using binary search instead of linear scans.

This was motivated by commit times on a store w/ hundreds of thousands of total manifest files reaching roughly 90 seconds. Even though the commits touched only a couple hundred manifest files, finding which manifests to update became the bottleneck. I tried to reduce the number of manifests by making splits larger, but that increases the amount of download, merge, upload work to do per commit, making that portion of work the bottleneck.

### DIFF
--- a/icechunk-format/src/snapshot.rs
+++ b/icechunk-format/src/snapshot.rs
@@ -732,16 +732,15 @@ impl Snapshot {
     ) -> IcechunkResult<Option<ManifestFileInfo>> {
         let root = self.root();
         if let Some(mf2) = root.manifest_files_v2() {
-            mf2.iter()
-                .find(|mf| mf.id().is_some_and(|mid| mid.0 == id.0))
-                .map(|mf| (&mf).try_into())
-                .transpose()
+            lookup_index_by_key(mf2, Some(id.0), |mf, id| {
+                mf.id().map(|mid| mid.0).cmp(id)
+            })
+            .map(|index| (&mf2.get(index)).try_into())
+            .transpose()
         } else {
-            Ok(root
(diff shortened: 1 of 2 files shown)


icechunk-format/src/snapshot.rs
```diff
@@ -731,19 +731,43 @@ impl Snapshot {
...
+                mf1.get(index).id().0.cmp(&id.0)
+            })
+            .map(|index| mf1.get(index).into()))
+        }
+    }
+}
+
+// Return the first matching entry when multiple manifests have the same ID.
+// Continuing left after a match keeps even an all-equal vector logarithmic.
+fn lookup_first_manifest_index(
```
paraseba, line 753: I thought we were doing binary search somewhere else, I wonder if we have this function already.

paraseba (approved):
This is great @aldenks !!

icechunk-format/src/snapshot.rs
aldenks: Good point, `lookup_index_by_key` (icechunk-format/src/lib.rs:632) already does this. The distinction is behavior on duplicate values -- it returns the mid point value, while the `.find` this is replacing picked the first instance.

That is probably a moot point since I don't believe we expect duplicate IDs -- these are UUIDs/hashes right? If so, I can use lookup_index_by_key here. But that would technically be a behavior change.

li-em (approved):

aldenks:
Thanks for the quick review n merge!

PR #2382, opened 2026-09-15
Test: Replace MinIO with rustfs 1.0.0-rc.6

li-em:
https://github.com/rustfs/rustfs/pull/4228 added the path normalization that #2242 required, so we can drop minio and use only `rustfs` for our tests


compose.yaml
```diff
@@ -51,7 +50,6 @@ services:
             {
               "Sid": "AllowGetObjectOnly",
               "Effect": "Allow",
-              "Principal": { "AWS": ["readonly"] },
```
li-em, line 54: Dropped these from config because RustFS doesn't accept them anymore

paraseba (approved):

PR #2385, opened 2026-09-16
GC: split object-id listings on GCS too

samantha-earthmover:
> Draft. Samantha will review every line before marking it ready.

#2371 made GC list object ids 32 ways in parallel, but only on the native S3 backend. GCS still did one slow listing.

The problem: `object_store` only lists by whole path segment. You can ask for everything under `chunks/`, but not for everything under `chunks/0`.

The fix: use `list_with_offset` instead. For each of the 32 possible first characters, start a listing right after `chunks/<char>` and stop as soon as a key with a different first character shows up. This works because GCS sends the offset to the server as `start-after` and returns keys in alphabetical order.

What changed:

- `ObjectStoreBackend` gets `ordered_offset_listing()`. It returns true for GCS and for the `object_store` S3 backend, false for everything else.
- `ObjectStorage::list_objects_with_id_first_chars` runs 32 offset listings in parallel when the backend says yes. Otherwise it does what it did before: one listing, filtered.
- The filter moved to `icechunk_storage::filter_ids_by_first_char`, so the trait default and the fallback share it.

Azure, local files, in-memory and HTTP keep the single filtered listing. They apply the offset on the client, or return keys out of order, so splitting would not help them.

One edge case: the offset is exclusive, so an object whose id is exactly one character would be skipped. Icechunk ids are always 20 characters, so that cannot happen.

Measured from a laptop, anonymous, on `gs://gcp-public-data-arco-era5/.../2m_temperature` (69,035 keys, only 10 distinct first characters): one listing took 18.3 s, the split listing 3.4 s, with the same ids.

Tests: `test_gcs_list_objects_with_id_first_chars` lists the public `al-public-test-bucket/verification-copy` repo both ways and checks the results match, for all four GC folders and for a single character. The existing `test_list_objects_with_id_first_chars` now covers the offset path on the `object_store` S3 backends.


icechunk-arrow-object-store/src/lib.rs
```diff
@@ -704,6 +705,38 @@ impl Storage for ObjectStorage {
...
+        &'a self,
+        settings: &Settings,
+        prefix: &str,
+        first_chars: &HashSet<char>,
+    ) -> StorageResult<BoxStream<'a, StorageResult<ListInfo<String>>>> {
+        if !self.backend.ordered_offset_listing() {
+            let listing = self.list_objects(settings, prefix).await?;
+            return Ok(filter_ids_by_first_char(listing, first_chars));
+        }
+        let root = ObjectPath::from(format!("{}/{}", self.backend.prefix(), prefix));
```
paraseba, line 719: Does this work if the backend path is empty?

li-em (approved):

icechunk-arrow-object-store/src/lib.rs
```diff
@@ -907,6 +940,12 @@ pub trait ObjectStoreBackend: Debug + Display + Sync + Send {
     /// The prefix for the object store.
     fn prefix(&self) -> String;
 
+    /// Whether `list_with_offset` runs server side and returns keys in lexicographic
+    /// order. Only then can [`ObjectStorage`] split an id listing by first character.
+    fn ordered_offset_listing(&self) -> bool {
+        false
```
paraseba, line 946: but does S3 support the offset part?

samantha-earthmover: yes, been using the same trick elsewhere for a while now :) , gets translated to `start-after`

icechunk-arrow-object-store/src/lib.rs
samantha-earthmover: > Yes. ObjectPath::from(format!("{}/{}", "", "chunks")) gives chunks, because object_store::path::Path drops empty segments when it parses (path/mod.rs:457-467 in 0.14.1). list_objects on the line above already builds its prefix the same way, so the two paths behave alike. I added test_list_objects_with_id_first_chars_at_bucket_root, which uses an object_store S3 storage with prefix: None and lists under a directory at the bucket root. It passes.

paraseba (approved):

PR #2387, opened 2026-09-16
Tests: draw `delete_chunk` keys from the storage grid in the compat test

li-em:
Zarr's `delete_chunk` rule draws the index from `cdata_shape`,
which counts inner chunks. A sharded array holds one key per shard.
The rule can name a key that cannot exist.

Icechunk 2 ignores such a key since #2312. The pinned `icechunk_v1`
wheel rejects it, and so our compat tests were broken.

The override draws from `storage_chunk_sizes`. It skips a zero-length
dimension, which doesn't have any keys.


ianhi:
> impossible key

should we upstream the rule to prevent it from generating impossible keys?

ianhi (approved):
seems rigth to me

li-em:
> should we upstream the rule to prevent it from generating impossible keys?

https://github.com/earth-mover/icechunk/pull/2312 has the longer version, but this is a limitation on our side, not upstream. This PR is needed because v1 doesn't have the fix

dcherian (approved):
Would be good to port upstream

icechunk-python/tests/test_stateful_compat.py
```diff
@@ -424,6 +426,24 @@ def switch_actor(self, data: st.DataObject) -> None:
     def commit_with_check(self, data: st.DataObject) -> None:
         return super().commit_with_check(data)
 
+    # `cdata_shape` counts inner chunks. A sharded array holds one key per shard.
```
dcherian, line 429: I like to annotate these things with `PORT(ED) UPSTREAM` or some such so we can remember to delete it at sometime.

PR #2388, opened 2026-09-17
Verify metadata files before writing them

paraseba:
Repo info accessors validated the buffer with flatbuffers' default options, which cap a file at one million tables. Repositories past that size could not be opened at all. They now read unchecked,  but a RepoInfo still always holds a verified buffer. max_tables goes from 5M to 50M.

Serializers don't validate what they build, so a file that tripped a limit was written happily and only failed when someone read it back. Metadata files at or above 128 MiB are now verified before upload, using the same checks the read path applies.

We have verified we can now open real-world repos with 0.5M snapshots and there is a test that tries with 10M fake ones.


icechunk-format/src/manifest.rs
```diff
@@ -1062,10 +1066,22 @@ fn mk_chunk_ref<'bldr>(
 static ROOT_OPTIONS: VerifierOptions = VerifierOptions {
     max_depth: 64,
     max_tables: 500_000_000,
-    max_apparent_size: 1 << 31, // taken from the default
+    // Large enough that any buffer the builder can produce is readable: the
+    // builder refuses to grow past FLATBUFFERS_MAX_BUFFER_SIZE, and the
+    // verifier's counter runs at most MAX_APPARENT_SIZE_INFLATION times the
+    // buffer it walks. Saturating because the product doesn't fit a 32 bit
```
paraseba, line 1072: @mpiannucci is this the right way to deal with the number being too large?

mpiannucci: Yes assuming we don't want to panic and just cap it

Changelog.python.md
```diff
@@ -2,6 +2,15 @@
...
 
+### Fixes
+
+- Repositories with more than one million snapshots are readable again ([#2388](https://github.com/earth-mover/icechunk/pull/2388)).
+- Metadata files that could not be read back are no longer written: large files are validated before upload ([#2388](https://github.com/earth-mover/icechunk/pull/2388)).
+
+### Performance
+
+- Repo info accessors no longer re-validate the whole file on every call ([#2388](https://github.com/earth-mover/icechunk/pull/2388)).
+
```
li-em: These should go under 2.2.1 below

paraseba: done

paraseba: good catch

icechunk-format/src/repo_info.rs
```diff
@@ -1311,21 +1340,29 @@ impl RepoInfo {
         }
     }
 
+    /// Check that `buffer` passes every check [`RepoInfo::from_buffer`] applies.
```
dcherian, line 1343: confusing comment.. our check is that flatbuffers can decode the buffer? Why would it let you create a thing it can't decode? 🤔

icechunk-format/src/repo_info.rs
```diff
@@ -1311,21 +1340,29 @@ impl RepoInfo {
...
     }
 
     pub fn bytes(&self) -> &[u8] {
         self.buffer.as_slice()
     }
 
+    #[expect(unsafe_code)]
     fn root(&self) -> IcechunkResult<generated::Repo<'_>> {
-        flatbuffers::root::<generated::Repo<'_>>(&self.buffer).capture()
+        // SAFETY: every RepoInfo holds a verified buffer: `from_buffer` and the
```
dcherian, line 1361: this assertion isn't true for existing repos? I'm OK with taking a calculated risk since we are presumable the only writers on the planet

icechunk-format/src/repo_info.rs
paraseba: Yeah, this is confusing. The checks are "limits" we impose when we fetch data from the buffer. Mostly intended as protection from DoS. Stuff of the type "If you see more than x tables don't even try to continue opening the file". So it was possible before to create flatbuffer files that were within the absolute limits of what a flatbuffer can hold, but beyond what we allowed at read time.

This now runs the same check at write time.

icechunk-format/src/repo_info.rs
paraseba: no, we are good. Previous code had one bug and one performance issue:
* we weren't applying the limit correctly so we were using a default which limited snaps to < 1M
* we were re-validating the full flatbuffer on each accessor.

This new code fixes the limit and only validates on the first read from the wire, this is the same behavior as in the other file types (snap, manifest, etc)

paraseba: New code more than "fixes the limit", it fixes it and makes it larger than it was.

icechunk-format/src/repo_info.rs
dcherian: yup i just scanned the `verifier.rs` in the flatbuffers crate and it makes more sense to me.

dcherian (approved):

icechunk-format/src/repo_info.rs
dcherian: nice!

PR #2389, opened 2026-09-17
Fix GC deadlock on hosts with few CPUs

samantha-earthmover:
> Draft. Samantha will review every line before marking it ready.

Garbage collection on 2.2.1 hangs on hosts with few CPUs. Within a minute it stops making progress and never returns: no error, no retry, no timeout. Reproduced 5 out of 5 times in a 2-CPU Kubernetes pod and 3 out of 3 times in `docker run --cpus=2`, against a private repo and against the public `s3://dynamical-noaa-hrrr/noaa-hrrr-forecast-48-hour/v0.1.0.icechunk`. The same run completes in 6.5 minutes with 8 CPUs, and completes with 2 CPUs on 2.2.0.

What goes wrong:

- `find_retained` feeds `pointed_snapshots_v2`, 25 concurrent snapshot fetches in a `buffer_unordered`, into a `try_buffer_unordered(8)` of manifest fetches. When the manifest buffer is full it stops polling the snapshot stream. That is normal `futures` behaviour.
- Every fetch ends in `fetch_and_decode`, which awaited `decode_gate().acquire()`, a tokio `Semaphore` with one permit per CPU, on the async task.
- A tokio semaphore is fair: a freed permit goes to the oldest waiter. The oldest waiters are the snapshot fetches that nobody is polling. With 2 permits, both end up assigned to futures that never run. The 8 manifest fetches wait for a permit forever, the manifest buffer never drains, and the snapshot stream is never polled again.
- 2.2.0 fetched snapshots one at a time, so at most one permit could get stuck and the other kept cycling. #2358 made the fetches concurrent, which exposed this.

The fix is one move: the permit is acquired inside the `spawn_blocking` closure, with `futures::executor::block_on`, instead of on the async task before it. A waiter that is a real thread is always running, so a freed permit is used the moment it is handed over.

The decode semaphore also moves from a process-wide static into `AssetManager`, sized by the new `RepositoryConfig.max_concurrent_decodes`. The default is unchanged: one permit per CPU available to the process. The option is exposed in the Python and JS configs and documented in the configuration guide. `AssetManager` pickles from older versions load with the default.

Test: `test_gc_completes_with_one_decode_slot` builds a 100-snapshot repo on in-memory storage with 5 ms read latency, opens it fresh with `max_concurrent_decodes=1` so nothing is cached, and runs GC. With the permit awaited on the async task it deadlocks and the 120 s timeout fires; with this change it finishes in about 5 s.

Repro without the test: `docker run --rm --cpus=2 python:3.13-slim`, `pip install icechunk==2.2.1`, then `Repository.garbage_collect_async(cutoff, dry_run=True, max_snapshots_in_memory=25, max_concurrent_manifest_fetches=8)` on the public repo above with anonymous credentials. It freezes after roughly 70 downloads, within 10 seconds.


icechunk/src/asset_manager.rs
```diff
@@ -1199,11 +1199,18 @@ fn binary_file_header(
 /// Caps concurrent CPU decodes so the blocking pool can't oversubscribe the cores.
 /// Absent under shuttle: a process-global semaphore would leak across shuttle
 /// executions, and the gate's core-oversubscription job is meaningless there.
+/// `ICECHUNK_DECODE_CONCURRENCY` overrides the size.
 #[cfg(not(feature = "shuttle"))]
 fn decode_gate() -> &'static Semaphore {
     static GATE: LazyLock<Semaphore> = LazyLock::new(|| {
-        let n = std::thread::available_parallelism().map_or(8, |n| n.get());
-        Semaphore::new(n)
+        let n = std::env::var("ICECHUNK_DECODE_CONCURRENCY")
```
paraseba: not a fan, we don't do this type of global env stuff anywhere, how hard is to thread this as a config

samantha-earthmover: done!

Changelog.python.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Python Icechunk Library [unreleased]
+
+### Fixes
+
+- Garbage collection no longer hangs on hosts with few CPUs. 2.2.1 fetched snapshots 25 at a time, and the decode gate, a tokio semaphore with one slot per CPU, handed freed slots to snapshot fetches that the full manifest buffer had stopped polling. On a 2-CPU host GC froze within a minute with no error. Decode slots are now taken on the blocking thread, and the new `RepositoryConfig.max_concurrent_decodes` sets the number of slots per repository ([#2389](https://github.com/earth-mover/icechunk/pull/2389)).
```
paraseba: Let's make this much shorter. Just informing that we fixed a deadlock in low CPU machines or something

paraseba (approved):

PR #2397, opened 2026-09-18
CI: run each hypothesis state machine in its own shard

samantha-earthmover:
## Summary

Each of the three hypothesis state machines now runs in a shard of its own, `test_store_properties.py` gets its own shard, and the catch-all shard stays last. The shard step also reports its ten slowest tests.

## Context

The shards run pytest-xdist with two workers, and xdist hands each worker two consecutive tests up front. In a shard anchored by a heavy state machine, the next test in collection order was queued on the same worker as the machine and could not start until it finished, while the other worker went idle.

Measured with `--durations=0` on `main`'s layout:

| shard | step | slowest test | queued behind it |
|---|---|---|---|
| repo-ops + store-properties | 344s | `test_stateful_repo_ops.py::VersionControlTest` 300s | `test_getitem` 46s |
| rest | 340s | `test_stateful_parallel_sessions.py::VersionControlTest` 254s | about 80s of `test_move.py` |
| zarr-hierarchy | 129s | `test_zarr_hierarchy` 123s | nothing |

In both slow shards the step time is the machine plus whatever sat behind it, not the total work divided across two workers.

With this change nothing queues behind a machine. Two runs of this branch:

| shard | run 1 | run 2 |
|---|---|---|
| repo-ops, `VersionControlTest` | 229s | 351s |
| parallel-sessions, `VersionControlTest` | 341s | 303s |
| zarr-hierarchy, `test_zarr_hierarchy` | 145s | 176s |
| store-properties, 9 tests | 158s step | 161s step |
| rest, 6 tests | 119s step | 113s step |

This does not measurably shorten `test-hypothesis`. From the end of `build-wheels` to the last shard finishing took 375s on `main`'s layout and 381s and 386s here. The machines themselves vary between 215s and 351s from run to run, since the CI profile does not derandomize, and that spread is larger than the queued test this removes. The gate is now bounded by whichever machine runs longest on its own.

What the change does buy: shard cost no longer depends on which test happens to land next to a machine, each shard's cost is visible in its log, and the layout is ready for splitting a machine's example budget across shards, which is the change that would actually shorten this job.

Adding `-n 4` does not help: the slow shards are a single test each, and the runner has two physical cores.

## Risk level

Low. Only the hypothesis shard matrix changes. The required `test-hypothesis` gate job keeps its name and still waits on every shard. Coverage fragments and hypothesis example artifacts are collected by glob, so the new shard ids need no other changes. Every hypothesis test still runs on every PR.

## Reviews

No required reviewers beyond the usual CI owners.

---


li-em:
Superseded by #2398

PR #2404, opened 2026-09-20
perf: A parallel walker for GC and stats manifest traversal

paraseba:
The old find_retained and repo_chunks_storage were too complex
which made them very hard to optimize. As a result, they were
making bad use of the network bandwidth: the per-manifest
work serialized behind the fetches and left the network idle
for seconds at a time on large manifests.

ops::walker now runs fetch, decode and per-manifest work
in worker pools connected by bounded channels. We also now have a
memory budget for manifests. Retained chunk ids are inserted
in parallel into a sharded set (ops::sharded_set).
Both GC and stats consume the same walker.

The sharded set is vibed but much faster than the OS alternatives
for this use case because it allows multiple insertions.
I added property testing that validates it behaves as a HashSet.

The walker needs raw fetches and decodes (no caches or limits)
so we added a method in the asset_manager.

Payload decode errors now fail the run instead of being logged and
skipped, which would have deleted live chunks.

Reviewers:
- Don't worry with the ShardedSet (see note above)
- The walker is pretty nice code
- GC is much clearer now and it shares much more with stats

---

**Stack:** 4 of 15 — based on #2403 (`gc-improvements-03`).


icechunk/src/ops/sharded_set.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Insert every key, taking each shard's lock at most once. Returns the
+    /// summed weight of the keys that were not already present.
+    pub fn extend_weighted(&self, items: impl IntoIterator<Item = (K, u64)>) -> u64 {
+        self.try_extend_weighted::<std::convert::Infallible>(items.into_iter().map(Ok))
+            .unwrap_or_else(|never| match never {})
+    }
```
li-em, line 146: This is a fun way to eat the error 😄

li-em (approved):

icechunk/src/ops/gc.rs
```diff
@@ -205,131 +207,59 @@ pub enum GCError {
...
-    .await
-    .capture()??;
-    Ok(minfo)
+        self.retained.try_extend_weighted(ids).inject()?;
+        Ok(())
+    }
+
+    // Nothing to fold: `consume` already inserts every id into the shared
+    // sharded set from the decode workers in parallel. Funnelling millions of ids
+    // was a bottleneck.
```
dcherian:
```suggestion
    // sharded set from the decode workers in parallel.
```

icechunk/src/ops/walker.rs
```diff
@@ -0,0 +1,729 @@
...
+            let _entered = span.entered();
+            let Fetched { bytes, _permit } = fetched;
+            let manifest = decode_manifest(&bytes)?;
+            drop(bytes);
+            consumer.consume(&manifest)
+            // manifest and _permit drop here, releasing the budget
+        })
+        .await
+        .capture()??;
+        if tx.send(output).await.is_err() {
```
dcherian, line 142: Same comment here would be helpful. 
```
// receivers are gone because a decode worker failed; that error
// is reported through the JoinSet, not from here
```

FWIW this is only readable to me, because i went through the same circus in the zax executor.  Is this a standard pattern?

dcherian (approved):
very cool

PR #2406, opened 2026-09-20
perf: Parallel obj listing, decoupled deletes, GC progress

paraseba:
The AssetManager now lists id sub-prefixes on a pool of tasks,
so parsing responses spreads over cores. The repo size is probed
and 32 or 1024 sub-prefixes (object id initial characters) are
used. A new GC configuration `max_concurrent_listings` limits
the pool of processors. On an 8-core in-region instance against
S3: 130k keys/s before, ~800k after.

GC deletes now run on a deleter task, so listing doesn't
stall while a batch of deletes is in flight.

The gc_bench list subcommand measures listing throughput in two modes,
the store's single-stream baseline and the production path, optionally
against a real S3 bucket.

Reviewers:
- focus on the listing logic if you want but it's somewhat tricky.
  There is good test coverage (previous and new). I have tested
  performance extensively and it's a big win. The important code
  is in listing.rs
- don't even look in benchmarks
- gc.rs optimizes the delete parts that used listing intensively,
  this became so much faster. Worth a review.
- walker.rs not much going on, mostly adding progress logs

---

**Stack:** 6 of 15 — based on #2405 (`gc-improvements-05`).


icechunk/src/ops/gc.rs
```diff
@@ -837,58 +870,170 @@ where
...
+                return Ok(());
+            }
+            if !should_delete(&candidate) {
+                continue;
+            }
+            progress.candidates_accepted.fetch_add(1, AtomicOrdering::Relaxed);
+            batch.push((candidate.id.to_string(), candidate.size_bytes));
+            if batch.len() == batch_size.get() {
+                let full =
+                    std::mem::replace(&mut batch, Vec::with_capacity(batch_size.get()));
```
li-em, line 908: so sneaky

li-em (approved):

PR #2412, opened 2026-09-20
perf: Use mimalloc as allocator in the Python extension

paraseba:
glibc struggles maintaining the memory budget during GC runs
and the process can sometimes get OOM-killed.

This is what Claude thinks is the reason: 

> glibc malloc retained two to three times the live memory: the ~28 MB compressed
> manifest buffers sit under its mmap threshold and are freed on a different
> thread than they were allocated on, so they went back to arenas that were
> never trimmed.

We tried jemalloc and fixed it; mimalloc as the extension's
global allocator is more compatible. It is a default-on
feature so wheels get it; `--no-default-features --features cli,otel`
keeps the system allocator.

Claude also say:

Local-dynamic TLS because the extension is dlopen'ed by Python.

Reviewers: Do we think this will work in al our targets? How do we know?

---

**Stack:** 12 of 15 — based on #2411 (`gc-improvements-11`).


samantha-earthmover:
@paraseba linked you the internal PR for glibc + docker, if that solves this it would be much nicer than messing with the allocator.

li-em (commented):
+0 on this one... the system allocator is a safer default, and `mimalloc` is a more available fallback than `jemalloc`.

From a quick search large Rust projects like `polars` tend to use `mimalloc` only for windows, `jemalloc` on unix. So if we are going for a system allocator change (and for perf) it might be worth doing a similar split?

li-em (approved):

PR #2413, opened 2026-09-20
feat: Warn when the fd limit is too low for GC and chunk stats

paraseba:
**Stack:** 13 of 15 — based on #2412 (`gc-improvements-12`).


icechunk/src/ops/mod.rs
```diff
@@ -67,6 +67,49 @@ pub(crate) async fn ensure_repo_writable(
...
+    let Some(soft) = nofile_soft_limit() else { return };
+    let needed = peak_requests + FD_HEADROOM;
+    if soft < needed {
+        warn!(
+            soft_limit = soft,
+            needed,
+            "{op_name} peaks at {peak_requests} concurrent requests, but this \
+             process can only open {soft} files and will likely fail with \
+             \"too many open files\". Raise the limit (ulimit -n {needed}) or \
+             lower the concurrency settings."
```
ianhi, line 108: Would it be better to just dynamically adjust the combination of those params down to the limit so we don't fail, but still warn?

ianhi: hmm or maybe even we just temporaily raise the soft limit?

li-em (approved):

icechunk/src/ops/mod.rs
```diff
@@ -67,6 +67,49 @@ pub(crate) async fn ensure_repo_writable(
...
+pub(crate) fn warn_on_low_fd_limit(peak_requests: u64, op_name: &str) {
+    let Some(soft) = nofile_soft_limit() else { return };
+    let needed = peak_requests + FD_HEADROOM;
+    if soft < needed {
+        warn!(
+            soft_limit = soft,
+            needed,
+            "{op_name} peaks at {peak_requests} concurrent requests, but this \
+             process can only open {soft} files and will likely fail with \
+             \"too many open files\". Raise the limit (ulimit -n {needed}) or \
```
ianhi:
```suggestion
             \"too many open files\". Raise the limit (ulimit -S -n {needed}) or \
```

add `-S` to only modify soft, instead of potentailyl soft and hard (depends on your shell impl)

icechunk/src/ops/mod.rs
paraseba: I didn't want to do that because there are so many levels to the whole thing. The machine may not allow to raise the limit, or the user may not want that. The concurrency may be carefully set and known to work. The only thing I feel confident doing is checking and warning, and even that is heavily operative system dependent...

PR #2423, opened 2026-09-21
perf: don't write a snapshot when forking a clean session

paraseba:
Also, and this has more behavior change and subtlety, don't register fork snapshots.


icechunk-python/python/icechunk/session.py
```diff
@@ -700,6 +700,13 @@ def fork(self) -> "ForkSession":
 
 
 class ForkSession(Session):
+    """A `Session` that can be pickled to a worker, written to, and merged back.
+
+    Fork sessions are based on a snapshot that is not part of the repository, so they can
```
dcherian: > based on a snapshot that is not part of the repository

sounds weird to me

paraseba: improved

dcherian (approved):

